PR #2, opened 2019-04-25
Add controller to manage storage cluster pods

piyush-nimbalkar:
- supports pod creation and reconciliation of storage cluster pods
  - given a CR controller will install the storage pods
  - will restart a pod if deleted
  - deleting cluster spec deletes the pods (does not clean up anything else)
- using latest operator-sdk model of deployment (0.7.0) and controller runtime 0.1.10

TODO:
- add watch on node so we can reach on node changes
  - this is not critical as we reconcile every 30 seconds and will react on the node changes
- add placement constraints from the spec to the pods
- support updating params in the cluster spec
- clean up on removing the spec


Gopkg.toml
```diff
@@ -33,41 +33,41 @@ required = [
...
-  name = "k8s.io/apiserver"
-  version = "kubernetes-1.11.0"
-
-[[constraint]]
   name = "k8s.io/client-go"
-  version = "8.0.0"
+  version = "kubernetes-1.13.1"
 
-[[constraint]]
+[[override]]
```
harsh-px: can we just work with constraints? `override` is more of a force option.

Makefile
```diff
@@ -15,7 +15,7 @@ ifeq ($(BUILD_TYPE),debug)
 BUILDFLAGS += -gcflags "-N -l"
 endif
 
-RELEASE_VER := 1.0.0
+RELEASE_VER := 0.0.1
```
harsh-px: let's do 0.1

0.0.1 will look too alpha'ish.

cmd/operator.go
```diff
@@ -106,96 +95,35 @@ func run(c *cli.Context) {
...
+		log.Fatal(err)
 	}
 
-	clusterController := cluster.Controller{
-		Driver:   d,
-		Recorder: recorder,
-	}
-	if err := clusterController.Init(); err != nil {
-		log.Fatalf("Error initializing storage cluster controller: %v", err)
+	log.Info("Registering Components...")
```
harsh-px: nit: `Registering  components`

drivers/storage/storage.go
```diff
@@ -1,8 +1,10 @@
 package storage
 
 import (
+	corev1alpha1 "github.com/libopenstorage/operator/pkg/apis/core/v1alpha1"
 	"github.com/libopenstorage/operator/pkg/errors"
 	"github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
```
harsh-px: just use v1. alias not required. Same for other places.

Gopkg.toml
piyush-nimbalkar: This is necessary for now because some of the other packages try to vendor in older version of k8s. Once all the dependent packages are upgraded, we can change it back to constraint.

harsh-px: Can you file an issue just so that we don't forget?

piyush-nimbalkar: Done

pkg/apis/core/v1alpha1/storagecluster.go
```diff
@@ -40,18 +40,27 @@ type StorageClusterSpec struct {
...
 	// Kvdb is the information of kvdb that storage driver uses
 	Kvdb *KvdbSpec `json:"kvdb"`
 	// CloudStorage details of storage in cloud environment.
 	// Keep this only at cluster level for now until support to change
 	// cloud storage at node level is added.
 	CloudStorage *CloudStorageSpec `json:"cloudStorage"`
 	// SecretsProvider is the name of secret provider that driver will connect to
-	SecretsProvider string `json:"secretsProvider"`
-	// CSIEndpoint is the csi endpoint for the driver
-	CSIEndpoint string `json:"csiEndpoint"`
```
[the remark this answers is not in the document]
piyush-nimbalkar: We don't need to take the csi endpoint as we construct one in the spec. In annotation we are taking the csi version.

drivers/storage/portworx/deployment.go
```diff
@@ -0,0 +1,636 @@
...
+		}
+		if t.cluster.Spec.Storage.JournalDevice != nil &&
+			*t.cluster.Spec.Storage.JournalDevice != "" {
+			args = append(args, "-j", *t.cluster.Spec.Storage.JournalDevice)
+		}
+		if t.cluster.Spec.Storage.SystemMdDevice != nil &&
+			*t.cluster.Spec.Storage.SystemMdDevice != "" {
+			args = append(args, "-metadata", *t.cluster.Spec.Storage.SystemMdDevice)
+		}
+	}
```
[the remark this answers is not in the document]
piyush-nimbalkar: Need a round of error handling for lot of things. Adding a TODO comment at the top to tackle that.
Changed to `else if`.

drivers/storage/portworx/deployment.go
```diff
@@ -0,0 +1,636 @@
...
+		if t.isPKS {
+			volumeMounts = append(volumeMounts,
+				v1.VolumeMount{
+					Name:      "pxlogs",
+					MountPath: "/var/lib/osd/log",
+				},
+			)
+		}
+	} else {
+		// TODO: older versions of k8s don't support mountPropagation
```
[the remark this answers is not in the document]
piyush-nimbalkar: Not sure. Just left a comment in case we decide to support pre 1.10 of k8s.
I guess openshift 3.8 was using k8s 1.9 which does not have mount propagation.

drivers/storage/portworx/deployment.go
```diff
@@ -0,0 +1,636 @@
...
+		}
+
+		volumeMounts = append(volumeMounts, srcMount, kubeletMount)
+	}
+
+	// TODO: Imp: add etcd certs to the volume mounts
+	return volumeMounts
+}
+
+func (t *template) getVolumes() []v1.Volume {
```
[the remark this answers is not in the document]
piyush-nimbalkar: Keeping it two different function but using a common list of volume info object. This will make it slightly easier to add volumes in future.

drivers/storage/portworx/deployment.go
```diff
@@ -0,0 +1,636 @@
+package portworx
```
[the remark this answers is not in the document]
piyush-nimbalkar: Nice suggestions. Adding basic UTs for validation.
Will add more cases as we go forward.

cmd/operator.go
```diff
@@ -3,31 +3,24 @@ package main
...
-	"os/signal"
-	"syscall"
+	"time"
 
 	"github.com/libopenstorage/operator/drivers/storage"
 	_ "github.com/libopenstorage/operator/drivers/storage/portworx"
-	"github.com/libopenstorage/operator/pkg/cluster"
-	"github.com/libopenstorage/operator/pkg/controller"
+	"github.com/libopenstorage/operator/pkg/apis"
+	storagecluster "github.com/libopenstorage/operator/pkg/controller/storagecluster"
```
harsh-px: nit alias needed?

drivers/storage/portworx/deployment.go
```diff
@@ -0,0 +1,670 @@
...
+	corev1alpha1 "github.com/libopenstorage/operator/pkg/apis/core/v1alpha1"
+	"github.com/sirupsen/logrus"
+	"k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+const (
+	pxContainerName          = "portworx"
+	pxServiceAccount         = "px-account"
+	annotationIsDocker       = "portworx/is-docker"
```
harsh-px: let's use `portworx.io` for all annotation prefixes. Use a prefix const and create all annotation keys from that.

drivers/storage/portworx/deployment.go
```diff
@@ -0,0 +1,670 @@
...
+	for _, env := range t.cluster.Spec.Env {
+		envCopy := env.DeepCopy()
+		envList = append(envList, *envCopy)
+	}
+
+	return envList
+}
+
+func (t *template) getVolumeMounts() []v1.VolumeMount {
+	// TODO: Imp: add etcd certs to the volume mounts
```
harsh-px: What will be current workaround without etcd cert support?

drivers/storage/portworx/deployment.go
```diff
@@ -0,0 +1,670 @@
...
+	if t.cluster.Annotations[annotationCSIVersion] != "" {
+		args = append(args, "--csiversion", t.cluster.Annotations[annotationCSIVersion])
+	}
+
+	if t.isPxDev {
+		args = append(args, "--dev")
+	}
+
+	if t.isPKS {
+		args = append(args, "--keep-px-up")
```
harsh-px: Can you checkout with @zoxpx. The latest oci-monitor might have --keep-px-up behavior as the default. So we may not needed this.

drivers/storage/portworx/deployment.go
```diff
@@ -0,0 +1,670 @@
...
+	t.isOpenshift = err == nil && enabled
+
+	enabled, err = strconv.ParseBool(cluster.Annotations[annotationIsDocker])
+	t.isDocker = err == nil && enabled
+
+	enabled, err = strconv.ParseBool(cluster.Annotations[annotationIsPxDev])
+	t.isPxDev = err == nil && enabled
+
+	enabled, err = strconv.ParseBool(cluster.Annotations[annotationRunOnMaster])
+	t.runOnMaster = err == nil && enabled
```
harsh-px: we should not continue support for runOnMaster. Running on master was a requirement during 1.6 timeframe and we have no longer see any use cases.

drivers/storage/portworx/feature.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+	"strconv"
+)
+
+// Feature is the enum type for different features
+type Feature string
+
+const (
+	// CSI is used to indicate CSI feature
+	CSI Feature = "CSI"
```
harsh-px: prefix all feature constants with "Feature".

Since the pkg name is portworx, the usages will look like portworx.CSI. So it isn't intuitive this is a feature flag.

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -0,0 +1,733 @@
...
+		// If storage pod isn't supposed to run on node, but it is, delete all storage pods on node.
+		for _, pod := range storagePods {
+			podsToDelete = append(podsToDelete, pod.Name)
+		}
+	}
+
+	return nodesNeedingStoragePods, podsToDelete, nil
+}
+
+func (c *Controller) nodeShouldRunStoragePod(
```
adityadani: nit: doc what this function does and what it returns

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -0,0 +1,733 @@
...
+		var storagePodsRunning []*v1.Pod
+		for _, pod := range storagePods {
+			if pod.DeletionTimestamp != nil {
+				continue
+			}
+			if pod.Status.Phase == v1.PodFailed {
+				msg := fmt.Sprintf("Found failed storage pod %s on node %s, will try to kill it", pod.Name, node.Name)
+				logrus.Warnf(msg)
+				c.recorder.Eventf(cluster, v1.EventTypeWarning, FailedStoragePodReason, msg)
+				podsToDelete = append(podsToDelete, pod.Name)
```
adityadani: A pod to replace the one we are deleting will get scheduled in the next sync loop?

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -0,0 +1,733 @@
...
+	if err != nil {
+		return fmt.Errorf("couldn't get list of nodes when syncing storage cluster %#v: %v",
+			cluster, err)
+	}
+
+	var nodesNeedingStoragePods, podsToDelete []string
+	for _, node := range nodeList.Items {
+		nodesNeedingStoragePodsOnNode, podsToDeleteOnNode, err := c.podsShouldBeOnNode(&node, nodeToStoragePods, cluster)
+		if err != nil {
+			continue
```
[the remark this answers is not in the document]
disrani-px: Also need a log message here with the error

drivers/storage/portworx/deployment.go
piyush-nimbalkar: As discussed added a comment. Will keep it for older version of oci-mon

drivers/storage/portworx/deployment.go
piyush-nimbalkar: Removing support for px dev and running on master.

pkg/controller/storagecluster/storagecluster.go
piyush-nimbalkar: The pods we are deleting here are the ones that are in failed state. 
Irrespective of that, as the pod is deleted the controller gives us a callback on pod changes. We don't have to wait till the next sync loop as we are watching over pod changes.

drivers/storage/portworx/deployment.go
```diff
@@ -0,0 +1,670 @@
...
+		},
+		{
+			name:      "dbusmount",
+			hostPath:  "/var/run/dbus",
+			mountPath: "/var/run/dbus",
+		},
+	}
+
+	// dockerVolumeInfoList is a list of volumes applicable when running in docker
+	dockerVolumeInfoList = []volumeInfo{
```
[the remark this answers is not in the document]
piyush-nimbalkar: Removing docker support

pkg/controller/storagecluster/storagecluster.go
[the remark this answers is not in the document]
piyush-nimbalkar: The error is coming from `simulate()` and we are logging it right there. If there is an error in there it is likely to not go away immediately and will flood the logs. Will let it retry after resync period.

drivers/storage/portworx/deployment.go
```diff
@@ -0,0 +1,670 @@
...
+	mountPath        string
+	readOnly         bool
+	mountPropagation *v1.MountPropagationMode
+	hostPathType     *v1.HostPathType
+	pks              *pksVolumeInfo
+	openshift        *openshiftVolumeInfo
+	coreos           *coreosVolumeInfo
+}
+
+type pksVolumeInfo struct {
```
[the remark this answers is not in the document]
piyush-nimbalkar: Had made 3 separate types because there could be overlaps in future. Anyway removed coreos and openshift types because they were needed only when running as docker. For runc it is the same mounts with or without the options enabled.

drivers/storage/portworx/deployment.go
piyush-nimbalkar: No workaround for etcd with auth. Going to do soon in the next couple of PRs.

PR #3, opened 2019-05-01
Update storage pods based on spec changes

piyush-nimbalkar:
- Adding update strategy similar to daemon sets
- Use controller revisions to store revisions of the storage cluster spec. In future we could even provide a way to rollback to certain revision like daemon sets
  - Create revision every time cluster spec changes
  - Try to get pod to the latest revision (in rolling update)
  - If the latest revision does not have changes that need a pod restart, no need to restart the pod 
- Add defaults to spec if not present. This will give better visibility to the user on what defaults are being used.
- Change labels to use `libopenstorage.org/name` and `libopenstorage.org/driver` and let the driver set it's own labels.
- [portworx] Add runtime options to the container args if present in the cluster spec


drivers/storage/portworx/deployment.go
```diff
@@ -386,6 +381,23 @@ func (t *template) getArguments() []string {
 		args = append(args, "--log", t.cluster.Annotations[annotationLogFile])
 	}
 
+	rtOpts := make([]string, 0)
+	for k, v := range t.cluster.Spec.RuntimeOpts {
+		key := strings.TrimSpace(k)
+		value := strings.TrimSpace(v)
+		_, err := strconv.Atoi(value)
+		if err != nil {
+			logrus.Warnf("Invalid integer value %v in runtime options", value)
```
harsh-px: log `err` here

drivers/storage/portworx/portworx.go
```diff
@@ -2,12 +2,15 @@ package portworx
 
 import (
 	storage "github.com/libopenstorage/operator/drivers/storage"
+	corev1alpha1 "github.com/libopenstorage/operator/pkg/apis/core/v1alpha1"
```
harsh-px: alias not needed

pkg/controller/storagecluster/util.go
```diff
@@ -0,0 +1,108 @@
...
+	v1helper "k8s.io/kubernetes/pkg/apis/core/v1/helper"
+	schedapi "k8s.io/kubernetes/pkg/scheduler/api"
+	hashutil "k8s.io/kubernetes/pkg/util/hash"
+)
+
+func addOrUpdateStoragePodTolerations(pod *v1.Pod) {
+	// StorageCluster pods shouldn't be deleted by NodeController in case of node problems.
+	// Add infinite toleration for taint notReady:NoExecute here to survive taint-based
+	// eviction enforced by NodeController when node turns not ready.
+	v1helper.AddOrUpdateTolerationInPod(pod, &v1.Toleration{
```
harsh-px: We should also set ourselves as critical pods: https://kubernetes.io/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods/

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -192,9 +230,50 @@ func (c *Controller) syncStorageCluster(
...
+
+	// TODO: Don't process a storage clsuter until all its previous creations and
+	// deletions have been processed.
+	err = c.manage(cluster, hash)
+	if err != nil {
+		return err
+	}
+
+	switch cluster.Spec.UpdateStrategy.Type {
+	case corev1alpha1.OnDeleteStorageClusterStrategyType:
```
harsh-px: we should warn that ondelelete isn't supported and that we are falling back on RollingUpdate.

harsh-px: Also add an issue to add support for this

drivers/storage/portworx/portworx.go
piyush-nimbalkar: Just keeping it consistent in all the code

pkg/controller/storagecluster/storagecluster.go
piyush-nimbalkar: It is actually supported. Basically we don't update the pods here until it is deleted.

pkg/apis/core/v1alpha1/storagecluster.go
```diff
@@ -139,6 +185,9 @@ type StorageSpec struct {
 	// UseAllWithPartitions use all available unformatted devices
 	// including partitions. This will be ignored if Devices is not empty.
 	UseAllWithPartitions *bool `json:"useAllWithPartitions"`
+	// Force use the drives even if there is file system present on it. Note
+	// that the drives will be wiped before using.
+	Force *bool `json:"force"`
```
disrani-px: Can we make this more descriptive like `ForceUseDisks`? `Force` could mean any number of things

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -192,9 +230,50 @@ func (c *Controller) syncStorageCluster(
...
+	case corev1alpha1.RollingUpdateStorageClusterStrategyType:
+		err = c.rollingUpdate(cluster, hash)
+	}
+	if err != nil {
```
disrani-px: Check is only required for the RollingUpdate case above?

pkg/controller/storagecluster/update.go
```diff
@@ -0,0 +1,604 @@
...
+			return err
+		}
+		toKill--
+	}
+
+	return nil
+}
+
+// isPodUpdated checks if pod contains label value that matches the hash
+func (c *Controller) isPodUpdated(cluster *corev1alpha1.StorageCluster, pod *v1.Pod, hash string) bool {
```
disrani-px: The name for this is a little confusing. Can it be renamed to `isPodUpdateNeeded`?

pkg/controller/storagecluster/update.go
```diff
@@ -0,0 +1,604 @@
...
+			Namespace:       cluster.Namespace,
+			Labels:          historyLabels,
+			Annotations:     cluster.Annotations,
+			OwnerReferences: []metav1.OwnerReference{*metav1.NewControllerRef(cluster, controllerKind)},
+		},
+		Data:     runtime.RawExtension{Raw: patch},
+		Revision: revision,
+	}
+
+	err = c.client.Create(context.TODO(), history)
```
adityadani: err from Create is not being used under the if condition.
outErr := c.client.Create()
if errors.IsAlreadyExists(outerErr)

pkg/controller/storagecluster/update.go
```diff
@@ -0,0 +1,604 @@
...
+		getErr := c.client.Get(
+			context.TODO(),
+			types.NamespacedName{
+				Name:      name,
+				Namespace: cluster.Namespace,
+			},
+			existedHistory,
+		)
+		if getErr != nil {
+			return nil, getErr
```
adityadani: may be you want to wrap the getErr with why we were doing Get in the first place.
I am assuming this will show up when someone does a get status of Storage Cluster.

adityadani: same for others in this if condition

pkg/controller/storagecluster/update.go
piyush-nimbalkar: It is actually the reverse of that. We return true when we the pod is updated or at least selected fields have not changed (matched the current cluster spec). 
For clarity adding more comments in the function.

drivers/storage/portworx/deployment.go
```diff
@@ -386,6 +381,23 @@ func (t *template) getArguments() []string {
 		args = append(args, "--log", t.cluster.Annotations[annotationLogFile])
 	}
 
+	rtOpts := make([]string, 0)
+	for k, v := range t.cluster.Spec.RuntimeOpts {
```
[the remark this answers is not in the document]
piyush-nimbalkar: Just keeping it generic for future proofing. Also this is not specific to Portworx where options could have non int values.

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -593,6 +607,45 @@ func (c *Controller) simulate(
...
+	if updateStrategy.Type == "" {
+		updateStrategy.Type = corev1alpha1.RollingUpdateStorageClusterStrategyType
+	}
+	if updateStrategy.Type == corev1alpha1.RollingUpdateStorageClusterStrategyType {
+		if updateStrategy.RollingUpdate == nil {
+			updateStrategy.RollingUpdate = &corev1alpha1.RollingUpdateStorageCluster{}
+		}
+		if updateStrategy.RollingUpdate.MaxUnavailable == nil {
+			// Set default MaxUnavailable as 1 by default.
+			maxUnavailable := intstr.FromInt(defaultMaxUnavailablePods)
```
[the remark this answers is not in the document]
piyush-nimbalkar: Yes that's correct. 
Either we can decide that based of a new custom policy or if it is internal to the driver, then we should at least not go beyond this maxUnavailablePods if user has given one.

pkg/controller/storagecluster/util.go
piyush-nimbalkar: Adding a TODO for this. Will explore more with system-cluster-critical and system-node-critical. The annotation is getting removed from k8s 1.14.

PR #4, opened 2019-05-02
Restructuring the files and directory according to operator guidelines

piyush-nimbalkar:
- made changes for operator certification
- add portworx config under olm-catalog that will be sent for certification and fields from it will be visible under the operator tab on openshift


deploy/olm-catalog/portworx/1.0.0/portworxoperator.v1.0.0.clusterserviceversion.yaml
```diff
@@ -0,0 +1,254 @@
...
+    alm-examples: |-
+      [
+        {
+          "apiVersion": "core.libopenstorage.org/v1alpha1",
+          "kind": "StorageCluster",
+          "metadata": {
+            "name": "portworx",
+            "namespace": "kube-system"
+          },
+          "spec": {
```
disrani-px: Also set internalKVDB to true?

build/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM registry.access.redhat.com/rhel7-atomic
+
+LABEL name="OpenStorage Operator" \
+      vendor="" \
+      version="v0.1.0" \
+      release="1" \
+      summary="OpenStorage Operator" \
+      description="This operator will deploy 3 instances of OpenStorage operator in the cluster."
```
disrani-px: I think the `3 instances` should be removed from the description here. This will be the docker image description which doesn't need that detail. It can be in the operator metadata.

deploy/operator.yaml
```diff
@@ -0,0 +1,46 @@
...
+        name: portworx-operator
+    spec:
+      containers:
+      - command:
+        - /operator
+        - --verbose
+        - --driver=portworx
+        - --leader-elect=true
+        imagePullPolicy: Always
+        image: portworx/openstorage-operator:latest
```
disrani-px: Use a tag instead of latest.

deploy/operator.yaml
```diff
@@ -0,0 +1,46 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  annotations:
+    scheduler.alpha.kubernetes.io/critical-pod: ""
+  name: portworx-operator
+  namespace: kube-system
```
disrani-px: Doesn't the catalog allow installing the operator in any namespace? We should not hard code this to kube-system.

deploy/operator.yaml
```diff
@@ -0,0 +1,46 @@
...
+  strategy:
+    rollingUpdate:
+      maxSurge: 1
+      maxUnavailable: 1
+    type: RollingUpdate
+  replicas: 1
+  template:
+    metadata:
+      annotations:
+        scheduler.alpha.kubernetes.io/critical-pod: ""
```
disrani-px: This annotation is being deprecated in 1.14. `priorityClassName` should be used instead. Also, I'm not sure if we need this for the operator itself. This is usually required to prevent eviction for pods that can affect the operation of a cluster if it is evicted. The operator will come up on another node in this case without any negative effect.

deploy/operator.yaml
piyush-nimbalkar: This namespace is not taken by the catalog. The catalog csv has a placeholder namespace. It takes only the pod template from this deployment.

deploy/olm-catalog/portworx/1.0.0/portworxoperator.v1.0.0.clusterserviceversion.yaml
piyush-nimbalkar: Had not added it because it is the default anyway. Will put it to make it explicit.

piyush-nimbalkar:
Could not resolve conflicts on this branch. Created another with cherrypicks - https://github.com/libopenstorage/operator/pull/11
Closing this and merging the other.

PR #5, opened 2019-05-03
Adding RBAC permissions for portworx before installing

piyush-nimbalkar:


drivers/storage/portworx/rbac.go
```diff
@@ -0,0 +1,284 @@
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+const (
+	pxServiceAccountName     = "px-account"
+	pxServiceName            = "portworx-service"
+	pxSecretsNamespace       = "portworx"
+	pxClusterRoleName        = "node-get-put-list-role"
```
harsh-px: at some point, we should rename these and handle upgrades. I had named these wrong 2 years back to start with :) The actual role now has very different permissions.

piyush-nimbalkar: Will add it in a separate PR.

PR #13, opened 2019-05-07
Adding generic pre exec command for rhel coreos 

piyush-nimbalkar:
- Added all spec configs to CR example

### DIFF
--- a/drivers/storage/portworx/deployment.go
+++ b/drivers/storage/portworx/deployment.go
@@ -466,12 +466,14 @@ func (t *template) getEnvList() []v1.EnvVar {
 		},
 	}
 
-	if t.isPKS {
-		envList = append(envList, v1.EnvVar{
-			Name:  "PRE-EXEC",
-			Value: "if [ ! -x /bin/systemctl ]; then apt-get update; apt-get install -y systemd; fi",
-		})
+	preExecEnv := v1.EnvVar{
+		Name:  "PRE-EXEC",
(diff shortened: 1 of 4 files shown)


drivers/storage/portworx/deployment.go
```diff
@@ -466,12 +466,14 @@ func (t *template) getEnvList() []v1.EnvVar {
...
+	preExecEnv := v1.EnvVar{
+		Name:  "PRE-EXEC",
+		Value: "if grep -q 'Red Hat Enterprise Linux CoreOS' /etc/os-release && /bin/ls -dZ /var/opt/ | grep -q :object_r:var_t: ; then semanage fcontext -a -t usr_t '/var/opt(/.*)?' ; restorecon -Rv /var/opt/ ; fi",
+	}
 	if t.isPKS {
-		envList = append(envList, v1.EnvVar{
-			Name:  "PRE-EXEC",
-			Value: "if [ ! -x /bin/systemctl ]; then apt-get update; apt-get install -y systemd; fi",
-		})
+		preExecEnv.Value = "if [ ! -x /bin/systemctl ]; then apt-get update; apt-get install -y systemd; fi"
```
adityadani: Can't there be more than one preExecEnv variables?
For now, this is OK, but what if we want the above preExec Value along with specifics for PKS

piyush-nimbalkar: Currently there is only one. We may not even need this if the issue if fixed in rhel. For now, PKS will not get this PRE-EXEC and continue with it's PRE-EXEC for simplicity.

PR #18, opened 2019-05-28
Correct validation for integer fields and add more validations

piyush-nimbalkar:

### DIFF
--- a/deploy/olm-catalog/portworx/1.0.3/core_v1alpha1_storagecluster_crd.yaml
+++ b/deploy/olm-catalog/portworx/1.0.3/core_v1alpha1_storagecluster_crd.yaml
@@ -33,35 +33,55 @@ spec:
               type: string
             imagePullPolicy:
               type: string
-            imagePullSecret: {}
-            secretsProvider: {}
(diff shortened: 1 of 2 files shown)


deploy/olm-catalog/portworx/1.0.3/core_v1alpha1_storagecluster_crd.yaml
```diff
@@ -33,35 +33,55 @@ spec:
...
-                systemMetadataDevice: {}
-                dataStorageType: {}
-                raidLevel: {}
+                journalDevice:
+                  type: string
+                systemMetadataDevice:
+                  type: string
+                dataStorageType:
+                  type: string
+                raidLevel:
```
harsh-px: Does `raidLevel` level values other than int? I see in the CRD it's defined as a string. Just checking if it needs to a string.

piyush-nimbalkar: Not that I know of but we store raid level as string. Will keep it as string for extensibility.

PR #22, opened 2019-06-25
CSI support for portworx driver

piyush-nimbalkar:


drivers/storage/portworx/deployment.go
```diff
@@ -559,33 +654,26 @@ func (t *template) getVolumes() []v1.Volume {
...
-	if strings.Contains(registryAndRepo, "/") && len(imgParts) > 1 {
-		// advance to the last element, skipping image's repository
-		imgParts = imgParts[len(imgParts)-1:]
-	}
-	return registryAndRepo + "/" + path.Join(imgParts...)
+	volumeInfoList := []volumeInfo{registrationVol}
+	volumeInfoList = append(volumeInfoList,
+		volumeInfo{
+			name:         "csi-driver-path",
+			hostPath:     "/var/lib/kubelet/plugins/com.openstorage.pxd",
```
harsh-px: @lpabon with the CSI driver name change, all of these paths will need to change too right?

What version is the driver name change effective from?

PR #23, opened 2019-07-03
Add cluster and node status from the driver

piyush-nimbalkar:
- Adding cluster status for portworx driver
- Creating and updating StorageNodeStatus objects for node statuses. This gives a k8s native way to access storage driver nodes.


drivers/storage/portworx/portworx.go
```diff
@@ -168,6 +182,239 @@ func (p *portworx) DeleteStorage(storageCluster *corev1alpha1.StorageCluster) (*
...
+				Network: corev1alpha1.NetworkStatus{
+					DataIP: nodeResp.Node.DataIp,
+					MgmtIP: nodeResp.Node.MgmtIp,
+				},
+				// TODO: Add a human readable reason with the status
+				Conditions: []corev1alpha1.NodeCondition{
+					{
+						Type:   corev1alpha1.NodeState,
+						Status: mapNodeStatus(nodeResp.Node.Status),
+					},
```
adityadani: We need StorageState as well here

drivers/storage/portworx/portworx.go
```diff
@@ -168,6 +182,239 @@ func (p *portworx) DeleteStorage(storageCluster *corev1alpha1.StorageCluster) (*
...
+	}
+
+	ownerRef := metav1.NewControllerRef(cluster, controllerKind)
+	for _, nodeID := range nodeEnumerateResponse.NodeIds {
+		nodeResp, err := nodeClient.Inspect(context.TODO(), &api.SdkNodeInspectRequest{NodeId: nodeID})
+		if err != nil {
+			logrus.Warnf("Failed to inspect node %v: %v", nodeID, err)
+			continue
+		} else if nodeResp.Node == nil {
+			logrus.Warnf("Empty NodeInspect response for nodeID %v: %v", nodeID, err)
```
disrani-px: No need to print `err` here, it'll always be `nil`

disrani-px: Also, if we get an empty response for a ndoe do we want to set a `StatusUnknown` status instead of not listing it at all?

drivers/storage/portworx/portworx.go
```diff
@@ -168,6 +182,239 @@ func (p *portworx) DeleteStorage(storageCluster *corev1alpha1.StorageCluster) (*
...
+						Type:   corev1alpha1.NodeState,
+						Status: mapNodeStatus(nodeResp.Node.Status),
+					},
+				},
+			},
+		}
+
+		err = k8sutil.CreateOrUpdateStorageNodeStatus(p.k8sClient, nodeStatus, ownerRef)
+		if err != nil {
+			logrus.Warnf("Failed to update status for nodeID %v: %v", nodeID, err)
```
disrani-px: Also log an event to the cluster object so that users can see it

drivers/storage/portworx/portworx.go
```diff
@@ -168,6 +182,239 @@ func (p *portworx) DeleteStorage(storageCluster *corev1alpha1.StorageCluster) (*
...
+
+	case api.Status_STATUS_NOT_IN_QUORUM:
+		fallthrough
+	case api.Status_STATUS_NOT_IN_QUORUM_NO_STORAGE:
+		return corev1alpha1.ClusterNotInQuorum
+
+	case api.Status_STATUS_OK:
+		fallthrough
+	case api.Status_STATUS_STORAGE_DOWN:
+		return corev1alpha1.ClusterRunning
```
disrani-px: Can we use `ClusterOnline` since we have `ClusterOffline`

drivers/storage/portworx/portworx.go
```diff
@@ -168,6 +182,239 @@ func (p *portworx) DeleteStorage(storageCluster *corev1alpha1.StorageCluster) (*
...
+		fallthrough
+	case api.Status_STATUS_INIT:
+		fallthrough
+	case api.Status_STATUS_OFFLINE:
+		fallthrough
+	case api.Status_STATUS_ERROR:
+		fallthrough
+	case api.Status_STATUS_DECOMMISSION:
+		fallthrough
+	case api.Status_STATUS_MAINTENANCE:
```
disrani-px: This is more a node status than a cluster status.

drivers/storage/portworx/portworx.go
```diff
@@ -168,6 +182,239 @@ func (p *portworx) DeleteStorage(storageCluster *corev1alpha1.StorageCluster) (*
...
+		fallthrough
+	case api.Status_STATUS_NEEDS_REBOOT:
```
disrani-px: Same as above

pkg/apis/core/v1alpha1/storagenodestatus.go
```diff
@@ -0,0 +1,100 @@
...
+	StorageState NodeConditionType = "StorageState"
+)
+
+// ConditionStatus is the enum type for node condition statuses
+type ConditionStatus string
+
+// These are valid statuses of different node conditions.
+const (
+	// NodeOnline means the node condition is online and healthy
+	NodeOnline ConditionStatus = "Online"
```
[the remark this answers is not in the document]
disrani-px: I think we can use the same statuses for both and just use the `Reason` field to say why it is in that state. For example for `StorageFull`, the status could be `Degraded` and `Reason` could say `Storage device/pool is full`

drivers/storage/portworx/portworx.go
piyush-nimbalkar: We don't have a way to get storage status right now. Once we have the internal SDK we can add that.

drivers/storage/portworx/portworx.go
piyush-nimbalkar: Now using Enumerate which gets all nodes. Won't be needing to update status to Unknown.

piyush-nimbalkar:
Thanks for the review guys!

PR #27, opened 2019-07-23
Add missing volumes mounts for node wiper

piyush-nimbalkar:
- Wipe metadata only if UninstallAndWipe strategy is selected
- Use k8s client for controller-runtime instead of sched ops in uninstall.go
- Do not error out in storagecluster controller if cluster object is already deleted
- Add unit tests for uninstall/delete workflows in portworx driver

### DIFF
--- a/pkg/util/k8s/k8s.go
+++ b/pkg/util/k8s/k8s.go
@@ -730,3 +730,40 @@ func CreateOrUpdateStorageNodeStatus(
 	}
 	return nil
 }
+
+// GetDaemonSetPods returns a list of pods for the given daemon set
+func GetDaemonSetPods(
+	k8sClient client.Client,
+	ds *appsv1.DaemonSet,
+) ([]*v1.Pod, error) {
+	return GetPodsByOwner(k8sClient, ds.UID, ds.Namespace)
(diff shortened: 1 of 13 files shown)


pkg/util/k8s/k8s.go
```diff
@@ -730,3 +730,40 @@ func CreateOrUpdateStorageNodeStatus(
...
+// GetDaemonSetPods returns a list of pods for the given daemon set
+func GetDaemonSetPods(
+	k8sClient client.Client,
+	ds *appsv1.DaemonSet,
+) ([]*v1.Pod, error) {
+	return GetPodsByOwner(k8sClient, ds.UID, ds.Namespace)
+}
+
+// GetPodsByOwner returns pods for a given owner and namespace
+func GetPodsByOwner(
```
harsh-px: Can we not use https://github.com/portworx/sched-ops/blob/master/k8s/k8s.go#L2564?

piyush-nimbalkar: This gets it from the cache. We are using the client from sdk everywhere to reduce api server calls as much as possible.

PR #28, opened 2019-07-26
Add env vars to stork and lighthouse specs

piyush-nimbalkar:
- Pass the envs in stork section to stork deployment
- Override the images for lighthouse sidecars if corresponding envs are set
- Override the px node wiper image if corresponding env is set
- PVC controller should not run if running in kube-system namespace in Openshift
- PVC controller should not run if explicitly disabled


drivers/storage/portworx/components.go
```diff
@@ -1136,14 +1143,38 @@ func (p *portworx) createLighthouseDeployment(
...
+	partitions := strings.Split(t.cluster.Spec.UserInterface.Image, ":")
+	if len(partitions) > 1 {
+		imageTag = partitions[len(partitions)-1]
+	}
 
-	if !p.lhDeploymentCreated || lhImageName != existingImage {
-		deployment := getLighthouseDeploymentSpec(t, ownerRef, lhImageName)
+	configSyncImage := getImageFromEnv(envKeyLhConfigSyncImage, t.cluster.Spec.UserInterface.Env)
+	if len(configSyncImage) == 0 {
+		configSyncImage = "portworx/lh-config-sync:" + imageTag
```
disrani-px: Use constants for the default image names

drivers/storage/portworx/deployment.go
```diff
@@ -231,12 +231,20 @@ func newTemplate(cluster *corev1alpha1.StorageCluster) (*template, error) {
 	enabled, err = strconv.ParseBool(cluster.Annotations[annotationIsOpenshift])
 	t.isOpenshift = err == nil && enabled
 
+	// Enable PVC controller for managed kubernetes services. Also enable it for openshift,
+	// only if Portworx service is node deployed in kube-system namespace.
```
disrani-px: `not deployed`

PR #31, opened 2019-08-05
Add node status with phase and version to display on the UI

piyush-nimbalkar:
- Adding the node status object to openshift UI
- Adding phase/status and version fields to the StorageNodeStatus object so it can be displayed in the Openshift list view
- Giving all permissions to the operator deployment as stork needs it anyway and will be easier to add permissions in the future.
- Adding more validations to the CRD yamls


drivers/storage/portworx/portworx.go
```diff
@@ -266,16 +268,30 @@ func (p *portworx) updateNodeStatuses(
...
 					{
 						Type:   corev1alpha1.NodeState,
-						Status: mapNodeStatus(node.Status),
+						Status: phase,
 					},
 				},
 			},
 		}
 
+		if version, ok := node.NodeLabels["PX Version"]; ok {
```
disrani-px: Make `'PX Version` a constant and use that here

drivers/storage/portworx/components.go
```diff
@@ -299,16 +299,41 @@ func (p *portworx) unsetInstallParams() {
...
-		Scope:      apiextensionsv1beta1.ClusterScoped,
-		Kind:       "VolumePlacementStrategy",
-		ShortNames: []string{"vps", "vp"},
-	}
-	err := k8s.Instance().CreateCRD(resource)
+		Plural: "volumeplacementstrategies",
+		Group:  "portworx.io",
+	}
+
+	crd := &apiextensionsv1beta1.CustomResourceDefinition{
```
disrani-px: This shouldn't be done by the operator. This needs to be maintained by Portworx since there is a dependency on which version of PX can use which version of this CRD.

piyush-nimbalkar: We discussed offline, we will move this to Portworx eventually. For now keeping it operator until Portworx registers it's own CRD.

PR #32, opened 2019-08-06
Add CloudStorageSpec to StorageCluster.

adityadani:
The changes are divided into the following commits:

*1st commit*
- Change the StorageCluster version from v1alpha1 to v1alpha2
- Remove the list of device specs from CloudStorage and use CapacitySpecs instead.
- Modify StorageNodeStatus to include cloud config that gets provisioned on the node.
This commit contains only the CRD changes.

*2nd commit*
- Generated code.

*3rd commit*
- Change references from v1alpha1 to v1alpha2

*4th commit*
-  Add a new Cloud Storage Manager interface.
- Adds an abstraction layer/interface that provides APIs to interact with the cloudops library.
- Currently only the Portworx driver implements it.
- There is no hard dependency on a driver to implement this interface.

*5th commit*
- Mock generated code for Driver and CloudStorageManager

*6th commit*
-  Changes to the StorageCluster controller to handle new CloudStorageSpec.
- Add a new UpdateDriver API to the Driver interface to pass in information from the controller
   to the driver which cannot be passed as a part of StorageCluster.
- Modify GetStoragePodSpec implementation to take in a node name.
- Use the CloudStorageManager interface to get per node specs.
- If the NodeStatus object already has the cloud config do not invoke the cloudops library.
- Modify the arguments generation logic to handle the new cloud storage spec.
- Add and modify existing UTs for cloud storage spec.

```
TODO
1. Handle upgrades from v1alpha1 to v1alpha2 StorageCluster objects.
2. Read the yaml decision matrix and invoke the cloudops library.
3. Persist the cloud storage config in NodeStatus object.
4. Determine the no. of zones and cloud provider from node labels.
```


adityadani:
Changes moved to #39

PR #38, opened 2019-08-08
Detect cloud provider and zone/instances information in controller loop.

adityadani:
- Pass on this information to the driver through the new UpdateDriver API.

- The UpdateDriver API can be used to keep updating the Driver with information that the controller has.

### DIFF
--- a/drivers/storage/portworx/portworx.go
+++ b/drivers/storage/portworx/portworx.go
@@ -51,6 +51,8 @@ type portworx struct {
 	lhDeploymentCreated               bool
 	csiStatefulSetCreated             bool
 	sdkConn                           *grpc.ClientConn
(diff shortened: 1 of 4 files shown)


drivers/storage/portworx/portworx.go
```diff
@@ -69,6 +71,12 @@ func (p *portworx) Init(k8sClient client.Client, recorder record.EventRecorder)
 	return nil
 }
 
+func (p *portworx) UpdateDriver(info *storage.UpdateDriverInfo) error {
+	p.zoneToInstancesMap = info.ZoneToInstancesMap
+	p.cloudProvider = info.CloudProvider
```
piyush-nimbalkar: We could convert this into Cloudops cloud provider type for future use. Can do it later when you start using it.

PR #39, opened 2019-08-10
Handle CapacitySpecs in Portworx Driver

adityadani:
- Update StorageCluster CRD to include CapacitySpecs in CloudStorageSpec. …
- Regenerate code.
- Change the Portworx driver to handle CapacitySpecs.
- Invoke the CloudStorage API to get per node drive config.
- Parse the storage distribution returned by cloudops into arguments.
- UTs for CapacitySpecs

### DIFF
--- a/drivers/storage/portworx/deployment.go
+++ b/drivers/storage/portworx/deployment.go
@@ -575,8 +575,8 @@ func (t *template) getArguments() []string {
 	} else if t.cluster.Spec.CloudStorage != nil {
 		if t.cloudConfig != nil && len(t.cloudConfig.CloudStorage) > 0 {
 			// CapacitySpecs have higher preference over DeviceSpecs
-			for _, cloudNodeSpec := range t.cloudConfig.CloudStorage {
-				args = append(args, "-s", t.getCloudStorageArguments(cloudNodeSpec))
+			for _, cloudDriveSpec := range t.cloudConfig.CloudStorage {
+				args = append(args, "-s", t.getCloudStorageArguments(cloudDriveSpec))
 			}
 		} else if t.cluster.Spec.CloudStorage.DeviceSpecs != nil {
 			for _, dev := range *t.cluster.Spec.CloudStorage.DeviceSpecs {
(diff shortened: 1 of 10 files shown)


drivers/storage/portworx/deployment.go
```diff
@@ -440,6 +482,27 @@ func (t *template) getSelectorRequirements() []v1.NodeSelectorRequirement {
 	return selectorRequirements
 }
 
+func (t *template) getCloudStorageArguments(cloudDeviceSpec cloudstorage.CloudDriveConfig) string {
+	devSpec := strings.Join(
+		[]string{
+			"type=" + cloudDeviceSpec.Type,
+			"size=" + strconv.FormatUint(cloudDeviceSpec.SizeInGiB, 10),
+			"iops=" + strconv.FormatUint(uint64(cloudDeviceSpec.IOPS), 10)},
```
piyush-nimbalkar: Question:
Will IOPS always be present in the output from cloudops?
For some providers or drive type, Portworx will ignore the iops param, right?

drivers/storage/portworx/deployment.go
```diff
@@ -510,26 +573,32 @@ func (t *template) getArguments() []string {
 		}
 
 	} else if t.cluster.Spec.CloudStorage != nil {
-		if t.cluster.Spec.CloudStorage.DeviceSpecs != nil {
+		if t.cloudConfig != nil && len(t.cloudConfig.CloudStorage) > 0 {
+			// CapacitySpecs have higher preference over DeviceSpecs
+			for _, cloudNodeSpec := range t.cloudConfig.CloudStorage {
```
piyush-nimbalkar: nit: Can we call this cloudDriveSpec? Node spec is a bit misleading.

drivers/storage/portworx/deployment.go
```diff
@@ -510,26 +573,32 @@ func (t *template) getArguments() []string {
...
 			args = append(args, "-metadata", *t.cluster.Spec.CloudStorage.SystemMdDeviceSpec)
 		}
-		if t.cluster.Spec.CloudStorage.MaxStorageNodes != nil {
+		if t.cluster.Spec.CloudStorage.MaxStorageNodes != nil &&
+			*t.cluster.Spec.CloudStorage.MaxStorageNodes > 0 {
 			args = append(args, "-max_drive_set_count",
 				strconv.Itoa(int(*t.cluster.Spec.CloudStorage.MaxStorageNodes)))
 		}
-		if t.cluster.Spec.CloudStorage.MaxStorageNodesPerZone != nil {
+		if t.cloudConfig != nil && t.cloudConfig.StorageInstancesPerZone > 0 {
```
piyush-nimbalkar: If cloudConfig is nil, we should still pass the `cluster.Spec.CloudStorage.MaxStorageNodesPerZone` for existing consumers.

drivers/storage/portworx/deployment.go
adityadani: Yes thats right. I think the operator should just pass on the parameters it gets from cloudops. The actual driver in Portworx can handle if it needs to ignore.

drivers/storage/portworx/deployment.go
adityadani: done!

drivers/storage/portworx/deployment.go
```diff
@@ -510,24 +573,36 @@ func (t *template) getArguments() []string {
...
```
piyush-nimbalkar: We could add a test for this case.

adityadani: done

adityadani:
The travis builds are failing. It seems like travis is pulling an older version of master which was force pushed later.
Ran the tests locally and those passed.

PR #41, opened 2019-08-13
Package storage distribution yamls

adityadani:
- Package libopenstorage/cloudop's storage distribution yamls into operator container.

- Added a new API CreateStorageDecisionMatrix in CloudManager.
- Read the cloud provider specific yamls packaged in the operator
  and create a config map if it does not exist.
- Added UTs for the new API.

### DIFF
--- a/drivers/storage/portworx/cloud_storage.go
+++ b/drivers/storage/portworx/cloud_storage.go
@@ -4,12 +4,18 @@ import (
 	"context"
 	"fmt"
 	"math"
+	"path"
 
 	"github.com/libopenstorage/cloudops"
 	"github.com/libopenstorage/cloudops/pkg/parser"
+
+	// importing the decisionmatrix package from cloudops to vendor
+	// the cloud specific decision matrices
+	_ "github.com/libopenstorage/cloudops/specs/decisionmatrix"
 	corev1alpha1 "github.com/libopenstorage/operator/pkg/apis/core/v1alpha1"
 	"github.com/libopenstorage/operator/pkg/cloudstorage"
 	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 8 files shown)


drivers/storage/portworx/cloud_storage.go
```diff
@@ -98,6 +105,46 @@ func (p *portworxCloudStorage) GetStorageNodeConfig(
...
+	decisionMatrixCM := &v1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      storageDecisionMatrixCMName,
+			Namespace: p.namespace,
+		},
+		Data: map[string]string{
+			storageDecisionMatrixCMKey: string(yamlBytes),
+		},
+	}
+	return p.k8sClient.Create(
```
harsh-px: set owner references so that the yaml gets deleted when the operator is deleted.

drivers/storage/portworx/deployment.go
```diff
@@ -287,6 +287,10 @@ func (p *portworx) generateCloudStorageSpecs(
 		p.k8sClient,
 	}
 
+	if err := cloudStorageManager.CreateStorageDistributionMatrix(); err != nil {
+		logrus.Warnf("Failed to generate storage distribution matrix config map: %v", err)
```
harsh-px: why is this just a warning?

adityadani: This is to not error out in scenarios where someone wants to create the config map out of band.

adityadani: If the config map does not exists, the pod creation will fail anyways as we cannot determine the node specific specs.

harsh-px: Won't users create the config map with the same name?

adityadani: Yes they will have to. This function creates a config map. If it fails to do so we don't error out. 
The actual API that reads the config map returns an error if it can't find a config map.

PR #53, opened 2019-09-04
OPERATOR-76: Fixed creating additional storages when increase amount …

trierra:
…of nodes it cluster


pkg/cloudstorage/manager.go
```diff
@@ -17,7 +17,7 @@ type CloudDriveConfig struct {
 }
 
 // Config is the cloud storage configuration for a node
-// TODO: This struct should eventually move under StorageNodeStatus
+// TODO: This struct should eventually move under StorageNode
```
trierra: Just noticed this. Looks like it's a good time to move

deploy/operator.yaml
```diff
@@ -27,7 +27,7 @@ spec:
         - --driver=portworx
         - --leader-elect=true
         imagePullPolicy: Always
-        image: openstorage/openstorage-operator:master
+        image: trierra/operator:dev
```
adityadani: same here

deploy/crds/core_v1alpha1_storagecluster_cr.yaml
```diff
@@ -4,9 +4,9 @@ metadata:
   name: portworx
   namespace: kube-system
   annotations:
-    portworx.io/is-openshift: "true"
+    portworx.io/is-aks: "true"
```
adityadani: Is this your local change? We want this to be is-openshift.

drivers/storage/portworx/deployment.go
```diff
@@ -318,11 +320,46 @@ func (p *portworx) GetStoragePodSpec(
...
+				logrus.Error(err)
+				return v1.PodSpec{}, err
+			}
+
+			nodes, err := p.getNodesList()
+			if err != nil {
+				return v1.PodSpec{}, err
+			}
+
+			for _, node := range nodes.Items {
```
adityadani: We should generate the StorageNode object only for one node here and not for all the nodes. The list returned by getNodesList contains all the nodes in the kubernetes cluster, but that does not mean storage provider (or Portworx) would be running on all of them. 

May be pass in the node name in GetStoragePodSpec() and only create StorageNode object for that node.

drivers/storage/portworx/deployment.go
```diff
@@ -318,11 +320,46 @@ func (p *portworx) GetStoragePodSpec(
...
+				configureStorageNodeSpec(storageNode, cloudConfig)
+				err := k8s2.CreateOrUpdateStorageNode(p.k8sClient, storageNode, ownerRef)
+				if err != nil {
+					msg := fmt.Sprintf("Failed to create node for nodeID %v: %v", node.Name, err)
+					p.warningEvent(cluster, failedSyncReason, msg)
+				}
+			}
+		} else {
+			// Cloud Environment
+			cloudConfig = p.storageNodeToCloudSpec(&storageNodes.Items[0])
```
adityadani: Create the StorageNode object for the new node here.

pkg/apis/core/v1alpha1/storagenode.go
```diff
@@ -37,6 +37,22 @@ type StorageNodeList struct {
 type StorageNodeSpec struct {
 	// Version of the storage driver on the node
 	Version string `json:"version,omitempty"`
+	// CloudConfig used to create pods
+	CloudConfig []StorageNodeCloudDriveConfig
+
```
adityadani: nit: comment for StorageInstancesPerZone

pkg/cloudstorage/manager.go
adityadani: Yes if this gets moved, then you don't need to do the conversion from StorageNodeCloudDriveConfig and CloudConfig structs

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -429,31 +430,24 @@ func (c *Controller) manage(
 	)
 	zoneMap := make(map[string]int)
 
-	// Get the cloud provider
+	cloudProvider := cloudprovider.New(cloudProviderName)
```
adityadani: cloudProviderName here is nil or empty.
The cloudProviderName gets populated in the next for loop. Once that gets generated then call 
cloudprovider.New(cloudProviderName)

drivers/storage/portworx/deployment.go
```diff
@@ -2,12 +2,15 @@ package portworx
 
 import (
 	"fmt"
+	k8s2 "github.com/libopenstorage/operator/pkg/util/k8s"
```
piyush-nimbalkar: nit: Let's call this `k8sutil`

drivers/storage/portworx/deployment.go
```diff
@@ -359,6 +396,32 @@ func (p *portworx) GetStoragePodSpec(
...
+			Options:   conf.Options,
+		}
+		node.Spec.CloudConfig = append(node.Spec.CloudConfig, sc)
+	}
+	node.Spec.StorageInstancesPerZone = config.StorageInstancesPerZone
+}
+
+func (p *portworx) getStorageNodes() (*corev1alpha1.StorageNodeList, error) {
+	nodeList := &corev1alpha1.StorageNodeList{}
+	err := p.k8sClient.List(context.TODO(), &client.ListOptions{}, nodeList)
```
piyush-nimbalkar: In the list options we should filter the storage nodes by the cluster's namespace and label selector from `p.GetSelectorLabels()`. Also return only those StorageNodes that are owned by the current StorageCluster..

drivers/storage/portworx/deployment.go
```diff
@@ -318,11 +320,46 @@ func (p *portworx) GetStoragePodSpec(
...
+
+				configureStorageNodeSpec(storageNode, cloudConfig)
+				err := k8s2.CreateOrUpdateStorageNode(p.k8sClient, storageNode, ownerRef)
+				if err != nil {
+					msg := fmt.Sprintf("Failed to create node for nodeID %v: %v", node.Name, err)
+					p.warningEvent(cluster, failedSyncReason, msg)
+				}
+			}
+		} else {
+			// Cloud Environment
```
piyush-nimbalkar: nit: This comments needs to be at the start of the outer if condition where it was before.

drivers/storage/portworx/deployment.go
```diff
@@ -318,11 +320,46 @@ func (p *portworx) GetStoragePodSpec(
...
 
 	if cluster.Spec.CloudStorage != nil && len(cluster.Spec.CloudStorage.CapacitySpecs) > 0 {
-		// Cloud Environment
-		cloudConfig, err := p.generateCloudStorageSpecs(cluster)
+		var cloudConfig *cloudstorage.Config
+		storageNodes, err := p.getStorageNodes()
 		if err != nil {
 			return v1.PodSpec{}, err
 		}
+		if len(storageNodes.Items) == 0 {
```
piyush-nimbalkar: Can you add unit tests for this code?

pkg/cloudstorage/manager.go
trierra: it causes import loop. Will check if I can fix it quickly

deploy/operator.yaml
trierra: yes, accidentally committed, thanks for noticing

drivers/storage/portworx/deployment.go
trierra: I thought the idea was to keep a separate spec for each storage node, wasn't it?

drivers/storage/portworx/deployment.go
trierra: need more details, looks like I didn't get your idea

drivers/storage/portworx/deployment.go
trierra: sure

pkg/apis/core/v1alpha1/storagenode.go
```diff
@@ -37,6 +37,22 @@ type StorageNodeList struct {
...
+	StorageInstancesPerZone int `json:"storageInstancesPerZone,omitempty"`
```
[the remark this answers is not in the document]
trierra: @adityadani @piyush-nimbalkar let's discuss this quickly today

drivers/storage/portworx/deployment.go
adityadani: Yes we want to create a StorageNode object only for those nodes where Portworx will be running. Lets say a user has provided node affinity rules where in Portworx needs to be installed on node1, node2, node3 and skip node4, node5.

The controller makes sure that Portworx pod does not start on node4 and node5. But this for loop will create 5 StorageNode objects. 

Which is why I am suggesting pass in the NodeName in GetStoragePodSpec and only create the StorageNode object for that node.

trierra:
done

drivers/storage/portworx/deployment.go
```diff
@@ -289,48 +302,65 @@ func (p *portworx) generateCloudStorageSpecs(
 		metav1.NewControllerRef(cluster, controllerKind),
 	}
 
-	if err := cloudStorageManager.CreateStorageDistributionMatrix(); err != nil {
-		logrus.Warnf("Failed to generate storage distribution matrix config map: %v", err)
-	}
+	if len(nodes) == 0 {
```
piyush-nimbalkar: For safety, should we check if at least one the node has the drives?

drivers/storage/portworx/deployment_test.go
```diff
@@ -702,7 +713,7 @@ func TestPodSpecWithCloudStorageSpec(t *testing.T) {
 		"-max_storage_nodes_per_zone", "2",
 	}
 
-	actual, err = driver.GetStoragePodSpec(cluster)
+	actual, err = driver.GetStoragePodSpec(cluster, nodeName)
```
piyush-nimbalkar: After this we should verify that a StorageNode object is created for `nodeName` node. Also that it has the expected cloud config.

drivers/storage/portworx/deployment.go
```diff
@@ -359,6 +389,60 @@ func (p *portworx) GetStoragePodSpec(
...
+			OwnerReferences: []metav1.OwnerReference{*ownerRef},
+			Labels:          p.GetSelectorLabels(),
+		},
+		Status: corev1alpha1.NodeStatus{},
+	}
+
+	configureStorageNodeSpec(storageNode, cloudConfig)
+	cluster.Status.StorageInstancesPerZone = cloudConfig.StorageInstancesPerZone
+
+	return k8s2.CreateOrUpdateStorageNode(p.k8sClient, storageNode, ownerRef)
```
piyush-nimbalkar: If we only want to create StorageNode here, we should just call `p.k8sClient.Create()` to avoid unwanted updates -
```
err := p.k8sClient(context.TODO(), storageNode)
if err != nil && !errors.IsAlreadyExists(err) {
  return err
}
return nil
```

drivers/storage/portworx/deployment_test.go
```diff
@@ -771,7 +746,7 @@ func TestPodSpecWithCloudStorageSpec(t *testing.T) {
 		"-max_storage_nodes_per_zone", "3",
 	}
 
-	actual, err = driver.GetStoragePodSpec(cluster)
+	actual, err = driver.GetStoragePodSpec(cluster, "nodeName1")
```
piyush-nimbalkar: Verify here that a new storage node object is created for `nodeName1`.

piyush-nimbalkar: Can we separate this out in multiple tests - 
- without any user provided `maxStorageNodesPerZone`, use the same node name, and the instances per zone increase, the `-max_storage_nodes_per_zone` should be still 2 because we have a storage node now from the previous run.
- without any user provided `maxStorageNodesPerZone`, use a new node name, and the increased instances per zone.. should have the same output
- user provided `maxStorageNodesPerZone`.. now the `-max_storage_nodes_per_zone` should change to what the user has passed.

trierra:
I will never complete this PR =)

drivers/storage/portworx/deployment.go
trierra: what kind of unwanted updates? this function is called only if 
`!storageNodeExists(nodeName, nodes) `

trierra: or you want to split `CreateOrUpdateStorageNode` into 2 functions?

piyush-nimbalkar: No need to split, you just call `p.k8sClient(context.TODO(), storageNode)`, which is what `CreateOrUpdateStorageNode` does internally. `CreateOrUpdateStorageNode` also updates the object if it is present, which is not our desired behavior here.

drivers/storage/portworx/deployment_test.go
trierra: how?

piyush-nimbalkar: Just do a get on the fake client and we should find it there because our code creates it.

drivers/storage/portworx/deployment.go
piyush-nimbalkar: That's why just call create, no need to call CreateOrUpdate if you are not planning to Update anyway.

drivers/storage/portworx/deployment_test.go
trierra: added a separate test for this case

drivers/storage/portworx/deployment_test.go
```diff
@@ -760,54 +776,41 @@ func TestPodSpecWithCloudStorageSpec(t *testing.T) {
...
+	// despite adding a new node, node spec should be pulled from the existing storage node for nodeName
+	// Storage node will be created for "nodeName1"
+	actual, err = driver.GetStoragePodSpec(cluster, nodeName)
 	assert.NoError(t, err, "Unexpected error on GetStoragePodSpec")
 	assert.ElementsMatch(t, expectedArgs, actual.Containers[0].Args)
 
+	list, err = driver.storageNodesList(cluster)
+	assert.NoError(t, err, "Unexpected error on GetStorageNodeList")
+	assert.Equal(t, 1, len(list), "expected storage nodes in list")
+	assert.ElementsMatch(t, list[0].Spec.CloudStorage.DriveConfigs, expectedCloudStorageSpec)
```
piyush-nimbalkar: This should be checked in the first case we well on line 756.

drivers/storage/portworx/deployment_test.go
```diff
@@ -1904,12 +1966,238 @@ func TestPodSpecForKvdbAuthErrorReadingSecret(t *testing.T) {
...
+					IOPS:             uint32(210),
+				},
+			},
+		}, nil)
+
+	actual, err = driver.GetStoragePodSpec(cluster, nodeName)
+	assert.NoError(t, err, "Unexpected error on GetStoragePodSpec")
+	assert.ElementsMatch(t, expectedArgs, actual.Containers[0].Args)
+
+	list, err = driver.storageNodesList(cluster)
```
piyush-nimbalkar: Let's also verify that the cluster.Status.Storage.StorageInstancesPerZone is also set to the right value.

drivers/storage/portworx/deployment_test.go
```diff
@@ -1904,12 +1966,238 @@ func TestPodSpecForKvdbAuthErrorReadingSecret(t *testing.T) {
...
+		"-c", "px-cluster",
+		"-x", "kubernetes",
+		"-s", "type=foo,size=120,iops=110,foo1=bar1",
+		"-s", "type=bar,size=220,iops=210,foo3=bar3",
+		"-max_storage_nodes_per_zone", "2",
+	}
+
+	// despite adding a new node, node spec should be pulled from the existing storage node for nodeName
+	// Storage node will be created for "nodeName1"
+	actual, err = driver.GetStoragePodSpec(cluster, nodeName2)
```
piyush-nimbalkar: Here we should also increase the instancesPerZone map in the driver. Basically add more instances there. Just to make sure that it still returns the same output.

drivers/storage/portworx/deployment.go
```diff
@@ -351,6 +378,79 @@ func (p *portworx) GetStoragePodSpec(
...
+		nodes)
+
+	if err != nil {
+		return storageNodes, err
+	}
+
+	for _, node := range nodes.Items {
+		ownerRef := metav1.NewControllerRef(cluster, controllerKind)
+		if ownerRef != nil {
+			for _, owner := range node.OwnerReferences {
```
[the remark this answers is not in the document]
trierra: That code didn't work for me for some reason, and it compared owner reference, which was created from cluster, to cluster.UID. I didn't get your idea how to check the node's owner ref

drivers/storage/portworx/deployment.go
```diff
@@ -351,6 +378,79 @@ func (p *portworx) GetStoragePodSpec(
...
+			Name:            nodeName,
+			Namespace:       cluster.Namespace,
+			OwnerReferences: []metav1.OwnerReference{*ownerRef},
+			Labels:          p.GetSelectorLabels(),
+		},
+		Status: corev1alpha1.NodeStatus{},
+	}
+
+	configureStorageNodeSpec(storageNode, cloudConfig)
+	cluster.Status.Storage.StorageInstancesPerZone = cloudConfig.StorageInstancesPerZone
```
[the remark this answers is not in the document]
trierra: that's a good point

drivers/storage/portworx/deployment_test.go
trierra: right

pkg/apis/core/v1alpha1/storagecluster.go
```diff
@@ -323,6 +323,14 @@ type StorageClusterStatus struct {
...
 	// Conditions describes the current conditions of the cluster
 	Conditions []ClusterCondition `json:"conditions,omitempty"`
+	// Storage represents cluster storage details
+	Storage Storage
+}
+
+// Storage represents cluster storage details
+type Storage struct {
+	// StorageInstancesPerZone describes the amount of instances per zone
+	StorageInstancesPerZone int `json:"storageInstancesPerZone,omitempty"`
```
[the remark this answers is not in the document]
trierra: requires too much changes

drivers/storage/portworx/deployment_test.go
trierra: As per my understanding, instancesPerZone should not be modified after a cluster has been initialized

pkg/apis/core/v1alpha1/storagecluster.go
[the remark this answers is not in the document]
piyush-nimbalkar: We should not keep it open-ended from our side. It should be a one time change wherever you do. Also for consistency can we rename at least the json tag to `storageNodesPerZone` because we have fields like `maxStorageNodesPerZone` and `maxStorageNodes`.

drivers/storage/portworx/deployment_test.go
piyush-nimbalkar: It keeps getting updated periodically. If more instances are added, it will have the latest `instancesPerZone counts`.

drivers/storage/portworx/deployment_test.go
```diff
@@ -1904,12 +1969,240 @@ func TestPodSpecForKvdbAuthErrorReadingSecret(t *testing.T) {
...
+					InstancesPerZone: inputInstancesPerZone,
+					IOPS:             uint32(210),
+				},
+			},
+		}, nil)
+
+	actual, err = driver.GetStoragePodSpec(cluster, nodeName)
+	assert.NoError(t, err, "Unexpected error on GetStoragePodSpec")
+	assert.ElementsMatch(t, expectedArgs, actual.Containers[0].Args)
+	assert.Equal(t, inputInstancesPerZone, cluster.Status.Storage.StorageInstancesPerZone, "wrong number of instances per zone in a the cluster status")
```
piyush-nimbalkar: We should do this for all below test cases, to ensure that the count hasn't changed in the status after different runs.

pkg/apis/core/v1alpha1/storagecluster.go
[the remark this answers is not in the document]
trierra: is there a difference between StorageInstancesPerZone and maxStorageNodesPerZone ?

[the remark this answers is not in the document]
trierra: if I change int to int32 here, I have to modify `	InstancesPerZone int `json:"instances_per_zone" yaml:"instances_per_zone"`
` in `vendor/github.com/libopenstorage/cloudops/cloud_storage_management.go` as well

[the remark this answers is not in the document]
trierra: I don't have to, nvm

PR #56, opened 2019-09-11
Exposing default operator metrics

piyush-nimbalkar:
- Added a new `stable` channel to portworx operator.
- Added `OPERATOR_NAME` and `POD_NAME` env vars to operator deployment as it is needed to deploy the service that exposes metrics from the operator. This is what operator-sdk expects.
- Vendored modified sched-ops without stork and autopilot clients. Stork and autopilot use operator-sdk:0.0.7 while for metrics we needed 0.2.0 and above.


cmd/operator.go
```diff
@@ -12,17 +14,27 @@ import (
...
-	defaultLockObjectNamespace = "kube-system"
-	defaultResyncPeriod        = 30 * time.Second
+	flagLeaderElect              = "leader-elect"
+	flagLeaderElectLockName      = "leader-elect-lock-name"
+	flagLeaderElectLockNamespace = "leader-elect-lock-namespace"
+	flagMetricsPort              = "metrics-port"
+	defaultLockObjectName        = "openstorage-operator"
+	defaultLockObjectNamespace   = "kube-system"
+	defaultResyncPeriod          = 30 * time.Second
+	defaultMetricsPort           = 8999
```
adityadani: Is this to align with 9001 port range ?

piyush-nimbalkar: Yes. But it doesn't really matter. It is used for the operator service. Also it can be changed via an argument.

PR #58, opened 2019-09-17
Update Portworx CSI specs

piyush-nimbalkar:
- Based on portworx version deploy the correct CSI sidecars
- Use deployment for CSI 1.0 and above
- Support upgrading from CSI stateful set to deployment


drivers/storage/portworx/deployment.go
```diff
@@ -877,6 +869,45 @@ func (t *template) loadKvdbAuth() map[string]string {
...
+	}
+
+	parts := strings.Split(pxImage, ":")
+	if len(parts) >= 2 {
+		pxVersionStr := parts[len(parts)-1]
+		pxVersion, err = version.NewSemver(pxVersionStr)
+		if err != nil {
+			logrus.Warnf("Invalid PX version %s extracted from image name: %v", pxVersionStr, err)
+			if pxVersionStr, exists := cluster.Annotations[annotationPXVersion]; exists {
+				pxVersion, err = version.NewSemver(pxVersionStr)
```
ggriffiths: Nice, this is much better than the installer version. We need to use the oci-monitor there and the px version is inferred from that, which isn't always accurate.

drivers/storage/portworx/csi.go
```diff
@@ -44,62 +66,109 @@ func (g *csiGenerator) getSidecarContainerVersions() (csiVersions, error) {
...
 	// Setup registration correctly
-	if g.kubeVer.LessThan(k8sVer1_13) {
+	if g.kubeVersion.LessThan(k8sVer1_13) {
 		cv.useOlderPluginsDirAsRegistration = true
 	}
 
+	// Set the CSI driver name
+	// - PX Versions <2.2.0 will always use the deprecated CSI Driver Name.
+	// - PX Versions >=2.2.0 will default to the new name
+	if g.pxVersion.LessThan(pxVer2_2) {
```
ggriffiths: We also need to check if the UseDeprecatedName flag is set:
https://github.com/portworx/px-installer/blob/master/px-spec-websvc/csi.go#L127

ggriffiths: This flag originates from the PX Installer page. If they select "Use Deprecated CSI Driver Name" (under Enable CSI), we alter the spec to use the deprecated name, as well as pass the UseDeprecatedName flag to PX. This will tell PX to use the old name.

piyush-nimbalkar:
- Fixed the deprecated driver name
- Will make the change to pass the env from installer if the `useDeprecated` flag in set in installer
- Also added the missing CSINodeInfo CRD

TODO: Need to create the CSIDriver object. The object is present in k8s version 1.14+, so will have to update the vendor to 1.14. Will do this as a separate change.

drivers/storage/portworx/testspec/csiDeploymentWithResizer_1.0.yaml
```diff
@@ -0,0 +1,96 @@
+kind: Deployment
```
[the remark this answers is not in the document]
piyush-nimbalkar: Yes these specs are from the installer output. I think going forward I should put the url used in the comment in this spec.

drivers/storage/portworx/csi.go
```diff
@@ -23,17 +30,38 @@ type csiVersions struct {
...
+	// driverName is decided based on the PX Version unless the deprecated flag is provided.
+	driverName string
+	// useDeployment decides whether to use a StatefulSet or Deployment for the CSI side cars.
+	useDeployment bool
+	// includeAttacher dictates whether we include the csi-attacher sidecar or not.
+	includeAttacher bool
+	// includeResizer dicates whether or not to include the resizer sidecar.
+	includeResizer bool
+	// includeCsiDriverInfo dictates whether or not to add the CsiDriverInfo object.
+	includeCsiDriverInfo bool
```
ggriffiths: It doesn't look like this is used.

If this is true, we must created a CSIDriver object. This happens in k8s 1.14+

drivers/storage/portworx/components.go
```diff
@@ -344,6 +366,107 @@ func createVolumePlacementStrategyCRD() error {
 	return k8s.Instance().ValidateCRD(resource, 1*time.Minute, 5*time.Second)
 }
 
+func createCSINodeInfoCRD() error {
```
ggriffiths: Currently we don't manually create a CSINodeInfo CRD. Our CSINodeInfo object is created by the Node Driver Registrar:

https://kubernetes-csi.github.io/docs/node-driver-registrar.html

This sidecar is in the Portworx daemonset and continually registers the CSINodeInfo depending on the status of that PX node.

PR #62, opened 2019-09-27
Use release manifest to group component versions

piyush-nimbalkar:
- Introduce a hardcoded portworx release manifest with related component versions to go with different portworx versions. In future we will have an online manifest that we can pull before using the hardcoded one. Will make Manifest an interface then.
- By default if portworx image version is changed, other component versions also change based on the release manifest. 
- Introduced `LockImage` boolean in all component specs to disallow default image override from manifest. 
- Release manifest will look something like this - 
```
defaultRelease: 2.1.3
releases:
  2.1.2:
    stork: openstorage/stork:2.2.5
    lighthouse: portworx/px-lighthouse:2.0.4
  2.1.3:
    stork: openstorage/stork:2.2.5
    lighthouse: portworx/px-lighthouse:2.0.4
```
- The portworx version is determined from the tag of `spec.image`. If `PX_IMAGE` env is present then it's tag take precedence over `spec.image`. For dev/testing, if custom tags are used in `spec.image` and/or `PX_IMAGE`, then we look at the annotation for version `portworx.io/px-version` which directly can have version (eg. 2.1.5) as it's value.
- If no matching tag is found in the release manifest, we pick up the default version from the release manifest to deploy.

### DIFF
--- a/drivers/storage/portworx/manifest/manifest.go
+++ b/drivers/storage/portworx/manifest/manifest.go
@@ -1,0 +1,136 @@
+package manifest
+
+import (
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"path"
+	"sort"
+
+	version "github.com/hashicorp/go-version"
+	yaml "gopkg.in/yaml.v2"
+)
+
+const (
+	// ManifestDir directory where release manifest is stored
+	ManifestDir = "manifests"
+	// ReleasesFilename name of the release manifest file
+	ReleasesFilename = "portworx-releases.yaml"
+)
+
+var (
+	// ErrEmptyReleases when releases list is empty
+	ErrEmptyReleases = errors.New("releases cannot be empty")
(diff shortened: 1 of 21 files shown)


drivers/storage/portworx/manifest/manifest.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+// Get returns a release object for given string version
+func (m *ReleaseManifest) Get(target string) (Release, error) {
+	if release, exists := m.Releases[target]; exists {
+		return release, nil
+	}
+	targetVersion, err := version.NewSemver(target)
+	if err != nil {
+		return Release{}, fmt.Errorf("invalid version %s: %v", target, err)
```
disrani-px: You can return `*Release` from the function, so that you can return `nil, err` instead of an empty Release struct

PR #64, opened 2019-10-08
Create ServiceMonitor and PrometheusRule for external prometheus

piyush-nimbalkar:
- Operator should create service monitor and prometheus rule for external prometheus by changing the StorageCluster spec
```
spec:
  monitoring:
    enableMetrics: true
```
- The specs get deployed in the same namespace as portworx / storagecluster object
- Do not fail if unable to deploy specs because they are not registered, just log and create an event so users know what is wrong.
- Instead of using sched-ops SetClient method, using the individual setters to avoid changing the tests every time SetClient's definition changes.


drivers/storage/portworx/components.go
```diff
@@ -206,6 +221,25 @@ func (p *portworx) setupLighthouse(t *template) error {
 	return nil
 }
 
+func (p *portworx) setupMonitoring(t *template) error {
+	ownerRef := metav1.NewControllerRef(t.cluster, controllerKind)
+	if err := p.createServiceMonitor(t.cluster, ownerRef); metaerrors.IsNoMatchError(err) {
+		p.warningEvent(t.cluster, failedComponentReason,
+			fmt.Sprintf("Failed to create ServiceMonitor for Portworx. Ensure Prometheus is deployed correctly. %v", err))
```
pault84: Do we detect if they have prometheus/operator installed at all? this would only fail if the CRD is not registered which means the operator isn't installed. Prometheus itself has very little to do with the service monitor not deploying. If we can detect that they have an operator installed at all we could maybe enhance the ui around enabling monitoring.

drivers/storage/portworx/components.go
```diff
@@ -206,6 +221,25 @@ func (p *portworx) setupLighthouse(t *template) error {
...
+		return nil
+	} else if err != nil {
+		return err
+	}
+	if err := p.createPrometheusRule(t.cluster, ownerRef); metaerrors.IsNoMatchError(err) {
+		p.warningEvent(t.cluster, failedComponentReason,
+			fmt.Sprintf("Failed to create PrometheusRule for Portworx. Ensure Prometheus is deployed correctly. %v", err))
```
pault84: Ditto to the above.

drivers/storage/portworx/components.go
```diff
@@ -2012,6 +2059,184 @@ func (p *portworx) createPortworxAPIDaemonSet(
...
+			},
+			OwnerReferences: []metav1.OwnerReference{*ownerRef},
+		},
+		Spec: monitoringv1.ServiceMonitorSpec{
+			Endpoints: []monitoringv1.Endpoint{
+				{
+					Port: pxRESTPortName,
+				},
+				{
+					Port: pxKVDBPortName,
```
pault84: We should only add the Kvdb endpoint if we are using internal ETCD.
I don't know how if this is always the case. But if they can use an external etcd these will show up as "n"  failing endpoints on the ui.

drivers/storage/portworx/components.go
piyush-nimbalkar: This will check if there is a kind for prometheus objects. If the object is not found, then we just raise a Warning event and don't fail the operator reconcile loop. We don't need to worry if prometheus operator is running as long as the CRD is present.

drivers/storage/portworx/components.go
piyush-nimbalkar: Done

pault84: Looks good

drivers/storage/portworx/components.go
```diff
@@ -2012,6 +2059,189 @@ func (p *portworx) createPortworxAPIDaemonSet(
...
+		svcMonitor.Spec.Endpoints = append(
+			svcMonitor.Spec.Endpoints,
+			monitoringv1.Endpoint{Port: pxKVDBPortName},
+		)
+	}
+
+	return k8sutil.CreateOrUpdateServiceMonitor(p.k8sClient, svcMonitor, ownerRef)
+}
+
+func (p *portworx) createPrometheusRule(
```
adityadani: wouldn't this be very difficult to manage. Whenever we want to add new rules, we will end up modifying at multiple places - here in operator, installer spec (for daemonset) and our docs.

Instead can we just maintain one spec at install.portworx.com. Operator tries to fetch this from yaml from install.portworx.com. If it cannot reach then it will have one spec baked in the container for each version.

drivers/storage/portworx/testspec/prometheusRule.yaml
```diff
@@ -0,0 +1,109 @@
...
+        summary: Portworx volume {{$labels.volumeid}} on {{$labels.host}} is at {{$value}}% used.
+      expr: 100 * (px_volume_usage_bytes / px_volume_capacity_bytes) > 70
+      for: 5m
+      labels:
+        issue: Portworx volume {{$labels.volumeid}} usage on {{$labels.host}} is critical.
+        severity: warning
+    - alert: PortworxVolumeWillFill
+      annotations:
+        description: Disk volume {{$labels.volumeid}} on {{$labels.host}} is over 70% full and has
+          been predicted to fill within 2 weeks for more than 10 minutes.
```
adityadani: You don't need for more than 10 minutes here.

drivers/storage/portworx/testspec/prometheusRule.yaml
```diff
@@ -0,0 +1,109 @@
...
+          more than 10 minutes.
+        summary: Portworx volume {{$labels.volumeid}} on {{$labels.host}} is at {{$value}}% used.
+      expr: 100 * (px_volume_usage_bytes / px_volume_capacity_bytes) > 70
+      for: 5m
+      labels:
+        issue: Portworx volume {{$labels.volumeid}} usage on {{$labels.host}} is critical.
+        severity: warning
+    - alert: PortworxVolumeWillFill
+      annotations:
+        description: Disk volume {{$labels.volumeid}} on {{$labels.host}} is over 70% full and has
```
adityadani: nit: Portworx volume

drivers/storage/portworx/testspec/prometheusRule.yaml
```diff
@@ -0,0 +1,109 @@
...
+          been predicted to fill within 2 weeks for more than 10 minutes.
+        summary: Portworx volume {{$labels.volumeid}} on {{$labels.host}} is over 70% full and is
+          predicted to fill within 2 weeks.
+      expr: (px_volume_usage_bytes / px_volume_capacity_bytes) > 0.7 and predict_linear(px_cluster_disk_available_bytes[1h],
+        14 * 86400) < 0
+      for: 10m
+      labels:
+        issue: Disk volume {{$labels.volumeid}} on {{$labels.host}} is predicted to fill within
```
adityadani: Portworx volume

drivers/storage/portworx/testspec/prometheusRule.yaml
```diff
@@ -0,0 +1,109 @@
...
+          2 weeks.
+        severity: warning
+    - alert: PortworxStorageUsageCritical
+      annotations:
+        description: Portworx storage {{$labels.volumeid}} on {{$labels.host}} is over 80% used
+          for more than 10 minutes.
+        summary: Portworx volume {{$labels.volumeid}} on {{$labels.host}} is at {{$value}}% used.
```
adityadani: nit: Portworx storage

adityadani: same at other places in this yaml. Lets use volumes for volume alert, disk/storage for storage alerts.

drivers/storage/portworx/components.go
pault84: That's fine we can also just add additional rule objects.

PR #68, opened 2019-10-29
Use default storage if nothing present

piyush-nimbalkar:
- Add status for deleting stages
- Allow all installation modes for the operator to support previous releases
- Remove the defaults from the example spec to keep the StorageCluster spec simple.

### DIFF
--- a/pkg/controller/storagecluster/storagecluster.go
+++ b/pkg/controller/storagecluster/storagecluster.go
@@ -403,13 +403,10 @@ func (c *Controller) deleteStorageCluster(
 
 	if deleteFinalizerExists(cluster) {
 		toDelete := cluster.DeepCopy()
-		deleteClusterCondition, err := c.Driver.DeleteStorage(toDelete)
-		if err != nil {
-			logrus.Errorf("Failed to delete storage: %v", err)
-			return fmt.Errorf("driver failed to delete storage: %v", err)
-		} else if deleteClusterCondition == nil {
-			// Return without error as the delete condition is not yet populated
-			return nil
+		deleteClusterCondition, driverErr := c.Driver.DeleteStorage(toDelete)
+		if driverErr != nil {
+			msg := fmt.Sprintf("Driver failed to delete storage. %v", driverErr)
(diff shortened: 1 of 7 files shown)


pkg/controller/storagecluster/storagecluster.go
```diff
@@ -420,17 +417,26 @@ func (c *Controller) deleteStorageCluster(
...
+		if err := k8sutil.UpdateStorageClusterStatus(c.client, toDelete); err != nil && !errors.IsNotFound(err) {
 			return fmt.Errorf("error updating delete status for StorageCluster %v/%v: %v",
 				toDelete.Namespace, toDelete.Name, err)
 		}
 
-		if deleteClusterCondition.Status == corev1alpha1.ClusterOperationCompleted {
+		if toDelete.Status.Conditions[foundIndex].Status == corev1alpha1.ClusterOperationCompleted {
 			newFinalizers := removeDeleteFinalizer(toDelete.Finalizers)
 			toDelete.Finalizers = newFinalizers
 			if err := c.client.Update(context.TODO(), toDelete); err != nil {
```
adityadani: check here for errors.IsNotFound() ?

pkg/util/k8s/k8s_test.go
```diff
@@ -901,6 +902,38 @@ func TestDeleteStatefulSet(t *testing.T) {
...
+	}
+
+	// Fail if cluster is not present
+	err := UpdateStorageClusterStatus(k8sClient, cluster)
+	require.True(t, errors.IsNotFound(err))
+
+	err = k8sClient.Create(context.TODO(), cluster)
+	require.NoError(t, err)
+
+	// Should update cluster if present with same resource version
```
adityadani: UpdateStorageClusterStatus is not checking for the same resource version, but simply overwriting the input with the existing resource version. 

May be you can add a test that actually updates the Status field of the StorageCluster and check if it got updated.

piyush-nimbalkar: Fixed the test to send input object with old resource version and verify that it gets updated to whatever is the latest in the api server.

PR #69, opened 2019-10-31
Add autopilot support

piyush-nimbalkar:


drivers/storage/portworx/components.go
```diff
@@ -714,6 +779,52 @@ func (p *portworx) createCSIDriver(
...
+			params += fmt.Sprintf("%s=%s,", k, provider.Params[k])
+		}
+		params = strings.TrimRight(params, ",")
+		config += fmt.Sprintf(`
+- name: %s
+  type: %s
+  params: %s`,
+			provider.Name, provider.Type, params)
+	}
+	for key, value := range autopilotConfigParams {
```
adityadani: You can get rid of this.

drivers/storage/portworx/components.go
```diff
@@ -714,6 +779,52 @@ func (p *portworx) createCSIDriver(
...
+	t *template,
+	ownerRef *metav1.OwnerReference,
+) error {
+	config := "providers:"
+	for _, provider := range t.cluster.Spec.Autopilot.Providers {
+		keys := make([]string, 0, len(provider.Params))
+		for k := range provider.Params {
+			keys = append(keys, k)
+		}
+		sort.Strings(keys)
```
adityadani: why sort ?

drivers/storage/portworx/components.go
```diff
@@ -93,9 +101,12 @@ const (
 )
 
 var (
-	kbVerRegex     = regexp.MustCompile(`^(v\d+\.\d+\.\d+).*`)
-	controllerKind = corev1alpha1.SchemeGroupVersion.WithKind("StorageCluster")
-	specsBaseDir   = getSpecsBaseDir
+	kbVerRegex            = regexp.MustCompile(`^(v\d+\.\d+\.\d+).*`)
+	controllerKind        = corev1alpha1.SchemeGroupVersion.WithKind("StorageCluster")
+	specsBaseDir          = getSpecsBaseDir
+	autopilotConfigParams = map[string]string{
```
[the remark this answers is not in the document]
piyush-nimbalkar: Keeping this helps people overwrite the value using args section in autopilot. If we remove this then it will simply be passed to autopilot as an arg (if someone sets it in the storage cluster spec). 

I will change the code to just use it for finding if any config param is present in the autopilot args. Will not pass in the default config.

drivers/storage/portworx/components.go
piyush-nimbalkar: Helps in comparing the params and see if anything has changed.

PR #70, opened 2019-11-07
Get release manifest from installer if available else use embedded manifest

piyush-nimbalkar:


drivers/storage/portworx/portworx.go
```diff
@@ -615,6 +626,32 @@ func isTLSEnabled() bool {
 	return err == nil && enabled
 }
 
+func componentVersions(releases *manifest.ReleaseManifest, pxVersion *version.Version) (manifest.Release, error) {
+	if releases == nil {
+		return manifest.Release{}, fmt.Errorf("release manifest is empty")
+	}
+	components, err := releases.GetFromVersion(pxVersion)
```
adityadani: pxVersion is being extracted from PX_IMAGE env variable as well. So in that case, if I use `px-enterprise:dev`, there will be no release manifest associated with it?

Also I think default should be given higher priority over edge portworx. May be if an annotation is set for edge then we should pick up edge version.

piyush-nimbalkar: Edge is given higher priority for internal builds. https://install.portworx.com/versions will not have edge release target. In that case we will fallback to the default release version. 
Edge will be present only for unreleased branches. This will ensure that portworx internal builds of form sha_sha will use edge release target for components.
Also for dev testing, you can always use `portworx.io/px-version` for explicit version override.

PR #74, opened 2019-12-10
Heterogenous storage configuration using node specs

piyush-nimbalkar:
- If the selector in node specs matches a k8s node, use the storage
   configuration from the matching node spec.
- If a partial storage config is used in node spec, populate the rest
   of the fields from cluster storage configuration.

A sample list of node specs will look something like below:
```
spec:
  storage:
    devices:
    - /dev/sda
    - /dev/sdb

  nodes:
  - selector:
      nodeName: k8s-node-1
    storage:
      useAll: true

  - selector:
      labelSelector:
        matchExpressions:
        - key: drive-type
          operator: In
          values:
          - ssd
          - hdd
    storage:
      devices:
      - /dev/xvdf

  - selector:
      labelSelector:
        matchLabels:
          storageless: "true"
    storage:
      devices: {}
```

### DIFF
--- a/drivers/storage/portworx/portworx.go
+++ b/drivers/storage/portworx/portworx.go
@@ -665,18 +665,29 @@ func setNodeSpecDefaults(toUpdate *corev1alpha1.StorageCluster) {
 		if nodeSpec.Storage == nil {
 			nodeSpecCopy.Storage = toUpdate.Spec.Storage.DeepCopy()
 		} else if toUpdate.Spec.Storage != nil {
-			if nodeSpecCopy.Storage.UseAll == nil && toUpdate.Spec.Storage.UseAll != nil {
-				nodeSpecCopy.Storage.UseAll = boolPtr(*toUpdate.Spec.Storage.UseAll)
+			// Devices, UseAll and UseAllWithPartitions should be set exclusive of each other, if not already
+			// set by the user in the node spec.
+			if nodeSpecCopy.Storage.Devices == nil &&
+				(nodeSpecCopy.Storage.UseAll == nil || !*nodeSpecCopy.Storage.UseAll) &&
+				(nodeSpecCopy.Storage.UseAllWithPartitions == nil || !*nodeSpecCopy.Storage.UseAllWithPartitions) &&
+				toUpdate.Spec.Storage.Devices != nil {
+				devices := append(make([]string, 0), *toUpdate.Spec.Storage.Devices...)
+				nodeSpecCopy.Storage.Devices = &devices
 			}
-			if nodeSpecCopy.Storage.UseAllWithPartitions == nil && toUpdate.Spec.Storage.UseAllWithPartitions != nil {
+			if nodeSpecCopy.Storage.UseAllWithPartitions == nil &&
+				(nodeSpecCopy.Storage.UseAll == nil || !*nodeSpecCopy.Storage.UseAll) &&
+				nodeSpecCopy.Storage.Devices == nil &&
+				toUpdate.Spec.Storage.UseAllWithPartitions != nil {
 				nodeSpecCopy.Storage.UseAllWithPartitions = boolPtr(*toUpdate.Spec.Storage.UseAllWithPartitions)
+			}
+			if nodeSpecCopy.Storage.UseAll == nil &&
+				(nodeSpecCopy.Storage.UseAllWithPartitions == nil || !*nodeSpecCopy.Storage.UseAllWithPartitions) &&
+				nodeSpecCopy.Storage.Devices == nil &&
+				toUpdate.Spec.Storage.UseAll != nil {
+				nodeSpecCopy.Storage.UseAll = boolPtr(*toUpdate.Spec.Storage.UseAll)
 			}
 			if nodeSpecCopy.Storage.ForceUseDisks == nil && toUpdate.Spec.Storage.ForceUseDisks != nil {
 				nodeSpecCopy.Storage.ForceUseDisks = boolPtr(*toUpdate.Spec.Storage.ForceUseDisks)
-			}
-			if nodeSpecCopy.Storage.Devices == nil && toUpdate.Spec.Storage.Devices != nil {
-				devices := append(make([]string, 0), *toUpdate.Spec.Storage.Devices...)
-				nodeSpecCopy.Storage.Devices = &devices
 			}
 			if nodeSpecCopy.Storage.JournalDevice == nil && toUpdate.Spec.Storage.JournalDevice != nil {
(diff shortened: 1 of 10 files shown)


drivers/storage/portworx/portworx.go
```diff
@@ -652,6 +654,42 @@ func defaultPortworxImageVersion(releases *manifest.ReleaseManifest) string {
...
+func setNodeSpecDefaults(toUpdate *corev1alpha1.StorageCluster) {
+	if len(toUpdate.Spec.Nodes) == 0 {
+		return
+	}
+
+	updatedNodeSpecs := make([]corev1alpha1.NodeSpec, 0)
+	for _, nodeSpec := range toUpdate.Spec.Nodes {
+		nodeSpecCopy := nodeSpec.DeepCopy()
+		if nodeSpec.Storage == nil {
+			nodeSpecCopy.Storage = toUpdate.Spec.Storage.DeepCopy()
```
adityadani: check here if toUpdate.Spec.Storage != nil

adityadani: Or this could be refactored to
```
if toUpdate.Spec.Storage != nil {
     		if nodeSpec.Storage == nil {
			nodeSpecCopy.Storage = toUpdate.Spec.Storage.DeepCopy()
                } else {
                          ........
               }
}
```

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -643,6 +640,108 @@ func (c *Controller) syncNodes(
...
+	nodeList := &v1.NodeList{}
+	err := c.client.List(context.TODO(), nodeList, &client.ListOptions{})
+	if err != nil {
+		return nil, nil, err
+	}
+
+	remainingNodes := make(map[string]*v1.Node)
+	for _, nodeName := range nodesNeedingStoragePods {
+		for _, node := range nodeList.Items {
+			if node.Name == nodeName {
```
adityadani: what if there is a node present in the nodesNeedingStoragePods, but not present in nodeList.Items ?
Even though the chances of that happening are slim, we should at least log it, as we are emptying out the nodesNeedingStoragePods slice later.

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -643,6 +640,108 @@ func (c *Controller) syncNodes(
...
+
+func (c *Controller) createPodTemplateForNodeGroup(
+	cluster *corev1alpha1.StorageCluster,
+	nodeGroup []*v1.Node,
+	nodesNeedingStoragePods *[]string,
+	podTemplates *[]*v1.PodTemplateSpec,
+	remainingNodes map[string]*v1.Node,
+	hash string,
+) error {
+	podTemplate, err := c.createPodTemplate(cluster, nodeGroup[0], hash)
```
adityadani: Shouldn't this be inside the for loop ?
Why use the 1st node for all pod templates?

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -643,6 +640,108 @@ func (c *Controller) syncNodes(
...
+			for _, node := range remainingNodes {
+				if nodeSelector.Matches(labels.Set(node.Labels)) {
+					nodeGroup = append(nodeGroup, node)
+				}
+			}
+		}
+
+		if len(nodeGroup) > 0 {
+			clusterForNodeGroup := cluster.DeepCopy()
+			updateClusterSpecWithNodeSpec(&clusterForNodeGroup.Spec, nodeSpec.DeepCopy())
```
adityadani: Is this update required for the hash check?
Also can we rename it to overwriteClusterSpecWithNodeSpec and a doc here

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -643,6 +640,108 @@ func (c *Controller) syncNodes(
...
+				clusterForNodeGroup, nodeGroup,
+				&nodesNeedingStoragePods, &podTemplates,
+				remainingNodes, hash,
+			); err != nil {
+				return nil, nil, err
+			}
+		}
+	}
+
+	if len(remainingNodes) > 0 {
```
adityadani: nit: Can you add some docs in this function describing what it does.

pkg/apis/core/v1alpha1/storagecluster.go
```diff
@@ -100,10 +100,8 @@ type NodeSpec struct {
 	// Selector rest of the attributes are applied to a node that matches
 	// the selector
 	Selector NodeSelector `json:"selector,omitempty"`
-	// Geo is topology information for the node
-	Geo *Geography `json:"geography,omitempty"`
-	// CommonConfig that is present at both cluster and node level
-	CommonConfig
+	// Storage details of storage used by the driver
+	Storage *StorageSpec `json:"storage,omitempty"`
```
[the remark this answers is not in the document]
adityadani: discussed offline

pkg/controller/storagecluster/storagecluster.go
piyush-nimbalkar: Not needed for hash check because we use the complete StorageCluster spec to calculate the hash. 
Will rename the method.

pkg/controller/storagecluster/storagecluster.go
piyush-nimbalkar: All the nodes in nodeGroup will have the same template, so no need to call createPodTemplate for each one of the nodes. That is why we are grouping them.
In most cases (without any node spec) there is going to be a single template for all pods, so wanted to optimize this.

pkg/controller/storagecluster/storagecluster.go
piyush-nimbalkar: If we don't find the node in nodeList, then we should not even create a template or the pod itself as the node does not exist. The nodesNeedingStoragePods is created by us is the same reconcile loop.  The final list is only logged just before creating pods, so what we do here does not matter much as it is an intermediate step.

drivers/storage/portworx/portworx.go
piyush-nimbalkar: The DeepCopy() handles a nil receiver. It returns nil which is what we want.

drivers/storage/portworx/portworx.go
```diff
@@ -652,6 +654,42 @@ func defaultPortworxImageVersion(releases *manifest.ReleaseManifest) string {
...
+		} else if toUpdate.Spec.Storage != nil {
+			if nodeSpecCopy.Storage.UseAll == nil && toUpdate.Spec.Storage.UseAll != nil {
```
adityadani: UseAll | UseAllWithPartitions | DeviceSpecs are exclusive and only one of them should be set.

PR #76, opened 2019-12-12
Extracting components that operator deploys

piyush-nimbalkar:
Creating a `PortworxComponent` interface to help implement portworx components easier. 

All the components in the operator project are auto registered and will be reconciled based on whether the component is enabled or not. The operator will just go over the registered components and reconcile them.

```
for component in RegisteredComponents:
  if component.IsEnabled() {
    component.Reconcile()
  } else {
    component.Delete()
  }
}
```

- This will make it easier to write and manage new components added to Portworx driver
- Also components can be written outside the operator repo and just vendored in the operator project.


drivers/storage/portworx/component/component.go
```diff
@@ -0,0 +1,61 @@
...
+	MarkDeleted()
+}
+
+var (
+	components = make(map[string]PortworxComponent)
+)
+
+// Register registers a PortworxComponent and stores in the global map of components
+func Register(name string, c PortworxComponent) {
+	logrus.Debugf("Registering component %v with Portworx driver", name)
```
adityadani: usually it is a good idea to add locks in the registry model

drivers/storage/portworx/component/storage_class.go
```diff
@@ -0,0 +1,198 @@
+package component
```
adityadani: nit: can rename this to portworx_sc.go to make it similar to others
OR
you can completely get rid of the portworx_ prefix as the pkg hierarchy already has portworx in it

drivers/storage/portworx/component/storage_class.go
```diff
@@ -0,0 +1,198 @@
...
+				api.SpecHaLevel:   "3",
+				api.SpecIoProfile: "db",
+			},
+		},
+		{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:            pxDbEncryptedStorageClass,
+				OwnerReferences: []metav1.OwnerReference{*ownerRef},
+				Annotations: map[string]string{
+					"params/note": "Ensure that you have a cluster-wide secret created in the configured secrets provider",
```
adityadani: why are there no docAnnotations to the rest of the storage classes ?

adityadani: We could even add only the relevant annotations to each storage class, instead of putting all of them

drivers/storage/portworx/components_test.go
```diff
@@ -2963,7 +2921,7 @@ func TestCSIInstallWithNewerCSIVersion(t *testing.T) {
 
 	expectedCR := testutil.GetExpectedClusterRole(t, "csiClusterRole_k8s_1.13.yaml")
 	actualCR := &rbacv1.ClusterRole{}
-	err = testutil.Get(k8sClient, actualCR, csiClusterRoleName, "")
+	err = testutil.Get(k8sClient, actualCR, "px-csi", "")
```
adityadani: nit feel free to ignore: you could make all of them global constants in pkg/components
So that you don't need to keep changing the tests if a name changes.

drivers/storage/portworx/component/storage_class.go
piyush-nimbalkar: The idea was to have a basic storage class which talks about all the supported fields and also points to our docs website. We have also added a note on the encrypted storage class about how using cluster wide secret. Adding to all storage classes will make it very verbose, especially when they cannot change any of these.

PR #81, opened 2020-01-21
Add prometheus component

piyush-nimbalkar:


drivers/storage/portworx/portworx.go
```diff
@@ -181,6 +181,15 @@ func (p *portworx) SetDefaultsOnStorageCluster(toUpdate *corev1alpha1.StorageClu
 		}
 	}
 
+	// If prometheus is enabled and metrics are not enabled explicitly, then
+	// enabled exporting of metrics
+	if toUpdate.Spec.Monitoring != nil &&
```
adityadani: What happens when Spec.Monitoring.EnableMetrics is set to true, but Prometheus is nil ?

drivers/storage/portworx/component/prometheus.go
```diff
@@ -0,0 +1,514 @@
...
+	cluster *corev1alpha1.StorageCluster,
+	ownerRef *metav1.OwnerReference,
+) *appsv1.Deployment {
+	replicas := int32(1)
+	runAsNonRoot := true
+	runAsUser := int64(65534)
+	labels := map[string]string{
+		"k8s-app": PrometheusOperatorDeploymentName,
+	}
+	operatorImage := util.GetImageURN(cluster.Spec.CustomImageRegistry, "quay.io/coreos/prometheus-operator:v0.29.0")
```
adityadani: shouldn't the image name be a part of the spec ?

adityadani: as prometheus is external to PX, I think we will eventually have to give more control to the users as to how they want to configure prometheus. May be just take the operator deployment spec in the spec.Monitoring.Prometheus.Spec value.

drivers/storage/portworx/component/prometheus.go
```diff
@@ -0,0 +1,514 @@
...
+							Args:  args,
+							Ports: []v1.ContainerPort{
+								{
+									Name:          "http",
+									ContainerPort: int32(8080),
+								},
+							},
+							Resources: v1.ResourceRequirements{
+								Limits: map[v1.ResourceName]resource.Quantity{
+									v1.ResourceCPU:    resource.MustParse("200m"),
```
adityadani: What happens if someone externally updates these resource requirements on the deployment? Will the operator reconcile it back to these values?

adityadani: It does an Update so it will be overwritten again.

pkg/apis/core/v1alpha1/storagecluster.go
```diff
@@ -348,7 +348,18 @@ type DataProviderSpec struct {
 type MonitoringSpec struct {
 	// EnableMetrics this exposes the storage cluster metrics to external
 	// monitoring solutions like Prometheus.
-	EnableMetrics bool `json:"enableMetrics,omitempty"`
+	EnableMetrics *bool `json:"enableMetrics,omitempty"`
+	// Prometheus contains the details of the Prometheus stack deployed to monitor
+	// metrics from the storage cluster.
+	Prometheus *PrometheusSpec `json:"prometheus,omitempty"`
```
adityadani: The spec is not explicit enough to distinguish between 
- deploying prometheus operator only
- deploying prometheus and creating a service monitor
- creating only service monitor

adityadani: May be it should be something like this
```
type MonitoringSpec struct {
    Prometheus *PrometheusSpec
}

type PrometheusSpec struct {
       EnableMetrics *bool -> enables exporting metrics to prometheus endpoint(which could be deployed by us)
       DeployComponent *bool -> deploy prometheus stack
       RemoteWriteEndpoint string
}

```

drivers/storage/portworx/component/prometheus.go
piyush-nimbalkar: For now, let's not expose it to the users as we also create Custom resources that are registered by the prometheus operator. Allowing change in version could result in incompatible CRD specs. If there is a need we can expose it through env var in the near future.

drivers/storage/portworx/component/prometheus.go
piyush-nimbalkar: As discussed offline, this will get overwritten as we don't store the deployment config. For now removing the resource completely and letting k8s allocate resources to the pod. If there is a need we can expose an annotation to override the cpu/mem like we do for pvc-controller, stork and autopilot deployments.

pkg/apis/core/v1alpha1/storagecluster.go
piyush-nimbalkar: We cannot remove EnableMetrics as it is a released flag. Marking it for deprecation. If present copying it to the new flag in Prometheus section and then removing it from the StorageCluster object.
Now the spec looks like this -
```
type MonitoringSpec struct {
    EnableMetrics *bool
    Prometheus *PrometheusSpec
}

type PrometheusSpec struct {
       ExportMetrics bool // new flag that gets set if EnableMetrics was set previously
       Enabled bool 
       RemoteWriteEndpoint string
}
```

deploy/crds/core_v1alpha1_storagecluster_crd.yaml
```diff
@@ -561,8 +561,21 @@ spec:
...
+                prometheus:
+                  type: object
+                  description: Contains configuration of Prometheus to monitor the storage cluster.
+                  properties:
+                    exportMetrics:
+                      type: boolean
+                      description: If this flag is enabled it will expose the storage cluster metrics to Prometheus.
+                    enabled:
+                      type: boolean
+                      description: Flag indicating whether Prometheus stack needs to be enabled.
```
adityadani: nit: Flag indicating whether Prometheus stack needs to be enabled and deployed by the Storage operator.

PR #82, opened 2020-01-22
Allow disabling of storage pods

piyush-nimbalkar:
- If portworx is disabled, then do not deploy
  portworx pods, portworx RBAC, API pods, portworx service, storage classes
- Do not update storage cluster status or create storage node objects
- Do not allow deleting the portworx with any delete strategy
- We should still be able to deploy components using the StorageCluster

### DIFF
--- a/pkg/controller/storagecluster/storagecluster.go
+++ b/pkg/controller/storagecluster/storagecluster.go
@@ -24,6 +24,7 @@ import (
 	"path"
 	"reflect"
 	"sort"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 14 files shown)


pkg/controller/storagecluster/storagecluster.go
```diff
@@ -64,7 +65,10 @@ import (
 
 const (
 	// ControllerName is the name of the controller
-	ControllerName                      = "storagecluster-controller"
+	ControllerName = "storagecluster-controller"
+	// AnnotationDisableStorage annotation to disable the storage pods from running.
+	// Defaults to false value.
+	AnnotationDisableStorage            = operatorPrefix + "/disable-storage-pods"
```
adityadani: nit: why not just "disable-storage "?

piyush-nimbalkar:
Added another commit to remove portworx api, service and rbac when portworx is disabled post installation. There is no need to keep the above components as portworx pods are not running anyway. 
Also addressed the comment.

PR #83, opened 2020-01-22
add ValidateStorageCluster

thiguetta:

### DIFF
--- a/pkg/util/test/util.go
+++ b/pkg/util/test/util.go
@@ -3,7 +3,12 @@ package test
 import (
 	"context"
 	"fmt"
+	"github.com/libopenstorage/openstorage/api"
+	"github.com/portworx/sched-ops/k8s"
+	"github.com/portworx/sched-ops/task"
+	"google.golang.org/grpc"
 	"io/ioutil"
+	"k8s.io/kubernetes/pkg/scheduler/algorithm/predicates"
 	"path"
 	"testing"
 	"time"
@@ -26,6 +31,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes/scheme"
+	schedulernodeinfo "k8s.io/kubernetes/pkg/scheduler/nodeinfo"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 )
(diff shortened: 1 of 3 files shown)


test/integration_test/common_test.go
```diff
@@ -70,11 +72,45 @@ func validateStorageCluster(
...
+		return err
+	}
+
+	svc, err := k8s.Instance().GetService("portworx-service", cluster.Namespace)
+	pxEndpoint := ""
+	if err == nil {
+		pxEndpoint = svc.Spec.ClusterIP
+	}
+
+	conn, err := grpc.Dial(pxEndpoint, grpc.WithInsecure())
```
piyush-nimbalkar: The endpoint would also require the sdk port. You can get that from the above service itself.

test/integration_test/common_test.go
```diff
@@ -70,11 +72,45 @@ func validateStorageCluster(
...
+	if err != nil {
+		return err
+	}
+	nodeClient := api.NewOpenStorageNodeClient(conn)
+	nodeEnumerateResp, err := nodeClient.Enumerate(context.Background(), &api.SdkNodeEnumerateRequest{})
+	if err != nil {
+		return err
+	}
+	expectedNodes := len(cluster.Spec.Nodes)
```
piyush-nimbalkar: The cluster nodes are only used to overwrite the node specific spec, so it may not always be populated. If there are no nodes in this section, it means that all nodes will use the configuration given at the cluster.Spec level.

We should validate the number of px nodes should be equal to `expectedPods()`. This method takes into consideration node affinity rules if there are any.

test/integration_test/common_test.go
```diff
@@ -70,11 +72,45 @@ func validateStorageCluster(
...
+	actualNodes := len(nodeEnumerateResp.GetNodeIds())
+	if actualNodes != expectedNodes {
+		return fmt.Errorf("expected nodes: %v. actual nodes: %v", expectedNodes, actualNodes)
+	}
 
-	// TODO: Validate expected number of portworx nodes are online (using sdk)
 	// TODO: Validate portworx is started with correct params. Check individual options
+	for _, n := range nodeEnumerateResp.GetNodeIds() {
+		nodeResp, err := nodeClient.Inspect(d.getContext(), &api.SdkNodeInspectRequest{NodeId: n})
```
piyush-nimbalkar: This is good. To start with, we should validate the storage cluster params against the response received from inspect. At the very least check the status is Online/Running from the response.

pkg/util/test/util.go
```diff
@@ -217,3 +218,8 @@ func ActivateCRDWhenCreated(fakeClient *fakeextclient.Clientset, crdName string)
 		return false, nil
 	})
 }
+
+// ValidateStorageCluster validates a StorageCluster spec
+func ValidateStorageCluster(cluster *corev1alpha1.StorageCluster) error {
+	return fmt.Errorf("not implemented yet")
```
piyush-nimbalkar: You can move the implementation here from the common_test file.
Also you will have to move - 
UninstallStorageCluster() and ValidateUninstallStorageCluster(), so we can use it in spawn.

nikolaypopov:
What does the validation function suppose to do? I was under the impression that it is suppose to compare PX StorageCluster yaml with the parameters PX actually got deployed with and check if they match.

pkg/util/test/util.go
```diff
@@ -217,3 +225,190 @@ func ActivateCRDWhenCreated(fakeClient *fakeextclient.Clientset, crdName string)
...
+}
+
+// ValidateStorageCluster validates a StorageCluster spec
+func ValidateStorageCluster(
+	cluster *corev1alpha1.StorageCluster,
+	timeout, interval time.Duration,
+) error {
+	var err error
+	cluster, err = k8s.Instance().GetStorageCluster(cluster.Name, cluster.Namespace)
+	if err != nil && !errors.IsNotFound(err) {
```
piyush-nimbalkar: We should fail even if it is not found. Otherwise all the following checks will fail or return a nil pointer error.

pkg/util/test/util.go
```diff
@@ -217,3 +225,190 @@ func ActivateCRDWhenCreated(fakeClient *fakeextclient.Clientset, crdName string)
...
+
+	svc, err := k8s.Instance().GetService("portworx-service", cluster.Namespace)
+	pxEndpoint := ""
+	if err != nil {
+		return err
+	}
+
+	servicePort := int32(9020)
+	for _, port := range svc.Spec.Ports {
+		if strings.Contains(port.Name, "sdk") {
```
piyush-nimbalkar: The port is called px-sdk. We can directly compare the name instead of doing Contains.
Should error out if there is no px-sdk port in the service.

test/integration_test/common_test.go
```diff
@@ -46,140 +38,8 @@ func createStorageCluster(cluster *corev1alpha1.StorageCluster) error {
 	return err
 }
 
-func uninstallStorageCluster(cluster *corev1alpha1.StorageCluster) error {
```
piyush-nimbalkar: We will have to change basic test to use the methods from test/util now that we have deleted it from here.

PR #84, opened 2020-01-28
Convert to go modules

rodrigc:
The following Portworx projects use go modules:

* https://github.com/portworx/sched-ops/blob/master/go.mod
* https://github.com/portworx/torpedo/blob/master/go.mod
* https://github.com/portworx/spawn/blob/master/go.mod

Kubernetes 1.15 and higher use go modules:

* https://github.com/kubernetes/kubernetes/blob/release-1.15/go.mod

It will be easier to interoperate between this projects if `libopenstorage/operator` also uses go modules.


thiguetta:
in general, it looks good to me, don't know if it might be a concern but we have other projects in libopenstorage using dep, such as:
https://github.com/libopenstorage/stork
https://github.com/libopenstorage/autopilot-api
https://github.com/libopenstorage/gossip

rodrigc:
Those can be converted over time.  I have an old PR for stork for example that needs to be brought up to date:
https://github.com/libopenstorage/stork/pull/296

rodrigc:
@thiguetta I updated my PR to stork for go modules: https://github.com/libopenstorage/stork/pull/296

hack/update-codegen.sh
```diff
@@ -21,7 +21,22 @@ set -o pipefail
...
+function cleanup {
+  mv ${CODEGEN_PKG}/generate-groups.sh.orig ${CODEGEN_PKG}/generate-groups.sh
+}
+trap cleanup EXIT
+
+go install ./${CODEGEN_PKG}/cmd/{defaulter-gen,client-gen,lister-gen,informer-gen,deepcopy-gen}
+
+
 # generate the code with:
+chmod +x ${CODEGEN_PKG}/generate-groups.sh
```
piyush-nimbalkar, line 39: Or we could just do `bash ${CODEGEN_PKG}/genere....`

rodrigc: Sure we could do that, that's probably better.

rodrigc: Fixed

harsh-px:
No longer required. Operator is already using go modules.

PR #91, opened 2020-01-30
Apply imagePullSecret to all components

piyush-nimbalkar:
- Also do the same for image pull policy


pkg/controller/storagecluster/stork.go
```diff
@@ -823,6 +820,15 @@ func getStorkServiceLabels() map[string]string {
 	}
 }
 
+func imagePullPolicy(cluster *corev1alpha1.StorageCluster) v1.PullPolicy {
```
adityadani: why this special handling for stork ?

piyush-nimbalkar: As discussed, we have the same method for all portworx components. So added the same for Stork.

PR #92, opened 2020-01-31
Apply imagePullSecret to all components (cherry-pick)

piyush-nimbalkar:
- Also do the same for image pull policy


drivers/storage/portworx/component/lighthouse.go
```diff
@@ -338,7 +338,7 @@ func getLighthouseDeploymentSpec(
 	maxSurge := intstr.FromInt(1)
 	imagePullPolicy := pxutil.ImagePullPolicy(cluster)
```
adityadani: this imagePullPolicy variable is not being used anywhere

piyush-nimbalkar: It is used below in all the lighthouse containers.

PR #101, opened 2020-02-10
PTX-2254 Validate components: stork, lighthouse, autopilot and pvc controller

thiguetta:

### DIFF
--- a/pkg/util/test/util.go
+++ b/pkg/util/test/util.go
@@ -4,7 +4,10 @@ import (
 	"context"
 	"fmt"
 	"io/ioutil"
+	"os"
 	"path"
+	"strconv"
+	"strings"
 	"testing"
 	"time"
 
@@ -13,6 +16,7 @@ import (
 	"github.com/libopenstorage/openstorage/api"
 	corev1alpha1 "github.com/libopenstorage/operator/pkg/apis/core/v1alpha1"
 	"github.com/libopenstorage/operator/pkg/mock"
+	appops "github.com/portworx/sched-ops/k8s/apps"
 	coreops "github.com/portworx/sched-ops/k8s/core"
 	k8serrors "github.com/portworx/sched-ops/k8s/errors"
 	operatorops "github.com/portworx/sched-ops/k8s/operator"
@@ -236,8 +240,11 @@ func ActivateCRDWhenCreated(fakeClient *fakeextclient.Clientset, crdName string)
 }
 
 // UninstallStorageCluster uninstalls and wipe storagecluster from k8s
-func UninstallStorageCluster(cluster *corev1alpha1.StorageCluster) error {
+func UninstallStorageCluster(cluster *corev1alpha1.StorageCluster, kubeconfig ...string) error {
 	var err error
+	if len(kubeconfig) != 0 && kubeconfig[0] != "" {
+		os.Setenv("KUBECONFIG", kubeconfig[0])
+	}
 	cluster, err = operatorops.Instance().GetStorageCluster(cluster.Name, cluster.Namespace)
 	if err != nil && !errors.IsNotFound(err) {
 		return err
@@ -260,8 +267,12 @@ func UninstallStorageCluster(cluster *corev1alpha1.StorageCluster) error {
 func ValidateStorageCluster(
 	cluster *corev1alpha1.StorageCluster,
 	timeout, interval time.Duration,
+	kubeconfig ...string,
 ) error {
 	var err error
+	if len(kubeconfig) != 0 && kubeconfig[0] != "" {
+		os.Setenv("KUBECONFIG", kubeconfig[0])
+	}
 	cluster, err = operatorops.Instance().GetStorageCluster(cluster.Name, cluster.Namespace)
 	if err != nil {
 		return err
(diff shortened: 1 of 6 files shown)


pkg/util/test/util.go
```diff
@@ -288,6 +288,42 @@ func ValidateStorageCluster(
 
 	pxEndpoint = fmt.Sprintf("%s:%d", svc.Spec.ClusterIP, servicePort)
 
+	if err := validatePodsByName(cluster.Namespace, "portworx-pvc-controller", timeout, interval); err != nil {
+		return err
+	}
+
+	if cluster.Spec.Stork != nil && cluster.Spec.Stork.Enabled {
```
piyush-nimbalkar: In case of stork, lighthouse and autopilot we should validate the image in the pod/deployment is same as storage cluster.
For instance stork deployment should have `cluster.Spec.Stork.Image`

pkg/util/test/util.go
```diff
@@ -288,6 +288,42 @@ func ValidateStorageCluster(
...
+		if err := validatePodsByName(cluster.Namespace, "stork", timeout, interval); err != nil {
```
piyush-nimbalkar: We can probably just use ValidateDeployment from sched-ops. It verifies that the number of pod is the same as given in the deployment. That way we are verifying the deployment is present and its pods are running.

pkg/util/test/util.go
```diff
@@ -288,6 +288,42 @@ func ValidateStorageCluster(
 
 	pxEndpoint = fmt.Sprintf("%s:%d", svc.Spec.ClusterIP, servicePort)
 
+	if err := validatePodsByName(cluster.Namespace, "portworx-pvc-controller", timeout, interval); err != nil {
```
piyush-nimbalkar: PVC controller is not always deployed. You can refer the following function to see when pvc controller is enabled - https://github.com/libopenstorage/operator/blob/master/drivers/storage/portworx/component/pvccontroller.go#L58

I would recommend duplicating a similar function in test.

pkg/util/test/util.go
```diff
@@ -288,6 +288,42 @@ func ValidateStorageCluster(
...
+		return err
+	}
+
+	if cluster.Spec.Stork != nil && cluster.Spec.Stork.Enabled {
+		if err := validatePodsByName(cluster.Namespace, "stork", timeout, interval); err != nil {
+			return err
+		}
+		if err := validatePodsByName(cluster.Namespace, "stork-scheduler", timeout, interval); err != nil {
+			return err
```
piyush-nimbalkar: For stork-scheduler and pvc-controller we should check that the image tag is same as the k8s version.

pkg/util/test/util.go
nikolaypopov: @thiguetta you can refer to this PR https://github.com/portworx/spawn/pull/518 in CreatePortworxStorageClusterSpec() in cmd/provision_px_k8s.go file to see what I'm fetching from the StorageCluster yaml and implement same checks here as well.

pkg/util/test/util.go
```diff
@@ -288,6 +288,42 @@ func ValidateStorageCluster(
...
+		}
+	}
+
+	if cluster.Spec.Autopilot != nil && cluster.Spec.Autopilot.Enabled {
```
nikolaypopov: Should also validate image, see example from my PR https://github.com/portworx/spawn/pull/518
`		// Autopilot
		if spec.Spec.Autopilot != nil {
			params.Autopilot = spec.Spec.Autopilot.Enabled
			if spec.Spec.Autopilot.Image != "" {
				params.AutopilotImage = spec.Spec.Autopilot.Image
			}
		}`

pkg/util/test/util.go
```diff
@@ -288,6 +288,42 @@ func ValidateStorageCluster(
...
+		if err := validatePodsByName(cluster.Namespace, "autopilot", timeout, interval); err != nil {
+			return err
+		}
+	}
+
+	if cluster.Spec.UserInterface != nil && cluster.Spec.UserInterface.Enabled {
```
nikolaypopov: Should also validate image, see example from my PR portworx/spawn#518
`		// Lighthouse
		if spec.Spec.UserInterface != nil {
			if spec.Spec.UserInterface.Enabled {
				params.EnableLighthouse = "true"
				if spec.Spec.UserInterface.Image != "" {
					params.EnableLighthouse = spec.Spec.UserInterface.Image
				}
			}
		}`

pkg/util/test/util.go
```diff
@@ -288,6 +288,42 @@ func ValidateStorageCluster(
...
+		if err := validatePodsByName(cluster.Namespace, "lighthouse", timeout, interval); err != nil {
+			return err
+		}
+	}
+
+	if csi, ok := cluster.Spec.FeatureGates["CSI"]; ok && csi == "true" {
+		if err := validatePodsByApp(cluster.Namespace, "px-csi-driver", timeout, interval); err != nil {
```
nikolaypopov: We should also probably validate `portworx` pods here, as they will have extra containers running if `CSI` is enabled.

pkg/util/test/util.go
thiguetta: Done

pkg/util/test/util.go
thiguetta: Done

pkg/util/test/util.go
thiguetta: Done

pkg/util/test/util.go
thiguetta: Done

pkg/util/test/util.go
```diff
@@ -277,32 +280,118 @@ func ValidateStorageCluster(
...
+		lighthouseDp.Namespace = cluster.Namespace
+		if err = appops.Instance().ValidateDeployment(lighthouseDp, timeout, interval); err != nil {
+			return err
+		}
+		if err = validateImageOnPods(cluster.Spec.UserInterface.Image, cluster.Namespace, map[string]string{"name": "lighthouse"}); err != nil {
+			return err
+		}
+	}
+
+	if csi, ok := cluster.Spec.FeatureGates["CSI"]; ok && csi == "true" {
```
piyush-nimbalkar: We should allow all truth values. Let's do it this way -
```
if csi, _ := strconv.ParseBool(cluster.Spec.FeatureGates["CSI"]); csi {
  ...
}
```

pkg/util/test/util.go
```diff
@@ -277,32 +280,118 @@ func ValidateStorageCluster(
...
+	if err != nil {
+		return err
+	}
+
+	if isPVCControllerEnabled(cluster) {
+		if err = validatePodsByName(cluster.Namespace, "portworx-pvc-controller", timeout, interval); err != nil {
+			return err
+		}
+
+		if err = validateImageTag(k8sVersion.String(), cluster.Namespace, map[string]string{"name": "portworx-pvc-controller"}); err != nil {
```
piyush-nimbalkar: We should ideally validate the complete image name. If people are using custom image registry, we should validate that we well with the tag.
Let's do that as part of a new change. For now this is fine.

pkg/util/test/util.go
```diff
@@ -277,32 +280,118 @@ func ValidateStorageCluster(
 		return err
 	}
 
+	k8sVersion, err := coreops.Instance().GetVersion()
+	if err != nil {
+		return err
+	}
+
+	if isPVCControllerEnabled(cluster) {
```
piyush-nimbalkar: nit: Can we extract all the component validations in a single separate function for readability?

PR #115, opened 2020-03-05
Create proxy pod and service in kube-system namespace for Portworx volume provisioning

piyush-nimbalkar:
- If running outside kube-system namespace and not using 9001 port, we create a proxy service to redirect in-tree driver traffic to portworx nodes.
- Pass PX_NAMESPACE to portworx pods.

### DIFF
--- a/drivers/storage/portworx/component/portworx_proxy.go
+++ b/drivers/storage/portworx/component/portworx_proxy.go
@@ -6,6 +6,7 @@ import (
 	"github.com/hashicorp/go-version"
 	pxutil "github.com/libopenstorage/operator/drivers/storage/portworx/util"
 	corev1alpha1 "github.com/libopenstorage/operator/pkg/apis/core/v1alpha1"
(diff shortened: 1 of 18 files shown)


drivers/storage/portworx/component/portworx_proxy.go
```diff
@@ -0,0 +1,203 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/record"
+	api "k8s.io/kubernetes/pkg/apis/core"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	// PortworxProxyComponent name of the Portworx Proxy component. This component
+	// runs in the kube-system namespace if the cluster is running outside. This
+	// ensures that k8s in-tree driver traffic gets routed the Portworx nodes.
```
adityadani: nit: routed to the Portworx nodes

PR #118, opened 2020-03-09
PTX-2253 Validate storage cluster arguments against portworx pod arguments

thiguetta:


pkg/util/test/util.go
```diff
@@ -643,6 +655,102 @@ func validateImageTag(tag, namespace string, listOptions map[string]string) erro
...
+			if cluster.Spec.Network != nil && cluster.Spec.Network.DataInterface != nil {
+				if *cluster.Spec.Network.DataInterface != args[i+1] {
+					return fmt.Errorf("data network interface does not match. expected: %s, got: %s",
+						*cluster.Spec.Network.DataInterface, args[i+1])
+				}
+			} else {
+				return fmt.Errorf("data network interface expected to be set")
+			}
+			i++
+		case "-s":
```
piyush-nimbalkar: We need to check against cloud drive specs as well. 
`-s` is used for both normal storage as well as cloud drive storage. So it could have a value from either `spec.storage.devices` or `spec.cloudStorage.deviceSpecs`.
Same logic applies to `-j` and `-metadata`

Also check these storage args -
spec.cloudStorage.MaxStorageNodes (-max_drive_set_count)
spec.cloudStorage.MaxStorageNodesPerZone (-max_storage_nodes_per_zone)

pkg/util/test/util.go
```diff
@@ -643,6 +655,102 @@ func validateImageTag(tag, namespace string, listOptions map[string]string) erro
...
+						*cluster.Spec.Network.MgmtInterface, args[i+1])
+				}
+			} else {
+				return fmt.Errorf("data network interface expected to be set")
+			}
+			i++
+		case "-k":
+			expectedEndpoints := strings.Split(args[i+1], ",")
+			sort.Strings(expectedEndpoints)
+			existingEndpoints := cluster.Spec.Kvdb.Endpoints
```
piyush-nimbalkar: Args `-cert`, `-key`, `-ca`, `-acltoken` and `-userpwd` should have the values from `spec.kvdb.authSecret`.
This is different from other params because here the authSecret is just the name of the secret in cluster.Namespace. We will have to read the secret and validate the above values correspond to a key from the k8s secret.
```
key in k8s secret (arg)
kvdb-ca.crt (-ca)
kvdb.crt (-cert)
kvdb.key (-key)
acl-token (-acltoken)
username
password (-userpwd username:password)
```

PR #133, opened 2020-03-25
Pass portworx service name to stork

piyush-nimbalkar:
- Pass PX_SERVICE_NAME env to stork

### DIFF
--- a/drivers/storage/portworx/portworx.go
+++ b/drivers/storage/portworx/portworx.go
@@ -104,6 +104,10 @@ func (p *portworx) GetStorkEnvList(cluster *corev1alpha1.StorageCluster) []v1.En
 			Name:  pxutil.EnvKeyPortworxNamespace,
 			Value: cluster.Namespace,
 		},
+		{
+			Name:  pxutil.EnvKeyPortworxServiceName,
+			Value: component.PxAPIServiceName,
+		},
 	}
 }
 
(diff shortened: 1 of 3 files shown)


drivers/storage/portworx/deployment.go
```diff
@@ -117,10 +117,12 @@ var (
...
 		{
-			name: "pxlogs",
+			name:             "libosd",
+			hostPath:         "/var/lib/osd",
+			mountPath:        "/var/lib/osd",
+			mountPropagation: mountPropagationModePtr(v1.MountPropagationBidirectional),
 			pks: &pksVolumeInfo{
-				hostPath:  "/var/vcap/store/lib/osd/log",
-				mountPath: "/var/lib/osd/log",
+				hostPath: "/var/vcap/store/lib/osd",
```
adityadani: This needs to be /var/vcap/store/lib/osd/log. Refer PWX-12002

drivers/storage/portworx/testspec/pks.yaml
```diff
@@ -114,6 +115,9 @@ spec:
         - name: etcpwx
           hostPath:
             path: /var/vcap/store/etc/pwx
+        - name: libosd
```
adityadani: same here

adityadani:
@piyush-nimbalkar We decided to not add this bind mount at all. You can close this PR. The relevant alert related files will now lie under /var/cores which is already bind mounted in oci-monitor

PR #135, opened 2020-03-26
Added px-backup service monitor selector into Prometheus Instance

tusharraut1994:


drivers/storage/portworx/component/monitoring.go
```diff
@@ -184,6 +186,7 @@ func (c *monitoring) retryCreate(
 func serviceMonitorLabels() map[string]string {
 	return map[string]string{
 		"name": PxServiceMonitor,
+		"app":  PxBackupServiceMonitor,
```
piyush-nimbalkar: These labels are added on the portworx service montior as well. Can you separate out this method into two, one for prometheus's service monitor selector labels and other for the labels on the service monitor itself?

tusharraut1994: Added.

PR #145, opened 2020-04-02
Give portworx permissions for custom secrets namespace

piyush-nimbalkar:
- Also override existing default env var values with
  user-passed values. Dedup env vars.

### DIFF
--- a/drivers/storage/portworx/component/portworx_basic.go
+++ b/drivers/storage/portworx/component/portworx_basic.go
@@ -1,14 +1,19 @@
 package component
 
 import (
+	"context"
+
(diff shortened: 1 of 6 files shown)


drivers/storage/portworx/component/portworx_basic.go
```diff
@@ -131,6 +182,7 @@ func (c *portworxBasic) createRole(clusterNamespace string, ownerRef *metav1.Own
 }
 
 func (c *portworxBasic) createRoleBinding(
+	namespace string,
```
adityadani, line 185: nit: could we rename this to roleBindingNamespace

PR #161, opened 2020-05-14
Add auth support to cluster status requests

ggriffiths:

### DIFF
--- a/drivers/storage/portworx/portworx.go
+++ b/drivers/storage/portworx/portworx.go
@@ -35,6 +35,7 @@ const (
 	defaultNodeWiperImage             = "portworx/px-node-wiper:2.1.4"
 	envKeyNodeWiperImage              = "PX_NODE_WIPER_IMAGE"
 	envKeyPortworxEnableTLS           = "PX_ENABLE_TLS"
+	envKeyPortworxSharedSecretKey     = "PORTWORX_AUTH_SYSTEM_KEY"
 	storageClusterDeleteMsg           = "Portworx service NOT removed. Portworx drives and data NOT wiped."
 	storageClusterUninstallMsg        = "Portworx service removed. Portworx drives and data NOT wiped."
 	storageClusterUninstallAndWipeMsg = "Portworx service removed. Portworx drives and data wiped."
@@ -241,6 +242,14 @@ func (p *portworx) markComponentsAsDeleted() {
 	for _, comp := range component.GetAll() {
 		comp.MarkDeleted()
 	}
+}
+
+func (p *portworx) normalEvent(
+	cluster *corev1alpha1.StorageCluster,
+	reason, message string,
+) {
+	logrus.Info(message)
+	p.recorder.Event(cluster, v1.EventTypeNormal, reason, message)
 }
 
 func (p *portworx) warningEvent(
--- a/drivers/storage/portworx/status.go
+++ b/drivers/storage/portworx/status.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/libopenstorage/openstorage/api"
+	"github.com/libopenstorage/openstorage/pkg/auth"
 	"github.com/libopenstorage/openstorage/pkg/grpcserver"
 	pxutil "github.com/libopenstorage/operator/drivers/storage/portworx/util"
 	corev1alpha1 "github.com/libopenstorage/operator/pkg/apis/core/v1alpha1"
@@ -20,7 +21,8 @@ import (
 	"github.com/sirupsen/logrus"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
-	"k8s.io/api/core/v1"
+	"google.golang.org/grpc/metadata"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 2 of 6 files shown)


drivers/storage/portworx/status.go
```diff
@@ -433,3 +450,70 @@ func isTLSEnabled() bool {
...
+	if err != nil {
+		return "", err
+	}
+
+	return token, nil
+}
+
+func (p *portworx) getAuthSecret() (string, error) {
+	// Get secret for operator system shared secret
+	pxSharedSecretName := os.Getenv(envKeyPortworxAuthSecretName)
```
piyush-nimbalkar, line 480: The user will not pass these variables to the operator. We will have to read it from the StorageCluster's env.

drivers/storage/portworx/status.go
```diff
@@ -433,3 +450,70 @@ func isTLSEnabled() bool {
...
+	pxSharedSecretNamespace := os.Getenv(envKeyPortworxAuthSecretNamespace)
+	if pxSharedSecretName == "" || pxSharedSecretNamespace == "" {
+		pxSharedSecretKey := os.Getenv(envKeyPortworxSharedSecretKey)
+		return pxSharedSecretKey, nil
+	}
+
+	// Get secret key
+	pxSharedSecret, err := coreops.Instance().GetSecret(pxSharedSecretName, pxSharedSecretNamespace)
```
piyush-nimbalkar, line 488: Let's use the p.k8sClient as it gets from k8s cache instead of hitting k8s API server.
```
pxSharedSecret := &v1.Secret{}
p.k8sClient.Get(
  context.TODO(), 
  types.NamespacedName{Name: pxSharedSecretName, Namespace: pxSharedSecretNamespace},
  pxSharedSecret,
)
```

drivers/storage/portworx/status.go
ggriffiths: Fixed this

drivers/storage/portworx/status.go
ggriffiths: Fixed this and corresponding tests

ggriffiths:
@piyush-nimbalkar 
```
spec:
    image: 'portworx/oci-monitor:b61d074_39e53e1'
    imagePullSecrets:
    - name: regcred
    stork:
        enabled: true
        image: 'openstorage/stork:2.2.5'
    userInterface:
        enabled: true
        image: 'portworx/px-lighthouse:2.0.4'
    storage:
        useAll: true
    env:
    - name: PORTWORX_AUTH_JWT_ISSUER
      value: "openstorage.io"
    - name: PORTWORX_AUTH_JWT_SHAREDSECRET
      value: "mysecret"
    - name: PORTWORX_AUTH_SYSTEM_KEY   # <--- this key already exists in PX startup, so we can reuse it for the operator.
      value: "mysecret"
    - name: PORTWORX_AUTH_SYSTEM_SECRET_NAME  # <--- new key for operator
      value: "opsecret"
    - name: PORTWORX_AUTH_SYSTEM_SECRET_NAMESPACE  # <--- new key for operator
      value: "default"
```

drivers/storage/portworx/status.go
```diff
@@ -433,3 +450,72 @@ func isTLSEnabled() bool {
...
+		return ctx, fmt.Errorf("failed to create operator token: %v", err.Error())
+	}
+	md := metadata.New(map[string]string{
+		"authorization": "bearer " + token,
+	})
+	return metadata.NewOutgoingContext(ctx, md), nil
+}
+
+// GetClusterSpecEnv gets an environment variable string from cluster spec
+func (p *portworx) GetClusterSpecEnv(ctx context.Context, cluster *corev1alpha1.StorageCluster, key string) (string, error) {
```
piyush-nimbalkar, line 518: nit: we can remove this now

drivers/storage/portworx/util/util.go
```diff
@@ -80,6 +84,7 @@ const (
 
 	pxAnnotationPrefix = "portworx.io"
 	labelKeyName       = "name"
+	labelPxSecurity    = "px-security"
```
piyush-nimbalkar, line 87: nit: I guess we don't use this anywhere

drivers/storage/portworx/portworx.go
```diff
@@ -33,8 +33,10 @@ const (
 	defaultSDKPort                    = 9020
 	defaultSecretsProvider            = "k8s"
 	defaultNodeWiperImage             = "portworx/px-node-wiper:2.1.4"
+	pxAuthSecretKey                   = "auth-secret"
 	envKeyNodeWiperImage              = "PX_NODE_WIPER_IMAGE"
 	envKeyPortworxEnableTLS           = "PX_ENABLE_TLS"
+	envKeyPortworxSharedSecretKey     = "PORTWORX_AUTH_SYSTEM_KEY"
```
piyush-nimbalkar, line 39: This is the same env variable that is passed to PX, right? Just want to make sure we are not adding any special env variables.

drivers/storage/portworx/portworx.go
```diff
@@ -33,8 +33,10 @@ const (
 	defaultSDKPort                    = 9020
 	defaultSecretsProvider            = "k8s"
 	defaultNodeWiperImage             = "portworx/px-node-wiper:2.1.4"
+	pxAuthSecretKey                   = "auth-secret"
```
piyush-nimbalkar, line 36: nit: I guess we don't need this either

ggriffiths: Fixed

drivers/storage/portworx/portworx.go
ggriffiths: Yes, confirmed this by testing if auth was enabled after each setup - direct env value, secretRef, and configMapRef. PX started up with auth enabled and the operator authenticated with the SDK too.

drivers/storage/portworx/status.go
ggriffiths: Removed

drivers/storage/portworx/util/util.go
ggriffiths: Removed

PR #165, opened 2020-05-20
PTX-2527 fix image validation when using custom or private registry

thiguetta:


pkg/util/test/util.go
```diff
@@ -623,7 +623,7 @@ func validateImageOnPods(image, namespace string, listOptions map[string]string)
 	}
 	for _, pod := range pods.Items {
 		for _, container := range pod.Spec.Containers {
-			if container.Image != image {
+			if sanitizeImage(container.Image) != sanitizeImage(image) {
```
piyush-nimbalkar, line 626: This would mean we won't be validating that the custom registry is being used or not.
Rather add the custom repo/registry to the expected image and send it for validation.
Use this method to correctly add the repo/registry to the image before calling this validateImageOnPods function.
https://github.com/libopenstorage/operator/blob/master/pkg/util/util.go#L38

PR #168, opened 2020-06-10
Add default configuration for PX Security and key generation

ggriffiths:
Adds security component


drivers/storage/portworx/component/security.go
```diff
@@ -0,0 +1,265 @@
...
+				cluster.Spec.Env[i].ValueFrom = nil
+			}
+
+			// found and updated, exit
+			return nil
+		}
+	}
+
+	// if not found, add default jwt issuer
+	cluster.Spec.Env = append(cluster.Spec.Env, v1.EnvVar{
```
piyush-nimbalkar, line 119: Instead of setting back the env, I think we should just create and pass the env directly to the pod in deployment.go
We should use the value in selfSigned.Issuer as it will always be present.

Basically we won't need this setJwtIssuer method.

drivers/storage/portworx/portworx.go
```diff
@@ -408,6 +410,47 @@ func setNodeSpecDefaults(toUpdate *corev1alpha1.StorageCluster) {
...
+		} else if *toUpdate.Spec.Security.Enabled {
+			if toUpdate.Spec.Security.Auth != nil {
+				// security enabled and auth provided, add defaults if needed.
+				if toUpdate.Spec.Security.Auth.GuestAccess == nil {
+					toUpdate.Spec.Security.Auth.GuestAccess = defaultAuthTemplate.GuestAccess
+				}
+				if toUpdate.Spec.Security.Auth.Authenticators == nil || toUpdate.Spec.Security.Auth.Authenticators.SelfSigned == nil {
+					toUpdate.Spec.Security.Auth.Authenticators = defaultAuthTemplate.Authenticators
+				} else {
+					if toUpdate.Spec.Security.Auth.Authenticators.SelfSigned.Issuer == nil {
```
piyush-nimbalkar, line 438: Extending the comment from setJwtIssuer - 
If selfSigned.Issuer == nil, check env var and set if the value is present, else set the default.
Also let's remove the env right here so there are no duplicates in the future. That way we won't need the setJwtIssuer method.

As said before, in deployment.go where we form env vars for the pod, let's just add the issuer env from this value if security is enabled.

drivers/storage/portworx/status.go
```diff
@@ -493,18 +502,31 @@ func (p *portworx) getAuthSecret(
...
 	if err != nil {
 		return ctx, fmt.Errorf("failed to get auth secret: %v", err.Error())
 	}
 	if pxAuthSecret == "" {
 		return ctx, nil
 	}
 
-	// Generate token and add to metadata
-	token, err := p.getOperatorToken(string(pxAuthSecret))
+	// Generate token and add to metadatagetAuthSecret
```
piyush-nimbalkar, line 528: nit: Is this a typo?

pkg/apis/core/v1alpha1/storagecluster.go
```diff
@@ -107,6 +109,43 @@ type NodeSpec struct {
...
+type SecuritySpec struct {
+	Enabled *bool     `json:"enabled,omitempty"`
+	Auth    *AuthSpec `json:"auth,omitempty"`
+}
+
+// AuthSpec lets the user define authorization configurations
+// for creating a PX Security enabled cluster
+type AuthSpec struct {
+	GuestAccess    *GuestAccessType    `json:"guestAccess,omitempty"`
+	Authenticators *AuthenticatorsSpec `json:"authenticators,omitempty"`
```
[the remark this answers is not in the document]
ggriffiths: We decided to only support one self signed for now

pkg/apis/core/v1alpha1/storagecluster.go
```diff
@@ -107,6 +109,43 @@ type NodeSpec struct {
...
+}
+
+// GuestAccessType lets the user choose the
+// level of permissions the system.user has
+type GuestAccessType string
+
+var (
```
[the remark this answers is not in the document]
ggriffiths: This needs to be a var because GuestAccessType in the spec is a pointer. This is so it can be nil if the user does not provide it. When the user does not provide it, we need to set it to true by default. This is hard to do with a standard boolean

drivers/storage/portworx/component/security.go
```diff
@@ -0,0 +1,265 @@
...
+				SecretKeyRef: &v1.SecretKeySelector{
+					LocalObjectReference: v1.LocalObjectReference{
+						Name: secretName,
+					},
+					Key: SecurityPXAuthSecretKey,
+				},
+			},
+		})
+	}
+
```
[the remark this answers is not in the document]
ggriffiths: Validated this is added
```
      PORTWORX_AUTH_JWT_SHAREDSECRET:  <set to the key 'shared-secret-key' in secret 'px-private-keys'>  Optional: false
      PX_NAMESPACE:                    kube-system
      PORTWORX_AUTH_SYSTEM_KEY:        <set to the key 'system-secret-key' in secret 'px-private-keys'>  Optional: false
```

drivers/storage/portworx/portworx.go
ggriffiths: Fixed this

pkg/apis/core/v1alpha1/storagecluster.go
```diff
@@ -107,6 +109,43 @@ type NodeSpec struct {
...
+	// GuestRoleEnabled sets the guest role to the default volume lifecycle permissions
+	GuestRoleEnabled = GuestAccessType("enabled")
```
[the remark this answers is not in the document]
ggriffiths: Removed guest role management for now.

pkg/apis/core/v1alpha1/storagecluster.go
```diff
@@ -107,6 +109,43 @@ type NodeSpec struct {
 	CommonConfig
 }
 
+// SecuritySpec is used to define the security configuration for a cluster.
+type SecuritySpec struct {
+	Enabled *bool     `json:"enabled,omitempty"`
```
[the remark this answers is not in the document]
ggriffiths: fixed

drivers/storage/portworx/component/security.go
ggriffiths: Fixed this

drivers/storage/portworx/component/security.go
```diff
@@ -0,0 +1,212 @@
...
+		SecuritySystemSecretKey,
+	)
+	if err != nil {
+		return err
+	}
+
+	privateKeysSecret := &v1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      SecurityPXPrivateKeysSecretName,
+			Namespace: cluster.Namespace,
```
piyush-nimbalkar, line 141: Any reason we are not adding owner reference here? 
Is the intention to not delete the secret when storage cluster is deleted?

drivers/storage/portworx/component/security.go
```diff
@@ -0,0 +1,212 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	// SecurityComponentName is the name for registering this component
+	SecurityComponentName = "Security"
+	// SecurityPXPrivateKeysSecretName is the admin secret name for PX security
+	SecurityPXPrivateKeysSecretName = "px-private-keys"
+	// SecuritySharedSecretKey is the data key for any secret containing an auth secret
+	SecuritySharedSecretKey = "shared-secret"
```
piyush-nimbalkar, line 25: Just wondering whether we should use `PORTWORX_AUTH_SYSTEM_KEY` and `PORTWORX_AUTH_JWT_SHAREDSECRET` as the keys even in the secret. That way there are not too many names for the same thing. 
What do you think?

drivers/storage/portworx/component/security.go
```diff
@@ -0,0 +1,212 @@
...
+	"k8s.io/client-go/tools/record"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	// SecurityComponentName is the name for registering this component
+	SecurityComponentName = "Security"
+	// SecurityPXPrivateKeysSecretName is the admin secret name for PX security
+	SecurityPXPrivateKeysSecretName = "px-private-keys"
+	// SecuritySharedSecretKey is the data key for any secret containing an auth secret
```
piyush-nimbalkar, line 24: nit: Is the comment still correct?

drivers/storage/portworx/component/security.go
```diff
@@ -0,0 +1,212 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/record"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	// SecurityComponentName is the name for registering this component
+	SecurityComponentName = "Security"
+	// SecurityPXPrivateKeysSecretName is the admin secret name for PX security
+	SecurityPXPrivateKeysSecretName = "px-private-keys"
```
piyush-nimbalkar, line 23: Can we call this something different? Not sure what but does not convey that this has secrets for auth. We can talk about this offline.

drivers/storage/portworx/components_test.go
```diff
@@ -2662,6 +2662,41 @@ func TestAutopilotInvalidCPU(t *testing.T) {
 		fmt.Sprintf("%v %v Failed to setup Autopilot.", v1.EventTypeWarning, util.FailedComponentReason))
 }
 
+func TestSecurityInstall(t *testing.T) {
```
piyush-nimbalkar, line 2665: Can we add couple more tests here - 
- When there are env variables already present in StorageCluster, ensure secrets don't have auto-generated values, but contain the ones from env variables.
- Disable security and verify that the secret is deleted (What would happen in this case? Can we really change PX to a non-secure mode? If not, do we want to delete the secrets?)

drivers/storage/portworx/components_test.go
```diff
@@ -2662,6 +2662,41 @@ func TestAutopilotInvalidCPU(t *testing.T) {
...
+	k8sClient := testutil.FakeK8sClient()
+	driver := portworx{}
+	driver.Init(k8sClient, runtime.NewScheme(), record.NewFakeRecorder(0))
+
+	cluster := &corev1alpha1.StorageCluster{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "px-cluster",
+			Namespace: "kube-test",
+			Annotations: map[string]string{
+				annotationPVCController: "true",
```
piyush-nimbalkar, line 2677: nit: No need for this annotation.

drivers/storage/portworx/portworx_test.go
```diff
@@ -829,6 +829,55 @@ func TestStorageClusterDefaultsForNodeSpecs(t *testing.T) {
...
+	driver.SetDefaultsOnStorageCluster(cluster)
+	require.NotNil(t, cluster.Spec.Security)
+	require.Equal(t, true, cluster.Spec.Security.Enabled)
+	require.NotNil(t, true, cluster.Spec.Security.Auth.Authenticators.SelfSigned.Issuer)
+	require.Equal(t, "portworx.io", *cluster.Spec.Security.Auth.Authenticators.SelfSigned.Issuer)
+	require.NotNil(t, true, cluster.Spec.Security.Auth.Authenticators.SelfSigned.TokenLifetime)
+	require.Equal(t, metav1.Duration{Duration: 24 * time.Hour}, *cluster.Spec.Security.Auth.Authenticators.SelfSigned.TokenLifetime)
+
+	// issuer, when manually set, is not overwritten.
+	cluster.Spec.Security.Auth.Authenticators.SelfSigned.Issuer = stringPtr("myissuer.io")
```
piyush-nimbalkar, line 870: Let's add the following test cases as well:
- spec.auth is not nil but empty
- spec.auth.authenticators is not nil but empty
- spec.auth.authenticators.selfSigned is not nil but empty
- issuer is not set but the issue env is present in StorageCluster.

Tip: I generally check codecov to see if there is any untested code path. It's quite handy.
https://codecov.io/gh/libopenstorage/operator/pull/168/diff?src=pr&el=tree#diff-ZHJpdmVycy9zdG9yYWdlL3BvcnR3b3J4L3BvcnR3b3J4Lmdv

drivers/storage/portworx/status.go
```diff
@@ -79,15 +73,15 @@ func (p *portworx) UpdateStorageClusterStatus(
 	cluster.Status.ClusterName = pxCluster.Cluster.Name
 	cluster.Status.ClusterUID = pxCluster.Cluster.Id
 
-	return p.updateStorageNodes(clientConn, cluster)
+	return p.updateStorageNodes(p.sdkConn, cluster)
```
piyush-nimbalkar, line 76: nit: Might as well not pass `p.sdkConn` as we have access to `p.sdkConn` inside updateStorageNodes now.

drivers/storage/portworx/util/util.go
```diff
@@ -304,3 +333,155 @@ func GetValueFromEnv(ctx context.Context, client client.Client, envVar *v1.EnvVa
...
+func GetAdminSecret(
+	ctx context.Context,
+	cluster *corev1alpha1.StorageCluster,
+	k8sClient client.Client,
+) (string, error) {
+	var authSecret string
+	var err error
+
+	// check for provided secret
+	for _, envVar := range cluster.Spec.Env {
```
piyush-nimbalkar, line 444: Even here we should get the shared secret from the k8s secret that operator created (either generated or from user provided env value).

drivers/storage/portworx/util/util.go
```diff
@@ -254,8 +271,20 @@ func StorageClusterKind() schema.GroupVersionKind {
 	return corev1alpha1.SchemeGroupVersion.WithKind("StorageCluster")
 }
 
-// GetValueFromEnv returns the value of v1.EnvVar Value or ValueFrom
-func GetValueFromEnv(ctx context.Context, client client.Client, envVar *v1.EnvVar, namespace string) (string, error) {
+// GetClusterEnvVarValue returns the environment variable value for a cluster.
+// Note: This does not
```
piyush-nimbalkar, line 275: nit: Incomplete comment?

drivers/storage/portworx/component/security.go
ggriffiths: @piyush-nimbalkar I was following the keys in px-docs example here:
https://docs.portworx.com/cloud-references/security/kubernetes/shared-secret-model-multitenancy/generating-tokens/

I think these keys are fine, since the user won't really be pulling these often. Having lowercase keys inside of the k8s secret also matches the `auth-token` key we have.

drivers/storage/portworx/components_test.go
ggriffiths: Just checked and the secret is not deleted when you disable security.

drivers/storage/portworx/components_test.go
```diff
@@ -2662,6 +2662,76 @@ func TestAutopilotInvalidCPU(t *testing.T) {
...
+	require.Equal(t, sharedSecretKey, securitySecret.StringData[component.SecuritySharedSecretKey])
+
+	// Disable security, secret should not be deleted.
+	cluster.Spec.Security.Enabled = false
+	err = driver.PreInstall(cluster)
+	require.NoError(t, err)
+	err = testutil.Get(k8sClient, securitySecret, pxutil.SecurityPXAuthKeysSecretName, cluster.Namespace)
+	require.NoError(t, err)
+
+	// Delete cluster (cluster wipe), secret should be deleted.
```
piyush-nimbalkar, line 2725: In this case secret should not be deleted (the assertion is correct, just the comment is incorrect)
Let's add a case where the deleteStrategy is set and ensure that the secret is deleted.

drivers/storage/portworx/deployment_test.go
```diff
@@ -2304,6 +2304,51 @@ func TestStorageNodeConfig(t *testing.T) {
...
+	for _, env := range actual.Containers[0].Env {
+		if env.Name == pxutil.EnvKeyPortworxAuthJwtIssuer {
+			jwtIssuer = env.Value
+			break
+		}
+	}
+	assert.Equal(t, expectedJwtIssuer, jwtIssuer)
+
+	expectedJwtIssuer = "test.io"
+	cluster.Spec.Security.Auth.Authenticators.SelfSigned.Issuer = stringPtr(expectedJwtIssuer)
```
piyush-nimbalkar, line 2340: Can we also test that the other two envs are also added from the secret?

pkg/util/k8s/k8s.go
```diff
@@ -1383,3 +1383,79 @@ func removeOwners(current, toBeDeleted []metav1.OwnerReference) []metav1.OwnerRe
...
+		existingSecret,
+	)
+	if errors.IsNotFound(err) {
+		logrus.Debugf("Creating %s/%s Secret", secret.Namespace, secret.Name)
+		return k8sClient.Create(context.TODO(), secret)
+	} else if err != nil {
+		return err
+	}
+
+	modified := !reflect.DeepEqual(secret.Data, existingSecret.Data)
```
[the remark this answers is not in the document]
ggriffiths: From https://github.com/libopenstorage/operator/blob/master/vendor/k8s.io/api/core/v1/types.go#L4986-L4992:

	// stringData allows specifying non-binary secret data in string form.
	// It is provided as a write-only convenience method.
	// All keys and values are merged into the data field on write, overwriting any existing values.
	// It is never output when reading from the API.

Looks like we only need to read from Data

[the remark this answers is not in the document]
piyush-nimbalkar: Ahh, good to know!

PR #171, opened 2020-06-21
Not own cross namespaces objects by storagecluster

piyush-nimbalkar:
- As per k8s docs - https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/ cross namespace ownership is disallowed. So removing all object ownership on cluster-scoped and other namespace objects.
- Removing ownership on already installed objects 
- Delete all object during cluster deletion. Mostly need to explicitly delete cross namespace objects only because they don't have ownership, but it is easier to deleted all objects once.
- Delete px storageclasses only when portworx is uninstalled

### DIFF
--- a/drivers/storage/portworx/component/csi.go
+++ b/drivers/storage/portworx/component/csi.go
@@ -74,17 +74,17 @@ func (c *csi) Reconcile(cluster *corev1alpha1.StorageCluster) error {
 	if err := c.createServiceAccount(cluster.Namespace, ownerRef); err != nil {
 		return err
 	}
-	if err := c.createClusterRole(cluster, csiConfig, ownerRef); err != nil {
+	if err := c.createClusterRole(cluster, csiConfig); err != nil {
 		return err
 	}
-	if err := c.createClusterRoleBinding(cluster, ownerRef); err != nil {
+	if err := c.createClusterRoleBinding(cluster); err != nil {
 		return err
 	}
 	if err := c.createService(cluster, ownerRef); err != nil {
 		return err
 	}
 	if csiConfig.IncludeCsiDriverInfo {
-		if err := c.createCSIDriver(csiConfig, ownerRef); err != nil {
+		if err := c.createCSIDriver(csiConfig); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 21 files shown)


pkg/util/k8s/k8s.go
```diff
@@ -358,31 +351,21 @@ func DeleteClusterRole(
...
 
-	newOwners := removeOwners(clusterRole.OwnerReferences, owners)
-
-	// Do not delete the object if it does not have the owner that was passed;
-	// even if the object has no owner
-	if (len(clusterRole.OwnerReferences) == 0 && len(owners) > 0) ||
-		(len(clusterRole.OwnerReferences) > 0 && len(clusterRole.OwnerReferences) == len(newOwners)) {
+	// Do not delete the object if it has an owner
+	if len(clusterRole.OwnerReferences) > 0 {
 		logrus.Debugf("Cannot delete ClusterRole %s as it is not owned", name)
```
adityadani, line 356: nit: cannot delete as it is owned

pkg/util/k8s/k8s.go
```diff
@@ -431,23 +412,14 @@ func DeleteClusterRoleBinding(
...
 
-	newOwners := removeOwners(crb.OwnerReferences, owners)
-
-	// Do not delete the object if it does not have the owner that was passed;
-	// even if the object has no owner
-	if (len(crb.OwnerReferences) == 0 && len(owners) > 0) ||
-		(len(crb.OwnerReferences) > 0 && len(crb.OwnerReferences) == len(newOwners)) {
+	// Do not delete the object if it has an owner
+	if len(crb.OwnerReferences) > 0 {
 		logrus.Debugf("Cannot delete ClusterRoleBinding %s as it is not owned", name)
```
adityadani, line 417: same here and other places in this file

drivers/storage/portworx/component/csi.go
```diff
@@ -114,12 +114,13 @@ func (c *csi) Reconcile(cluster *corev1alpha1.StorageCluster) error {
 
 func (c *csi) Delete(cluster *corev1alpha1.StorageCluster) error {
 	ownerRef := metav1.NewControllerRef(cluster, pxutil.StorageClusterKind())
-	// We don't delete the service account for CSI because it is part of CSV. If
-	// we disable CSI then the CSV upgrades would fail as requirements are not met.
-	if err := k8sutil.DeleteClusterRole(c.k8sClient, CSIClusterRoleName, *ownerRef); err != nil {
+	if err := k8sutil.DeleteServiceAccount(c.k8sClient, CSIServiceAccountName, cluster.Namespace, *ownerRef); err != nil {
```
adityadani, line 117: Does the older comment not hold true anymore?

piyush-nimbalkar: In 1.3 we removed the service accounts (lighthouse, pvccontroller, csi) from the operator CSV as it was needed there and started handling deletion of SA ourselves. We missed SA for csi.

PR #172, opened 2020-06-23
Add auto generation and refresh of admin/user tokens

ggriffiths:

### DIFF
--- a/drivers/storage/portworx/component/security.go
+++ b/drivers/storage/portworx/component/security.go
@@ -4,14 +4,20 @@ import (
 	"context"
 	"crypto/rand"
 	"encoding/base64"
+	"fmt"
+	"time"
 
+	"github.com/dgrijalva/jwt-go"
 	"github.com/hashicorp/go-version"
+	"github.com/libopenstorage/openstorage/pkg/auth"
 	pxutil "github.com/libopenstorage/operator/drivers/storage/portworx/util"
 	corev1alpha1 "github.com/libopenstorage/operator/pkg/apis/core/v1alpha1"
 	k8sutil "github.com/libopenstorage/operator/pkg/util/k8s"
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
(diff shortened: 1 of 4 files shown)


drivers/storage/portworx/component/security.go
```diff
@@ -138,10 +185,9 @@ func (c *security) createPrivateKeysSecret(
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      pxutil.SecurityPXAuthKeysSecretName,
 			Namespace: cluster.Namespace,
-		},
-		StringData: map[string]string{
-			SecuritySharedSecretKey: sharedSecretKey,
-			SecuritySystemSecretKey: internalSystemSecretKey,
+		}, Data: map[string][]byte{
```
ggriffiths, line 188: This makes testing easier. It seems our fake driver doesn't understand StringData vs. Data, so sticking to one field helps this confusion.

drivers/storage/portworx/component/security.go
```diff
@@ -138,10 +185,9 @@ func (c *security) createPrivateKeysSecret(
...
+			SecuritySharedSecretKey: []byte(sharedSecretKey),
+			SecuritySystemSecretKey: []byte(internalSystemSecretKey),
```
piyush-nimbalkar, line 190: We will be separating these into two secrets later, right?

drivers/storage/portworx/util/util.go
```diff
@@ -508,7 +482,15 @@ func SetupContextWithToken(ctx context.Context, cluster *corev1alpha1.StorageClu
...
-	token, err := GetOperatorToken(cluster, string(pxAuthSecret))
+	claims := &auth.Claims{
+		Issuer:  *cluster.Spec.Security.Auth.Authenticators.SelfSigned.Issuer,
+		Subject: "operator@portworx.io",
+		Name:    "operator communications",
+		Email:   "operator@portworx.io",
+		Roles:   []string{"system.admin"},
+		Groups:  []string{"*"},
+	}
+	token, err := GenerateToken(cluster, pxAuthSecret, claims)
```
piyush-nimbalkar, line 493: Should we be using the admin token from the secret instead of creating a new one?

drivers/storage/portworx/components_test.go
```diff
@@ -2677,43 +2679,90 @@ func TestSecurityInstall(t *testing.T) {
...
+	adminSecret := &v1.Secret{}
+	err = testutil.Get(k8sClient, adminSecret, component.SecurityPXAdminTokenSecretName, cluster.Namespace)
+	require.NoError(t, err)
+	_, _, err = new(jwt.Parser).ParseUnverified(string(adminSecret.Data[component.SecurityAuthTokenKey]), &jwt.StandardClaims{})
+	require.NoError(t, err)
+
+	userSecret := &v1.Secret{}
+	err = testutil.Get(k8sClient, userSecret, component.SecurityPXUserTokenSecretName, cluster.Namespace)
+	require.NoError(t, err)
+	_, _, err = new(jwt.Parser).ParseUnverified(string(userSecret.Data[component.SecurityAuthTokenKey]), &jwt.StandardClaims{})
```
piyush-nimbalkar, line 2729: Should we check the contents of the token to ensure we are setting the right values?
Things like expiry, role, groups, etc?

drivers/storage/portworx/component/security.go
ggriffiths: Yes I'll do this in a separate PR

drivers/storage/portworx/util/util.go
ggriffiths: Yeah, for now we can use that. Once we bring in the apps issuer change though, we'll probably store the token in memory per app.

ggriffiths:
Ready for re-review.

drivers/storage/portworx/util/util.go
```diff
@@ -499,21 +473,36 @@ func SetupContextWithToken(ctx context.Context, cluster *corev1alpha1.StorageClu
 		return ctx, nil
 	}
 
-	pxAuthSecret, err := GetAdminSecret(ctx, cluster, k8sClient)
+	pxAdminToken, err := GetSecretValue(ctx, cluster, k8sClient, "px-admin-token", "auth-token")
```
piyush-nimbalkar, line 476: nit: Can we use the constants for px-admin-token and auth-token?

ggriffiths: Will address in a future PR

PR #176, opened 2020-07-01
Add option to run stork in host network

piyush-nimbalkar:
Adding a flag under stork section in the StorageCluster spec to toggle host network for stork pods. 
From what I understand, this is needed for px-backup to work in AWS.
```
spec:
  stork:
    enabled: true
    hostNetwork: true
```

### DIFF
--- a/pkg/controller/storagecluster/stork.go
+++ b/pkg/controller/storagecluster/stork.go
@@ -520,6 +520,7 @@ func (c *Controller) createStorkDeployment(
 		!reflect.DeepEqual(existingCommand, command) ||
 		!reflect.DeepEqual(existingEnvs, envVars) ||
 		existingCPUQuantity.Cmp(targetCPUQuantity) != 0 ||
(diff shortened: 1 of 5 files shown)


piyush-nimbalkar:
@adityadani @disrani-px @harsh-px Can one of you review this and check if the `hostNetwork` option seems fine in the stork section?

pkg/controller/storagecluster/stork.go
```diff
@@ -635,6 +636,10 @@ func (c *Controller) getStorkDeploymentSpec(
 		)
 	}
 
+	if cluster.Spec.Stork.HostNetwork != nil {
```
disrani-px, line 639: If that is nil (ie been removed from the storagecluster spec) we should reset it to `false` in the deployment

piyush-nimbalkar: The default value of deployment's hostNetwork is false. On line 523 when we detect the previous value does not match current (nil), we redeploy the spec. It will result in having a deployment spec with host network as false (default). 
Will update the test to cover `true -> nil` scenario.

piyush-nimbalkar:
Thanks for the review @disrani-px!

PR #177, opened 2020-07-01
Split px-auth-keys into two secrets, move constants, remove authenticators

ggriffiths:
I broke this up into 3 commits if you want to look at the changes individually.

Nothing functionally changed other than splitting the secret into two and managing separately.

I spoke with @lpabon and we decided that authenticators is redundant and can be removed. Open to your thoughts on this one too.


pkg/apis/core/v1alpha1/storagecluster.go
```diff
@@ -118,18 +118,21 @@ type SecuritySpec struct {
...
 type SelfSignedSpec struct {
-	Issuer        *string        `json:"issuer,omitempty"`
-	TokenLifetime *meta.Duration `json:"tokenLifetime,omitempty"`
+	Issuer        *string             `json:"issuer,omitempty"`
+	TokenLifetime *meta.Duration      `json:"tokenLifetime,omitempty"`
+	SharedSecret  *SecretKeyReference `json:"sharedSecretRef,omitempty"`
+}
+
+// SecretKeyReference is a user-friendly reference to a value based on a secret name and key
+type SecretKeyReference struct {
```
ggriffiths, line 132: I made this new object instead of using the k8s built in ones. I think having a custom one with secretName/secretNamespace would be better.

piyush-nimbalkar: I think we should only take the secret name, just like we do for image pull secret and kvdb secret. Kubernetes does the same for image pull secrets in pods. Just trying to keep the spec as simple as possible.
We can mandate the secret to be present in the StorageCluster's namespace. 
For the key, we can assume a constant which the operator knows and use the same in our docs. If we don't find the correct key in the secret we can raise an event, that way users know what is the expected key in-case they created the secret incorrectly.
For most cases the operator would be creating the secret, so the key should be the one expected.

drivers/storage/portworx/util/util.go
```diff
@@ -99,8 +99,20 @@ const (
 	// EnvKeyPortworxAuthJwtIssuer is an environment variable defining the PX Security JWT Issuer
 	EnvKeyPortworxAuthJwtIssuer = "PORTWORX_AUTH_JWT_ISSUER"
 
-	// SecurityPXAuthKeysSecretName is the admin secret name for PX security
-	SecurityPXAuthKeysSecretName = "px-auth-keys"
+	// SecurityPXSystemSecretsSecretName is the secret name for PX security system secrets
+	SecurityPXSystemSecretsSecretName = "px-system-secrets"
```
piyush-nimbalkar, line 103: Any reason this is plural while the other is not?

piyush-nimbalkar:
Removing the authenticator seems fine to me. It's better to have a simple spec.

drivers/storage/portworx/util/util.go
ggriffiths: @piyush-nimbalkar I was thinking this would include the apps issuer secret key as well.

pkg/apis/core/v1alpha1/storagecluster.go
ggriffiths: Ah yeah good point. That simplifies things a lot.

ggriffiths:
Ready for re-review. Tested w/ a custom sharedSecret:

spec:
```
    security:
        enabled: true
        auth:
            selfSigned:
                tokenLifetime: '10m'
                issuer: "portworx.io"
                sharedSecret: "gg-secret"
```

result:
```
[grant@PDC4-SM26-N8 stc]$ kn describe pod portworx-pnl6x | grep gg-secret
      PORTWORX_AUTH_JWT_SHAREDSECRET:  <set to the key 'shared-secret' in secret 'gg-secret'>          Optional: false
```

drivers/storage/portworx/util/util.go
piyush-nimbalkar: Oh right, sounds good then.

PR #181, opened 2020-07-06
Add management of system.guest RBAC

ggriffiths:


Makefile
```diff
@@ -120,6 +120,11 @@ get-release-manifest: clean-release-manifest
 	mkdir -p manifests
 	wget -q '$(PX_INSTALLER_HOST)/versions' -O manifests/portworx-releases-local.yaml
 
+mockgen:
```
piyush-nimbalkar, line 123: Thanks for adding this!

drivers/storage/portworx/util/util.go
```diff
@@ -376,6 +376,9 @@ func GetPortworxConn(sdkConn *grpc.ClientConn, k8sClient client.Client, namespac
 
 	// Get the ports from service
 	for _, pxServicePort := range pxService.Spec.Ports {
+		// if pxServicePort.Name == PortworxSDKPortName && pxServicePort.TargetPort.IntValue() != 0 {
```
piyush-nimbalkar, line 379: nit: Remove this

pkg/apis/core/v1alpha1/storagecluster.go
```diff
@@ -128,6 +129,19 @@ type SelfSignedSpec struct {
 	SharedSecret  *string        `json:"sharedSecret,omitempty"`
 }
 
+// GuestAccessType lets the user choose the
+// level of permissions the system.user has
+type GuestAccessType string
+
+var (
```
piyush-nimbalkar, line 136: I feel we should keep this const so there are no accidental changes to the value of the enums at runtime.
We can always a pointer to the copy of this constant. That's what is done in kubernetes for types like HostPathType or MountPropagationMode in volumes.

drivers/storage/portworx/portworx.go
```diff
@@ -456,6 +457,10 @@ func setSecuritySpecDefaults(toUpdate *corev1alpha1.StorageCluster) {
 	if toUpdate.Spec.Security != nil {
 		if toUpdate.Spec.Security.Enabled {
 			if toUpdate.Spec.Security.Auth != nil && (*toUpdate.Spec.Security.Auth != corev1alpha1.AuthSpec{}) {
+				if toUpdate.Spec.Security.Auth.GuestAccess == nil || (*toUpdate.Spec.Security.Auth.GuestAccess == "") {
```
piyush-nimbalkar, line 460: Can we have a test for these conditions? May be just under the existing test `TestStorageClusterDefaultsForSecurity` in portworx_test.go

drivers/storage/portworx/components_test.go
```diff
@@ -2794,6 +2803,128 @@ func TestSecurityInstall(t *testing.T) {
...
+	}
+	// Initial run
+	setSecuritySpecDefaults(cluster)
+	securityComponent, ok := component.Get(component.SecurityComponentName)
+	require.True(t, ok)
+	component.DeregisterAllComponents()
+	component.Register(component.SecurityComponentName, securityComponent)
+	securityComponent.Initialize(k8sClient, hversion.Version{}, runtime.NewScheme(), record.NewFakeRecorder(10))
+	driver.UpdateStorageClusterStatus(cluster)
+	require.NoError(t, err)
```
piyush-nimbalkar, line 2876: Do we need any of this? Why do we need to call UpdateStorageClusterStatus? I guess, if we don't call that, we don't even need Node and Cluster services.

drivers/storage/portworx/components_test.go
```diff
@@ -2794,6 +2803,128 @@ func TestSecurityInstall(t *testing.T) {
...
+		Return(&osdapi.SdkRoleUpdateResponse{
+			Role: &component.GuestRoleDisabled,
+		}, nil).
+		Times(1)
+	err = driver.PreInstall(cluster)
+	require.NoError(t, err)
+
+	// Enable guest access, should be updated again
+	cluster.Spec.Security.Auth.GuestAccess = &corev1alpha1.GuestRoleEnabled
+	expectedRole = component.GuestRoleEnabled
```
piyush-nimbalkar, line 2904: Shouldn't this be GuestRoleDisabled? 
Also how is the test passing of both expected and desired are the same?

drivers/storage/portworx/components_test.go
```diff
@@ -2794,6 +2803,128 @@ func TestSecurityInstall(t *testing.T) {
...
+		}, nil).
+		Times(1)
+	err = driver.PreInstall(cluster)
+	require.NoError(t, err)
+
+	// set to managed to avoid more calls without a corresponding mock expect
+	cluster.Spec.Security.Auth.GuestAccess = &corev1alpha1.GuestRoleManaged
+	err = driver.PreInstall(cluster)
+	require.NoError(t, err)
+	// no expected mock calls
```
piyush-nimbalkar, line 2927: Can we add a test case for invalid GuestRole and see if an event is raised for it?

drivers/storage/portworx/component/security.go
```diff
@@ -324,6 +372,70 @@ func (c *security) deleteSecret(
...
+		return fmt.Errorf("invalid guest access type: %s", *cluster.Spec.Security.Auth.GuestAccess)
+	}
+
+	// managed, do not interfere with system.guest role
+	if *cluster.Spec.Security.Auth.GuestAccess == corev1alpha1.GuestRoleManaged {
+		return nil
+	}
+
+	var err error
+	c.sdkConn, err = pxutil.GetPortworxConn(c.sdkConn, c.k8sClient, cluster.Namespace)
```
piyush-nimbalkar, line 400: In this method, the default grpcserver connection waits for max 1 min for the connection to be ready. That means during the initial install, the operator will get stuck for a 1 min and not proceed with other component installation. That's because the SDK server is not yet up.
So far it was fine because the only thing dependent on SDK server was status update, which is not critical compared to deploying other components.

Can we add a new method on openstorage to take connection timeout from the client and use that in GetPortworxConn? This is out of the scope of this PR, but we will have to do it in this release.

https://github.com/libopenstorage/openstorage/blob/master/pkg/grpcserver/grpcutil.go#L59

drivers/storage/portworx/components_test.go
ggriffiths: Yeah, the test was failing with other components running. I think due to the mock server. I just wanted to isolate this test to only run the security operator.

ggriffiths: Just removed the node/cluster servers as we don't need them.

drivers/storage/portworx/components_test.go
ggriffiths: In this section, I'm setting GuestRoleEnabled in the cluster spec, and I'm expecting that during a PreInstall, we will make a call to the Mock server with the GuestRoleEnabled RBAC.

drivers/storage/portworx/components_test.go
ggriffiths: Added this

drivers/storage/portworx/component/security.go
ggriffiths: Created an issue for this.

drivers/storage/portworx/portworx.go
ggriffiths: Added a test for default guest access values

pkg/apis/core/v1alpha1/storagecluster.go
ggriffiths: Fixed

drivers/storage/portworx/util/util.go
ggriffiths: Removed

drivers/storage/portworx/portworx_test.go
```diff
@@ -912,6 +912,22 @@ func TestStorageClusterDefaultsForSecurity(t *testing.T) {
...
+	}
+	driver.SetDefaultsOnStorageCluster(cluster)
+	cluster.Spec.Security = &corev1alpha1.SecuritySpec{
+		Enabled: true,
+		Auth: &corev1alpha1.AuthSpec{
+			GuestAccess: nil,
+		},
+	}
+	driver.SetDefaultsOnStorageCluster(cluster)
+	assertDefaultSecuritySpec(t, cluster)
```
piyush-nimbalkar, line 929: We will have to add default guest access assertion  to the `assertDefaultSecuritySpec`

PR #182, opened 2020-07-07
Use prometheus images from version manifest

piyush-nimbalkar:

### DIFF
--- a/drivers/storage/portworx/portworx.go
+++ b/drivers/storage/portworx/portworx.go
@@ -189,6 +189,16 @@ func (p *portworx) SetDefaultsOnStorageCluster(toUpdate *corev1alpha1.StorageClu
 			toUpdate.Status.DesiredImages.CSIResizer = release.Components.CSIResizer
 			toUpdate.Status.DesiredImages.CSISnapshotter = release.Components.CSISnapshotter
 		}
+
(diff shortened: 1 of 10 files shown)


drivers/storage/portworx/portworx.go
```diff
@@ -574,6 +594,13 @@ func hasCSIChanged(cluster *corev1alpha1.StorageCluster) bool {
 		cluster.Status.DesiredImages.CSIProvisioner == ""
 }
 
+func hasPrometheusChanged(cluster *corev1alpha1.StorageCluster) bool {
+	return cluster.Spec.Monitoring != nil &&
+		cluster.Spec.Monitoring.Prometheus != nil &&
+		cluster.Spec.Monitoring.Prometheus.Enabled &&
+		cluster.Status.DesiredImages.PrometheusOperator == ""
```
adityadani, line 601: So basically we will only allow overriding PrometheusOperator image and no other prometheus components?

piyush-nimbalkar: No we will be overriding all prometheus images. This just checks one of the image because when prometheus is disabled we remove all the images from Status.DesiredImages. So in future if we enable it and find that any prometheus images is not present we will populate the DesiredImages.Prometheus* images.
Basically we can check any one of the images.

PR #184, opened 2020-07-07
Add support for stork and apps secrets

ggriffiths:

### DIFF
--- a/drivers/storage/portworx/components_test.go
+++ b/drivers/storage/portworx/components_test.go
@@ -2711,11 +2711,15 @@ func TestSecurityInstall(t *testing.T) {
 	systemSecret := &v1.Secret{}
 	err = testutil.Get(k8sClient, systemSecret, pxutil.SecurityPXSystemSecretsSecretName, cluster.Namespace)
 	require.NoError(t, err)
-	require.Empty(t, sharedSecret.OwnerReferences)
+	require.Empty(t, systemSecret.OwnerReferences)
 	require.Equal(t, 64, len(systemSecret.Data[pxutil.SecuritySystemSecretKey]))
 	oldSystemSecret := systemSecret.Data[pxutil.SecuritySystemSecretKey]
+	require.Equal(t, 64, len(systemSecret.Data[pxutil.SecurityAppsSecretKey]))
+	oldAppsSecret := systemSecret.Data[pxutil.SecurityAppsSecretKey]
 
 	require.NotEqual(t, systemSecret.Data[pxutil.SecuritySystemSecretKey], sharedSecret.Data[pxutil.SecuritySharedSecretKey])
+	require.NotEqual(t, sharedSecret.Data[pxutil.SecuritySharedSecretKey], systemSecret.Data[pxutil.SecurityAppsSecretKey])
+	require.NotEqual(t, systemSecret.Data[pxutil.SecurityAppsSecretKey], systemSecret.Data[pxutil.SecuritySystemSecretKey])
 
 	// No changes should happen to the auto-generated secrets
 	err = driver.PreInstall(cluster)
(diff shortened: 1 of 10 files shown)


drivers/storage/portworx/portworx_test.go
```diff
@@ -94,15 +94,28 @@ func TestGetStorkEnvList(t *testing.T) {
...
 	require.Equal(t, pxutil.EnvKeyPortworxServiceName, envVars[1].Name)
 	require.Equal(t, component.PxAPIServiceName, envVars[1].Value)
+	require.Equal(t, pxutil.SecurityPXSystemSecretsSecretName, envVars[2].ValueFrom.SecretKeyRef.Name)
+	require.Equal(t, pxutil.SecurityAppsSecretKey, envVars[2].ValueFrom.SecretKeyRef.Key)
+
+	cluster.Spec.Image = "portworx/image:2.5.0"
+	envVars = driver.GetStorkEnvList(cluster)
+	require.Equal(t, pxutil.EnvKeyStorkPXSharedSecret, envVars[2].Name)
+	require.Equal(t, pxutil.SecurityPXSystemSecretsSecretName, envVars[2].ValueFrom.SecretKeyRef.Name)
+	require.Equal(t, pxutil.SecurityAppsSecretKey, envVars[2].ValueFrom.SecretKeyRef.Key)
```
piyush-nimbalkar, line 118: Can we add a quick test to see only two envs are passed (without stork shared secret) when security is disabled?

drivers/storage/portworx/components_test.go
```diff
@@ -2791,6 +2797,10 @@ func TestSecurityInstall(t *testing.T) {
 		Name:  pxutil.EnvKeyPortworxAuthJwtSharedSecret,
 		Value: string(sharedSecretKey),
 	})
+	cluster.Spec.Env = append(cluster.Spec.Env, v1.EnvVar{
```
piyush-nimbalkar, line 2800: nit: Can we duplicate this case and also check that if user has given only one env (either system-secret or apps-secret), then we don't overwrite the other? I see from the code we do handle that, but just to make sure we don't regress in future and overwrite one because of the other.

pkg/controller/storagecluster/update.go
```diff
@@ -584,6 +584,8 @@ func matchSelectedFields(
 		return false, nil
 	} else if !reflect.DeepEqual(oldSpec.RuntimeOpts, currentSpec.RuntimeOpts) {
 		return false, nil
+	} else if !reflect.DeepEqual(oldSpec.Security, currentSpec.Security) {
```
piyush-nimbalkar, line 587: For security spec we will have to handle this a bit differently/selectively.
For instance, we should not restart the pods on guestMode change or token lifetime change. I think for the rest of the spec we will have to restart the pods.
May be just extract this out in a method just like `isEnvEqual`.

piyush-nimbalkar: Also add a test similar to `TestUpdateStorageCluster**` for security spec change.

pkg/controller/storagecluster/controller_test.go
```diff
@@ -4842,6 +4842,163 @@ func TestUpdateStorageClusterShouldRestartPodIfItsHistoryHasInvalidSpec(t *testi
...
+	k8sClient.Update(context.TODO(), cluster)
+
+	podControl.DeletePodName = nil
+
+	result, err = controller.Reconcile(request)
+	require.NoError(t, err)
+	require.Empty(t, result)
+	require.Equal(t, []string{oldPod.Name}, podControl.DeletePodName)
+
+	// TestCase: no change, no pod to delete.
```
piyush-nimbalkar, line 4978: Similar test for GuestAccessMode as well?

ggriffiths: Ah, yea will add that.

PR #185, opened 2020-07-08
Add an option to deploy essentials only cluster

piyush-nimbalkar:
- Operator takes `PORTWORX_ESSENTIALS` env var which tells it to only deploy an essentials cluster
- Pass `-marketplace_name` argument to Portworx (>= 2.5.5) which operator gets from `MARKETPLACE_NAME` env var
- Raise a warning event on StorageCluster object if `px-essential` secret is not present. Do not deploy portworx pods until the secret is created.

Here are some sample error events wrt the secret - 
```
Events:
  Type     Reason      Age                   From                       Message
  ----     ------      ----                  ----                       -------
  Warning  FailedSync  60s (x2 over 61s)     storagecluster-controller  pre-install hook failed: secret kube-system/px-essential does not have Essentials Entitlement ID (px-essen-user-id)
  Warning  FailedSync  30s (x20 over 2m43s)  storagecluster-controller  pre-install hook failed: secret kube-system/px-essential should be present to deploy a Portworx Essentials cluster
  Warning  FailedSync  0s (x2 over 1s)       storagecluster-controller  pre-install hook failed: secret kube-system/px-essential does not have Portworx OSB endpoint (px-osb-endpoint)
```

### DIFF
--- a/drivers/storage/portworx/portworx.go
+++ b/drivers/storage/portworx/portworx.go
@@ -20,7 +20,9 @@ import (
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
+	api "k8s.io/kubernetes/pkg/apis/core"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
@@ -216,6 +218,10 @@ func (p *portworx) SetDefaultsOnStorageCluster(toUpdate *corev1alpha1.StorageClu
 }
 
 func (p *portworx) PreInstall(cluster *corev1alpha1.StorageCluster) error {
+	if err := p.validateEssentials(); err != nil {
+		return err
+	}
+
 	for componentName, comp := range component.GetAll() {
 		if comp.IsEnabled(cluster) {
 			err := comp.Reconcile(cluster)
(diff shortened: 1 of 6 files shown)


cmd/operator.go
```diff
@@ -71,6 +74,10 @@ func main() {
 			Usage: "Port on which the operator metrics are to be exposed",
 			Value: defaultMetricsPort,
 		},
+		cli.BoolFlag{
```
sharma-tapas, line 77: instead of having a bool flag, can we take in the `essentials entitlement id`? This will solve 2 things
1. We do not need to ask the user to create the secret for metering
2. We know that the user has registered on the central SaaS portal before trying to install PX Essentials.
Also, taking the ID, will make sure they actually unlink any clusters that are running essentials for their user ID.

piyush-nimbalkar: These are the operator arguments and not the StorageCluster itself. In operator model users deploy two things - 
- Install the operator
- Create StorageCluster CR which has the portworx cluster configuration. Operator looks at StorageCluster and then creates a Portworx cluster accordingly

This boolean is meant to tell the operator that even if the user does not say that this has to be an essentials cluster, we create one because this is essentials operator.

piyush-nimbalkar:
For the secret I was thinking of adding the marketplace information in the spec generator itself. Operatorhub does not have a form like Rancher to fill in the essentials id. Most likely people will go to central to generate a spec, something like this -
https://install.portworx.com/2.5?mc=false&kbver=1.17.0&oem=esse&user=f9279a8d-75e2-11ea-97e6-f6e09c7a4e5e&b=true&cache=undefined&c=px-cluster-2a37baeb-cf39-4210-96a5-cdb2e69ff260&osft=true&operator=true&stork=true&lh=true&st=k8s
This spec has both the StorageCluster and the secret.

We could have passed the essentials user id in plain text to StorageCluster directly and the operator could have created the secret for the user. The only problem I thought was the user id will remain visible to anyone who has access to StorageCluster object.

piyush-nimbalkar:
- Have updated the review to use environment variables to enforce px essentials. 
- Also added -marketplace_name argument to portworx starting 2.5.5

drivers/storage/portworx/util/util.go
```diff
@@ -46,6 +46,12 @@ const (
 	PortworxSDKPortName = "px-sdk"
 	// PortworxKVDBPortName name of the Portworx internal KVDB port
 	PortworxKVDBPortName = "px-kvdb"
+	// EssentialsSecretName name of the Portworx Essentials secret
+	EssentialsSecretName = "px-essentials"
```
adityadani, line 50: The secret name is called `px-essential`

drivers/storage/portworx/portworx.go
```diff
@@ -236,6 +242,34 @@ func (p *portworx) PreInstall(cluster *corev1alpha1.StorageCluster) error {
...
+func (p *portworx) validateEssentials() error {
+	if pxutil.EssentialsEnabled() {
+		resource := types.NamespacedName{
+			Name:      pxutil.EssentialsSecretName,
+			Namespace: api.NamespaceSystem,
+		}
+		secret := &v1.Secret{}
+		err := p.k8sClient.Get(context.TODO(), resource, secret)
+		if errors.IsNotFound(err) {
+			return fmt.Errorf("secret %s/%s should be present to deploy a "+
```
adityadani, line 254: should we add a message saying the user should go to PX-Central to generate one?

piyush-nimbalkar: There is no clear page on how/where they can generate the secret. I think once we have a doc page it would be easier to link it to that.
Looking at the above message hopefully they will look at the prereq section until then.

PR #186, opened 2020-07-09
OPERATOR-183 Add StorageNode controller

harsh-px:
1. Add a separate controller for StorageNode. This is responsible for
    * Reacting to status updates of StorageNode and syncing storage driver and kvdb pods. The updates of the actual StorageNode objects is still done by the StorageCluster controller
2. If a Kubernetes is controller by a Machine (Cluster API) and the machine is being deleted, don't create new pods
3. Add Storage status in the StorageNode to reflect total and used size on the node
4. Add KVDB status condition to the StorageNode to reflect if the node is a KVDB node


pkg/controller/storagenode/storagenode.go
```diff
@@ -0,0 +1,393 @@
...
+	podSpec.NodeName = storageNode.Name
+
+	trueVar := true
+	newPod := &v1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: fmt.Sprintf("%s-kvdb-", c.Driver.String()),
+			Namespace:    storageNode.Namespace,
+			Labels:       c.kvdbPodLabels(cluster),
+			OwnerReferences: []metav1.OwnerReference{
+				{
```
piyush-nimbalkar, line 332: nit: We could use this - `metav1.NewControllerRef(storageNode, corev1alpha1.SchemeGroupVersion.WithKind("StorageNode"))`, instead of manually creating the owner ref.

pkg/controller/storagenode/storagenode.go
```diff
@@ -0,0 +1,393 @@
...
+) error {
+	// sync the storage labels on pods
+	portworxPodList := &v1.PodList{}
+	pxLabels := c.Driver.GetSelectorLabels()
+	err := c.client.List(
+		context.TODO(),
+		portworxPodList,
+		&client.ListOptions{
+			Namespace:     storageNode.Namespace,
+			LabelSelector: labels.SelectorFromSet(pxLabels),
```
piyush-nimbalkar, line 273: Why not use FieldSelector here as well to filter by the spec.nodeName?

pkg/controller/storagenode/storagenode.go
```diff
@@ -0,0 +1,393 @@
...
+	)
+	if err != nil {
+		return fmt.Errorf("failed to get list of portworx pods. %v", err)
+	}
+
+	for _, pod := range portworxPodList.Items {
+		podCopy := pod.DeepCopy()
+		controllerRef := metav1.GetControllerOf(podCopy)
+		if controllerRef != nil && controllerRef.UID == cluster.UID &&
+			len(pod.Spec.NodeName) != 0 && storageNode.Name == podCopy.Spec.NodeName {
```
piyush-nimbalkar, line 284: We should ignore pod which is marked for deletion - `pod.DeletionTimestamp != nil`
Although storage cluster controller only starts a pod if the previous is completely terminated, just to be safe we can add that check, so update the newer pod and not the one in terminating state.

pkg/controller/storagecluster/controller_test.go
```diff
@@ -332,6 +290,8 @@ func TestStorageClusterDefaults(t *testing.T) {
 		Driver: driver,
 	}
 
+	controller.log(cluster).Debugf("testing default cluster")
```
piyush-nimbalkar, line 293: nit: Remove this?

pkg/controller/storagenode/storagenode_test.go
```diff
@@ -0,0 +1,615 @@
...
+			Name:      testStorageNode,
+			Namespace: testNS,
+		},
+	}
+	result, err := controller.Reconcile(request)
+	require.NoError(t, err)
+	require.NotNil(t, result)
+
+	checkStoragePod := &v1.Pod{}
+	err = controller.client.Get(context.TODO(), client.ObjectKey{
```
piyush-nimbalkar, line 321: nit: We can use https://github.com/libopenstorage/operator/blob/master/pkg/util/test/util.go#L70 to do objects gets.

pkg/controller/storagenode/storagenode_test.go
```diff
@@ -0,0 +1,615 @@
...
+	// reconcile storageless node
+	request = reconcile.Request{
+		NamespacedName: types.NamespacedName{
+			Name:      testStoragelessNode,
+			Namespace: testNS,
+		},
+	}
+	result, err = controller.Reconcile(request)
+	require.NoError(t, err)
+	require.NotNil(t, result)
```
piyush-nimbalkar, line 357: Shouldn't we check that the storagelessPod (podNode2)still does not have the labels?

pkg/controller/storagenode/storagenode_test.go
```diff
@@ -0,0 +1,615 @@
...
+		NamespacedName: types.NamespacedName{
+			Name:      "no-such-node",
+			Namespace: testNS,
+		},
+	}
+	_, err = controller.Reconcile(request)
+	require.NoError(t, err)
+}
+
+// TestReconcileKVDB focuses on reconcilign a StorageNode which is running KVDB
```
piyush-nimbalkar, line 370: nit: reconciling

pkg/controller/storagenode/storagenode_test.go
```diff
@@ -0,0 +1,615 @@
...
+
+	request := reconcile.Request{
+		NamespacedName: types.NamespacedName{
+			Name:      testStorageNode,
+			Namespace: testNS,
+		},
+	}
+	result, err := controller.Reconcile(request)
+	require.NoError(t, err)
+	require.NotNil(t, result)
```
piyush-nimbalkar, line 318: nit: result is not a pointer so it will always be non-nil even in case of error. If you want, you can check that it is empty.

drivers/storage/portworx/status.go
```diff
@@ -99,6 +108,29 @@ func (p *portworx) updateStorageNodes(
...
 	}
 
+	// If cluster is running internal kvdb, get current bootstrap nodes
+	kvdbNodeMap := make(map[string]*kvdb_api.BootstrapEntry)
+	if cluster.Spec.Kvdb != nil && cluster.Spec.Kvdb.Internal {
+		cmName := k8s.GetBootstrapConfigMapName(cluster.GetName())
+		cm := &v1.ConfigMap{}
+		err = p.k8sClient.Get(context.TODO(), types.NamespacedName{
+			Name:      cmName,
+			Namespace: cluster.GetNamespace(),
```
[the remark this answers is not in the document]
adityadani: Thats right. PX has been creating this config map in kube-system since Day 1.

pkg/controller/storagenode/storagenode_test.go
harsh-px: added check. It was already being checked indirectly above due to transition of storage node to storageless node.

pkg/util/k8s/k8s_test.go
```diff
@@ -2019,3 +2020,45 @@ func TestServiceChangeSelector(t *testing.T) {
...
+	recorder := record.NewFakeRecorder(10)
+	n1 := &v1.Node{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "n1",
+		},
+		Status: v1.NodeStatus{
+			Phase: v1.NodeRunning,
+		},
+	}
+	WarningEvent(recorder, n1, "test reason", "test message")
```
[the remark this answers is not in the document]
harsh-px: added check

drivers/storage/portworx/status.go
```diff
@@ -270,6 +304,43 @@ func (p *portworx) updateStorageNodeStatus(
...
+		TotalSize: *resource.NewQuantity(totalSizeInBytes, resource.BinarySI),
+		UsedSize:  *resource.NewQuantity(usedSizeInBytes, resource.BinarySI),
+	}
+
+	kvdbEntry, present := kvdbNodeMap[storageNode.Status.NodeUID]
+	if present && kvdbEntry != nil {
+		nodeKVDBCondition := &corev1alpha1.NodeCondition{
+			Type:   corev1alpha1.NodeKVDBCondition,
+			Status: mapKVDBState(kvdbEntry.State),
+			Message: fmt.Sprintf("node is kvdb %s listening on %s",
```
adityadani, line 326: the node.Type is an int, so this message will be confusing for the user. May be change this to
`node is running kvdb and is listening on port %s`

harsh-px: The %s in the message is `mapKVDBNodeType(kvdbEntry.Type)`. So it will print either "leader" or "member" in string format

PR #188, opened 2020-07-15
Detect k3s cluster and install portworx accordingly

piyush-nimbalkar:
- add extra mount for k3s
- enable csi

### DIFF
--- a/drivers/storage/portworx/deployment.go
+++ b/drivers/storage/portworx/deployment.go
@@ -173,6 +173,8 @@ type template struct {
 	cluster         *corev1alpha1.StorageCluster
 	isPKS           bool
 	isOpenshift     bool
+	isK3s           bool
+	runOnMaster     bool
 	imagePullPolicy v1.PullPolicy
 	serviceType     v1.ServiceType
 	startPort       int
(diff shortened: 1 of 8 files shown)


drivers/storage/portworx/deployment.go
```diff
@@ -878,6 +895,10 @@ func (t *template) getEnvList() []v1.EnvVar {
 
 func (t *template) getVolumeMounts() []v1.VolumeMount {
 	volumeInfoList := append([]volumeInfo{}, defaultVolumeInfoList...)
+	if t.isK3s {
+		volumeInfoList = append(volumeInfoList, t.getK3sVolumeInfoList()...)
+	}
+
 	volumeMounts := make([]v1.VolumeMount, 0)
```
zoxpx, line 902: nit -- the following allocation is a little bit better:
```go
volumeMounts := make([]v1.VolumeMount, 0, len(volumeInfoList))
```
* I normally always take advantage of the fact that we know exactly how large will the produced array be
* if we leave it up to `append()` to extend the array-capacity, then it'll _double_ the array capacity (also _copy_ all the elements) every time it extends the capacity

piyush-nimbalkar:
Addressed both the comments.
Introduced an internal annotation on StorageCluster `portworx.io/run-on-master` which users can use to toggle running on master nodes. With k3s we will run on master by default without user adding the annotation.

PR #193, opened 2020-07-17
Use REGISTRY_SECRET instead of REGISTRY_CONFIG

piyush-nimbalkar:
Starting oci-monitor:2.3.2, we support taking in the image pull
secret name directly in oci-mon.

### DIFF
--- a/drivers/storage/portworx/deployment.go
+++ b/drivers/storage/portworx/deployment.go
@@ -796,16 +796,24 @@ func (t *template) getEnvList() []v1.EnvVar {
 	}
 
 	if t.cluster.Spec.ImagePullSecret != nil && *t.cluster.Spec.ImagePullSecret != "" {
-		envMap["REGISTRY_CONFIG"] = &v1.EnvVar{
-			Name: "REGISTRY_CONFIG",
-			ValueFrom: &v1.EnvVarSource{
-				SecretKeyRef: &v1.SecretKeySelector{
-					Key: ".dockerconfigjson",
-					LocalObjectReference: v1.LocalObjectReference{
-						Name: *t.cluster.Spec.ImagePullSecret,
+		pxVer2_3_2, _ := version.NewVersion("2.3.2")
+		if t.pxVersion.LessThan(pxVer2_3_2) {
+			envMap["REGISTRY_CONFIG"] = &v1.EnvVar{
+				Name: "REGISTRY_CONFIG",
+				ValueFrom: &v1.EnvVarSource{
+					SecretKeyRef: &v1.SecretKeySelector{
+						Key: ".dockerconfigjson",
+						LocalObjectReference: v1.LocalObjectReference{
+							Name: *t.cluster.Spec.ImagePullSecret,
+						},
 					},
 				},
-			},
+			}
+		} else {
+			envMap["REGISTRY_SECRET"] = &v1.EnvVar{
+				Name:  "REGISTRY_SECRET",
+				Value: *t.cluster.Spec.ImagePullSecret,
+			}
 		}
 	}
 
--- a/drivers/storage/portworx/deployment_test.go
+++ b/drivers/storage/portworx/deployment_test.go
@@ -92,7 +92,7 @@ func TestPodSpecWithImagePullSecrets(t *testing.T) {
 	cluster := &corev1alpha1.StorageCluster{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "px-cluster",
-			Namespace: "kube-system",
+			Namespace: "kube-test",
 		},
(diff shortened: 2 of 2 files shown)


drivers/storage/portworx/deployment.go
```diff
@@ -796,16 +796,24 @@ func (t *template) getEnvList() []v1.EnvVar {
...
+		pxVer2_3_2, _ := version.NewVersion("2.3.2")
+		if t.pxVersion.LessThan(pxVer2_3_2) {
+			envMap["REGISTRY_CONFIG"] = &v1.EnvVar{
+				Name: "REGISTRY_CONFIG",
+				ValueFrom: &v1.EnvVarSource{
+					SecretKeyRef: &v1.SecretKeySelector{
+						Key: ".dockerconfigjson",
+						LocalObjectReference: v1.LocalObjectReference{
+							Name: *t.cluster.Spec.ImagePullSecret,
+						},
```
zoxpx, line 808: Do we actually need this block  (ie. is the current version of Px-Operator supposed to handle the px-2.3.2, released Dec 18, 2019)?

drivers/storage/portworx/deployment.go
```diff
@@ -796,16 +796,24 @@ func (t *template) getEnvList() []v1.EnvVar {
...
 					},
 				},
-			},
+			}
+		} else {
+			envMap["REGISTRY_SECRET"] = &v1.EnvVar{
+				Name:  "REGISTRY_SECRET",
+				Value: fmt.Sprintf("%s/%s", t.cluster.Namespace, *t.cluster.Spec.ImagePullSecret),
```
zoxpx, line 815: Note about namespace here:
* Kubernetes own `imagePullSecrets` directive does not normally allow Pull-Secrets to span between the namespaces (i.e. one can use only "local namespace" secrets)
* our implementation of `REGISTRY_SECRET` is a bit different, since it'll allow fetching secrets also from a different namespace  (note, we needed this to support specific IBM-K8s-cloud default setups)
* the newer versions of OCI-Monitor's do not need the `REGISTRY_SECRET` setting at all -- by default they'll try to reuse the pull-secret that's been assigned to OCI-Monitor POD itself
  - so, in the future PX-Operator releases we'll likely drop the `REGISTRY_SECRET` env. variable altogether

Because of this, it might be better to align with K8s limitation (and our future default), and not use Namespace in `REGISTRY_SECRET`  (i.e. use `<pullSecret>` notation, instead of our custom `<namespace>/<pullSecret>`)

drivers/storage/portworx/deployment.go
piyush-nimbalkar: There could be customers who are using 2.3.2. Don't want to break this for them. Will keep it around for a while.

drivers/storage/portworx/deployment.go
piyush-nimbalkar: @zoxpx So you recommend not passing REGISTRY_SECRET at all? Or asking to pass only the secret name?
I have to pass the namespace because from the ticket it seemed like oci-mon defaults to kube-system if namespace is not passed. Even operator mandates user to have pull secret in the same namespace. But if cluster itself if deployed in non-kube-system ns, I need to pass the namespace for oci-mon.

drivers/storage/portworx/deployment.go
zoxpx: Oh.. didn't realize the latest PX-Operator has to support _all_ Portworx versions:
* code might get busy dealing with the backward-compatibility tweaks, and the new features
* alternatively, we could just introduce support-matrix for PX-Operator, so a certain PX-Operator would support certain set of PX-versions
   - in fact, I think it's very likely we'll end up with such version-specific deployments in the future

drivers/storage/portworx/deployment.go
zoxpx: If namespace is not specified in `REGISTRY_SECRET`, the OCI-Monitor will use its own namespace to access the secrets.
Apropos removing the `REGISTRY_SECRET` variable altogether, I think it's too early for this  (we've introduced this default fairly recently, if I remember correctly)

piyush-nimbalkar: @zoxpx has this always been the case? Can I safely assume even for 2.3.2 we don't need to pass the namespace in REGISTRY_SECRET?

drivers/storage/portworx/deployment.go
piyush-nimbalkar: Sure we can have that in future when it gets unmanageable.

drivers/storage/portworx/deployment.go
zoxpx: yes and yes  :)

PR #194, opened 2020-07-19
Update CRD on operator restarts

piyush-nimbalkar:
- Updating CRD for validations changes if any on operator restart. If a new version of operator has new spec fields they will get applied to the CRD.
- Removing the deprecated StorageNodeStatusCRD
- Adding missing specs to the StorageCluster CRD
- Minor fix - When PX deployment is disabled (px central uses this), do not fetch release manifest. Also no need to populate image, version when px is not even enabled.


drivers/storage/portworx/portworx.go
```diff
@@ -149,16 +150,16 @@ func (p *portworx) SetDefaultsOnStorageCluster(toUpdate *corev1alpha1.StorageClu
 
 	toUpdate.Spec.Image = strings.TrimSpace(toUpdate.Spec.Image)
 	toUpdate.Spec.Version = pxutil.GetImageTag(toUpdate.Spec.Image)
-	pxVersionChanged := toUpdate.Spec.Version == "" ||
-		toUpdate.Spec.Version != toUpdate.Status.Version
+	pxVersionChanged := pxEnabled &&
+		(toUpdate.Spec.Version == "" || toUpdate.Spec.Version != toUpdate.Status.Version)
```
ggriffiths, line 154: what if `toUpdate.Status.Version` is already `""` ? Is an update required in this case?

piyush-nimbalkar: Yes. That will happen only the first time this version of operator is started (spec.Version present but status.Version absent), because the status.Version field was newly introduced. The status.Version will be set in the first reconcile loop and after that it  won't require the update until spec.Version changes.

PR #196, opened 2020-07-23
Create core v1 API and generate its clients

piyush-nimbalkar:
Signed-off-by: Piyush Nimbalkar <piyush@portworx.com>

PR #197, opened 2020-07-27
Upgrade API objects to v1 from v1alpha1

piyush-nimbalkar:
- Have followed [this Kubernetes doc](https://v1-15.docs.kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning/#writing-reading-and-updating-versioned-customresourcedefinition-objects) for CRD version upgrade.
- Had already copied the API objects in v1 folder and generated clients - https://github.com/libopenstorage/operator/pull/196
- Tested this in a vanilla Kubernetes and OpenShift cluster with a new StorageCluster/StorageNodes as well existing ones. Eventually all stored v1alpha1 objects get moved v1 in the k8s api-server.


piyush-nimbalkar:
Thinking of deleting the old v1alpha1 CRDs and its clients as there won't be any consumer.

PR #198, opened 2020-07-29
Add support for old stork issuer with newer versions of PX

ggriffiths:

### DIFF
--- a/drivers/storage/portworx/deployment_test.go
+++ b/drivers/storage/portworx/deployment_test.go
@@ -2627,7 +2627,6 @@ func TestSecuritySetEnv(t *testing.T) {
 	k8sClient := coreops.New(fakek8sclient.NewSimpleClientset())
 	coreops.SetInstance(k8sClient)
 	nodeName := "testNode"
-
 	cluster := &corev1alpha1.StorageCluster{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "px-cluster",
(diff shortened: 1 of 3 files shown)


drivers/storage/portworx/util/util.go
```diff
@@ -303,6 +305,37 @@ func GetPortworxVersion(cluster *corev1alpha1.StorageCluster) *version.Version {
 	return pxVersion
 }
 
+// GetStorkVersion returns the stork version based on the image provided.
+// We first look at spec.Image. If that is not present or invalid semvar, then we fallback to an
```
piyush-nimbalkar, line 309: nit: `spec.Stork.Image`
Also update the comment as per the below review comment.

drivers/storage/portworx/util/util.go
```diff
@@ -303,6 +305,39 @@ func GetPortworxVersion(cluster *corev1alpha1.StorageCluster) *version.Version {
...
+// annotation portworx.io/stork-version; else we return int max as the version.
+func GetStorkVersion(cluster *corev1alpha1.StorageCluster) *version.Version {
+	defaultVersion, _ := version.NewVersion(strconv.FormatInt(math.MaxInt64, 10))
+	if cluster.Spec.Stork == nil && cluster.Annotations[AnnotationStorkVersion] == "" {
+		logrus.Warnf("could not find stork version from cluster spec")
+		return defaultVersion
+	}
+
+	parts := strings.Split(cluster.Spec.Stork.Image, ":")
```
piyush-nimbalkar, line 318: Starting operator 1.4, `cluster.Spec.Stork.Image` may not always be present. If it is present, it will be only to override the operator selected image. The operator selected image will be found in `cluster.Status.DesiredImages.Stork`.
Basically it should be - 
```
if cluster.Spec.Stork.Image == "" {
  // get version from cluster.Status.DesiredImages.Stork
  // if this is semver, return it
  // else get it from the annotation and return it
} else {
  // if this is semver, return it
  // else get it from the annotation and return it
}
```

drivers/storage/portworx/util/util.go
```diff
@@ -303,6 +305,39 @@ func GetPortworxVersion(cluster *corev1alpha1.StorageCluster) *version.Version {
 	return pxVersion
 }
 
+// GetStorkVersion returns the stork version based on the image provided.
+// We first look at spec.Image. If that is not present or invalid semvar, then we fallback to an
+// annotation portworx.io/stork-version; else we return int max as the version.
+func GetStorkVersion(cluster *corev1alpha1.StorageCluster) *version.Version {
+	defaultVersion, _ := version.NewVersion(strconv.FormatInt(math.MaxInt64, 10))
+	if cluster.Spec.Stork == nil && cluster.Annotations[AnnotationStorkVersion] == "" {
```
piyush-nimbalkar, line 313: I think if stork is disabled, we can just return the default, ignoring the annotation, right?
We should check both stork == nil and stork.enabled == false to check stork is disabled.

drivers/storage/portworx/util/util.go
```diff
@@ -303,6 +305,46 @@ func GetPortworxVersion(cluster *corev1alpha1.StorageCluster) *version.Version {
...
+					logrus.Warnf("Invalid Stork version %s extracted from annotation: %v", storkVersionStr, err)
+				}
+			}
+		}
+	} else {
+		logrus.Warnf("could not find stork version from cluster spec")
+		return defaultVersion
+	}
+
+	return storkVersion
```
piyush-nimbalkar, line 345: We should check for nil here and return default version. 
If the image tag and annotation are both not semantic versions, this would be nil.

drivers/storage/portworx/deployment_test.go
```diff
@@ -2697,6 +2696,92 @@ func TestSecuritySetEnv(t *testing.T) {
...
+	actual, err = driver.GetStoragePodSpec(cluster, nodeName)
+	assert.NoError(t, err, "Unexpected error on GetStoragePodSpec")
+	appsSecretSet = false
+	for _, env := range actual.Containers[0].Env {
+		if env.Name == pxutil.EnvKeyPortworxAuthSystemAppsKey {
+			appsSecretSet = true
+		}
+	}
+	assert.True(t, storkSecretSet)
+
```
piyush-nimbalkar, line 2784: Add test case for 
- when spec.stork.image and annotation both have invalid semantic version, should use apps key
- when spec.stork.image (or status.desiredImage.stork) have no tag, should use apps key

PR #199, opened 2020-07-29
Add support for security TokenLifetime in days and years

ggriffiths:


drivers/storage/portworx/component/security.go
```diff
@@ -293,7 +293,11 @@ func (c *security) createToken(
 		Roles:   []string{role},
 		Groups:  groups,
 	}
-	token, err := pxutil.GenerateToken(cluster, authSecret, &claims, cluster.Spec.Security.Auth.SelfSigned.TokenLifetime.Duration)
+	tokenDuration, err := pxutil.ParseExtendedDuration(*cluster.Spec.Security.Auth.SelfSigned.TokenLifetime)
+	if err != nil {
+		return "", fmt.Errorf("failed to parse token lifetime: %v", err.Error())
```
piyush-nimbalkar, line 298: Add a test case for invalid duration.
Also check if event is raised.

ggriffiths: Added test. will fail until the regex fix is added.

PR #200, opened 2020-07-30
Make Stork JWT issuer depend on PX version

ggriffiths:

### DIFF
--- a/drivers/storage/portworx/portworx.go
+++ b/drivers/storage/portworx/portworx.go
@@ -119,6 +119,25 @@ func (p *portworx) GetStorkEnvList(cluster *corev1alpha1.StorageCluster) []v1.En
 					Key: pxutil.SecurityAppsSecretKey,
 				},
 			}})
+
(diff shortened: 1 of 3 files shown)


drivers/storage/portworx/portworx.go
```diff
@@ -119,6 +119,30 @@ func (p *portworx) GetStorkEnvList(cluster *corev1alpha1.StorageCluster) []v1.En
...
 				},
 			}})
+
+		pxVersion := pxutil.GetPortworxVersion(cluster)
+		storkVersion := pxutil.GetStorkVersion(cluster)
+		pxAppsIssuerVersion, err := version.NewVersion("2.6.0")
+		if err != nil {
+			logrus.Errorf("failed to create PX version variable 2.6.0: %s", err.Error())
+		}
+		storkIssuerVersion, err := version.NewVersion("2.5.0")
```
piyush-nimbalkar, line 129: Just to simplify, we don't need to check stork version. Technically passing an extra env should for older stork version should not be harmful.

PR #217, opened 2020-09-15
Let users specify the service account to be used by portworx pods

piyush-nimbalkar:
In AWS marketplace, portworx has to use a service account associated
with the IAM role so that the metering would work for Portworx.
The users can create that service account and ask the operator to use
that service account instead of the default one.


pault84 (approved):
LGTM

PR #220, opened 2020-10-02
Security: fix an issue where existing secrets were overwritten

ggriffiths:

### DIFF
--- a/drivers/storage/portworx/component/security.go
+++ b/drivers/storage/portworx/component/security.go
@@ -251,12 +251,12 @@ func (c *security) createPrivateKeysSecret(
 		},
 	}
 
-	err = k8sutil.CreateOrUpdateSecret(c.k8sClient, sharedSecret, nil)
+	err = k8sutil.CreateOrAppendToSecret(c.k8sClient, sharedSecret, nil)
 	if err != nil {
 		return err
 	}
 
-	err = k8sutil.CreateOrUpdateSecret(c.k8sClient, systemKeysSecret, nil)
+	err = k8sutil.CreateOrAppendToSecret(c.k8sClient, systemKeysSecret, nil)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


drivers/storage/portworx/components_test.go
```diff
@@ -3078,6 +3078,33 @@ func TestSecurityInstall(t *testing.T) {
...
+	err = k8sClient.Create(context.TODO(), sharedSecret)
+	require.NoError(t, err)
+	cluster.Spec.Security.Auth.SelfSigned.SharedSecret = &existingSharedSecretName
+
+	err = driver.PreInstall(cluster)
+	require.NoError(t, err)
+
+	err = testutil.Get(k8sClient, sharedSecret, existingSharedSecretName, cluster.Namespace)
+	require.NoError(t, err)
+	s1, ok := sharedSecret.Data["key1"]
```
piyush-nimbalkar, line 3102: Should we check that the actual shared-secret key is added to the secret?

drivers/storage/portworx/component/security.go
```diff
@@ -251,12 +251,12 @@ func (c *security) createPrivateKeysSecret(
 		},
 	}
 
-	err = k8sutil.CreateOrUpdateSecret(c.k8sClient, sharedSecret, nil)
+	err = k8sutil.CreateOrAppendToSecret(c.k8sClient, sharedSecret, nil)
```
piyush-nimbalkar, line 254: Not related to this change, but general question - When user is passing their own system and app secret as env variable, we will just end up creating a `px-system-secrets` secret and not using it, right?

piyush-nimbalkar (approved):

drivers/storage/portworx/components_test.go
ggriffiths: Added this check.

drivers/storage/portworx/component/security.go
ggriffiths: Answered offline - 
This is almost true. We always create and maintain `px-system-secrets`. We directly use the `px-system-secrets` for getting the apps secret for the operator to communicate w/ PX.

PR #221, opened 2020-10-06
PTX-2725 Taking images from list incase StorageCluster doesn't have images defined

nikolaypopov:
Taking images from list incase StorageCluster doesn't have images defined

### DIFF
--- a/pkg/util/test/util.go
+++ b/pkg/util/test/util.go
@@ -522,17 +522,6 @@ func expectedPods(cluster *corev1.StorageCluster) (int, error) {
 	return podCount, nil
 }
 
-func getImageNameFromMap(pxImageList map[string]string, component string) string {
-	var componentImageName string
-
-	for key, value := range pxImageList {
-		if key == component {
-			componentImageName = value
-		}
-	}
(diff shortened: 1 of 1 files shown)


pkg/util/test/util.go
```diff
@@ -521,7 +522,18 @@ func expectedPods(cluster *corev1.StorageCluster) (int, error) {
 	return podCount, nil
 }
 
-func validateComponents(cluster *corev1.StorageCluster, timeout, interval time.Duration) error {
+func getImageNameFromMap(pxImageList map[string]string, component string) string {
```
thiguetta, line 525: why do we need this? can we simple use `image := pxImageList[component]`?

thiguetta: if you need to validade if the key is present you can use
```
if value, ok := pxImageList[component]; ok {
         image  =  value
}
```

nikolaypopov: Thats a good idea @thiguetta let me try this out!

thiguetta (approved):
LGTM

PR #223, opened 2020-10-21
Create pod disruption budgets for storage and kvdb pods

piyush-nimbalkar:
Changes:
- Creating pod disruption budgets for portworx storage nodes and internal kvdb nodes
- By default PDBs will be created. Providing an annotation to disable their creation.
- Also waiting for cluster to be up before making SDK calls from PDB and security components. The calls used to fail when cluster was not yet initialized and raise warning events.

### DIFF
--- a/drivers/storage/portworx/component/disruption_budget.go
+++ b/drivers/storage/portworx/component/disruption_budget.go
@@ -1,0 +1,191 @@
+package component
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/go-version"
+	"github.com/libopenstorage/openstorage/api"
+	pxutil "github.com/libopenstorage/operator/drivers/storage/portworx/util"
+	corev1 "github.com/libopenstorage/operator/pkg/apis/core/v1"
+	"github.com/libopenstorage/operator/pkg/constants"
+	k8sutil "github.com/libopenstorage/operator/pkg/util/k8s"
+	"github.com/sirupsen/logrus"
+	"google.golang.org/grpc"
+	policyv1beta1 "k8s.io/api/policy/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/client-go/tools/record"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	// DisruptionBudgetComponentName name of the DisruptionBudget component
+	DisruptionBudgetComponentName = "DisruptionBudget"
+	// StoragePodDisruptionBudgetName name of the PodDisruptionBudget for portworx storage pods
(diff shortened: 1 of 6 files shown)


drivers/storage/portworx/component/disruption_budget.go
```diff
@@ -0,0 +1,191 @@
...
+func (c *disruptionBudget) createKVDBPodDisruptionBudget(
+	cluster *corev1.StorageCluster,
+	ownerRef *metav1.OwnerReference,
+) error {
+	// No need to create PDB for KVDB when using external KVDB
+	if cluster.Spec.Kvdb != nil && !cluster.Spec.Kvdb.Internal {
+		return nil
+	}
+
+	minAvailable := intstr.FromInt(2)
```
adityadani, line 152: We do support kvdb cluster size of 5. We can check from the StorageCluster object if the "--kvdb_cluster_size" argument is provided and then decide whether to use 2 or 3.

drivers/storage/portworx/component/disruption_budget.go
```diff
@@ -0,0 +1,191 @@
...
+					MatchLabels: map[string]string{
+						constants.LabelKeyClusterName: cluster.Name,
+						constants.LabelKeyStoragePod:  constants.LabelValueTrue,
+					},
+				},
+			},
+		}
+		return k8sutil.CreateOrUpdatePodDisruptionBudget(c.k8sClient, pdb, ownerRef)
+	}
+	return nil
```
adityadani, line 140: need to close sdk connection?
In a rare scenario, if someone deletes a PDB object, we will create another sdkConn object without closing the previous one.

adityadani (approved):
looks good

ggriffiths (approved):

drivers/storage/portworx/component/disruption_budget.go
piyush-nimbalkar: You mean check the argument from misc-args?

drivers/storage/portworx/component/disruption_budget.go
piyush-nimbalkar: We don't need to close the connection. We store the sdkConn in the disruptionBudget component. It lives for the lifetime of the operator container.
`GetPortworxConn` method does not create a new one if the current one is not nil.

drivers/storage/portworx/component/disruption_budget.go
adityadani: Yes

drivers/storage/portworx/component/disruption_budget.go
adityadani: Ah saw the GetPortworxConn implementation. Ignore my comment

piyush-nimbalkar:
@adityadani, for the kvdb cluster size change I am going to create a new PR as it has gotten slightly bigger. Will avoid adding it to this approved PR.

piyush-nimbalkar:
https://github.com/libopenstorage/operator/pull/225

PR #230, opened 2020-11-25
Removed http from remote write endpoint url

tusharraut1994:
From pxcentral-api server side, install module api will take care of adding http or https based on either ssl is required/enabled in cortex endpoint.

### DIFF
--- a/drivers/storage/portworx/component/prometheus.go
+++ b/drivers/storage/portworx/component/prometheus.go
@@ -571,7 +571,7 @@ func (c *prometheus) createPrometheusInstance(
 		cluster.Spec.Monitoring.Prometheus.RemoteWriteEndpoint != "" {
 		prometheusInst.Spec.RemoteWrite = []monitoringv1.RemoteWriteSpec{
 			{
-				URL: fmt.Sprintf("http://%s/api/prom/push", cluster.Spec.Monitoring.Prometheus.RemoteWriteEndpoint),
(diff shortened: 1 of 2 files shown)


drivers/storage/portworx/component/prometheus.go
```diff
@@ -571,7 +571,7 @@ func (c *prometheus) createPrometheusInstance(
 		cluster.Spec.Monitoring.Prometheus.RemoteWriteEndpoint != "" {
 		prometheusInst.Spec.RemoteWrite = []monitoringv1.RemoteWriteSpec{
 			{
-				URL: fmt.Sprintf("http://%s/api/prom/push", cluster.Spec.Monitoring.Prometheus.RemoteWriteEndpoint),
+				URL: fmt.Sprintf("%s/api/prom/push", cluster.Spec.Monitoring.Prometheus.RemoteWriteEndpoint),
```
sharma-tapas, line 574: does it get redirected according to the protocol, i.e 80 for HTTP and 443 for https?
how does prometheus figure it out?

tusharraut1994: will set this protocol in endpoint form pxcentral api-server, as per this: https://github.com/portworx/pxcentral-onprem/pull/35/files#diff-202b2517a7852bf6ae601a7c789581917fdded59b8758165cef13993e852ca6cR517-R520

sharma-tapas (approved):

piyush-nimbalkar (approved):

PR #231, opened 2020-12-10
PWX-17604: fix OpenShift upgrades

zoxpx:
* adding a pre-exec rule to fix protection on cached px-configs

Signed-off-by: Zoran Rajic <zox@portworx.com>

### DIFF
--- a/drivers/storage/portworx/testspec/openshift_runc.yaml
+++ b/drivers/storage/portworx/testspec/openshift_runc.yaml
@@ -47,6 +47,8 @@ spec:
               value: "1500"
             - name: "PX_TEMPLATE_VERSION"
               value: "v4"
+            - name: "PRE-EXEC"
+              value: "if [ -f /etc/pwx/.private.json ] && [ -d /ostree/deploy/rhcos/deploy ]; then chattr -i /etc/pwx/.private.json /ostree/deploy/rhcos/deploy/*/etc/pwx/.private.json || /bin/true; fi"
           livenessProbe:
             periodSeconds: 30
             initialDelaySeconds: 840 # allow image pull in slow networks
--- a/drivers/storage/portworx/deployment.go
+++ b/drivers/storage/portworx/deployment.go
@@ -852,7 +852,7 @@ func (t *template) getEnvList() []v1.EnvVar {
 	if t.isOpenshift {
 		envMap["PRE-EXEC"] = &v1.EnvVar{
 			Name:  "PRE-EXEC",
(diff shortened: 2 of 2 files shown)


zoxpx:
Hey @piyush-nimbalkar -- looks like this will require a follow-up fix;  the updated px-operator is _not_ fixing up already deployed OCI-Monitors  (have to bounce/recreate OCI-Monitor's PODs for the change to apply).
* any tips how to fix this ?

piyush-nimbalkar (approved):

zoxpx:
Re-tested the latest Px-Operator version  (it's un-chatterring the files OK).
Also had a discussion w/ @piyush-nimbalkar, Kent and Luay -- we decided the 2nd / follow-up fix will not be required  -- we'll ask people to manually recycle the OCI-Mon pods to apply the change.

drivers/storage/portworx/testspec/openshift_runc.yaml
```diff
@@ -47,6 +47,8 @@ spec:
               value: "1500"
             - name: "PX_TEMPLATE_VERSION"
               value: "v4"
+            - name: "PRE-EXEC"
+              value: "if [ -f /etc/pwx/.private.json ] && [ -d /ostree/deploy/rhcos/deploy ]; then chattr -i /etc/pwx/.private.json /ostree/deploy/rhcos/deploy/*/etc/pwx/.private.json || /bin/true; fi"
```
cgwalters, line 51: I'd use `[ -f /run/ostree-booted ]` which is a generic OSTree chekc.  The path above (`.../rhcos/...`) is specific to RHEL CoreOS but there also exists e.g. FCOS and in general other ostree-based systems.

zoxpx: > I'd use `[ -f /run/ostree-booted ]` which is a generic OSTree check. The path above (`.../rhcos/...`) is specific to RHEL CoreOS but there also exists e.g. FCOS and in general other ostree-based systems.

Hi @cgwalters
Thanks for your feedback.  We'll actually have to revert this fix, since OpenShift-4.7 started write-protecting the partition, so `chattr -i ...` commands fail with "Read only partition..." errors.  Instead, we'll do a more robust check (and possible chattr-adjustment) during the `portworx.service` startup with our latest px-2.8.0 release.

cgwalters: Yes, we made `/sysroot` read only because really it shouldn't be modified in general.  This is a special case in trying to fix a previous deployment which we do need.

Try changing it to do:
`unshare -m /bin/sh -c 'chattr -i /etc/pwx/.private.json /ostree/deploy/rhcos/deploy/*/etc/pwx/.private.json'`

PR #234, opened 2021-01-22
PTX-2845 Adding first initial node affinity test

nikolaypopov:
Adding first initial node affinity test

### DIFF
--- a/test/integration_test/common_test.go
+++ b/test/integration_test/common_test.go
@@ -9,8 +9,12 @@ import (
 
 	corev1 "github.com/libopenstorage/operator/pkg/apis/core/v1"
 	k8sutil "github.com/libopenstorage/operator/pkg/util/k8s"
+	"github.com/portworx/sched-ops/k8s/core"
 	"github.com/portworx/sched-ops/k8s/operator"
+	"github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
 )
 
 const (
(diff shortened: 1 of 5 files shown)


test/integration_test/common_test.go
```diff
@@ -9,18 +9,56 @@ import (
...
+	}
+
+	k8sOps := core.Instance()
+	k8sOps.SetConfig(config)
+
+	operatorOps := operator.Instance()
+	operatorOps.SetConfig(config)
+
+	return nil
+}
```
piyush-nimbalkar, line 60: You don't need to do any of this. Just using `core.Instance()` or `operator.Instance()` anywhere in the code automatically initialize the kubernetes client to the current cluster. sched-ops handles that.
You need to call SetConfig only if you want to talk to a different Kubernetes cluster from this one. Stork has to perform migration tests, that's why it needs it. We don't need it here.

The assumption here is that we will be running the test pod in the same cluster as the operator. If you want to run the test locally, you can set the KUBECONFIG variable and it should talk to that k8s cluster.

test/integration_test/common_test.go
```diff
@@ -9,18 +9,56 @@ import (
...
 
 const (
 	specDir = "./specs"
 )
 
 func TestMain(m *testing.M) {
+	if err := setup(); err != nil {
+		logrus.Errorf("Setup failed with error: %v", err)
+		os.Exit(1)
+	}
```
piyush-nimbalkar, line 28: You can remove this as well, as there is no setup needed as of now. We can add it later if there is setup needed.

test/integration_test/node_affinity_test.go
```diff
@@ -0,0 +1,82 @@
...
+			},
+		},
+	}
+
+	// Get K8S nodes
+	nodeList, err := core.Instance().GetNodes()
+	require.NoError(t, err)
+
+	// Set affinity Label one of the K8S nodes
+	for _, node := range nodeList.Items {
```
piyush-nimbalkar, line 62: Make sure this is not a master node, else the test will fail. 
You can add this later when you add the validation for the test.

piyush-nimbalkar (approved):
LGTM after addressing the comments.

test/integration_test/common_test.go
nikolaypopov: Good to know, removed.

test/integration_test/common_test.go
nikolaypopov: Removed.

test/integration_test/node_affinity_test.go
nikolaypopov: Added check for master label.

test/integration_test/node_affinity_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+	// Get K8S nodes
+	nodeList, err := core.Instance().GetNodes()
+	require.NoError(t, err)
+
+	// Set affinity Label one of the K8S nodes
+	var nodeNameWithLabel string
+	for _, node := range nodeList.Items {
+		for key := range node.Labels {
+			if strings.Contains(key, "master") { // Skip master node, we don't need to label it
```
piyush-nimbalkar, line 66: Use `node-role.kubernetes.io/master`

piyush-nimbalkar (approved):
Squash and merge your changes

test/integration_test/node_affinity_test.go
nikolaypopov: Done.

Rohit-PX (approved):

PR #236, opened 2021-01-25
PTX-2840 Add node affinity validation and utils

nikolaypopov:
1) Add validation utils
2) Add validation for Node Affinity rules
3) Add more verbose output for Node Affinity test

### DIFF
--- a/pkg/util/test/util.go
+++ b/pkg/util/test/util.go
@@ -456,13 +456,24 @@ func validateStorageClusterPods(
 		}
 
 		if len(pods) != len(expectedPxNodeNameList) {
-			return "", true, fmt.Errorf("expected pods: %v. actual pods: %v", expectedPodCount, len(pods))
+			return "", true, fmt.Errorf("expected pods: %v. actual pods: %v", len(expectedPxNodeNameList), len(pods))
 		}
 
+		var pxNodeNameList []string
+		var podsNotReady []string
 		for _, pod := range pods {
 			if !coreops.Instance().IsPodReady(pod) {
-				return "", true, fmt.Errorf("pod %v/%v is not yet ready", pod.Namespace, pod.Name)
+				podsNotReady = append(podsNotReady, pod.Name)
 			}
+			pxNodeNameList = append(pxNodeNameList, pod.Spec.NodeName)
+		}
+
+		if len(podsNotReady) > 0 {
+			return "", true, fmt.Errorf("pods are not Ready: %v", podsNotReady)
+		}
+
+		if !assert.ElementsMatch(&testing.T{}, expectedPxNodeNameList, pxNodeNameList) {
+			return "", false, fmt.Errorf("expected Portworx nodes: %+v, got %+v", expectedPxNodeNameList, pxNodeNameList)
 		}
 
 		return "", false, nil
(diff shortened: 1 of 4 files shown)


pkg/util/test/util.go
```diff
@@ -646,6 +661,44 @@ func validateComponents(pxImageList map[string]string, cluster *corev1.StorageCl
 		return err
 	}
 
+	if cluster.Spec.Placement != nil {
```
piyush-nimbalkar, line 664: Let's fix this as discussed on the call.

pkg/util/test/util.go
```diff
@@ -842,3 +895,109 @@ func getK8SVersion() (string, error) {
...
+		URL:      pxReleaseManifestURL,
+		Content:  "application/json",
+		Auth:     "",
+		Insecure: true,
+	}
+
+	htmlData, err := DoRequest(request)
+	if err != nil {
+		return nil, fmt.Errorf("failed to send GET request to %s, Err: %v", pxReleaseManifestURL, err)
+	}
```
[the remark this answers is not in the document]
nikolaypopov: Removed, I initially added it in case we want to use it to make other various HTTP requests, but I guess we can add it later as needed.

pkg/util/test/util.go
```diff
@@ -491,10 +471,70 @@ func validateStorageClusterPods(
...
+			return fmt.Errorf("node %s is not online. Current: %v", nodeResp.Node.SchedulerNodeName,
+				nodeResp.Node.Status)
+		}
+
+	}
 	return nil
 }
 
-func expectedPods(cluster *corev1.StorageCluster) (int, error) {
+func getExpectedPodCount(cluster *corev1.StorageCluster) (int, error) {
```
piyush-nimbalkar, line 537: I thought we decided to compare the actual node lists instead of count.

pkg/util/test/util.go
```diff
@@ -505,6 +545,11 @@ func expectedPods(cluster *corev1.StorageCluster) (int, error) {
 		dummyPod.Spec.Affinity = &v1.Affinity{
 			NodeAffinity: cluster.Spec.Placement.NodeAffinity.DeepCopy(),
 		}
+	} else {
+		newCluster := addDefaultNodeAffinityRules(cluster)
```
piyush-nimbalkar, line 549: You are setting the node affinity on the passed cluster object as well as returning the same cluster object.
Do you want to change the passed cluster object or just return a new one with affinity rules? 
If you don't want to change the passed cluster object, why not just return the nodeaffinity object here and use that?

nikolaypopov: Done. I'm now using that func to create a Node Affinity object, without passing cluster to the func or passing the object back to the cluster.

pkg/util/test/util.go
nikolaypopov: Done, added more check that we talked about in the morning.

pkg/util/test/util.go
```diff
@@ -475,14 +455,25 @@ func validateStorageClusterPods(
...
 			}
+			pxNodeNameList = append(pxNodeNameList, pod.Spec.NodeName)
+		}
+
+		if len(podsNotReady) > 0 {
+			return "", true, fmt.Errorf("pods are not Ready: %v", podsNotReady)
+		}
+
+		if !assert.ElementsMatch(&testing.T{}, expectedPxNodeNameList, pxNodeNameList) {
+			return "", false, fmt.Errorf("expected Portworx nodes: %+v, got %+v", expectedPxNodeNameList, pxNodeNameList)
```
piyush-nimbalkar, line 476: This should be `return "", true, ...` , right? We do want to retry if there are not enough pods, correct?

pkg/util/test/util.go
```diff
@@ -491,35 +482,98 @@ func validateStorageClusterPods(
...
 	}
 
 	dummyPod := &v1.Pod{}
 	if cluster.Spec.Placement != nil && cluster.Spec.Placement.NodeAffinity != nil {
 		dummyPod.Spec.Affinity = &v1.Affinity{
 			NodeAffinity: cluster.Spec.Placement.NodeAffinity.DeepCopy(),
 		}
+	} else {
+		dummyPod.Spec.Affinity = &v1.Affinity{
+			NodeAffinity: defaultPxNodeAffinityRules().DeepCopy(),
```
piyush-nimbalkar, line 560: nit: No need to deepcopy here. It will return a new object every time anyway.

pkg/util/test/util.go
```diff
@@ -491,35 +482,98 @@ func validateStorageClusterPods(
...
 			continue
 		}
 		nodeInfo := schedulernodeinfo.NewNodeInfo()
 		nodeInfo.SetNode(&node)
 		if ok, _, _ := predicates.PodMatchNodeSelector(dummyPod, nil, nodeInfo); ok {
-			podCount++
+			nodeNameListWithPxPods = append(nodeNameListWithPxPods, node.Name)
 		}
 	}
+	fmt.Printf("KOKADBG: nodeNameListWithPxPods: %d nodes: %+v\n", len(nodeNameListWithPxPods), nodeNameListWithPxPods)
```
piyush-nimbalkar, line 574: nit: Remove this print statement.

piyush-nimbalkar (approved):
LGTM after addressing the comments

pkg/util/test/util.go
nikolaypopov: No, we do not want to retry here, if those node arrays do not match, it's game over.

pkg/util/test/util.go
nikolaypopov: Got it, fixed.

pkg/util/test/util.go
nikolaypopov: Forgot to remove this debug, removed.

PR #237, opened 2021-01-30
PTX-2858 Implement ability to deploy/validate PX version based on Spec Generator URL

nikolaypopov:
Implement ability to deploy/validate PX version based on Spec Generator URL which is passed from the command line.

### DIFF
--- a/test/integration_test/common_test.go
+++ b/test/integration_test/common_test.go
@@ -22,7 +22,6 @@ const (
 	specDir = "./operator-test"
 
 	defaultPxNamespace             = "kube-system"
-	defaultPxImageName             = "portworx/oci-monitor"
 	pxReleaseManifestURLEnvVarName = "PX_RELEASE_MANIFEST_URL"
 	pxRegistryUserEnvVarName       = "REGISTRY_USER"
 	pxRegistryPasswordEnvVarName   = "REGISTRY_PASS"
--- a/test/integration_test/common_test.go
+++ b/test/integration_test/common_test.go
@@ -3,25 +3,73 @@
 package integrationtest
 
 import (
+	"flag"
 	"os"
 	"path"
+	"strings"
 	"testing"
+	"time"
 
 	corev1 "github.com/libopenstorage/operator/pkg/apis/core/v1"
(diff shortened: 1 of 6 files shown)


test/integration_test/common_test.go
```diff
@@ -3,27 +3,75 @@
...
+	defaultPxNamespace              = "kube-system"
+	pxReleaseManifestURLEnvVariable = "PX_RELEASE_MANIFEST_URL"
+
+	defaultValidateStorageClusterTimeout       = 900 * time.Second
+	defaultValidateStorageClusterRetryInterval = 30 * time.Second
+	defaultValidateUninstallTimeout            = 900 * time.Second
+	defaultValidateUninstallRetryInterval      = 30 * time.Second
 )
 
+var pxSpecGenURL string
```
piyush-nimbalkar, line 33: nit: use 
```
var (
  pxSpecGenURL string
  pxEndpoint string
)
```

test/integration_test/common_test.go
```diff
@@ -3,27 +3,75 @@
...
+	flag.StringVar(&pxEndpoint,
+		"portworx-endpoint",
+		"",
+		"Portworx Endpoint that defines the version of Portworx and its components")
+	flag.Parse()
 	os.Exit(m.Run())
 }
 
+// Here we make StorageCluster object and add all the common basic parameters that all StorageCluster should have
+func constructStorageCluster() (*corev1.StorageCluster, error) {
```
piyush-nimbalkar, line 50: If we want to test a released version of Portworx, how would this work? 
I assume, we would only pass the endpoint in that case. But here we are not setting the portworx version, so the operator will end up installing the latest available version every time, right?

test/integration_test/common_test.go
nikolaypopov: Done

test/integration_test/common_test.go
nikolaypopov: Added setting imageName for all deployments as well as EnvVar in case of `edge`.

piyush-nimbalkar (approved):

PR #238, opened 2021-02-01
Add NODE_NAME to the portworx pod env variables

piyush-nimbalkar:

### DIFF
--- a/drivers/storage/portworx/deployment.go
+++ b/drivers/storage/portworx/deployment.go
@@ -816,6 +816,15 @@ func (t *template) getEnvList() []v1.EnvVar {
 			Name:  pxutil.EnvKeyPortworxSecretsNamespace,
 			Value: t.cluster.Namespace,
 		},
+		"NODE_NAME": {
+			Name: "NODE_NAME",
+			ValueFrom: &v1.EnvVarSource{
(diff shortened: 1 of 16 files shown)


drivers/storage/portworx/deployment.go
```diff
@@ -816,6 +816,15 @@ func (t *template) getEnvList() []v1.EnvVar {
 			Name:  pxutil.EnvKeyPortworxSecretsNamespace,
 			Value: t.cluster.Namespace,
 		},
+		"NODE_NAME": {
+			Name: "NODE_NAME",
+			ValueFrom: &v1.EnvVarSource{
+				FieldRef: &v1.ObjectFieldSelector{
+					APIVersion: "v1",
+					FieldPath:  "spec.nodeName",
```
adityadani, line 824: Do you need the quotes "" around spec.nodeName ?

adityadani (approved):

drivers/storage/portworx/deployment.go
piyush-nimbalkar: Yes, because this is a field reference. Kubernetes populates the env value in the pod using this field reference. The nodename is not taken from the code here.

PR #242, opened 2021-02-08
OPERATOR-239: Add PodSecurityPolicy rules to the serviceaccounts

saheienko:
Add podsecuritypolicy resources to the deployment to tolerate k8s cluster with enabled PodSecurityPolicy plugin.

Signed-off-by: Serhii Aheienko <serhii.aheienko@gmail.com>


deploy/podsecuritypolicy.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: operator
```
piyush-nimbalkar, line 4: Let's rename this to px-operator

deploy/podsecuritypolicy.yaml
```diff
@@ -0,0 +1,25 @@
...
+spec:
+  allowPrivilegeEscalation: false
+  allowedCapabilities:
+  - '*'
+  volumes:
+  - '*'
+  hostNetwork: true
```
piyush-nimbalkar, line 11: Does this need hostNetwork, hostPorts, hostIPC, hostPID?
Also can we restrict the volumes and allowedCapabilities?

drivers/storage/portworx/component/prometheus.go
```diff
@@ -264,6 +265,12 @@ func (c *prometheus) createOperatorClusterRole() error {
 					ResourceNames: []string{"anyuid"},
 					Verbs:         []string{"use"},
 				},
+				{
+					APIGroups:     []string{"policy"},
+					Resources:     []string{"podsecuritypolicies"},
+					ResourceNames: []string{constants.RestrictedPSPName},
+					Verbs:         []string{"use"},
+				},
```
piyush-nimbalkar, line 273: Don't we need the PSP for prometheus's cluster role created below on line 279? Did you try to deploy prometheus and check if it comes up without PSPs?

drivers/storage/portworx/podsecuritypolicies.go
```diff
@@ -0,0 +1,96 @@
+package portworx
+
+import (
+	"github.com/libopenstorage/operator/pkg/constants"
+	policyv1beta1 "k8s.io/api/policy/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func portworxPodSecurityPolicies() []policyv1beta1.PodSecurityPolicy {
```
piyush-nimbalkar, line 9: We should create a component for PSPs instead of having it here. That way the Reconcile and Delete loops will automatically handle creation and removal of these PSPs.

drivers/storage/portworx/podsecuritypolicies.go
```diff
@@ -0,0 +1,96 @@
...
+				},
+				SELinux: policyv1beta1.SELinuxStrategyOptions{
+					Rule: policyv1beta1.SELinuxStrategyRunAsAny,
+				},
+				SupplementalGroups: policyv1beta1.SupplementalGroupsStrategyOptions{
+					Rule: policyv1beta1.SupplementalGroupsStrategyRunAsAny,
+				},
+			},
+		},
+		{
```
piyush-nimbalkar, line 67: We should create this only if Lighthouse is enabled. Nowadays, we mostly don't run Lighthouse so don't want an extra PSP created when LH is not being used.

drivers/storage/portworx/portworx.go
```diff
@@ -271,6 +271,10 @@ func (p *portworx) PreInstall(cluster *corev1.StorageCluster) error {
 		return err
 	}
 
+	if err := p.createPodPolicies(); err != nil {
+		return err
+	}
+
```
piyush-nimbalkar, line 277: We won't need this if we create a component for the PSPs

drivers/storage/portworx/component/prometheus.go
saheienko: Haven't enabled the prometheus operator, going to check this. Also if PodSecurityPolicy plugin is enabled each pod should have an assigned psp to be accepted by kube-apiserver.

drivers/storage/portworx/podsecuritypolicies.go
saheienko: Added this separately to be sure that psp is created before other resources as components aren't ordered.
Create a component for PSP.

deploy/podsecuritypolicy.yaml
saheienko: Updated

deploy/podsecuritypolicy.yaml
saheienko: No, this parameters aren't required. Updated.

drivers/storage/portworx/component/prometheus.go
saheienko: Prometheus needs some psp too:
```
2s          Warning   FailedCreate                        replicaset/px-prometheus-operator-655df66dfd                     Error creating: pods "px-prometheus-operator-655df66dfd-" is forbidden: unable to validate against any pod security policy: []
```

drivers/storage/portworx/podsecuritypolicies.go
saheienko: Removing psp for lighthouse.

drivers/storage/portworx/component/podsecuritypolicies.go
```diff
@@ -0,0 +1,126 @@
...
+	return nil
+}
+
+func (p *podsecuritypolicies) MarkDeleted() {
+}
+
+func portworxPodSecurityPolicies() []policyv1beta1.PodSecurityPolicy {
+	return []policyv1beta1.PodSecurityPolicy{
+		{
+			ObjectMeta: metav1.ObjectMeta{
```
piyush-nimbalkar, line 70: We should add owner reference of the cluster to the PSPs. That way if the StorageCluster is force deleted, Kubernetes will remove the PSPs. We do the same for all almost all other components created by the operator.

drivers/storage/portworx/component/autopilot.go
```diff
@@ -189,6 +190,12 @@ func (c *autopilot) createClusterRole() error {
 					Resources: []string{"*"},
 					Verbs:     []string{"*"},
 				},
+				{
```
piyush-nimbalkar, line 193: Does autopilot really need a PSP?

pkg/controller/storagecluster/stork.go
```diff
@@ -354,6 +360,12 @@ func (c *Controller) createStorkSchedClusterRole() error {
 					Resources: []string{"leases"},
 					Verbs:     []string{"get", "list", "watch", "create", "update"},
 				},
+				{
```
piyush-nimbalkar, line 363: Does the stork scheduler need a PSP?

drivers/storage/portworx/component/prometheus.go
piyush-nimbalkar: If we need it, then don't we need to add it to the prometheus role as well?

piyush-nimbalkar (commented):
Are all the unit tests passing (`make test`)?
Can you fix the tests and add new tests wherever required?

drivers/storage/portworx/component/podsecuritypolicies.go
saheienko: Updated.

drivers/storage/portworx/component/autopilot.go
saheienko: Yes, if PodSecurityPolicy admission plugin is enabled each controller have to use any PSP to create pods.
https://kubernetes.io/docs/concepts/policy/pod-security-policy/#authorizing-policies

saheienko:
Added unit tests for the podsecuritypolicy component.

piyush-nimbalkar (approved):
Is the prometheus image pull error happening consistently?

saheienko:
Yes, prometheus hasn't been started.

PR #247, opened 2021-03-05
Add snapshot controller and update images

ggriffiths:


ggriffiths:
Breaking this PR into three smaller ones

PR #254, opened 2021-03-11
Generic volume support to StorageCluster

piyush-nimbalkar:
- Giving an option to users to add extra volumes for portworx, stork and autopilot volumes.
- Operator should reconcile immediately on any input changes from the users to the volumes and restart the pods with the new volumes.
- Adding a prefix `user-` to user provided volume names to avoid future conflicts with existing volumes.
- Supports hostPath, secret, configMap and projected volumes. Other volume sources may work (as we use k8s objects) but not tested or handled explicitly.

Sample volume spec:
```
spec:
  volumes:
  - name: s3certs
    mountPath: /etc/ecscert
    secret:
      secretName: px-s3-certs
      items:
          - key: ecss3.pem
            path: ecss3.pem
  - name: s3config
    mountPath: /etc/testingconfig
    configMap:
      name: test-cm
      items:
          - key: some.config
            path: some.config
  - name: hostpath
    mountPath: /etc/testinghostpath
    hostPath:
      path: test-hostpath
```

TODO: unit tests


adityadani (approved):
lgtm

PR #258, opened 2021-03-17
Allow writing to root file system inside the container

piyush-nimbalkar:
This is needed by the node-wiper container as it writes its
done status in a file under /tmp


adityadani (approved):
Let's file a bug on Talisman to remove this dependency on /tmp

PR #259, opened 2021-03-17
Reconcile components in order of their priority

piyush-nimbalkar:
We should install certain components like PSP, security
before other components start to resolve run time
dependencies.

### DIFF
--- a/drivers/storage/portworx/component/autopilot.go
+++ b/drivers/storage/portworx/component/autopilot.go
@@ -75,6 +75,14 @@ type autopilot struct {
 	k8sClient client.Client
 }
 
+func (c *autopilot) Name() string {
+	return AutopilotComponentName
+}
+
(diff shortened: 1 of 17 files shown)


harsh-px:
what was the actual issue here? Looking at the change, we are creating PSPs and security components first. Did the reverse order cause the autopilot deployment to get stuck in blocked state?

piyush-nimbalkar:
Yes, autopilot was getting blocked. As Kubernetes doesn't retry starting the pod, it remains in blocked state. 
Restarting the pod works, but it is inconvenient.

harsh-px:
Makes sense.

> As Kubernetes doesn't retry starting the pod, it remains in blocked state.

So much for being eventually consistent :)

adityadani (approved):
lgtm

PR #260, opened 2021-03-19
PWX-19064-pt5: PX-Operator fixes for BottleRocket

zoxpx:
Introduces a feature "BottleRocketAMI" that adds a few tweaks to the Portworx spec.

Signed-off-by: Zoran Rajic <zox@portworx.com>

### DIFF
--- a/drivers/storage/portworx/deployment_test.go
+++ b/drivers/storage/portworx/deployment_test.go
@@ -1882,6 +1882,54 @@ func TestPodSpecForK3s(t *testing.T) {
 	assertPodSpecEqual(t, expected, &actual)
 }
 
+func TestPodSpecForBottleRocketAMI(t *testing.T) {
+	fakeClient := fakek8sclient.NewSimpleClientset()
+	coreops.SetInstance(coreops.New(fakeClient))
+	fakeClient.Discovery().(*fakediscovery.FakeDiscovery).FakedServerVersion = &version.Info{
+		GitVersion: "v1.13.2",
+	}
+
+	nodeName := "testNode"
+
+	cluster := &corev1.StorageCluster{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "px-cluster",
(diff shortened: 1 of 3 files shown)


drivers/storage/portworx/util/feature.go
```diff
@@ -10,6 +10,8 @@ type Feature string
 const (
 	// FeatureCSI is used to indicate CSI feature
 	FeatureCSI Feature = "CSI"
+	// FeatureBottleRocketAMI is used to activate BottleRocket AMI tweaks
```
piyush-nimbalkar, line 13: Instead of feature flag, can we put this as an annotation here - https://github.com/libopenstorage/operator/blob/master/drivers/storage/portworx/util/util.go#L58?

Feature flags was reserved mostly for portworx features like CSI, may be dm-thin, etc.
Also, eventually we want to remove these annotations and auto-detect most things about the environment.

drivers/storage/portworx/deployment_test.go
```diff
@@ -1882,6 +1882,54 @@ func TestPodSpecForK3s(t *testing.T) {
 	assertPodSpecEqual(t, expected, &actual)
 }
 
+func TestPodSpecForBottleRocketAMI(t *testing.T) {
```
piyush-nimbalkar, line 1885: Nice. Thanks for the UTs!

piyush-nimbalkar (commented):
@zoxpx we will also have to add the toggle to the StorageCluster template in px-installer based on the `ami=bt` argument.

harsh-px (changes requested):
Let's first investigate if we can do this by default (i.e detect bottlerocket and do this automatically). We want to avoid adding more features flags or annotations.

The case is stronger for this since if the user forgets to specify it, we will fail to come up. This is not an optional enablement of an integration.

zoxpx:
Abandoning this fix in favor of https://github.com/libopenstorage/operator/pull/286

PR #261, opened 2021-03-22
OPERATOR-284: Add tls section to StorageCluster security spec

puremudassir:
Supported scenario: user creates+installs tls certs

See https://docs.google.com/document/d/1piY8Pt79IhNrPmZuC0YbdOV82pb_fYNNDlJc0s0m7xU/edit#

supported spec
```
tls:
   enabled: true
   advancedOptions
     apirootCA:
       filename:
     certKey:
       filename:
     serverCert:
       filename:
```

Added `StorageCluster.spec.security.auth.enabled` flag
Passes the following args to oci-monitor:
  `-apirootca <filename> -apicert <filename> -apikey <filename>`

For now, setting both `PX_ENFORCE_TLS` (for portworx)
	and `PX_ENABLE_TLS` (for all other clients using `openstorage`)
	Eventually, `PX_ENFORCE_TLS` will be removed from portworx, but until
	that happens, operator needs to set it

Rename method: `SecurityEnabled` -> `AuthEnabled`
  because security now has 2 components: auth & tls
  
TODO: some more unit tests


drivers/storage/portworx/portworx_test.go
```diff
@@ -6235,6 +6379,111 @@ func manifestSetup() {
...
+	}
+	// test
+	s, _ = json.MarshalIndent(cluster.Spec.Security, "", "\t")
+	t.Logf("Security spec under test = \n, %v", string(s))
+	setTLSSpecDefaults(cluster)
+	// verify
+	verifyTLSSpecFileNames(t, cluster, stringPtr(defaultTLSCACertFilename), stringPtr(defaultTLSServerCertFilename), stringPtr(defaultTLSServerKeyFilename))
+}
+
+func verifyTLSSpecFileNames(t *testing.T, cluster *corev1.StorageCluster, CACertFileName, serverCertFileName, serverKeyFileName *string) {
```
piyush-nimbalkar, line 6466: nit: *CACertFileName -> caCertFileName

pkg/util/test/util.go
```diff
@@ -1040,3 +1040,56 @@ func validateStorageClusterIsOnline(cluster *corev1.StorageCluster, timeout, int
 
 	return cluster, nil
 }
+
+// CreatePodSpecWithTLS is a helper method
+func CreatePodSpecWithTLS(CACertFileName, serverCertFileName, serverKeyFileName *string) *corev1.StorageCluster {
```
piyush-nimbalkar, line 1045: This is creating and returning a StorageCluster spec and not pod spec. Should we call this `CreateClusterWithTLS`?

drivers/storage/portworx/portworx_test.go
```diff
@@ -6235,6 +6379,111 @@ func manifestSetup() {
...
+	// all filenames supplied
+	// setup
+	CACertFileName := stringPtr("testCA.crt")
+	serverCertFileName := stringPtr("testServer.crt")
+	serverKeyFileName := stringPtr("testServer.key")
+	// test
+	cluster := testutil.CreatePodSpecWithTLS(CACertFileName, serverCertFileName, serverKeyFileName)
+	s, _ := json.MarshalIndent(cluster.Spec.Security, "", "\t")
+	t.Logf("Security spec under test = \n, %v", string(s))
+	setTLSSpecDefaults(cluster)
```
piyush-nimbalkar, line 6392: Can we call `driver.SetDefaultsOnStorageCluster` instead of this private method?
I think it's better to avoid testing private methods (if possible). Private methods tend to get refactored internally which shouldn't require a change in UTs.

drivers/storage/portworx/util/util_test.go
```diff
@@ -0,0 +1,194 @@
...
+
+func TestAuthEnabled(t *testing.T) {
+	// security.enabled		security.auth.enabled		Auth enabled?
+	// =============================================================
+	// true					true						true
+	// true					false						false
+	// true					nil							true
+	// false				true						true
+	// false				false						false
+	// false				nil							false
```
piyush-nimbalkar, line 56: nit: indentation fix

drivers/storage/portworx/util/util_test.go
```diff
@@ -0,0 +1,194 @@
...
+
+func TestIsTLSEnabledOnCluster(t *testing.T) {
+	// security.enabled		security.tls.enabled		TLS enabled?
+	// =============================================================
+	// true					true						true
+	// true					false						false
+	// true					nil							true
+	// false				true						true
+	// false				false						false
+	// false				nil							false
```
piyush-nimbalkar, line 106: nit: indentation fix

drivers/storage/portworx/util/util_test.go
```diff
@@ -0,0 +1,194 @@
...
+	cluster := createPodSpecWithAuth()
+	cluster.Spec.Security.Enabled = securityEnabled
+	cluster.Spec.Security.Auth.Enabled = authEnabled
+	s, _ := json.MarshalIndent(cluster.Spec.Security, "", "\t")
+	t.Logf("Security spec under test = \n, %v", string(s))
+	actual := AuthEnabled(&cluster.Spec)
+	assert.Equal(t, actual, expectedResult)
+}
+
+func createPodSpecWithAuth() *corev1.StorageCluster {
```
piyush-nimbalkar, line 170: Should we call this `createClusterWithAuth`?

pkg/controller/storagecluster/controller_test.go
```diff
@@ -2534,7 +2534,20 @@ func TestUpdateStorageClusterWithRollingUpdateStrategy(t *testing.T) {
 	err = testutil.List(k8sClient, revisions)
 	require.NoError(t, err)
 	require.Len(t, revisions.Items, 2)
-	require.Equal(t, int64(2), revisions.Items[1].Revision)
+	// s, _ := json.MarshalIndent(revisions.Items, "", "\t")
+	// t.Logf(string(s))
```
piyush-nimbalkar, line 2538: nit: Let's remove commented code, if not needed

pkg/controller/storagecluster/controller_test.go
```diff
@@ -2534,7 +2534,20 @@ func TestUpdateStorageClusterWithRollingUpdateStrategy(t *testing.T) {
...
+	for i, rev := range revisions.Items {
+		if rev.Revision == int64(1) {
+			revision1 = &revisions.Items[i]
+		}
+		if rev.Revision == int64(2) {
+			revision2 = &revisions.Items[i]
+		}
+	}
+	require.NotNil(t, revision1, "Unable to find revision 1")
+	require.NotNil(t, revision2, "Unable to find revision 2")
```
piyush-nimbalkar, line 2550: nit: We could simply write this like below - 
```
require.ElementsMatch(t, []int64{1, 2}, []int64{revisions.Items[0].Revision, revisions.Items[1].Revision})
```

Same could be used below

deploy/crds/core_v1_storagecluster_crd.yaml
```diff
@@ -866,6 +870,42 @@ spec:
                           type: string
                           description: Shared secret is the name of the Kubernetes secret containing the shared key
                             used for signing RBAC tokens. The secret has to be present in the StorageCluster namespace.
+                tls:
+                  type: object
+                  description: Authorization configurations for a TLS enabled storage cluster
```
ggriffiths, line 875:
```suggestion
                  description: Security configurations for a TLS enabled storage cluster
```

drivers/storage/portworx/component/security.go
```diff
@@ -89,11 +89,20 @@ func (c *security) Initialize(
...
 // IsEnabled checks if the components needs to be enabled based on the StorageCluster
 func (c *security) IsEnabled(cluster *corev1.StorageCluster) bool {
-	return pxutil.SecurityEnabled(cluster)
+	return pxutil.AuthEnabled(&cluster.Spec) || pxutil.IsTLSEnabledOnCluster(&cluster.Spec)
 }
 
 // Reconcile reconciles the component to match the current state of the StorageCluster
 func (c *security) Reconcile(cluster *corev1.StorageCluster) error {
+	if pxutil.AuthEnabled(&cluster.Spec) {
+		logrus.Infof("Auth is enabled ... creating secrets")
```
ggriffiths, line 98:
```suggestion
```

ggriffiths: This log will cause a lot of spam in the operator logs, I think we can remove this. This log happens everytime the storagecluster obj is edited or checked.

drivers/storage/portworx/component/security.go
```diff
@@ -89,11 +89,20 @@ func (c *security) Initialize(
...
+		return c.reconcileAuth(cluster)
+	}
+	logrus.Infof("Auth is not enabled ... skipping secret creation")
```
ggriffiths, line 101: same as above

drivers/storage/portworx/components_test.go
```diff
@@ -3128,6 +3120,42 @@ func TestSecurityInstall(t *testing.T) {
...
+
+	// validate with security disabled, but auth enabled
+	cluster = &corev1.StorageCluster{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "px-cluster",
+			Namespace: "kube-test",
+		},
+		Spec: corev1.StorageClusterSpec{
+			Security: &corev1.SecuritySpec{
+				Enabled: false,
```
ggriffiths, line 3133: I think in this case, security is disabled, and therefore TLS should also be disabled.

Parent enabled should take precedence over child enabled.

pkg/apis/core/v1/storagecluster.go
```diff
@@ -122,6 +122,28 @@ type VolumeSpec struct {
...
+	APIRootCA  *CertLocation `json:"apirootCA,omitempty"`
+	ServerKey  *CertLocation `json:"certKey,omitempty"`
+	ServerCert *CertLocation `json:"serverCert,omitempty"`
+}
+
+// CertLocation specifies where portworx should pick up the certificate.
+// Certificate can be in a file on a fixed location or in a secret
+type CertLocation struct {
+	// filename on the node for the cert file. All files must be installed on a fixed location (/etc/pwx)
+	FileName *string `json:"filename,omitempty"`
```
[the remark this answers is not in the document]
puremudassir: > Does this need to be a pointer?

Yes, because in future we will have either a k8s secret or a file as source. Something like:
```
serverCert:
  filename: <could be blank if k8s secret supplied as source>
  secretKeyRef:
    name: 
    key: 
}
```

drivers/storage/portworx/component/security.go
```diff
@@ -89,11 +89,20 @@ func (c *security) Initialize(
 
 // IsEnabled checks if the components needs to be enabled based on the StorageCluster
 func (c *security) IsEnabled(cluster *corev1.StorageCluster) bool {
-	return pxutil.SecurityEnabled(cluster)
+	return pxutil.AuthEnabled(&cluster.Spec) || pxutil.IsTLSEnabledOnCluster(&cluster.Spec)
```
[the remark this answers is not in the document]
puremudassir: > I think we should make TLS a separate component to keep it simple and call this component as Auth instead of Security. It would be only needed when we start auto-generating certs.

As discussed in slack: renamed `security` --> `auth`. Did not create a `tls' component because it would be empty right now. Will add it in the future when we do auto-generating tls certs and there is work for tls component.

ggriffiths (approved):

deploy/crds/core_v1_storagecluster_crd.yaml
```diff
@@ -866,6 +870,42 @@ spec:
...
+                  properties:
+                    enabled:
+                      type: boolean
+                      description: Flag indicating whether TLS features need to be enabled for the storage 
+                        cluster. If omited take the value of the parent flag (security.enabled)
+                    advancedOptions:
+                      type: object
+                      description: Overrides for the default values in the parent tls section.
+                      properties:
+                        apirootCA:
```
piyush-nimbalkar, line 885: *rootCA

drivers/storage/portworx/util/util.go
```diff
@@ -36,6 +37,8 @@ const (
 	DefaultOpenshiftStartPort = 17001
 	// PortworxSpecsDir is the directory where all the Portworx specs are stored
 	PortworxSpecsDir = "/configs"
+	// PortworxTLSCertsDir is where all tls certs are to be placed
+	PortworxTLSCertsDir = "/etc/pwx/"
```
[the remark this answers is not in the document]
piyush-nimbalkar: Once we go this route it won't be easy to change and remove the default base path for existing customers. We won't know whether we need to add the base path or not.
I feel keeping it open-ended is safer in case the users are unwilling to move/copy their certificates in /etc/pwx. Also, it could be some automation in their infra that puts certificates in a location which is not /etc/pwx. The only downside  is that users will have to provide the full path every time.

piyush-nimbalkar (approved):
LGTM after a minor comment in the storagecluster schema.

This is a very neatly done change @puremudassir!
Thanks for adding all the UTs.

drivers/storage/portworx/util/util.go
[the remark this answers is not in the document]
puremudassir: > Once we go this route it won't be easy to change and remove the default base path for existing customers. We won't know whether we need to add the base path or not.
> I feel keeping it open-ended is safer in case the users are unwilling to move/copy their certificates in /etc/pwx. Also, it could be some automation in their infra that puts certificates in a location which is not /etc/pwx. The only downside is that users will have to provide the full path every time.

As discussed on slack: we will merge it into a feature branch separate from master until we've made a decision to ship TLS changes, so customer backwards compat issues will not arise. Will address https://portworx.atlassian.net/browse/OPERATOR-299  in a separate PR

PR #263, opened 2021-03-24
OPERATOR-285: get autopilot working with TLS enabled

puremudassir:
Prerequisite: https://github.com/libopenstorage/operator/pull/261

Operator adds the following env variables to the autopilot pod:
```
env:
- name: PX_CA_CERT_SECRET
  value: <default to portworx-api-root-ca if not supplied by user>
- name: PX_CA_CERT_SECRET_KEY
  value: <default to root-ca if not supplied by user>
- name: PX_ENABLE_TLS
  value: "true"
```
These tell autopilot that TLS is enabled, and the location (k8s secret) to pick up the CA cert to use to connect to portworx services


ggriffiths:
@puremudassir I noticed you moved this to be a draft. Do still want a review on this?

puremudassir:
> @puremudassir I noticed you moved this to be a draft. Do still want a review on this?

Yes I do want a review, but waiting for the parent PR to merge. This one adds only 2 commits to #261 - the last 2 commits. When the "parent" PR #261 gets merged this PR will show the correct diff.

drivers/storage/portworx/component/autopilot.go
```diff
@@ -386,6 +387,64 @@ func (c *autopilot) getAutopilotDeploymentSpec(
...
+		//      value: <default>
+		//    - name: PX_ENABLE_TLS
+		//	    value: "true"
+		CASecretNameValue := ""
+		CASecretKeyValue := ""
+		EnableTLSValue := ""
+		if cluster.Spec.Autopilot != nil && cluster.Spec.Autopilot.Env != nil {
+			CASecretNameValue = k8sutil.GetValueFromEnv(pxutil.EnvKeyAutopilotCASecretName, cluster.Spec.Autopilot.Env)
+			CASecretKeyValue = k8sutil.GetValueFromEnv(pxutil.EnvKeyAutopilotCASecretKey, cluster.Spec.Autopilot.Env)
+			EnableTLSValue = k8sutil.GetValueFromEnv(pxutil.EnvKeyPortworxEnableTLS, cluster.Spec.Autopilot.Env)
```
piyush-nimbalkar, line 412: When TLS is enabled in Portworx, the clients will have to use TLS, right? Or is that optional?
Assuming that it is necessary for clients to use TLS, we should not ask users to pass the environment variables twice in the outer env vars as well autopilot envs. In future in stork envs as well.

drivers/storage/portworx/util/util.go
```diff
@@ -150,6 +150,15 @@ const (
 	// EnvKeyMarketplaceName env var for the name of the source marketplace
 	EnvKeyMarketplaceName = "MARKETPLACE_NAME"
 
+	// EnvKeyAutopilotCASecretName env var for the name of the k8s secret containing the CA cert needed to connect to portworx when TLS is enabled
+	EnvKeyAutopilotCASecretName = "PX_CA_CERT_SECRET"
+	// EnvKeyAutopilotCASecretKey env var for the name of the key in the k8s secret which will retrieve the CA cert needed to connect to portworx when TLS is enabled
+	EnvKeyAutopilotCASecretKey = "PX_CA_CERT_SECRET_KEY"
+	// DefaultCASecretName is the default value for EnvKeyAutopilotCASecretName
+	DefaultCASecretName = "portworx-api-root-ca"
```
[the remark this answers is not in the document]
piyush-nimbalkar: In case of operator generated certs, would we be generating just server key and certs? Who provides the root CA?
If all of them are generated by the operator and only portworx clients are going to consume them, why not store then in the same secret.
If we store them in separate secrets, we will have to take two secret names at the top level - 
```
spec:
  tls:
    enabled: bool
    rootCASecret: string
    serverCertSecret: string
    advancedOptions: {}
```
Is this how we would do it?

drivers/storage/portworx/component/autopilot.go
piyush-nimbalkar: Should we be passing the secret name and key to the clients or just mount the secret directly as a file in them?

puremudassir: > Should we be passing the secret name and key to the clients or just mount the secret directly as a file in them?

That's a good idea. We will probably do that for the "server cert" and the "server key" that needs to be passed only to porx (i.e. mount the secret either as file or env)

For the CA cert, `openstorage`expects the name of the secret and doesn't read any environment variable
(https://github.com/libopenstorage/openstorage/blob/baba77246a2578575885654556cb538d2869fbbe/volume/drivers/pwx/connection.go#L73)
Therefore all application which need to connect to porx using tls (autopilot, stork) will need to be passed the secret name.

To make the change you suggest, we will need to change this here: https://github.com/libopenstorage/openstorage/blob/master/volume/drivers/pwx/connection.go#L178 ,
push the change to `openstorage`, update vendor packages in multiple repos etc etc.

drivers/storage/portworx/util/util.go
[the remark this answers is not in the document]
puremudassir: > In case of operator generated certs, would we be generating just server key and certs? Who provides the root CA?
> If all of them are generated by the operator and only portworx clients are going to consume them, why not store then in the same secret.
> If we store them in separate secrets, we will have to take two secret names at the top level -
> 
> ```
> spec:
>   tls:
>     enabled: bool
>     rootCASecret: string
>     serverCertSecret: string
>     advancedOptions: {}
> ```
> 
> Is this how we would do it?

For now, we assume client has generated everything, and placed cert files on the nodes ... therefore advancedOptions does not support cert as the source instead of file (yet). We can discuss this question in a later PR.

As for consuming the certs, there are the following certs:
- The CA certificate (which needs to be consumed by porx (it acts as both server and client for secure node-node communications over TLS) as well as clients like autopilot, stork.
- The Server cert (public key) and Server key (private key) pair that are needed only by porx (server).



What I had in mind for that future PR was this:
```
spec:
  tls:
    enabled: bool
    advancedOptions: {}
```
The ca cert is assumed to be in a "default" secret, so is the server cert and key.


```
spec:
  tls:
    enabled: bool
    advancedOptions:
      serverKey:
	    secretKeyRef:
	      name:
	      key:
      serverCert:
	    secretKeyRef:
	      name:
	      key:

```
`serverKey` are `serverCert` are not in the "default" secret ... so we provide reference to the secret

drivers/storage/portworx/component/autopilot.go
piyush-nimbalkar: I don't have a strong preference over passing the values as secret names or files. Just thought that it would be easier to mount it as a file as most ssl libraries consume files, so the client/server does not have to talk to k8s every time to get the CA or server certs.

@puremudassir @lpabon On a different note, not directly related to this PR - I am assuming we are enabling TLS for Kubernetes only, at least initially. But reading from a k8s secret from openstorage means it would work only for k8s. I am not sure if we ever plan to support other orchestrators, but you never know. In such a case, would it be better to read the certs as a file?

drivers/storage/portworx/util/util.go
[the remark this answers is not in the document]
piyush-nimbalkar: I am not completely clear on how we will be simplifying the workflow for operator generated certs. 
What you have done is this PR is good and we don't need to block this PR as we will be adding changes incrementally. But let's discuss offline on how we plan to simplify TLS experience for most users.

piyush-nimbalkar (approved):

PR #265, opened 2021-03-26
[CSI] Update images and fix breaking images changes

ggriffiths:


pkg/util/util.go
```diff
@@ -30,6 +31,7 @@ var (
 	commonDockerRegistries = map[string]bool{
 		"docker.io":                   true,
 		"quay.io":                     true,
+		"k8s.gcr.io":                  true,
```
piyush-nimbalkar, line 34: Let's hold onto this change.
The issue here is that we have customers using private registries and storing their images with the complete path. For instance, `private.reg.io/k8s.gcr.io/pause`, etc. 
Until now we used to remove the other common registries from our images, when prefixing to the image. But we missed `k8s.gcr.io` and `gcr.io`. Now adding them here will break the images in existing customer envs.

piyush-nimbalkar: We have a ticket to track this issue - https://portworx.atlassian.net/browse/OPERATOR-240

ggriffiths: Ok, will remove this

drivers/storage/portworx/components_test.go
```diff
@@ -3944,6 +3944,92 @@ func TestCSIInstallWithNewerCSIVersion(t *testing.T) {
...
+	require.Len(t, deployment.OwnerReferences, 1)
+	require.Equal(t, cluster.Name, deployment.OwnerReferences[0].Name)
+	require.Equal(t, len(expectedDeployment.Spec.Template.Spec.Containers), len(deployment.Spec.Template.Spec.Containers))
+	for i := 0; i < len(expectedDeployment.Spec.Template.Spec.Containers); i++ {
+		require.Equal(t, expectedDeployment.Spec.Template.Spec.Containers[i], deployment.Spec.Template.Spec.Containers[i])
+	}
+	for i := 0; i < len(deployment.Spec.Template.Spec.Containers); i++ {
+		require.Equal(t, expectedDeployment.Spec.Template.Spec.Containers[i], deployment.Spec.Template.Spec.Containers[i])
+	}
+	require.Equal(t, expectedDeployment.Spec.Template.Spec.Containers, deployment.Spec.Template.Spec.Containers)
```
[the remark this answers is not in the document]
ggriffiths: Ah yeah, this is leftover from fixing this test. The diff between containers was easier to read. Removing now

drivers/storage/portworx/components_test.go
```diff
@@ -3944,6 +3944,91 @@ func TestCSIInstallWithNewerCSIVersion(t *testing.T) {
...
+	require.Equal(t, expectedDeployment.Namespace, deployment.Namespace)
+	require.Len(t, deployment.OwnerReferences, 1)
+	require.Equal(t, cluster.Name, deployment.OwnerReferences[0].Name)
+	require.Equal(t, len(expectedDeployment.Spec.Template.Spec.Containers), len(deployment.Spec.Template.Spec.Containers))
+	for i := 0; i < len(expectedDeployment.Spec.Template.Spec.Containers); i++ {
+		require.Equal(t, expectedDeployment.Spec.Template.Spec.Containers[i], deployment.Spec.Template.Spec.Containers[i])
+	}
+	for i := 0; i < len(deployment.Spec.Template.Spec.Containers); i++ {
+		require.Equal(t, expectedDeployment.Spec.Template.Spec.Containers[i], deployment.Spec.Template.Spec.Containers[i])
+	}
```
piyush-nimbalkar, line 4021: nit: We still don't need these as we are comparing the whole deployment specs.
Same below.

piyush-nimbalkar (approved):

drivers/storage/portworx/component/csi.go
```diff
@@ -460,6 +460,24 @@ func getCSIDeploymentSpec(
 	}
 	imagePullPolicy := pxutil.ImagePullPolicy(cluster)
 
+	var args []string
+	if util.GetImageMajorVersion(provisionerImage) >= 2 {
```
piyush-nimbalkar, line 464: Would it be possible to add a test for this case where the provisioner image version is greater than 2.x?

piyush-nimbalkar (approved):

drivers/storage/portworx/component/csi.go
ggriffiths: Yeah I'll add one - I think maybe via the desiredimages field.

ggriffiths: I don't think this is possible for now via the desired images, seems to just get reset. 

Once we need to start support 2.x csi provisioner, this will have coverage.

PR #267, opened 2021-04-05
OPERATOR-297: Support supplying tls certs through k8s secrets

puremudassir:
Added secretRef to tls.advancedOptions.rootCA etc

any certs supplied through secrets are mapped on the runC container here:
/api-tls-certs/ca-cert/
/api-tls-certs/server-cert/
/api-tls-certs/server-key/

A volume and volumeMount is created for each secret that is presented
  through k8s secret.
Example:
```
volumes:
- name: apirootca
  secret:
    secretName: <secret name>
    items:
    - key: <secret key>
      path: <secret key>
volumeMounts:
- name: apirootca
  mountPath: /api-tls-certs/ca-cert/
```
Oci-Mon argument: `-apirootca /api-tls-certs/ca-cert/<secret key>`

Allows the user to stage tls certs in secret in addition to folder location on node.

Closes # OPERATOR-297


drivers/storage/portworx/util/util.go
```diff
@@ -635,32 +652,72 @@ func AppendTLSEnv(clusterSpec *corev1.StorageClusterSpec, envMap map[string]*v1.
...
+		if IsEmptyOrNilCertLocation(advancedOptions.ServerKey) {
+			return nil, fmt.Errorf("spec.security.tls.advancedOptions.serverKey is required")
 		}
-		apicertFilename := advTLSOptions.ServerCert.FileName
-		if advTLSOptions.ServerKey == nil || advTLSOptions.ServerKey.FileName == nil {
-			return nil, fmt.Errorf("spec.security.tls.advancedOptions.serverKey.filename is required")
+
+		apirootca, apicert, apikey := "", "", ""
+		if !IsEmptyOrNilSecretReference(advancedOptions.RootCA.SecretRef) {
+			apirootca = path.Join(DefaultTLSCACertMountPath, *advancedOptions.RootCA.SecretRef.SecretKey)
```
ggriffiths, line 669: For the values for `apirootca`, `apicert`, and `apikey`, I think you need to get the value stored inside of the secret?

This logic will add the secret key to the path, instead of the secret value. 

For example, we get the auth secret _value_ here:
https://github.com/libopenstorage/operator/blob/master/drivers/storage/portworx/component/security.go#L346-L349

ggriffiths: The secret has a name in k8s, and inside are k/v pairs. So we need to get the value inside of the secret for a given key that we define, maybe `root-ca` for example.

ggriffiths: `GetSecretValue` might be a better utility to use, given we haven't pulled the secret from k8s yet:
https://github.com/libopenstorage/operator/blob/5946543b9edd96b1d82952fd88c79c09a1324b25/drivers/storage/portworx/util/util.go#L613

puremudassir: > The secret has a name in k8s, and inside are k/v pairs. So we need to get the value inside of the secret for a given key that we define, maybe `root-ca` for example.

We are mounting the secrets, so porx will read it as files. See `GetVolumeInfoForTLSCerts` here: https://github.com/libopenstorage/operator/pull/267/files#diff-6407420d21ace61f620208f9f8115a20d3dc55065ee4b61185b5fa465de0001aR1183

Example:
```
        rootCA:
          secretRef:
            secretName: insecureca
            secretKey: insecure_ca.crt
```
This will mount result in secret mounted to `/api-tls-certs/ca-cert/insecure_ca.crt`

Why mount as files instead of getting porx to read from the environment? This will simplify porx code. Porx is supplied a single argument ( `-apirootca <file location` ) irrespective of how the secret is supplied (secret installed as a file on the node or supplied as a k8s secret)

puremudassir (commented):
I think I will schedule a quick meeting to go over and get feedback. I will also add the future owners of operator (@pure-jliao  and Haibin) to the meeting and the PR

ggriffiths (commented):
Looks great, nice test coverage! Just one comment on how we're getting the secret values.

drivers/storage/portworx/util/util.go
ggriffiths: Yes that makes sense to mount the secret. What I don't see is where we're using k8s secret value. `secretRef.secretKey` seems to behave more as a `secretRef.certPath`

piyush-nimbalkar (approved):

puremudassir:
Closing this for now ... if there's any further feedback in tomorrow's meeting, will send another PR

PR #268, opened 2021-04-06
OPERATOR-256 Add telemetry component

harsh-px:

Adds telemetry component which will create CCM configmap and also add CCM sidecar container to oci-monitor

Closes #


drivers/storage/portworx/component/telemetry.go
```diff
@@ -0,0 +1,170 @@
...
+        "path": "/dev/null"
+      }
+    }
+`, cluster.Namespace, TelemetryArcusLocationFilename)
+
+	data := map[string]string{
+		TelemetryPropertiesFilename: config,
+	}
+
+	if location, present := cluster.Annotations[pxutil.AnnotationTelemetryArcusLocation]; present && len(location) > 0 {
```
harsh-px, line 150: this is used to switch between staging and production arcus. Since we don't use end users to use staging, using an annotation to drive this.

drivers/storage/portworx/deployment.go
```diff
@@ -337,6 +336,20 @@ func (p *portworx) GetStoragePodSpec(
 		}
 	}
 
+	if cluster.Spec.Telemetry != nil && cluster.Spec.Telemetry.Enabled {
+		telemetryContainer := t.telemetryContainer()
+		if telemetryContainer != nil {
+			if len(telemetryContainer.Image) == 0 {
```
harsh-px, line 342: The CCM container hasn't been pushed to a public docker hub yet. Hence we don't have a default. So we need to provide an image in the spec

drivers/storage/portworx/component/portworx_basic.go
```diff
@@ -236,7 +236,7 @@ func (c *portworxBasic) createClusterRole() error {
 				{
 					APIGroups: []string{""},
 					Resources: []string{"secrets"},
-					Verbs:     []string{"get", "list", "watch"},
+					Verbs:     []string{"get", "list", "watch", "create", "delete", "update"},
```
adityadani, line 239: Can you use the namespace scope Role `portworx`

drivers/storage/portworx/component/telemetry.go
```diff
@@ -0,0 +1,170 @@
...
+      "server": {
+        "hostname": "0.0.0.0"
+      },
+      "logupload": {
+        "logfile_patterns": [
+            "/var/cores/*diags*",
+            "/var/cores/*px-cores*",
+            "/var/cores/*.heap",
+            "/var/cores/*.stack",
+            "/var/cores/.alerts/alerts*"
```
adityadani, line 118: Add `/var/cores/*.log` here instead of as additional_files ?

drivers/storage/portworx/component/telemetry.go
adityadani: Do we need a default location if this annotation is not set ?

drivers/storage/portworx/deployment.go
```diff
@@ -1120,7 +1191,95 @@ func (t *template) getCSIVolumeInfoList() []volumeInfo {
...
+			},
+			configVolume,
+		}
+		return volumeInfoList
+	}
+	return []volumeInfo{}
+}
+
 func (t *template) getK3sVolumeInfoList() []volumeInfo {
+	if !t.isIKS {
```
adityadani, line 1279: this should be `t.isK3S`

drivers/storage/portworx/manifest/manifest.go
```diff
@@ -33,7 +33,9 @@ const (
 	defaultPrometheusOperatorImage        = "quay.io/coreos/prometheus-operator:v0.34.0"
 	defaultPrometheusConfigMapReloadImage = "quay.io/coreos/configmap-reload:v0.0.1"
 	defaultPrometheusConfigReloaderImage  = "quay.io/coreos/prometheus-config-reloader:v0.34.0"
-	defaultManifestRefreshInterval        = 3 * time.Hour
+	// TODO change this once phone home team is done uploading image: https://jira.purestorage.com/browse/CLOUD-57162
```
adityadani, line 36: nit: add an OPERATOR bug as well so that we won't miss this

deploy/crds/core_v1_storagecluster_crd.yaml
```diff
@@ -816,6 +816,16 @@ spec:
                       params:
                         type: object
                         description: Map of key-value params for the provider.
+            telemetry:
```
piyush-nimbalkar, line 819: Should this be part of the monitoring section?

drivers/storage/portworx/component/telemetry.go
```diff
@@ -0,0 +1,170 @@
...
+	// TelemetryConfigMapName is the name of the config map that stores the telemetry configuration for CCM
+	TelemetryConfigMapName = "px-telemetry-config"
+	// TelemetryPropertiesFilename is the name of the CCM properties file
+	TelemetryPropertiesFilename = "ccm.properties"
+	// TelemetryArcusLocationFilename is name of the file storing the location of arcus endpoint to use
+	TelemetryArcusLocationFilename = "location"
+)
+
+type telemetry struct {
+	isCreated bool
```
piyush-nimbalkar, line 29: We don't need this for our use case. 
It is added for components involving Deployments as taking a diff of deployments is difficult, so we use this flag.

drivers/storage/portworx/manifest/manifest.go
```diff
@@ -57,6 +59,7 @@ type Release struct {
 	PrometheusOperator        string `yaml:"prometheusOperator,omitempty"`
 	PrometheusConfigMapReload string `yaml:"prometheusConfigMapReload,omitempty"`
 	PrometheusConfigReloader  string `yaml:"prometheusConfigReloader,omitempty"`
+	Telemetry                 string `yaml:"telemetry,omitempty"`
```
piyush-nimbalkar, line 62: Let's also create a px-installer ticket to add it there, if not already created.

drivers/storage/portworx/portworx.go
```diff
@@ -754,6 +766,12 @@ func hasCSIChanged(cluster *corev1.StorageCluster) bool {
 		cluster.Status.DesiredImages.CSIProvisioner == ""
 }
 
+func hasTelemetryChanged(cluster *corev1.StorageCluster) bool {
+	return cluster.Spec.Telemetry != nil &&
```
piyush-nimbalkar, line 770: This should be - 
```
  return autoUpdateTelemetry(cluster) && cluster.Status.DesiredImages.Telemetry == ""
```

pkg/apis/core/v1/storagecluster.go
```diff
@@ -446,6 +448,14 @@ type MonitoringSpec struct {
 	Prometheus *PrometheusSpec `json:"prometheus,omitempty"`
 }
 
+// TelemetrySpec contains details of an telemetry component
```
piyush-nimbalkar, line 451: nit: *an -> a

drivers/storage/portworx/deployment.go
```diff
@@ -1120,7 +1191,95 @@ func (t *template) getCSIVolumeInfoList() []volumeInfo {
...
+				hostPath:  "/var/run/log",
+				mountPath: "/var/run/log",
+				readOnly:  true,
+			},
+			{
+				name:      "journalmount2",
+				hostPath:  "/var/log",
+				mountPath: "/var/log",
+				readOnly:  true,
+			},
```
piyush-nimbalkar, line 1255: For the common volumes, can we extract it in a common place? That will avoid duplication and chance of error, if someone modifies the volume only in one place.

drivers/storage/portworx/component/telemetry.go
harsh-px: removed

drivers/storage/portworx/component/telemetry.go
harsh-px: anything in `logfile_patterns` expects these files to end with a specific format. The files need to be gzipped and have the ‘YYYYMMDD’ in the filename.

Hence the *.log files haven't been kept in additional_files.

drivers/storage/portworx/component/telemetry.go
harsh-px: Yes, default is external. If not specified, it uses external, which will be the default for external customers.

drivers/storage/portworx/deployment.go
harsh-px: done

drivers/storage/portworx/deployment.go
harsh-px: yeah copy paste error.. the UT caught this.. 

I guess with you reviewing the code, I can remove the UTs now.

drivers/storage/portworx/manifest/manifest.go
harsh-px: added

drivers/storage/portworx/manifest/manifest.go
harsh-px: https://portworx.atlassian.net/browse/PSG-200 for next sprint.

drivers/storage/portworx/portworx.go
harsh-px: fixed

deploy/crds/core_v1_storagecluster_crd.yaml
harsh-px: done

drivers/storage/portworx/portworx.go
```diff
@@ -203,6 +203,13 @@ func (p *portworx) SetDefaultsOnStorageCluster(toUpdate *corev1.StorageCluster)
 			toUpdate.Status.DesiredImages.UserInterface = release.Components.Lighthouse
 		}
 
+		if autoUpdateTelemetry(toUpdate) &&
```
piyush-nimbalkar, line 206: Can we add UTs for this part similar to what we have for other components? You can use this for reference - https://github.com/libopenstorage/operator/blob/master/drivers/storage/portworx/portworx_test.go#L507?

drivers/storage/portworx/deployment_test.go
```diff
@@ -1882,6 +1882,43 @@ func TestPodSpecForK3s(t *testing.T) {
...
+		},
+	}
+	driver := portworx{}
+	driver.SetDefaultsOnStorageCluster(cluster)
+
+	actual, err := driver.GetStoragePodSpec(cluster, nodeName)
+	assert.NoError(t, err, "Unexpected error on GetStoragePodSpec")
+
+	assertPodSpecEqual(t, expected, &actual)
+}
```
piyush-nimbalkar, line 1920: May be a test for when the telemetry image is empty and we get a warning?

drivers/storage/portworx/components_test.go
```diff
@@ -9737,6 +9738,46 @@ func TestDisablePodSecurityPolicies(t *testing.T) {
...
+	expectedConfigMap := testutil.GetExpectedConfigMap(t, "ccmconfig.yaml")
+	telemetryConfigMap := &v1.ConfigMap{}
+	err = testutil.Get(k8sClient, telemetryConfigMap, component.TelemetryConfigMapName, cluster.Namespace)
+	require.NoError(t, err)
+	require.Equal(t, expectedConfigMap.Name, telemetryConfigMap.Name)
+	require.Equal(t, expectedConfigMap.Namespace, telemetryConfigMap.Namespace)
+	require.Len(t, telemetryConfigMap.OwnerReferences, 1)
+	require.Equal(t, cluster.Name, telemetryConfigMap.OwnerReferences[0].Name)
+	require.Equal(t, expectedConfigMap.Data, telemetryConfigMap.Data)
+}
```
piyush-nimbalkar, line 9779: Can we add tests for
- when arcus location annotation is not passed?
- when telemetry is disabled, the configmap should go away (checkout [other component disable tests](https://github.com/libopenstorage/operator/blob/master/drivers/storage/portworx/components_test.go#L8299))

piyush-nimbalkar (approved):
LGTM after adding some more unit tests

drivers/storage/portworx/components_test.go
harsh-px: added

drivers/storage/portworx/portworx.go
harsh-px: added

drivers/storage/portworx/deployment_test.go
harsh-px: this is just a temporary situation since next sprint will add the default telemetry image. So not adding the empty image UT. Tried adding one but it's tricky since I add a dummy image in the fake manifest so the GetStoragePodSpec ends up using the image from manifest

PR #269, opened 2021-04-07
Add actions for stale-in-review, stale, size, and area labels

ggriffiths:

Closes #

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,21 @@
+area/specs:
+- deploy/**/*
+area/build:
+- build/**/*
+area/ci:
+- .travis
+- .github/**/*
+area/components:
+- drivers/storage/portworx/component/**/*
+area/manifest:
+- drivers/storage/portworx/manifest/**/*
+area/portworx:
+- drivers/storage/portworx/*
+area/controller:
+- pkg/controller/**/*
+area/integration-test:
+- test/**/*
+area/vendor:
+- vendor/**/*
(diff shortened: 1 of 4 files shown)


.github/labeler.yml
```diff
@@ -0,0 +1,19 @@
+area/build:
```
piyush-nimbalkar, line 1: Can you add the following
```
area/specs:
- deploy/**/*

File: .github/labeler.yml [hunk 2]
```diff
@@ -0,0 +1,19 @@
...
+- build/**/*
+area/ci:
+- .travis
+- .github/**/*
+area/components:
+- drivers/storage/portworx/component/**/*
```
piyush-nimbalkar, line 7: In future we could create area for different components like CSI, Security, etc. I think it's okay for now.

.github/labeler.yml
```diff
@@ -0,0 +1,19 @@
...
+area/manifest:
+- drivers/storage/portworx/manifest/**/*
+area/portworx:
+- drivers/storage/portworx/*
+area/controller:
+- pkg/controller/**/*
+area/test:
```
piyush-nimbalkar, line 14: Let's call this `area/integration-test`

piyush-nimbalkar (approved):
Thanks for doing this, Grant!

.github/labeler.yml
ggriffiths: added

PR #272, opened 2021-04-08
OPERATOR-298: commercially signed certs

puremudassir:

When customer is using tls certificates that are signed by a globally recognized certification authority (CA), they don't need to specify a CA cert.
- CA cert is optional if Server cert and key is specified .. in this scenario we assume CA cert is not needed if not specified. We do not apply defaults to the CA cert value.
- stork and autopilot don't need a CA cert in such case

Minor changes:
GetVolumeInfoForTLSCerts -> getVolumeInfoForTLSCerts
SecretName and SecretKey changed to string from pointers

Closes # OPERATOR-298, OPERATOR-296


drivers/storage/portworx/portworx.go
```diff
@@ -675,19 +675,31 @@ func setTLSSpecDefaults(toUpdate *corev1.StorageCluster) {
 		return
 	}
 
+	caSpecified, certSpecified, keySpecified :=
```
ggriffiths, line 678: nice, this is much cleaner

drivers/storage/portworx/util/util.go
```diff
@@ -657,38 +662,41 @@ func AppendTLSEnv(clusterSpec *corev1.StorageClusterSpec, envMap map[string]*v1.
...
 			"-apicert", apicert,
 			"-apikey", apikey,
 			"-apidisclientauth",
-		}, nil
+		}
+		if !IsEmptyOrNilCertLocation(advancedOptions.RootCA) {
+			if !IsEmptyOrNilSecretReference(advancedOptions.RootCA.SecretRef) {
+				args = append(args, "-apirootca", path.Join(DefaultTLSCACertMountPath, advancedOptions.RootCA.SecretRef.SecretKey))
+			} else if !IsEmptyOrNilStringPtr(advancedOptions.RootCA.FileName) {
+				args = append(args, "-apirootca", path.Join(DefaultTLSCertsFolder, *advancedOptions.RootCA.FileName))
```
ggriffiths, line 693: This still assumes the default base path - will this take a direct full path in a future PR? (as per our discussion today)

puremudassir: > This still assumes the default base path - will this take a direct full path in a future PR? (as per our discussion today)

Yes, there was a bit of feedback from the meeting today. I'm going to create a JIRA (for this and possibly other changes). I'll validate with other stakeholders to get an agreement from them.

This is still in the feature branch, so we should be ok for now

ggriffiths (approved):

PR #275, opened 2021-04-13
OPERATOR-309: Do not apply defaults on the TLS spec

puremudassir:
We have decided that defaults should not be applied on the TLS spec - the user is expected to provide the source for all TLS certs. The StorageCluster controller will validate the TLS Spec.

Additional feedback incorporated: filenames are full path for the sake of backwards compatibility in the future.
  Controller will validate that all files are under the `/etc/pwx/` mounted folder

Other changes:
- remove tls.advancedOptions ... reduce one level in spec
- Fix the  deploy/crds/core_v1_storagecluster_crd.yaml ... some updates were missed in previous PRs

https://portworx.atlassian.net/browse/OPERATOR-309


drivers/storage/portworx/util/util.go
```diff
@@ -721,6 +713,20 @@ func IsEmptyOrNilSecretReference(sref *corev1.SecretRef) bool {
 	return false
 }
 
+// IsPartialSecretRef is a helper method that checks if a SecretRef is partially specified (i.e. only one of the needed cert name and key specified)
+func IsPartialSecretRef(sref *corev1.SecretRef) bool {
+	if sref == nil {
+		return false
+	}
+	X := len(sref.SecretName) > 0
+	Y := len(sref.SecretKey) > 0
```
piyush-nimbalkar, line 722: nit: Let's use lower camelCase for private variables `x` and `y`

pkg/apis/core/v1/storagecluster.go
```diff
@@ -143,7 +137,7 @@ type AdvancedTLSOptions struct {
 // CertLocation specifies where portworx should pick up the certificate.
 // Certificate can be in a file on a fixed location or in a secret
 type CertLocation struct {
-	// filename on the node for the cert file. All files must be installed on a fixed location (/etc/pwx)
+	// file name with path on the node for the cert file. Currently only all files must be installed on a subfolder under /etc/pwx
```
piyush-nimbalkar, line 140: nit: *Currently all files ...

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -30,6 +30,7 @@ import (
 
 	"github.com/hashicorp/go-version"
 	"github.com/libopenstorage/operator/drivers/storage"
+	pxutil "github.com/libopenstorage/operator/drivers/storage/portworx/util"
```
piyush-nimbalkar, line 33: Let's not add/reference driver specific code from the controllers. The `IsPartialSecretRef` function can be moved to the package`pkg/util` as it has nothing to do with portworx driver.

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -221,6 +223,53 @@ func (c *Controller) validate(cluster *corev1.StorageCluster) error {
...
+		return fmt.Errorf("%s: can not specify both filename and secretRef as source for tls certs. ", certName)
+	}
+	// file must be under mounted folder /etc/pwx
+	if !pxutil.IsEmptyOrNilStringPtr(certLocation.FileName) {
+		if strings.HasPrefix(*certLocation.FileName, supportedRootFolderForTLSCerts) {
+			return nil // all good
+		}
+		return fmt.Errorf("%s: file path (%s) not under folder %s", certName, *certLocation.FileName, supportedRootFolderForTLSCerts)
+
+	}
```
piyush-nimbalkar, line 272: Let's not add this check here as it is specific to Portworx. We should create a TLS component under portworx driver and add this validation there.

drivers/storage/portworx/components_test.go
```diff
@@ -2677,6 +2677,15 @@ func TestAutopilotWithTLSEnabled(t *testing.T) {
...
 					Enabled: boolPtr(true),
+					RootCA: &corev1.CertLocation{
+						FileName: stringPtr("/etc/pwx/ca.crt"),
+					},
+					ServerCert: &corev1.CertLocation{
+						FileName: stringPtr("/etc/pwx/server.crt"),
+					},
+					ServerKey: &corev1.CertLocation{
+						FileName: stringPtr("/etc/pwx/server.key"),
+					},
```
piyush-nimbalkar, line 2688: In case of filenames, we expect the CA cert to be part of default secret as well for autopilot?
If that is the expectation, then why not have a default CA cert for portworx as well? Why expect the user to have both the secret and filename?

puremudassir: > In case of filenames, we expect the CA cert to be part of default secret as well for autopilot?
> If that is the expectation, then why not have a default CA cert for portworx as well? Why expect the user to have both the secret and filename?

That is a valid point. Yes that is extra work for the user if they choose to use a file.... but that's a choice. The customer has 2 choices

Choice 1: Customer places the CA cert in a k8s secret of their choice - they're done, we do not need them to create a `px-api-root-ca` k8s secret
^^^^^^^
I make code change based on your feedback to support this.

Choice 2: user chooses to create a CA file and install it on the node. Now they have to do extra work to also upload this into a k8s secret named `px-api-root-ca` for clients to access.
Maybe in the future we can figure out a way to make the choice2 experience better.

pkg/util/util.go
```diff
@@ -175,3 +175,17 @@ func ExtractVolumesAndMounts(volumeSpecs []corev1.VolumeSpec) ([]v1.Volume, []v1
...
+	if sref == nil {
+		return false
+	}
+	x := len(sref.SecretName) > 0
+	y := len(sref.SecretKey) > 0
+	// X xor Y -> (X || Y) && !(X && Y)
+	if (x || y) && !(x && y) {
+		return true
+	}
+	return false
```
piyush-nimbalkar, line 190: nit: We could simply write it like this -
```
return (x || y) && !(x && y)
```

drivers/storage/portworx/util/util.go
```diff
@@ -638,17 +630,32 @@ func AppendTLSEnv(clusterSpec *corev1.StorageClusterSpec, envMap map[string]*v1.
...
+				// The user installed a CA cert on the host file system. We assume that user has also uploaded it to the default k8s secret with the default key
+				logrus.Infof("Secret name containing CA cert: %v", DefaultCASecretName)
+				envMap[EnvKeyCASecretName] = &v1.EnvVar{
+					Name:  EnvKeyCASecretName,
+					Value: DefaultCASecretName,
+				}
+				envMap[EnvKeyCASecretKey] = &v1.EnvVar{
+					Name:  EnvKeyCASecretKey,
+					Value: DefaultCASecretKey,
+				}
```
piyush-nimbalkar, line 658: Sorry to go back and forth on this. As you mentioned the certs could be signed by a known CA in which case the clients don't need the CA cert. Which also means user won't create a secret with any CA cert. What would happen in this above case when we pass the default secret location to autopilot?

puremudassir: Yes, in that case we don't pass any CA information to autopilot and stork ... see https://github.com/libopenstorage/operator/pull/275/files/528f8c09601e7bb81a43718e2c13cedf18fe5aa1#diff-55d055b011d6ef2d0e392b7906d7716e28b33693712baf32368b3f74e3754597L643
```
		if clusterSpec.Security.TLS != nil && !IsEmptyOrNilCertLocation(clusterSpec.Security.TLS.RootCA) {
			// if no Root CA is specified, we assume server certs are signed by a well known cert authority. Do not need a CA
```
			
I validated this by manually testing as part of https://portworx.atlassian.net/browse/OPERATOR-296

piyush-nimbalkar: Oh right, makes sense!

piyush-nimbalkar (approved):

PR #283, opened 2021-04-20
OPERATOR-303 Add default telemetry image

harsh-px:

Closes #

### DIFF
--- a/drivers/storage/portworx/manifest/manifest_test.go
+++ b/drivers/storage/portworx/manifest/manifest_test.go
@@ -35,6 +35,7 @@ func TestManifestWithNewerPortworxVersion(t *testing.T) {
 			PrometheusOperator:        "image/prometheus-operator:2.6.0",
 			PrometheusConfigMapReload: "image/configmap-reload:2.6.0",
 			PrometheusConfigReloader:  "image/prometheus-config-reloader:2.6.0",
+			Telemetry:                 defaultTelemetryImage,
 		},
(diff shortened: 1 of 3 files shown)


drivers/storage/portworx/manifest/manifest_test.go
```diff
@@ -35,6 +35,7 @@ func TestManifestWithNewerPortworxVersion(t *testing.T) {
 			PrometheusOperator:        "image/prometheus-operator:2.6.0",
 			PrometheusConfigMapReload: "image/configmap-reload:2.6.0",
 			PrometheusConfigReloader:  "image/prometheus-config-reloader:2.6.0",
+			Telemetry:                 defaultTelemetryImage,
```
piyush-nimbalkar, line 38: I would rather use something custom (`"image/telemetry:x.x.x"`) here instead of default, just to verify we don't always end up returning default image instead of what we receive from the install.portworx.com endpoint. Same for the tests below. 

To check we actually return default when not specified by install.portworx.com, add a default image check here - https://github.com/libopenstorage/operator/blob/master/drivers/storage/portworx/manifest/manifest_test.go#L385

piyush-nimbalkar (approved):

drivers/storage/portworx/manifest/manifest_test.go
harsh-px: Good idea. Done !

PR #286, opened 2021-04-21
OPERATOR-304: Add BottleRocket AMI support

zoxpx:
* will collect the OSImage before generating template (i.e. "Bottlerocket OS 1.0.8" or "Ubuntu 18.04.5 LTS")
* based on OSImage, turns on Bottlerocket AMI customizations

Closes # OPERATOR-304

### DIFF
--- a/drivers/storage/portworx/deployment.go
+++ b/drivers/storage/portworx/deployment.go
@@ -21,6 +21,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
@@ -116,6 +117,7 @@ type template struct {
 	csiConfig       *pxutil.CSIConfiguration
 	kvdb            map[string]string
 	cloudConfig     *cloudstorage.Config
+	osImage         string
 }
 
 func newTemplate(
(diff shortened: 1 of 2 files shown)


drivers/storage/portworx/deployment.go
```diff
@@ -1250,6 +1308,19 @@ func (t *template) getDesiredTelemetryImage(cluster *corev1.StorageCluster) stri
...
+	logrus.Infof("Template OSImage set to %q", osImage)
+	t.osImage = osImage
+}
+
+func (t *template) isBottleRocketOS() bool {
+	if t.osImage == "" {
+		return false
+	}
+	lc := strings.ToLower(t.osImage)
+	return strings.HasPrefix(lc, "bottlerocket os")
```
harsh-px, line 1321: we can keep this more open by just checking for case insensitive bottlerocket. (in case they change the string slightly)

harsh-px (approved):

drivers/storage/portworx/deployment.go
```diff
@@ -206,6 +223,12 @@ func (p *portworx) GetKVDBPodSpec(
 		return v1.PodSpec{}, err
 	}
 
+	if node, err := p.getNodeByName(nodeName); err != nil {
+		logrus.WithError(err).Warnf("Could not get OSImage for node %s", nodeName)
+	} else {
+		t.setOSImage(node.Status.NodeInfo.OSImage)
+	}
+
```
piyush-nimbalkar, line 231: Do we need this here? We don't do anything for the kvdb pod wrt bottle rocket.

piyush-nimbalkar (approved):

drivers/storage/portworx/deployment.go
zoxpx: @piyush-nimbalkar -- I thought this was applied when the POD actually _includes_ the KVDB deployment (so, PX/OCI-Monitor + KVDB), in which case it would be needed.
However, if this deploys only the KVDB (without PX), then this might introduce a bug  (i.e. etcd getting extra `-disable-log-proxy --install-uncompress` args would fail to start).

Could you advise how to test this code-path?

piyush-nimbalkar: The `GetKVDBPodSpec` is called by [StorageNode controller](https://github.com/libopenstorage/operator/blob/master/pkg/controller/storagenode/storagenode.go#L322), where it just creates place holder pods where internal kvdb runs. It is useful for pod disruption budgets.

The above code you added is kind of a dead code as the `osImage` field won't be used to construct the kvdb container at all. It is only used to construct portworx pod which is created only in `GetPortworxPodSpec`.

zoxpx:
Thanks guys - both concerns addressed in ffe2267

PR #287, opened 2021-04-21
Add image bundle docker file

haibinxie:

Closes #


piyush-nimbalkar:
We generally don't commit this file because it is auto-generated. Also, the Dockerfile copies the manifest/ and metadata/ directories which are not checked in, so this is incomplete. We should either check-in everything or nothing.

haibinxie:
> We generally don't commit this file because it is auto-generated. Also, the Dockerfile copies the manifest/ and metadata/ directories which are not checked in, so this is incomplete. We should either check-in everything or nothing.

Ok, i updated the document, just so the openshift version list is documented.

PR #292, opened 2021-04-23
[CSI] Install by default when Portworx is enabled

ggriffiths:

* Installs CSI by default unless explicitly disabled.
* Fixes many unit tests that assumed CSI was disabled by default.

n/a

n/a

### DIFF
--- a/drivers/storage/portworx/component/csi.go
+++ b/drivers/storage/portworx/component/csi.go
@@ -72,13 +72,17 @@ func (c *csi) Initialize(
 }
 
 func (c *csi) IsEnabled(cluster *corev1.StorageCluster) bool {
-	return pxutil.FeatureCSI.IsEnabled(cluster.Spec.FeatureGates)
+	return pxutil.IsCSIEnabled(cluster)
 }
 
 func (c *csi) Reconcile(cluster *corev1.StorageCluster) error {
 	ownerRef := metav1.NewControllerRef(cluster, pxutil.StorageClusterKind())
 	pxVersion := pxutil.GetPortworxVersion(cluster)
 	csiConfig := c.getCSIConfiguration(cluster, pxVersion)
+
+	if cluster.Status.DesiredImages == nil {
+		cluster.Status.DesiredImages = &corev1.ComponentImages{}
+	}
 
 	if err := c.createServiceAccount(cluster.Namespace, ownerRef); err != nil {
 		return err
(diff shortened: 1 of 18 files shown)


drivers/storage/portworx/component/csi.go
```diff
@@ -72,14 +72,18 @@ func (c *csi) Initialize(
...
-	return pxutil.FeatureCSI.IsEnabled(cluster.Spec.FeatureGates)
+	return pxutil.IsCSIEnabled(cluster)
 }
 
 func (c *csi) Reconcile(cluster *corev1.StorageCluster) error {
 	ownerRef := metav1.NewControllerRef(cluster, pxutil.StorageClusterKind())
 	pxVersion := pxutil.GetPortworxVersion(cluster)
 	csiConfig := c.getCSIConfiguration(cluster, pxVersion)
 
+	if cluster.Status.DesiredImages == nil {
```
ggriffiths, line 83: This normally won't be called, as we usually set defaults in another function. However, for tests, this is needed.

drivers/storage/portworx/util/util.go
```diff
@@ -216,6 +216,18 @@ func IsPortworxEnabled(cluster *corev1.StorageCluster) bool {
...
+// IsCSIEnabled returns true if CSI is not disabled by the feature flag
+func IsCSIEnabled(cluster *corev1.StorageCluster) bool {
+	enableCSI := true
+	if FeatureCSI.IsSet(cluster.Spec.FeatureGates) {
+		enableCSI = FeatureCSI.IsEnabled(cluster.Spec.FeatureGates)
+	}
+
+	// Only enable CSI when Portworx is enabled
+	enableCSI = IsPortworxEnabled(cluster) && enableCSI
+	return enableCSI
```
piyush-nimbalkar, line 228: nit: The `feature.IsEnabled` method in feature.go could be simply written like this - 
```
func (feature Feature) IsEnabled(featureMap map[string]string) bool {
	enabled, err := strconv.ParseBool(featureMap[string(feature)])
	return err != nil || enabled
}
```
The above also makes sure, if the user has not properly disabled (by giving non-boolean or empty value) the feature, then we still return true.

Then we can simply write this helped method as follows - 
```
return IsPortworxEnabled(cluster) && FeatureCSI.IsEnabled(cluster.Spec.FeatureGates)
```

piyush-nimbalkar (approved):
Are we planning to make CSI default from operator 2.0.0 and px 3.0.0? Is there are ticket for this in Operator jira? If not, can we create one for tracking?

drivers/storage/portworx/util/feature.go
```diff
@@ -12,8 +12,13 @@ const (
...
 func (feature Feature) IsEnabled(featureMap map[string]string) bool {
 	enabled, err := strconv.ParseBool(featureMap[string(feature)])
-	return err == nil && enabled
+	return err != nil || enabled
+}
+
+// IsSet checks if the feature is set in the given feature map
+func (feature Feature) IsSet(featureMap map[string]string) bool {
+	_, ok := featureMap[string(feature)]
+	return ok
```
piyush-nimbalkar, line 23: We don't need this now.

piyush-nimbalkar (approved):

piyush-nimbalkar (approved):

PR #294, opened 2021-04-28
OPERATOR-256 Add subscription section to use appliance ID

harsh-px:
Needed as per CCM change: https://jira.purestorage.com/browse/CLOUD-58502

This allows CCM to just use the appliance ID (cluster uuid for PX) for subscription
and does not require to specify it again in the spec

Closes #

### DIFF
--- a/drivers/storage/portworx/component/telemetry.go
+++ b/drivers/storage/portworx/component/telemetry.go
@@ -135,6 +135,9 @@ func (t *telemetry) createConfigMap(
         "product_name": "portworx",
         "appliance_id_path": "/etc/pwx/cluster_uuid"
       },
+      "subscription": {
+        "use_appliance_id": "true",
+      },
       "proxy": {
         "path": "/dev/null"
       }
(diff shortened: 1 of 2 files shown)


drivers/storage/portworx/component/telemetry.go
```diff
@@ -135,6 +135,9 @@ func (t *telemetry) createConfigMap(
         "product_name": "portworx",
         "appliance_id_path": "/etc/pwx/cluster_uuid"
       },
+      "subscription": {
+        "use_appliance_id": "true",
```
piyush-nimbalkar, line 139: Will have to remove the comma at the end. This will fail to parse in json.

piyush-nimbalkar (approved):
LGTM after the minor fix.

harsh-px:
CI is failing due to docker rate limiting. All tests are passing.

drivers/storage/portworx/component/telemetry.go
harsh-px: good catch. Fixed.

PR #295, opened 2021-04-29
Operator 268 - Block installing portworx from StorageCluster when Por…

haibinxie:
…tworx Daemonset is present

Fix https://portworx.atlassian.net/browse/OPERATOR-268?src=confmacro

Closes #

The installation is blocked, and you will see a k8s event attached to StorageCluster object, also visible in operator log.

Events:
  Type     Reason      Age               From                       Message
  ----     ------      ----              ----                       -------
  Warning  FailedSync  7s (x16 over 1m)  storagecluster-controller  daemonset portworx exists, please delete it first. Operator will not proceed with installation

### DIFF
--- a/pkg/controller/storagecluster/storagecluster.go
+++ b/pkg/controller/storagecluster/storagecluster.go
@@ -77,6 +77,8 @@ const (
 	crdBasePath                         = "/crds"
 	storageClusterCRDFile               = "core_v1_storagecluster_crd.yaml"
 	minSupportedK8sVersion              = "1.12.0"
+	// portworxDaemonSetName is the name of DaemonSet when portworx is installed by DaemonSet.
+	portworxDaemonSetName = "portworx"
 )
 
 var _ reconcile.Reconciler = &Controller{}
@@ -308,6 +310,22 @@ func (c *Controller) RegisterCRD() error {
 	return nil
 }
 
+func (c *Controller) isDaemonSetPresent(name string) (bool, error) {
+	daemonSetList := &apps.DaemonSetList{}
+	if err := c.client.List(context.TODO(), daemonSetList, &client.ListOptions{}); err != nil {
+		logrus.Errorf("Failed to list DaemonSet: %v", err)
+		return false, err
+	}
+
+	for _, daemonSet := range daemonSetList.Items {
+		if daemonSet.Name == name {
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
+
 func (c *Controller) syncStorageCluster(
 	cluster *corev1.StorageCluster,
 ) error {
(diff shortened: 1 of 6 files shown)


pkg/controller/storagecluster/storagecluster.go
```diff
@@ -317,6 +335,19 @@ func (c *Controller) syncStorageCluster(
 		return c.deleteStorageCluster(cluster)
 	}
 
+	// If portworx has been installed by daemonset, before it's uninstalled, operator
+	// should not proceed with installation.
+	if exists, err := c.isDaemonSetPresent(portworxDaemonSetName); err != nil || exists {
```
piyush-nimbalkar, line 340: Let's put this with other validations in validate() function (line 219).

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -317,6 +335,19 @@ func (c *Controller) syncStorageCluster(
...
+		if err != nil {
+			return err
+		}
+
+		if exists {
+			return fmt.Errorf("daemonset %s exists, please delete it first. Operator will not proceed with installation",
```
piyush-nimbalkar, line 346: Let's not ask users to delete the daemonset now. We should just warn them that both installation methods won't work simultaneously and that's why we cannot proceed.

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -308,6 +310,22 @@ func (c *Controller) RegisterCRD() error {
 	return nil
 }
 
+func (c *Controller) isDaemonSetPresent(name string) (bool, error) {
```
piyush-nimbalkar, line 313: So far, we have maintained all portworx specific code under drivers/storage/portworx and the storagecluster controller only contains generic logic. It would be nice to continue and have that separation.
Can we create a Validate() function under drivers/storage.Driver interface and then implement this logic in the portworx.go file? From storagecluster we just call Driver.Validate() and return the error.

pkg/controller/storagecluster/storagecluster.go
haibinxie: the problem with putting it in Validate() is, the operator uninstallation code will not be hit. when both DaemonSet and Operator are present, we should still let operator continue with uninstallation.

haibinxie:
@piyush-nimbalkar I resolved your comments in a new commit. However I have concern that uninstallation with operator is also blocked when DaemonSet is present, hence I want to move the uninstallation code to before validate(). let me know what do you think?

piyush-nimbalkar (approved):
I was thinking that the finalizer on the storage cluster won't be set until the validation has passed. Also, as the reconciliation never goes past validate, nothing has been installed that needs cleanup in the deleteStorageCluster function.
Basically when the StorageCluster is deleted, it should just go away from k8s api server. Did you try it? Was there a problem when deleting?

haibinxie:
> I was thinking that the finalizer on the storage cluster won't be set until the validation has passed. Also, as the reconciliation never goes past validate, nothing has been installed that needs cleanup in the deleteStorageCluster function.
> Basically when the StorageCluster is deleted, it should just go away from k8s api server. Did you try it? Was there a problem when deleting?

@piyush-nimbalkar what if user installs with operator then installs with DaemonSet and uninstalls operator. It might not be big deal as it is a corner case.

haibinxie:
> > I was thinking that the finalizer on the storage cluster won't be set until the validation has passed. Also, as the reconciliation never goes past validate, nothing has been installed that needs cleanup in the deleteStorageCluster function.
> > Basically when the StorageCluster is deleted, it should just go away from k8s api server. Did you try it? Was there a problem when deleting?
> 
> @piyush-nimbalkar what if user installs with operator then installs with DaemonSet and uninstalls operator. It might not be big deal as it is a corner case.

Actually anything that fails in Validate would block uninstallation, I will merge this PR and add to the weekly meeting agenda to discuss it, and create a different ticket if needed after the discussion. For example, if we increase minK8SVersion in the future, cluster with old minK8SVersion will not be able to uninstall after upgrading operator.

piyush-nimbalkar:
That makes sense. We should move the delete handling code before the validate. 
Thanks for catching this!

PR #297, opened 2021-05-04
OPERATOR-329 Update pod spec on telemetry spec changes

harsh-px:

Closes #

### DIFF
--- a/pkg/controller/storagecluster/controller_test.go
+++ b/pkg/controller/storagecluster/controller_test.go
@@ -6249,130 +6249,221 @@ func TestDoesTelemetryMatch(t *testing.T) {
 			new:   &corev1.StorageClusterSpec{},
 			match: true,
 		},
-		{
-			old: &corev1.StorageClusterSpec{},
-			new: &corev1.StorageClusterSpec{
-				Monitoring: &corev1.MonitoringSpec{},
-			},
-			match: true,
-		},
-		{
-			old: &corev1.StorageClusterSpec{
-				Monitoring: &corev1.MonitoringSpec{},
-			},
-			new:   &corev1.StorageClusterSpec{},
-			match: true,
-		},
-		{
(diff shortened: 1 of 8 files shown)


pkg/controller/storagecluster/controller_test.go
```diff
@@ -6238,6 +6238,145 @@ func TestNodeShouldRunStoragePod(t *testing.T) {
 	require.True(t, shouldContinueRunning)
 }
 
+func TestDoesTelemetryMatch(t *testing.T) {
```
piyush-nimbalkar, line 6241: Instead of testing a private method, we could add a test like this - https://github.com/libopenstorage/operator/blob/master/pkg/controller/storagecluster/controller_test.go#L3322
This tests the complete reconcile flow and validates that the pod is marked for deletion when change is detected.

piyush-nimbalkar (approved):

pkg/controller/storagecluster/controller_test.go
harsh-px: Kept existing for loop as wanted a UT that targets that function directly. Easier to focus incremental changes.

Also added a Reconcile test now for the same cases.

pkg/controller/storagecluster/controller_test.go
```diff
@@ -6238,6 +6238,236 @@ func TestNodeShouldRunStoragePod(t *testing.T) {
...
+		//	new: &corev1.StorageClusterSpec{
+		//		Monitoring: &corev1.MonitoringSpec{
+		//			Telemetry: &corev1.TelemetrySpec{
+		//				Enabled: true,
+		//				Image:   "bar",
+		//			}},
+		//	},
+		//	match: false,
+		//},
+	}
```
piyush-nimbalkar, line 6371: Was this left commented accidentally?
Can you update this as well as the cherry-pick>

PR #299, opened 2021-05-04
Operator 193 -  Add support for cache drive

haibinxie:
Fix https://portworx.atlassian.net/browse/OPERATOR-193?src=confmacro

Closes #

Unit test added, tested on my cluster too.

### DIFF
--- a/drivers/storage/portworx/deployment.go
+++ b/drivers/storage/portworx/deployment.go
@@ -693,6 +693,12 @@ func (t *template) getArguments() []string {
 	}
 
 	if t.cluster.Spec.Storage != nil {
+		if t.cluster.Spec.Storage.CacheDevices != nil {
+			for _, dev := range *t.cluster.Spec.Storage.CacheDevices {
+				args = append(args, "-cache", dev)
+			}
+		}
+
 		if t.cluster.Spec.Storage.Devices != nil {
 			for _, dev := range *t.cluster.Spec.Storage.Devices {
 				args = append(args, "-s", dev)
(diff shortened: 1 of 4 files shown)


piyush-nimbalkar (changes requested):
- We copy the cluster level storage specs in the node specs (if no storage devices are overridden at the node level) to minimize confusion on what storage specs are being used for a particular node group. For that we will have to add some code here - https://github.com/libopenstorage/operator/blob/master/drivers/storage/portworx/portworx.go#L602
- Similar to [this test](https://github.com/libopenstorage/operator/blob/master/pkg/controller/storagecluster/controller_test.go#L3692), we will have to add test for cache device. This tests that on changing the cache devices, we bounce px pods.
- Add the cache devices spec to the CRD schema - https://github.com/libopenstorage/operator/blob/master/deploy/crds/core_v1_storagecluster_crd.yaml#L315

haibinxie:
@piyush-nimbalkar resolved your comments (1 and 3) in a new commit. the test (item 2) was in the PR.

piyush-nimbalkar (approved):
Thanks for making the changes. 
Sorry I missed the test the first time. Currently the test only checks addition of cache drives. Can we add cases for changing the original array of cache drives or removal of all cache drives?

haibinxie:
> Thanks for making the changes.
> Sorry I missed the test the first time. Currently the test only checks addition of cache drives. Can we add cases for changing the original array of cache drives or removal of all cache drives?

@piyush-nimbalkar sounds good, added the unit tests.

PR #303, opened 2021-05-07
OPERATOR-338 enable telemetry by default

harsh-px:

Closes #

### DIFF
--- a/drivers/storage/portworx/manifest/manifest.go
+++ b/drivers/storage/portworx/manifest/manifest.go
@@ -33,7 +33,7 @@ const (
 	defaultPrometheusOperatorImage        = "quay.io/coreos/prometheus-operator:v0.34.0"
 	defaultPrometheusConfigMapReloadImage = "quay.io/coreos/configmap-reload:v0.0.1"
 	defaultPrometheusConfigReloaderImage  = "quay.io/coreos/prometheus-config-reloader:v0.34.0"
-	defaultTelemetryImage                 = "purestorage/ccm-service:3.0.0-rc1"
+	defaultTelemetryImage                 = "purestorage/ccm-service:3.0.0-rc2"
 	defaultManifestRefreshInterval        = 3 * time.Hour
 )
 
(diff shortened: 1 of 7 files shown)


piyush-nimbalkar (approved):
This would work for older versions of PX as well, right?
Also, as this affects the px pod, this change will restart the px pod on operator upgrade. I would say discuss this with PM and take a call.

harsh-px:
> This would work for older versions of PX as well, right?
Also, as this affects the px pod, this change will restart the px pod on operator upgrade. I would say discuss this with PM and take a call.

Good point. The sidecar isn't useful for versions lower than 2.8.0. so it would be good to add a >= 2.8.0 version check. Do we have an example of a similar version check that I can follow?

piyush-nimbalkar:
Here is an example of this - https://github.com/libopenstorage/operator/blob/master/drivers/storage/portworx/deployment.go#L889

To test this with an internal image, you can either use PX_IMAGE env with a 2.8+ tag or use `portworx.io/px-version` annotation to explicitly tell the px version for non-semver image tags. Or just use a custom oci-mon image with 2.8+ tag.

drivers/storage/portworx/manifest/manifest.go
```diff
@@ -33,7 +33,7 @@ const (
 	defaultPrometheusOperatorImage        = "quay.io/coreos/prometheus-operator:v0.34.0"
 	defaultPrometheusConfigMapReloadImage = "quay.io/coreos/configmap-reload:v0.0.1"
 	defaultPrometheusConfigReloaderImage  = "quay.io/coreos/prometheus-config-reloader:v0.34.0"
-	defaultTelemetryImage                 = "purestorage/ccm-service:3.0.0-rc1"
+	defaultTelemetryImage                 = "purestorage/ccm-service:3.0.0-rc2"
```
harsh-px, line 36: bumped this to rc2 as it has many phone home fixes needed to make it default

PR #305, opened 2021-05-07
Remove github_api_url from labeler action

ggriffiths:
Removes github_api_url variable - seems like this action was either misconfigured or they removed this variable.

Has no effect other than removing a warning from the logs.

n/a

n/a


haibinxie (approved):
LGTM

PR #311, opened 2021-05-11
[cherry-pick] OPERATOR-338 enable telemetry by default

harsh-px:

Closes #


pault84 (approved):
LGTM

harsh-px:
> is the assumption here that we will be releasing operator 1.5.0 around 2.8

yes, we have to release operator 1.5.0 around PX 2.8.0.

> no existing customer would be affected (px pod getting bounced) by making telemetry default because they will be on older px version?

yes, don't want oci-monitor to bounce on operator upgrade for older versions. For e.g 2.7.0. I will add that test case in testrail.

PR #318, opened 2021-05-19
[OPERATOR-241] Pass custom annotations to Portworx pods

pure-jliao:
**What this PR does / why we need it**: Allow to pass custom annotations from StorageCluster to Portworx Pods

Closes # [OPERATOR-241](https://portworx.atlassian.net/browse/OPERATOR-241)


deploy/crds/core_v1_storagecluster_crd.yaml
```diff
@@ -77,6 +77,13 @@ spec:
             autoUpdateComponents:
               type: string
               description: A strategy to determine how component versions are to be updated automatically.
+            annotations:
+              type: object
+              description: >-
+                The annotations section of spec is a map of map to pass custom annotations to different
+                Portworx components. The key is a locators to locate specific Portworx components in format of
```
haibinxie, line 84: locators -> locator, or the key specifies component in format of ...

pkg/apis/core/v1/storagecluster.go
```diff
@@ -611,6 +613,21 @@ const (
 	ClusterOperationTimeout ClusterConditionStatus = "Timeout"
 )
 
+// AnnotationKind is the k8s object type of Portworx components
+// it will be part of the keys to locate a specific Portworx component
+type AnnotationKind string
```
haibinxie, line 618: Use enum so we can extend later, enum for both AnnotationKind and Component name

pkg/controller/storagecluster/update.go
```diff
@@ -529,6 +561,18 @@ func (c *Controller) isPodUpdated(
 		logrus.Warnf("Could not check the diff between current pod and latest spec. %v", err)
 		return false
 	}
+
+	// Pod not matched will be deleted then will be annotated upon restart.
+	// Updated pod will be annotated directly with known and custom annotations without restart,
+	// this case pod hash needs to be updated as well to avoid inconsistency
+	if matched {
```
haibinxie, line 568: move to outside of isPodUpdated function, the name isPodUpdated indicate checking if pod is updated instead of updating the pod.

pkg/controller/storagecluster/update.go
```diff
@@ -84,6 +85,37 @@ func (c *Controller) rollingUpdate(cluster *corev1.StorageCluster, hash string)
...
+		for k, v := range customAnnotations {
+			annotations[k] = v
+		}
+	}
+	if len(annotations) == 0 {
+		pod.SetAnnotations(nil)
+	} else {
+		pod.SetAnnotations(annotations)
+	}
+	err := c.client.Update(context.TODO(), pod)
```
haibinxie, line 111: I suggest check if custom annotations changed, if it is not changed, don't update the pod.

haibinxie:
Please also add unit tests

pkg/apis/core/v1/storagecluster.go
pure-jliao: IMO there is no need to use real "enum" here and it's probably an overkill to add all the map and int values here. 
this strong type of string already has a consistent format compare to other strong type strings in this file already.

pkg/controller/storagecluster/update.go
pure-jliao: able to move it to `getAllStorageClusterPods` same it's not a good place to make an update. 
update stc spec annotation will change the stc hash, and we don't want a pod restart, then in logic putting it in `isPodUpdated` after checking the condition pod's hash is equal to stc hash, but before returning the pod is updated should be the best place IMO.

pkg/controller/storagecluster/update.go
pure-jliao: with this change this function is more like a reconciler for custom annotations, then it makes more sense to put it in isPodUpdated.

pkg/apis/core/v1/storagecluster.go
haibinxie: string type enum is ok, try to encapsulate the isValid function. 
func (t AnnotationKind) IsValid() error {
}

pkg/controller/storagecluster/update.go
haibinxie: hash should not change with custom annotations right? i am not sure if i get it. in rollingUpdate function, it returned newPods here, we can update annotations here.

_, oldPods := c.getAllStorageClusterPods(cluster, nodeToStoragePods, hash)

haibinxie: return two flags for caller to handle in isPodUpdated, one flag for restarting, and one for update without restarting. we can rename isPodUpdated to podNeedsUpdate()

pkg/controller/storagecluster/update.go
```diff
@@ -84,6 +85,50 @@ func (c *Controller) rollingUpdate(cluster *corev1.StorageCluster, hash string)
...
+	if pod.Labels[defaultStorageClusterUniqueLabelKey] != hash {
+		pod.Labels[defaultStorageClusterUniqueLabelKey] = hash
+	}
+	pod.SetAnnotations(annotations)
+	err := c.client.Update(context.TODO(), pod)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
piyush-nimbalkar, line 129: nit: Could just simplify this by
```
  return c.client.Update(..)
```

pkg/util/util.go
```diff
@@ -186,3 +189,37 @@ func IsPartialSecretRef(sref *corev1.SecretRef) bool {
...
+	if !objKind.IsValid() {
+		logrus.Warnf("%s: %s", InvalidCustomAnnotationKind, objKind)
+		return nil
+	}
+	if !component.IsValid() {
+		logrus.Warnf("%s: %s", InvalidCustomAnnotationComponent, component)
+		return nil
+	}
+	if cluster.Spec.Annotations == nil {
+		return nil
```
piyush-nimbalkar, line 218: We could write a test for this case

pkg/util/util.go
```diff
@@ -186,3 +189,37 @@ func IsPartialSecretRef(sref *corev1.SecretRef) bool {
...
+	}
+	key := fmt.Sprintf("%s/%s", objKind, component)
+	if annotations, ok := cluster.Spec.Annotations[key]; ok && len(annotations) != 0 {
+		return annotations
+	}
+	return nil
```
piyush-nimbalkar, line 224: Same for this one.

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -278,6 +286,35 @@ func (c *Controller) validateSingleCluster(current *corev1.StorageCluster) error
...
+			}
+			component := corev1.AnnotationComponent(split[1])
+			if !component.IsValid() {
+				return fmt.Errorf("%s: %s", util.InvalidCustomAnnotationComponent, component)
+			}
+			// TODO: OPERATOR-343 handle validation for different Portworx components and strict syntax validation
+			// So far custom annotation is only supported for Portworx storage pod
+			switch {
+			case kind == corev1.PodAnnotations && component == corev1.Portworx:
+			default:
```
piyush-nimbalkar, line 309: IMHO we don't need to be very strict on checking the supported annotations. It feels unnecessary. Rather just use the annotation if it matches a component, else ignore the extra annotations.
We have a lot of components and each component has several k8s objects. Regularly maintaining a list of these feels a bit redundant. Also, it spills the component details outside of the component package.

pkg/controller/storagecluster/controller_test.go
```diff
@@ -5724,6 +5724,170 @@ func TestUpdateStorageClusterSecurity(t *testing.T) {
...
+
+	testutil.Get(k8sClient, oldPod, oldPod.Name, oldPod.Namespace)
+	require.NotEmpty(t, oldPod.Annotations)
+	val, ok := oldPod.Annotations[customAnnotationKey]
+	require.True(t, ok)
+	require.Equal(t, customAnnotationVal1, val)
+	_, ok = oldPod.Annotations[knownAnnotationKey]
+	require.True(t, ok)
+
+	// TestCase: Update existing custom annotations
```
piyush-nimbalkar, line 5808: Can we add a test case for removal of an annotation?

pkg/controller/storagecluster/controller_test.go
```diff
@@ -5043,13 +5043,6 @@ func TestUpdateStorageClusterNodeSpec(t *testing.T) {
 	err = k8sClient.Update(context.TODO(), cluster)
 	require.NoError(t, err)
 
-	// Change existing pod's hash to latest revision, to simulate new pod with latest spec
```
pure-jliao, line 5046: when pod is updated without restart, the hash update will be handled in the code, so no simulation is required.

pkg/controller/storagecluster/controller_test.go
```diff
@@ -5078,13 +5071,6 @@ func TestUpdateStorageClusterNodeSpec(t *testing.T) {
...
```
pure-jliao, line 5081: same here

pkg/controller/storagecluster/controller_test.go
```diff
@@ -6269,7 +6433,7 @@ func TestHistoryCleanup(t *testing.T) {
 
 	runningPod2, err := k8scontroller.GetPodFromTemplate(&podControl.Templates[0], cluster, clusterRef)
 	require.NoError(t, err)
-	require.Equal(t, revisions.Items[0].Labels[defaultStorageClusterUniqueLabelKey],
```
pure-jliao, line 6272: seems the expected revision was not correct, don't know how that passed. but after review, after cluster update and for new created runningPod2, it should be using the latest revision.

haibinxie (approved):
LGTM

haibinxie:
Please wait for Piyush to double check before merging it. thanks.

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -63,11 +63,12 @@ import (
 
 const (
 	// ControllerName is the name of the controller
-	ControllerName                      = "storagecluster-controller"
+	ControllerName = "storagecluster-controller"
+	// ComponentName is the component name of the storage cluster
+	ComponentName                       = "portworx"
```
piyush-nimbalkar, line 68: We don't have any portworx specific logic in the controller. All of it is in the driver/portworx package. To keep this portworx agnostic, can we call the annotation locator `pod/storage`?  Customers are used to storagenodes anyway, so storage pod would be on the same lines. Wdyt?
Sorry for pointing this out so late.
cc: @haibinxie

piyush-nimbalkar (commented):
The change looks pretty good now. Few minor comments.

pkg/controller/storagecluster/update.go
```diff
@@ -84,6 +87,37 @@ func (c *Controller) rollingUpdate(cluster *corev1.StorageCluster, hash string)
...
+	if customAnnotations := operatorutil.GetCustomAnnotations(cluster, k8s.Pod, ComponentName); customAnnotations != nil {
+		for k, v := range customAnnotations {
+			annotations[k] = v
+		}
+	}
+
+	// Skip cases when no action is needed
+	if len(annotations) == 0 && pod.Annotations == nil || reflect.DeepEqual(annotations, pod.Annotations) {
+		return
+	}
```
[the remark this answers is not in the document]
pure-jliao: true, makes sense.

pkg/controller/storagecluster/controller_test.go
```diff
@@ -5724,6 +5710,185 @@ func TestUpdateStorageClusterSecurity(t *testing.T) {
...
+			Name:      cluster.Name,
+			Namespace: cluster.Namespace,
+		},
+	}
+	result, err := controller.Reconcile(context.TODO(), request)
+	require.NoError(t, err)
+	require.Empty(t, result)
+	require.Empty(t, podControl.DeletePodName)
+
+	oldPod = createStoragePod(cluster, "old-pod", k8sNode.Name, storageLabels)
```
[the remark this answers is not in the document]
pure-jliao: that's the problem I mentioned after getting the oldPod it still keeps some of the old fields due to some reason, this create function just return a template and get again could fix it.

pkg/controller/storagecluster/storagecluster.go
pure-jliao: sounds good.

piyush-nimbalkar (approved):
LGTM. This looks really good now @pure-jliao! 
Feel free to merge after addressing the minor comments.

PR #319, opened 2021-05-21
OPERATOR-240 [Mastercard] Unable to set image urls when using operator in air gapped environment

haibinxie:
https://portworx.atlassian.net/browse/OPERATOR-240

Closes #

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -40,9 +40,7 @@ var (
 	}
 )
 
-// UpdateCommonRegistries add additional registries to common registries.
 func getMergedCommonRegistries(cluster *corev1.StorageCluster) map[string]bool {
-
 	val, ok := cluster.Annotations[constants.AnnotationCommonImageRegistries]
 
 	if !ok {
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -2,6 +2,7 @@ package util
 
 import (
 	"fmt"
+	"github.com/libopenstorage/operator/pkg/constants"
 	"path"
 	"reflect"
 	"strings"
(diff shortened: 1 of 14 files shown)


pkg/util/util_test.go
```diff
@@ -76,6 +76,21 @@ func TestGetImageURN(t *testing.T) {
...
+	out = GetImageURN("registry.io//", "k8s.gcr.io/pause:3.1")
+	require.Equal(t, "registry.io/pause:3.1", out)
+
+	// Update it again, now k8s.gcr.io should be deleted from common registries.
+	UpdateCommonRegistries([]string{"gcr.io"})
+	out = GetImageURN("registry.io", "k8s.gcr.io/pause:3.1")
+	require.Equal(t, "registry.io/k8s.gcr.io/pause:3.1", out)
+
+	out = GetImageURN("registry.io//", "k8s.gcr.io/pause:3.1")
+	require.Equal(t, "registry.io/pause:3.1", out)
```
piyush-nimbalkar, line 90: Should this have been `registry.io/k8s.gcr.io/pause` as common registries now only has `gcr.io`?

pkg/util/util.go
```diff
@@ -35,8 +35,28 @@ var (
 		"registry-1.docker.io":        true,
 		"registry.connect.redhat.com": true,
 	}
+
+	savedCommonDockerRegistries map[string]bool
 )
 
+// UpdateCommonRegistries add additional registries to common registries.
+func UpdateCommonRegistries(additionalRegistries []string) {
```
piyush-nimbalkar, line 43: This function will eventually forget about the original common registries and keep on adding to the list.

I think we should avoid saving such a list in utils package in a variable. It is error prone to keep such a singleton util.
I would suggest that in the GetImageURN function we can just pass the list of additional registries (or the whole cluster object and then get the annotation) and replace the common registry if present in the additional list on top of the common registries.

pkg/util/util_test.go
haibinxie: With double slashes as suffix in customRegistry (registry.io//), it always replace the rep, it does not look at common registries, this was old behavior.

pkg/util/util.go
haibinxie: Sounds good, I can make it as another parameter of GetImageURN

pkg/util/util_test.go
piyush-nimbalkar: Oh right, I forgot that part.

haibinxie:
@piyush-nimbalkar I resolved your comments.

pkg/util/util_test.go
```diff
@@ -9,74 +10,95 @@ import (
...
 
-	out = GetImageURN("registry.io//", "test/this/image")
+	out = getImageURN("", "registry.io//", "test/this/image")
 	require.Equal(t, "registry.io/image", out)
+
+	out = getImageURN("k8s.gcr.io", "registry.io//", "k8s.gcr.io/pause:3.1")
+	require.Equal(t, "registry.io/pause:3.1", out)
+
+	// Update it again, now k8s.gcr.io should be deleted from common registries.
+	out = getImageURN("gcr.io", "registry.io", "k8s.gcr.io/pause:3.1")
```
piyush-nimbalkar, line 86: Can we add a test for comma-separated multiple registries?

pkg/util/util.go
```diff
@@ -39,12 +40,37 @@ var (
 	}
 )
 
+// UpdateCommonRegistries add additional registries to common registries.
```
piyush-nimbalkar, line 43: Update the comment

piyush-nimbalkar (approved):

haibinxie:
@piyush-nimbalkar I resolved your last comments and merged the PR.

PR #339, opened 2021-06-18
OPERATOR-369: Mounts checks and override

zoxpx:
* will check the Portworx POD volumes, and make sure there are no overlapping destinations  (last mount w/ same dest wins)
* also removes POD's volumes that are not mounted in any container

The user-mounts can now override the OCI-Monitor's mounts.
* main use-case -- can use user-defined volumes to override where Portworx is installed on the host system

Closes # OPERATOR-369

Quick test:

* add the `/opt/pwx/` mount override into StorageCluster:

```yaml
  volumes:
    - hostPath:
        path: /usr/local/pwx
      mountPath: /opt/pwx
      name: install-override
```
* this causes Portworx to get installed into `/usr/local/pwx` directory:
```
# ls -al /usr/local/bin/pxctl
lrwxrwxrwx 1 root root 24 Jun 17 23:51 /usr/local/bin/pxctl -> /usr/local/pwx/bin/pxctl

# ls -al /usr/local/pwx/oci
total 32
drwx------  4 root root  4096 Jun 17 23:43 .
drwxr-xr-x  4 root root  4096 Jun 17 23:43 ..
-rw-r--r--  1 root root 16090 Jun 17 23:43 config.json
drwx------  5 root root  4096 Jun 17 23:43 mounts
drwxr-xr-x 29 root root  4096 Jun 17 23:51 rootfs
```

### DIFF
--- a/drivers/storage/portworx/deployment.go
+++ b/drivers/storage/portworx/deployment.go
@@ -361,7 +361,41 @@ func (p *portworx) GetStoragePodSpec(
 		}
 	}
 
+	p.pruneVolumes(&podSpec)
+
 	return podSpec, nil
+}
+
+func (p *portworx) pruneVolumes(spec *v1.PodSpec) {
+	seen := make(map[string]bool)
+	for ci, container := range spec.Containers {
+		seenDest := make(map[string]bool)
+		vm := container.VolumeMounts
+		for i := len(vm) - 1; i >= 0; i-- {
+			if seenDest[vm[i].MountPath] {
+				logrus.Warnf("Removed mount %s:%s for container %s due to non-unique destination",
+					vm[i].Name, vm[i].MountPath, container.Name)
+				copy(vm[i:], vm[i+1:])
+				spec.Containers[ci].VolumeMounts = vm[:len(vm)-1]
+			} else {
+				seen[vm[i].Name] = true
+				seenDest[vm[i].MountPath] = true
+			}
+		}
+	}
+	if len(seen) != len(spec.Volumes) {
+		for i := len(spec.Volumes) - 1; i >= 0; i-- {
+			if n := spec.Volumes[i].Name; !seen[n] {
+				hp := "???"
+				if spec.Volumes[i].VolumeSource.HostPath != nil {
+					hp = spec.Volumes[i].VolumeSource.HostPath.Path
+				}
+				logrus.Warnf("Removed unused Volume %s:%s from Spec", spec.Volumes[i].Name, hp)
+				copy(spec.Volumes[i:], spec.Volumes[i+1:])
+				spec.Volumes = spec.Volumes[:len(spec.Volumes)-1]
+			}
+		}
+	}
 }
 
 func (p *portworx) createStorageNode(cluster *corev1.StorageCluster, nodeName string, cloudConfig *cloudstorage.Config) error {
--- a/drivers/storage/portworx/deployment.go
+++ b/drivers/storage/portworx/deployment.go
@@ -367,34 +367,47 @@ func (p *portworx) GetStoragePodSpec(
 }
 
 func (p *portworx) pruneVolumes(spec *v1.PodSpec) {
-	seen := make(map[string]bool)
+	seenName := make(map[string]bool)
+	updatedMounts := false
 	for ci, container := range spec.Containers {
 		seenDest := make(map[string]bool)
 		vm := container.VolumeMounts
+		vmNew := make([]v1.VolumeMount, 0, len(vm))
 		for i := len(vm) - 1; i >= 0; i-- {
-			if seenDest[vm[i].MountPath] {
+			mp := path.Clean(vm[i].MountPath)
+			if !seenDest[mp] {
+				vmNew = append(vmNew, vm[i])
+				seenName[vm[i].Name] = true
+				seenDest[mp] = true
+			} else {
 				logrus.Warnf("Removed mount %s:%s for container %s due to non-unique destination",
 					vm[i].Name, vm[i].MountPath, container.Name)
-				copy(vm[i:], vm[i+1:])
-				spec.Containers[ci].VolumeMounts = vm[:len(vm)-1]
-			} else {
-				seen[vm[i].Name] = true
-				seenDest[vm[i].MountPath] = true
+				updatedMounts = true
 			}
+		}
+		if updatedMounts {
+			// reverse the array (since mounts were appended in reverse order)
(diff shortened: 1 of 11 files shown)


harsh-px:
@haibinxie can you help with a review here?

drivers/storage/portworx/deployment.go
```diff
@@ -361,9 +361,43 @@ func (p *portworx) GetStoragePodSpec(
...
 	return podSpec, nil
 }
 
+func (p *portworx) pruneVolumes(spec *v1.PodSpec) {
+	seen := make(map[string]bool)
+	for ci, container := range spec.Containers {
+		seenDest := make(map[string]bool)
+		vm := container.VolumeMounts
+		for i := len(vm) - 1; i >= 0; i-- {
+			if seenDest[vm[i].MountPath] {
```
haibinxie, line 375: you can use path.Clean function to convert the paths, which handles extra slashes in the end, ../, ./ etc in the path that may cause same paths to have different string

drivers/storage/portworx/deployment.go
```diff
@@ -361,9 +361,43 @@ func (p *portworx) GetStoragePodSpec(
...
+				logrus.Warnf("Removed mount %s:%s for container %s due to non-unique destination",
+					vm[i].Name, vm[i].MountPath, container.Name)
+				copy(vm[i:], vm[i+1:])
+				spec.Containers[ci].VolumeMounts = vm[:len(vm)-1]
```
haibinxie, line 379: this does not look correct, if there are 5 duplicate entries, this will still use len-1. suggest use a different array to save the elements instead of changing array inside of iteration.

drivers/storage/portworx/deployment.go
zoxpx: I am aware of the `path.Clean()` function -- but this is not what we are doing here:
* we will scan the volume mounts in the containers, and record all mounts into `seen`, and all mount destinations into `seenDest`
* if we find any mounts destination repeated (e.g. `-v /dir1:/dest -v /dir2:/dest ...`), we will simply exclude the first mount, and "let the last mount win"

drivers/storage/portworx/deployment_test.go
```diff
@@ -3347,6 +3347,83 @@ func TestIKSEnvVariables(t *testing.T) {
...
+					{
+						Name:      "foo",
+						MountPath: "/mnt/foo",
+					},
+					{
+						Name:      "bar",
+						MountPath: "/mnt/bar",
+					},
+					{
+						Name:      "foo-override",
```
haibinxie, line 3384: add a test case to overwrite foo twice.

drivers/storage/portworx/deployment.go
haibinxie: I got that, with string comparison mounts destination /dest and /dest/ (with extra slash) will be taken as different destination.

drivers/storage/portworx/deployment.go
zoxpx: This should be working also for cases when all 5 entries are duplicates:
* we are traversing the array right-to-left
* if we need to remove an element from the array, we'll copy the elements and "trim" the last element from the array
* the code "activates" only when the duplicate entry has been identified -- so we won't go "out of array bounds" with `[:len(vm)-1]`

Granted, the way this is done is not very efficient  (element's `copy` takes time), and we could achieve similar effect by building a brand new array as a replacement.  However, please consider:
* the code does not need to be very efficient, as the occurrence of the volume-mount removal should be _exceedingly rare_
* building a replacement array (`spec.Containers[#].VolumeMounts`) every time would be inefficient, since the elements are full struct's (note, they are not pointers to struct's)


So, sadly -- golang does not have a more elegant way how to remove elements from the arrays, and I wrote my reasons why I think doing "in place" element removals is more appropriate here.
Is this (still) going to be an issue?

drivers/storage/portworx/deployment.go
zoxpx: Oh, I see your point -- the `/dest`, `/dest/` and `/xx/../dest` would all be treated as a different mount directory; we should `path.Clean()` the paths before comparison, so they evaluate to the same directory,

drivers/storage/portworx/deployment.go
piyush-nimbalkar: Did not want to nit-pick here, but now that we are discussing this, even I feel creating a separate array is better for two reasons - easy to understand and more time efficient than copying array every time. I understand it is not space efficient but there are not a lot of volumes either.

haibinxie: @zoxpx I got that, for example if there are 5 duplicates, the array should have one in the end, but this assignment will save 4 because it uses len(vm) which is 5, did you want to trim the last element after copying the array one element ahead?

spec.Containers[ci].VolumeMounts = vm[:len(vm)-1]

drivers/storage/portworx/deployment.go
haibinxie: yes

drivers/storage/portworx/deployment.go
zoxpx: @haibinxie -- sorry, still not sure where the problem is.  Note that with the trimming at the line you pointed out, we are changing the array _slice_ without modifying the array.  This would happen 4 times if there are 5 duplicates, but the array "re-slicing" should be very cheap in golang.
If your concern that the backing array will still be 5 elements long, where the "useful" array slice is only 1 element?


@piyush-nimbalkar -- please check the 3953cb8 for updated code:
* this version uses a new array instead of "in-place" elements removal
* note that now we're probably copying more data, also we have to revert the "replacement" mounts' array since the elements were added in reverted order  (e.g. "last guy wins")

drivers/storage/portworx/deployment.go
zoxpx: see 3953cb8

drivers/storage/portworx/deployment_test.go
zoxpx: see 3953cb8

zoxpx:
Note -- also reverted my changes to the test-data @ 125c00e
* this is no longer required, as we'll check/update volumes only if we updated the volume-mounts  (e.g. excess volumes that were not mounted are OK again)

drivers/storage/portworx/deployment.go
haibinxie: i see it is fixed in new commit.

haibinxie (approved):
lgtm

zoxpx:
Thanks for the review Haibin and Piyush -- merging the PR

PR #340, opened 2021-06-21
Remove cross-namespace ownership

piyush-nimbalkar:

To access encryption secrets, we create role and rolebinding
in the encryption secrets namespace. We used to add storagecluster
as the owner for these objects. K8s does not like having cross-
namespace owner references. If done so it garbage collects
all the objects owned by the StorageCluster when the garbage
collector restarts.

Closes # https://portworx.atlassian.net/browse/OPERATOR-368

### DIFF
--- a/drivers/storage/portworx/component/portworx_basic.go
+++ b/drivers/storage/portworx/component/portworx_basic.go
@@ -92,15 +92,27 @@ func (c *portworxBasic) Delete(cluster *corev1.StorageCluster) error {
 	if err := k8sutil.DeleteClusterRoleBinding(c.k8sClient, PxClusterRoleBindingName); err != nil {
 		return err
 	}
-	if err := k8sutil.DeleteRole(c.k8sClient, PxRoleName, cluster.Namespace, *ownerRef); err != nil {
-		return err
-	}
-	if err := k8sutil.DeleteRoleBinding(c.k8sClient, PxRoleBindingName, cluster.Namespace, *ownerRef); err != nil {
-		return err
-	}
 	if err := k8sutil.DeleteService(c.k8sClient, pxutil.PortworxServiceName, cluster.Namespace, *ownerRef); err != nil {
 		return err
 	}
+
+	secretsNamespace := getSecretsNamespace(cluster)
+	if secretsNamespace == cluster.Namespace {
+		if err := k8sutil.DeleteRole(c.k8sClient, PxRoleName, secretsNamespace, *ownerRef); err != nil {
+			return err
+		}
+		if err := k8sutil.DeleteRoleBinding(c.k8sClient, PxRoleBindingName, secretsNamespace, *ownerRef); err != nil {
+			return err
+		}
+	} else {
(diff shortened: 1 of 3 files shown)


drivers/storage/portworx/component/portworx_basic.go
```diff
@@ -173,56 +179,106 @@ func (c *portworxBasic) createNamespace(namespace string) error {
...
+		types.NamespacedName{
+			Name:      role.Name,
+			Namespace: role.Namespace,
+		},
+		existingRole,
 	)
+	if err == nil && len(existingRole.OwnerReferences) > 0 {
+		existingRole.OwnerReferences = nil
+		logrus.Debugf("Removing ownership from %s/%s Role", role.Namespace, role.Name)
+		if err := c.k8sClient.Update(context.TODO(), existingRole); err != nil {
```
haibinxie, line 221: if we don't have the code here, later in this function k8sutil.CreateOrUpdateRole(c.k8sClient, role, nil) should do the same, right?

drivers/storage/portworx/component/portworx_basic.go
```diff
@@ -173,56 +179,106 @@ func (c *portworxBasic) createNamespace(namespace string) error {
...
+		types.NamespacedName{
+			Name:      roleBinding.Name,
+			Namespace: roleBinding.Namespace,
+		},
+		existingRoleBinding,
 	)
+	if err == nil && len(existingRoleBinding.OwnerReferences) > 0 {
+		existingRoleBinding.OwnerReferences = nil
+		logrus.Debugf("Removing ownership from %s/%s RoleBinding", roleBinding.Namespace, roleBinding.Name)
+		if err := c.k8sClient.Update(context.TODO(), existingRoleBinding); err != nil {
```
haibinxie, line 275: same here.

drivers/storage/portworx/component/portworx_basic.go
piyush-nimbalkar: The CreateOrUpdate* function will not update the object if we remove the ownerReference because of this line - 
```
if modified || len(role.OwnerReferences) > len(existingRole.OwnerReferences) {
```

Just trying to update the CreateOrUpdate* functions temporarily. We will remove the above code few releases down the line.

haibinxie: Yes, I think it will work. we don't have to change it. Just to discuss the CreateOrUpdate function, I don't see other places using it, currently the function only merges owner references from exiting role, it does not handle removing owner references. Is there a use case for that?

haibinxie (approved):
lgtm

drivers/storage/portworx/component/portworx_basic.go
piyush-nimbalkar: Generally we have never needed to remove owner references. Rather we only wanted to add ownerReferences. This is a one-off requirement because of the regression. That's why we are removing owner ref. Wanted to keep the CreateOrUpdateRole function consistent with other CreateOrUpdate* functions on that front. Hence kept the above update out of the CreateOrUpdateRole

PR #348, opened 2021-06-29
OPERATOR-268 Mark the cluster as failed if validation fails

piyush-nimbalkar:

Closes #

### DIFF
--- a/pkg/controller/storagecluster/storagecluster.go
+++ b/pkg/controller/storagecluster/storagecluster.go
@@ -207,6 +207,8 @@ func (c *Controller) Reconcile(ctx context.Context, request reconcile.Request) (
 
 	if err := c.validate(cluster); err != nil {
 		k8s.WarningEvent(c.recorder, cluster, util.FailedValidationReason, err.Error())
+		cluster.Status.Phase = string(corev1.ClusterOperationFailed)
+		k8s.UpdateStorageClusterStatus(c.client, cluster)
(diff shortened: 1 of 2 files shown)


pkg/controller/storagecluster/storagecluster.go
```diff
@@ -207,6 +207,8 @@ func (c *Controller) Reconcile(ctx context.Context, request reconcile.Request) (
 
 	if err := c.validate(cluster); err != nil {
 		k8s.WarningEvent(c.recorder, cluster, util.FailedValidationReason, err.Error())
+		cluster.Status.Phase = string(corev1.ClusterOperationFailed)
+		k8s.UpdateStorageClusterStatus(c.client, cluster)
```
haibinxie, line 211: return the error if the function fails.

haibinxie (approved):
you can merge after the minor change to return error if updating status fails.

pkg/controller/storagecluster/storagecluster.go
piyush-nimbalkar: It is returning the original validation error. In case of update failure, let's just log it and let the original error be returned.

piyush-nimbalkar: @haibinxie does the change work with you?

haibinxie: Yes, that sounds good.

PR #358, opened 2021-07-19
Implement px repo in operator

haibinxie:
Implement PX repo for portworx dependency module installation.
https://portworx.atlassian.net/browse/OPERATOR-340

Closes #


deploy/crds/core_v1_storagecluster_crd.yaml
```diff
@@ -279,6 +279,16 @@ spec:
                           the taint. By default, it is not set, which means tolerate the taint forever
                           (do not evict). Zero and negative values will be treated as 0 (evict
                           immediately) by the system.
+            pxRepo:
```
piyush-nimbalkar, line 282: Why do we need this to be a section in the StorageCluster spec. I think we can always keep this enabled and running. Is there is need to expose this to the user?

pkg/controller/storagecluster/pxrepo.go
```diff
@@ -0,0 +1,121 @@
+package storagecluster
```
piyush-nimbalkar, line 1: Can we move this under components package in 'driver/storage/portworx'?

haibinxie: fixed.

deploy/crds/core_v1_storagecluster_crd.yaml
haibinxie: after some considerations, i think it's better to keep it, we can make default as enabled. there are cases users may want to disable it.
- the image is too big
- there is unexpected issue in the feature

haibinxie:
@piyush-nimbalkar I have made the changes, I found imagepullpolicy is not considered in the deployment "modified" flag, thus imagePullPolicy update does not work in current version, i handled it in pxrepo deployment. we may need to fix it for other deployments later.

drivers/storage/portworx/component/pxrepo.go
```diff
@@ -0,0 +1,203 @@
...
+
+	// If existingDeployment does not exist, modified will be true.
+	modified :=
+		k8sutil.GetImageFromDeployment(existingDeployment, pxRepoContainerName) != k8sutil.GetImageFromDeployment(deployment, pxRepoContainerName) ||
+			k8sutil.GetImagePullPolicyFromDeployment(existingDeployment, pxRepoContainerName) != k8sutil.GetImagePullPolicyFromDeployment(deployment, pxRepoContainerName) ||
+			util.HasPullSecretChanged(cluster, existingDeployment.Spec.Template.Spec.ImagePullSecrets) ||
+			util.HasNodeAffinityChanged(cluster, existingDeployment.Spec.Template.Spec.Affinity) ||
+			util.HaveTolerationsChanged(cluster, existingDeployment.Spec.Template.Spec.Tolerations)
+
+	if modified {
```
piyush-nimbalkar, line 150: As we discussed offline, we will need the 'isDeploymentCreated' flag for future changes to the deployment spec in operator. We may miss out on adding that in future, so it is better to add it now just like other deployment components.

pkg/util/k8s/k8s.go
```diff
@@ -1511,6 +1511,21 @@ func GetImageFromDeployment(deployment *appsv1.Deployment, containerName string)
...
+		if c.Name == containerName {
+			return c.ImagePullPolicy
+		}
+	}
+	for _, c := range deployment.Spec.Template.Spec.InitContainers {
+		if c.Name == containerName {
+			return c.ImagePullPolicy
+		}
+	}
+	return v1.PullIfNotPresent
```
piyush-nimbalkar, line 1526: Why not return empty here if the container is not found?

drivers/storage/portworx/components_test.go
```diff
@@ -265,6 +266,71 @@ func TestBasicComponentsInstall(t *testing.T) {
...
+	err = driver.PreInstall(cluster)
+	require.NoError(t, err)
+
+	err = testutil.Get(driver.k8sClient, service, component.PxRepoServiceName, cluster.Namespace)
+	require.True(t, errors.IsNotFound(err))
+
+	err = testutil.Get(driver.k8sClient, deployment, component.PxRepoDeploymentName, cluster.Namespace)
+	require.True(t, errors.IsNotFound(err))
+}
+
```
piyush-nimbalkar, line 333: Can we add px-repo to the following tests for testing change in image, pull secret, affinity or tolerations -
https://github.com/libopenstorage/operator/blob/master/drivers/storage/portworx/components_test.go#L5871
https://github.com/libopenstorage/operator/blob/master/drivers/storage/portworx/components_test.go#L6596
https://github.com/libopenstorage/operator/blob/master/drivers/storage/portworx/components_test.go#L7454
https://github.com/libopenstorage/operator/blob/master/drivers/storage/portworx/components_test.go#L7759
https://github.com/libopenstorage/operator/blob/master/drivers/storage/portworx/components_test.go#L8150

drivers/storage/portworx/component/pxrepo.go
```diff
@@ -0,0 +1,203 @@
...
+	k8sClient client.Client,
+	_ version.Version,
+	_ *runtime.Scheme,
+	_ record.EventRecorder,
+) {
+	p.k8sClient = k8sClient
+}
+
+func (p *pxrepo) IsEnabled(cluster *corev1.StorageCluster) bool {
+	return cluster.Spec.PxRepo != nil && cluster.Spec.PxRepo.Enabled
```
piyush-nimbalkar, line 56: Can we also add one more check -
`&& pxutil.IsPortworxEnabled(cluster)`

piyush-nimbalkar (commented):
- Image pull policy was left out on purpose. None of the components including the portworx pods  bounce on changing the image pull policy. If the image is already pulled and the pods are running then do we need to bounce them on changing the pull policy?

- About the question of having pxrepo in the spec or not, let's keep the discussion offline. We can revisit and change this if needed. For the scope of this PR let us continue to keep it in the spec.

- Once we have the image ready from portworx side, we will have to get it from porworx/manifest package. Again this can be part of a separate PR, but let's create a ticket to track it.

drivers/storage/portworx/component/pxrepo.go
haibinxie: After looking at other components, some components do not have this flag. the flag is not needed anyway, i feel it's easier to understand without it.

piyush-nimbalkar: I believe all components that have a deployment or daemon set have the flag. It is not about being easier to understand, it is needed for future modifications to this deployment spec. For instance, in the next release of the operator if you add an env to the deployment in the code, it won't modify the existing deployment, if we don't add the flag.

drivers/storage/portworx/components_test.go
```diff
@@ -7550,6 +7708,12 @@ func TestCompleteInstallWithImagePullSecretChange(t *testing.T) {
 	require.Len(t, csiDeployment.Spec.Template.Spec.ImagePullSecrets, 1)
 	require.Equal(t, imagePullSecret, csiDeployment.Spec.Template.Spec.ImagePullSecrets[0].Name)
 
+	pxRepoDeployment := &appsv1.Deployment{}
+	err = testutil.Get(k8sClient, pxRepoDeployment, component.PxRepoDeploymentName, cluster.Namespace)
+	require.NoError(t, err)
+	require.Len(t, pxRepoDeployment.Spec.Template.Spec.Containers, 1)
```
piyush-nimbalkar, line 7714: nit: no need to check number of containers in this test

drivers/storage/portworx/components_test.go
```diff
@@ -7857,6 +8049,12 @@ func TestCompleteInstallWithTolerationsChange(t *testing.T) {
 	require.NoError(t, err)
 	require.ElementsMatch(t, tolerations, csiDeployment.Spec.Template.Spec.Tolerations)
 
+	pxRepoDeployment := &appsv1.Deployment{}
+	err = testutil.Get(k8sClient, pxRepoDeployment, component.PxRepoDeploymentName, cluster.Namespace)
+	require.NoError(t, err)
+	require.Len(t, pxRepoDeployment.Spec.Template.Spec.Containers, 1)
```
piyush-nimbalkar, line 8055: nit: No need to check number of containers in this test too

drivers/storage/portworx/components_test.go
```diff
@@ -8255,6 +8493,12 @@ func TestCompleteInstallWithNodeAffinityChange(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, nodeAffinity, csiDeployment.Spec.Template.Spec.Affinity.NodeAffinity)
 
+	pxRepoDeployment := &appsv1.Deployment{}
+	err = testutil.Get(k8sClient, pxRepoDeployment, component.PxRepoDeploymentName, cluster.Namespace)
+	require.NoError(t, err)
+	require.Len(t, pxRepoDeployment.Spec.Template.Spec.Containers, 1)
```
piyush-nimbalkar, line 8499: nit: Same as above

piyush-nimbalkar (approved):
LGTM

drivers/storage/portworx/components_test.go
haibinxie: resolved with a few other places

PR #365, opened 2021-08-02
[CSI] Lower pod security policy to Restricted

ggriffiths:

Closes #

### DIFF
--- a/drivers/storage/portworx/component/csi.go
+++ b/drivers/storage/portworx/component/csi.go
@@ -296,7 +296,7 @@ func (c *csi) createClusterRole(
 			{
 				APIGroups:     []string{"policy"},
 				Resources:     []string{"podsecuritypolicies"},
-				ResourceNames: []string{constants.PrivilegedPSPName},
+				ResourceNames: []string{constants.RestrictedPSPName},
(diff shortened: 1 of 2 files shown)


piyush-nimbalkar (approved):
LGTM after the tests pass.
Did you test this in an environment where PSPs are enabled?

ggriffiths:
I tried out CSI volume creation which uses the provisioner sidecar on k8s 1.20 w/ this change. Are PSP enabled on openshift only?

piyush-nimbalkar:
> I tried out CSI volume creation which uses the provisioner sidecar on k8s 1.20 w/ this change. Are PSP enabled on openshift only?

OpenShift uses SCC/Security context constraints instead of PSP, but it is more or less the same. We need to try this in a vanilla k8s cluster where PSP is enforced. To enable you need to enable the PSP admission controller plugin. You can mostly do this using kubeadm. If using Rancher, there is an easy toggle in the UI to enable PSP.

You can verify if they are enforced by creating a random privileged pod and it should fail to come up. Then just try creating StorageCluster with CSI enabled and verify that the csi ext deployment comes up.

PR #366, opened 2021-08-09
Use apiextension v1 apis to manage CRDs instead of v1beta1

piyush-nimbalkar:
With k8s 1.22, the v1beta1 apiextensions api is getting removed and clients needs to use v1 api. 
This PR changes the CRD APIs to use apiextensions v1.

- Moving the v1beta1 objects to deprecated folder to  support older k8s clusters
- Adding x-kubernetes-preserve-unknown-fields=true flag  for unstructured objects like maps

Closes # [OPERATOR-379](https://portworx.atlassian.net/browse/OPERATOR-379)


pkg/controller/storagecluster/storagecluster.go
```diff
@@ -306,6 +308,23 @@ func (c *Controller) validateCustomAnnotations(current *corev1.StorageCluster) e
 
 // RegisterCRD registers and validates CRDs
 func (c *Controller) RegisterCRD() error {
+	k8sVersion, err := k8s.GetVersion()
+	if err != nil {
+		return fmt.Errorf("error parsing Kubernetes version '%s'. %v", k8sVersion, err)
+	}
+
+	k8s1_16, err := version.NewVersion("1.16")
```
haibinxie, line 316: I am thinking whether we should use 1.22 or 1.16 here, if we use 1.22 then we only need to test new operator with 1.22, if we use 1.16 we may need to test the new operator with 1.16-1.21 to be safe. I am ok with using 1.16, just want to see if anyone foresee any issue. we need to test it prior to release.

haibinxie:
We need to consider test and upgrade matrix, would scenario 2 work?
1. 1.16- k8s with old operator, upgrade to new operator
2. 1.16+ k8s with old operator, upgrade to new operator
3. 1.16- k8s with new operator, upgrade to 1.16+ k8s
4. 1.16+ k8s with new operator, upgrade to k8s 1.22

pkg/controller/storagecluster/storagecluster.go
piyush-nimbalkar: Testing effort is a very good point. I assumed we have a kubernetes matrix job where we test multiple k8s versions and also k8s upgrades. Let me confirm with the QA team. 
Honestly, the only reason I added 1.16 here is to get the 'kubectl explain storagecluster|storagenode' working (https://portworx.atlassian.net/browse/OPERATOR-210). It did not work before because our schema was not structural.

If we are not covering the below scenarios in the automation, then we can do this only for 1.22 and above. Wdyt?

piyush-nimbalkar:
> We need to consider test and upgrade matrix, would scenario 2 work?
> 
> 1. 1.16- k8s with old operator, upgrade to new operator
> 2. 1.16+ k8s with old operator, upgrade to new operator
> 3. 1.16- k8s with new operator, upgrade to 1.16+ k8s
> 4. 1.16+ k8s with new operator, upgrade to k8s 1.22

Thanks for listing the scenarios. 
Yes, scenario 2 works. I have testing 1 and 2 on local clusters. Planning to run 3 and 4 using the k8s-matrix job.

haibinxie (approved):
lgtm

PR #370, opened 2021-08-19
OPERATOR-377 fix portworx quorum loss issue during openshift upgrade

haibinxie:
Previous we only look at storage pod to decide number of unavailable storage nodes, this is inaccurate because at some point we may think storage pod should not exist (due to node taint, affinity etc). the fix is to also look at storage node status to decide number of unavailable storage nodes. this number is used to do rolling update.

Closes #

Unit tests added, also verified on OCP cluster.

### DIFF
--- a/pkg/controller/storagecluster/controller_test.go
+++ b/pkg/controller/storagecluster/controller_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	storageapi "github.com/libopenstorage/openstorage/api"
 	"reflect"
 	"strconv"
 	"testing"
@@ -704,6 +705,7 @@ func TestFailureDuringStorkInstallation(t *testing.T) {
 	driver.EXPECT().GetStorkDriverName().Return("mock", nil).AnyTimes()
 	driver.EXPECT().SetDefaultsOnStorageCluster(gomock.Any())
 	driver.EXPECT().UpdateDriver(gomock.Any()).Return(nil)
+	driver.EXPECT().GetStorageNodes(gomock.Any()).Return(nil, nil).AnyTimes()
 	driver.EXPECT().UpdateStorageClusterStatus(gomock.Any()).Return(nil)
 	driver.EXPECT().GetSelectorLabels().Return(nil).AnyTimes()
 	driver.EXPECT().String().Return(driverName).AnyTimes()
@@ -803,6 +805,7 @@ func TestStoragePodsShouldNotBeScheduledIfDisabled(t *testing.T) {
 	driver.EXPECT().GetSelectorLabels().Return(nil).AnyTimes()
 	driver.EXPECT().String().Return(driverName).AnyTimes()
 	driver.EXPECT().UpdateDriver(gomock.Any()).Return(nil)
+	driver.EXPECT().GetStorageNodes(gomock.Any()).Return(nil, nil).AnyTimes()
 	driver.EXPECT().UpdateStorageClusterStatus(gomock.Any()).Return(nil)
 	driver.EXPECT().SetDefaultsOnStorageCluster(gomock.Any())
 	driver.EXPECT().IsPodUpdated(gomock.Any(), gomock.Any()).Return(true).AnyTimes()
@@ -929,6 +932,7 @@ func TestStoragePodGetsScheduled(t *testing.T) {
 	driver.EXPECT().GetSelectorLabels().Return(nil).AnyTimes()
 	driver.EXPECT().String().Return(driverName).AnyTimes()
 	driver.EXPECT().UpdateDriver(gomock.Any()).Return(nil)
+	driver.EXPECT().GetStorageNodes(gomock.Any()).Return(nil, nil).AnyTimes()
 	driver.EXPECT().UpdateStorageClusterStatus(gomock.Any()).Return(nil)
 	driver.EXPECT().SetDefaultsOnStorageCluster(gomock.Any()).
 		Do(func(c *corev1.StorageCluster) {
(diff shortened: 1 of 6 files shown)


pkg/controller/storagecluster/update.go
```diff
@@ -393,31 +395,67 @@ func (c *Controller) getUnavailableNumbers(
...
+			if !available {
+				numUnavailable++
 			}
 		}
-		if !available {
+	}
+
+	// For the storage nodes that do not have a corresponding k8s node.
+	for _, storageNode := range storageNodeMap {
+		if storageNode.Status != storageapi.Status_STATUS_OK {
```
piyush-nimbalkar, line 457: Should we increment the `desiredNumberScheduled` too? Or you are thinking of keeping it conservative and calculate the maxUnavailable number based only on kubernetes nodes and not portworx nodes?
I think that's fine too. Just thinking out loud if ever numUnavailable will always exceed maxUnavailable resulting in no way for users to override and force upgrade other online nodes. I mean they won't be able to go over 100%. This may happen if few non-k8s portworx nodes are offline for a known reason and users still want to continue with an upgrade for rest of the nodes. Having said that, they can always user numbers instead of percentage in that case and let the upgrade continue. So not a big deal, just wanted to know your reason too.

pkg/controller/storagecluster/controller_test.go
```diff
@@ -1341,6 +1347,102 @@ func TestStoragePodGetsScheduledWithCustomNodeSpecs(t *testing.T) {
 	require.Equal(t, *clusterRef, podControl.ControllerRefs[2])
 }
 
+// When a StorageNode is unhealthy, and the storage node is not a k8s node. we should not update storage pod as
+// it may cause storage cluster to lose quorum.  This could happen if a user removes a k8s node from the cluster
+// for maintenance, and meanwhile tries to uppgrade portworx.
```
piyush-nimbalkar, line 1352: nit: *upgrade

pkg/controller/storagecluster/controller_test.go
```diff
@@ -1341,6 +1347,102 @@ func TestStoragePodGetsScheduledWithCustomNodeSpecs(t *testing.T) {
...
+		recorder:            recorder,
+		kubernetesVersion:   k8sVersion,
+		lastPodCreationTime: make(map[string]time.Time),
+	}
+
+	var storageNodes []*storageapi.StorageNode
+	storageNodes = append(storageNodes, createStorageNode("k8s-node-1", true))
+	storageNodes = append(storageNodes, createStorageNode("k8s-node-2", true))
+	storageNodes = append(storageNodes, createStorageNode("k8s-node-3", true))
+	storageNodes = append(storageNodes, createStorageNode("not-k8s-node", false))
```
piyush-nimbalkar, line 1381: This is a nice test!
Can we fit in the scenarios in the same or a different test:
- where a storage node is offline, but on a valid kubernetes node (we are only checking offline storage node on non-k8s node)
- a k8s node for which`wantToRun` comes as false, but storage node is present (online or offline)

piyush-nimbalkar (commented):
The change looks pretty good. Great effort in investigation and fixing these complicated upgrade cases! 👏 

Most of the comments are not critical, except for the empty scheduler node name. Once we handle that, we can merge this change.

pkg/controller/storagecluster/update.go
```diff
@@ -393,31 +395,67 @@ func (c *Controller) getUnavailableNumbers(
...
+		storageNodeMap[storageNode.SchedulerNodeName] = storageNode
+	}
+
 	var numUnavailable, desiredNumberScheduled int
 	for _, node := range nodeList.Items {
+		// If the node has a storage node and it is not healthy.
+		storageNode, storageNodeExists := storageNodeMap[node.Name]
+		if storageNodeExists {
+			delete(storageNodeMap, node.Name)
+			if storageNode.Status != storageapi.Status_STATUS_OK {
```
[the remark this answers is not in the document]
haibinxie: DOWN or maintenance, should still be considered as unavailable.

pkg/controller/storagecluster/controller_test.go
```diff
@@ -7109,6 +7253,33 @@ func createStoragePod(
...
+		MemUsed:              0,
+		MemFree:              0,
+		AvgLoad:              0,
+		Status:               status,
+		Disks:                nil,
+		Pools:                nil,
+		MgmtIp:               "",
+		DataIp:               "",
+		Hostname:             "",
+		NodeLabels:           nil,
```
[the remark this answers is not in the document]
haibinxie: this is auto generated, i will remove the unused fields.

pkg/controller/storagecluster/update.go
haibinxie: Yes, in this case the external nodes will be upgraded externally and to be conservative admin can finish and ensure that is successful first.

pkg/controller/storagecluster/update.go
```diff
@@ -393,31 +395,67 @@ func (c *Controller) getUnavailableNumbers(
...
 	}
 
+	storageNodeList, err := c.Driver.GetStorageNodes(cluster)
+	if err != nil {
+		return -1, -1, fmt.Errorf("couldn't get list of storage nodes during rolling "+
+			"update of storage cluster  %#v: %v", cluster, err)
+	}
+	storageNodeMap := make(map[string]*storageapi.StorageNode)
+	for _, storageNode := range storageNodeList {
+		storageNodeMap[storageNode.SchedulerNodeName] = storageNode
```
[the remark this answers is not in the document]
haibinxie: Good point

pkg/controller/storagecluster/controller_test.go
haibinxie: @piyush-nimbalkar  we actually check offline storage nodes on k8s nodes and non-k8s nodes, so if a storage node is offline and it is also a k8s node, we will also not proceed with upgrade. 

`we are only checking offline storage node on non-k8s node`

haibinxie:
@piyush-nimbalkar i have resolved all comments

PR #371, opened 2021-08-20
Pause portworx upgrades if an openshift upgrade is in progress 

piyush-nimbalkar:
- The default behavior will be to pause Portworx upgrade in its current state when we detect an openshift upgrade is in progress. Basically the operator will not bounce any more portworx pods which usually results in portworx service restarts. 
- This will not prevent 'on-delete' kind of updates, if the user decides to restart the portworx pod by themselves. Also, if openshift restarts nodes during its upgrades, portworx service will automatically get updated. This is a good thing as there is only one portworx node that is getting updated at a given point of time.
- Operator will raise in event if an upgrade is paused
```
  Normal  UpgradePaused     84s (x71 over 10m)  storagecluster-controller  The rolling update of the storage cluster has been paused due to an ongoing OpenShift upgrade. Storage nodes will only be updated on pod deletion or after the OpenShift upgrade is completed.
```
- If the user wants to disable this behavior, then can add `operator/libopenstorage.org/force-continue-update: "true"` annotation to force continue the portworx updates.

Closes # [OPERATOR-371](https://portworx.atlassian.net/browse/OPERATOR-371)


pkg/controller/storagecluster/util.go
```diff
@@ -124,3 +126,10 @@ func isEmpty(object interface{}) bool {
 	}
 	return false
 }
+
+func forceContinueUpgrade(
+	cluster *corev1.StorageCluster,
+) bool {
+	force, err := strconv.ParseBool(cluster.Annotations[constants.AnnotationForceContinueUpdate])
+	return err == nil && force
```
haibinxie, line 134: please write a log if err is not nil as it is swallowed

pkg/util/k8s/cluster.go
```diff
@@ -0,0 +1,42 @@
...
+	metaerrors "k8s.io/apimachinery/pkg/api/meta"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// IsClusterBeingUpgraded checks if the Kubernetes cluster is being upgraded.
+// Currently this only checks for OpenShift clusters.
+func IsClusterBeingUpgraded(k8sClient client.Client) (bool, error) {
+	cvList := &ocp_configv1.ClusterVersionList{}
+	err := k8sClient.List(context.TODO(), cvList, &client.ListOptions{})
+	if metaerrors.IsNoMatchError(err) {
```
haibinxie, line 16: nit: when is this going to be executed? even if this is executed it is still ok to not have the if clause, and return false err.

pkg/util/k8s/cluster.go
```diff
@@ -0,0 +1,42 @@
...
+
+	cv := cvList.Items[0].DeepCopy()
+	if cv.Spec.DesiredUpdate == nil || cv.Spec.DesiredUpdate.Version == "" {
+		// No upgrade has been started
+		return false, nil
+	}
+
+	for _, h := range cv.Status.History {
+		if h.Version == cv.Spec.DesiredUpdate.Version {
+			return h.State != ocp_configv1.CompletedUpdate, nil
```
haibinxie, line 35: I am a little concerned, openshift upgrade may be in bad state in some cases, and portworx upgrade will be stuck forever, although we have that annotation to force continue update, users may not know it.

During my past week testing, i have seen many times openshift upgrade was stuck and failed in the middle. and sometimes the failure is not related to updating master/worker node. in that case we are still ok to perform portworx upgrade.

haibinxie (commented):
In general the code looks good, however I am not sure how often openshift upgrade fails, which will block portworx upgrade forever. I have seen openshift upgrade failed many times in the past week during my testing.

if there is a way we shall detect openshift upgrade failed, and resume portworx upgrade.

pkg/util/k8s/cluster.go
piyush-nimbalkar: This error happens when the k8s api server does have the `ClusterVersion` kind registered. Which would be almost all k8s clusters except OpenShift. Not returning error to avoid logging this as a warning in the caller as we should ignore this silently in non-openshift cases.

pkg/util/k8s/cluster.go
piyush-nimbalkar: Yes, even I thought about that and it is a valid concern. I saw that in case of a failed ocp upgrade, if you re-trigger the upgrade to a different version, the status in history is changed again and the failed status goes away. Basically if a user has some issues with ocp upgrade then they will either retry or try to fix the upgrade.
I think conservatively we should block the portworx upgrade until they contact either the redhat support or portworx support. One case is that redhat support fixes the ocp upgrade problem, then portworx upgrade goes fine. Another case is that portworx support determines that it is safe to let portworx upgrade and asks them to continue by adding the annotation.
Ideally we want users to not even run ocp and portworx upgrades in parallel and if either one is failing and stuck, they should resolve it and then move ahead with the other one.
Anyway these are my 2 cents on this. Let's check with our support team as well on slack and see what they think about this.

piyush-nimbalkar:
> In general the code looks good, however I am not sure how often openshift upgrade fails, which will block portworx upgrade forever. I have seen openshift upgrade failed many times in the past week during my testing.
> 
> if there is a way we shall detect openshift upgrade failed, and resume portworx upgrade.

There is no clear status of the reason of upgrade failure in the ClusterVersion. It just has a message in the status.conditions. The reason is important because if the issue in upgrade affects portworx upgrade, it would have been better to not bounce portworx in that case. Worst case, we leave a backdoor for the users to proceed, but ideally we want them to fix their ocp upgrade before they trigger portworx upgrade.

haibinxie (approved):
Sounds good.

PR #381, opened 2021-08-28
OPERATOR-415 fix storage node stuck initializing

haibinxie:
https://portworx.atlassian.net/browse/OPERATOR-415

Closes #

After the change, we will only look at the status in NodeInitCondition and NodeStateCondition and not look at timestamp.

### DIFF
--- a/deploy/crds/core_v1_storagenode_crd.yaml
+++ b/deploy/crds/core_v1_storagenode_crd.yaml
@@ -123,8 +123,16 @@ spec:
                       description: Status of the condition.
                     reason:
                       type: string
-                      description: Reason is the human readable message indicating details about the
+                      description: Reason is a unique one-word reason about the current state
+                        of the cluster.
+                    message:
+                      type: string
+                      description: Message is the human readable message indicating details about the
                         current state of the cluster.
+                    lastTransitionTime:
+                      type: string
+                      format: date-time
+                      description: Time at which the condition changed.
               geography:
                 type: object
                 description: Contains topology information for the storage node.
--- a/deploy/crds/deprecated/core_v1_storagenode_crd.yaml
+++ b/deploy/crds/deprecated/core_v1_storagenode_crd.yaml
@@ -126,8 +126,16 @@ spec:
                     description: Status of the condition.
                   reason:
                     type: string
-                    description: Reason is the human readable message indicating details about the
+                    description: Reason is a unique one-word reason about the current state
+                      of the cluster.
+                  message:
+                    type: string
+                    description: Message is the human readable message indicating details about the
                       current state of the cluster.
+                  lastTransitionTime:
+                    type: string
+                    format: date-time
+                    description: Time at which the condition changed.
             geography:
               type: object
               description: Contains topology information for the storage node.
--- a/drivers/storage/portworx/status.go
+++ b/drivers/storage/portworx/status.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"reflect"
 	"strings"
-	"time"
 
 	"github.com/libopenstorage/openstorage/api"
 	pxutil "github.com/libopenstorage/operator/drivers/storage/portworx/util"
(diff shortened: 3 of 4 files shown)


drivers/storage/portworx/status.go
```diff
@@ -468,30 +466,25 @@ func mapNodeStatus(status api.Status) corev1.NodeConditionStatus {
...
-	if latestCondition == nil || latestCondition.Status == "" ||
-		(latestCondition.Type == corev1.NodeInitCondition &&
-			latestCondition.Status == corev1.NodeSucceededStatus) {
-		return string(corev1.NodeInitStatus)
+	if nodeInitCondition == nil || nodeInitCondition.Status == corev1.NodeSucceededStatus {
+		if nodeStateCondition == nil {
+			return string(corev1.NodeInitStatus)
+		}
+
+		return string(nodeStateCondition.Status)
```
piyush-nimbalkar, line 484: This may not be true too. When a node is initializing and has NodeInitCondition's status == Succeeded, the SDK may not have started. This would mean the NodeStatusCondition is empty. The status reported by operator should be Initializing in such case and not empty.

drivers/storage/portworx/status.go
```diff
@@ -468,30 +466,25 @@ func mapNodeStatus(status api.Status) corev1.NodeConditionStatus {
...
-	// Also if the InitCondition is the latest condition and it has succeeded,
-	// then keep the node phase as Initializing
-	if latestCondition == nil || latestCondition.Status == "" ||
-		(latestCondition.Type == corev1.NodeInitCondition &&
-			latestCondition.Status == corev1.NodeSucceededStatus) {
-		return string(corev1.NodeInitStatus)
+	if nodeInitCondition == nil || nodeInitCondition.Status == corev1.NodeSucceededStatus {
+		if nodeStateCondition == nil {
+			return string(corev1.NodeInitStatus)
+		}
```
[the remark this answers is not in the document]
dsupure: Do we have a state machine to describe the state transitions of node? In essence, the operator is a state transition problem. I feel it's challenging to use "if-else" expression to check all conditions and state transitions.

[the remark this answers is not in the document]
piyush-nimbalkar: Currently we don't have a state machine, but I think we should add that in future. This will very soon get complicated and difficult to maintain. 
As part of a future ticket where we want to add node update transitions as well to the conditions, we can implement that. I will update the ticket description.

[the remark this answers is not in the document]
dsupure: I wrote a PoC operator which uses a state machine framework. All state machine's state transition APIs are ready to use. I can give a talk to the operator team when we get a chance. It's not urgent anyway.

deploy/crds/core_v1_storagenode_crd.yaml
```diff
@@ -122,9 +122,17 @@ spec:
                       type: string
                       description: Status of the condition.
                     reason:
+                      type: string
+                      description: Reason is a unique one-word reason about the current state
+                        of the cluster.
+                    message:
                       type: string
                       description: Reason is the human readable message indicating details about the
```
Pure-AdamuKaapan, line 130:
```suggestion
                      description: Message is the human readable message indicating details about the
```

deploy/crds/deprecated/core_v1_storagenode_crd.yaml
```diff
@@ -125,9 +125,17 @@ spec:
                     type: string
                     description: Status of the condition.
                   reason:
+                    type: string
+                    description: Reason is a unique one-word reason about the current state
+                      of the cluster.
+                  message:
                     type: string
                     description: Reason is the human readable message indicating details about the
```
Pure-AdamuKaapan, line 133:
```suggestion
                    description: Message is the human readable message indicating details about the
```

pure-jliao (approved):
LGTM!

drivers/storage/portworx/status.go
```diff
@@ -468,30 +467,27 @@ func mapNodeStatus(status api.Status) corev1.NodeConditionStatus {
...
-	if latestCondition == nil || latestCondition.Status == "" ||
-		(latestCondition.Type == corev1.NodeInitCondition &&
-			latestCondition.Status == corev1.NodeSucceededStatus) {
+	if nodeInitCondition == nil || nodeInitCondition.Status == corev1.NodeSucceededStatus {
+		if nodeStateCondition != nil && nodeStateCondition.Status != "" {
+			return string(nodeStateCondition.Status)
+		}
 		return string(corev1.NodeInitStatus)
+	} else if nodeStateCondition == nil ||
+		nodeStateCondition.LastTransitionTime.Before(&nodeInitCondition.LastTransitionTime) {
```
dsupure, line 487: Is it possible the evaluation of the expression be true? 
`nodeStateCondition.LastTransitionTime.Before(&nodeInitCondition.LastTransitionTime)`

drivers/storage/portworx/status_test.go
```diff
@@ -177,3 +184,283 @@ func TestSetupContextWithToken(t *testing.T) {
...
+				},
+			},
+			expectedPhase: "Initializing",
+		},
+		{
+			name: "with-old-succeeded-init-condition-and-valid-state-condition",
+			pxNode: &api.StorageNode{
+				Id:                "id",
+				SchedulerNodeName: "k8s-node",
+				Status:            api.Status_STATUS_NOT_IN_QUORUM,
```
dsupure, line 280: It might be a naive question, but why we don't need to test other statuses? such as Status_STATUS_MAINTENANCE.

drivers/storage/portworx/status.go
piyush-nimbalkar: This was the main issue in the previous code. 
The NodeInit condition is added by portworx, while the NodeState is added by operator. There was a race when portworx restarted quickly and updated the NodeInit condition, but as operator thought px was Online it never updated the NodeState condition. So the NodeInit condition had a later timestamp than the NodeState condition.

drivers/storage/portworx/status_test.go
piyush-nimbalkar: There is no reason as such. Just wanted to take something other than ONLINE/OK because that is what we end up checking in most places.
Here we wanted the check that NodeState is reflected in the final Phase, instead of NodeInit. So it does not matter what is the Portworx node status is. Anything other than Initializing would have done the job.

drivers/storage/portworx/status.go
dsupure: I see. It's due to the px restart. It's confusing when I first see the equation. I would suggest you put a comment to explain why the NodeInit condition had a later timestamp than the NodeState condition.

Again, the state machine can avoid checking the timestamp. Both PX and Operator can change state, and the operator just needs to do a periodical check in reconciling loop to take action based on the state and conditions.  
Our PSO cockroachDB operator also has similar problems, this is why I wrote a PoC operator.

drivers/storage/portworx/status_test.go
dsupure: Thanks for the explaination.

dsupure (approved):
LGTM!

drivers/storage/portworx/status.go
piyush-nimbalkar: Sounds good. I will add a comment and we can integrate the state machine for conditions later.

piyush-nimbalkar:
Thanks for the reviews, guys!

PR #384, opened 2021-09-01
OP-441 Keep CSI disabled for existing cluster, enable for new install

piyush-nimbalkar:

Closes # https://portworx.atlassian.net/browse/OPERATOR-441

### DIFF
--- a/drivers/storage/portworx/portworx.go
+++ b/drivers/storage/portworx/portworx.go
@@ -629,12 +629,15 @@ func setPortworxDefaults(toUpdate *corev1.StorageCluster) {
 		}
 	}
 
-	// Enable CSI flag by default
-	if _, ok := toUpdate.Spec.FeatureGates[string(pxutil.FeatureCSI)]; !ok {
-		if toUpdate.Spec.FeatureGates == nil {
-			toUpdate.Spec.FeatureGates = make(map[string]string)
+	if t.isK3s || toUpdate.Spec.Version == "" {
+		// Enable CSI if running in k3s environment or
+		// if this is a new Portworx installation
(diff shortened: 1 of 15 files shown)


adityadani (approved):

drivers/storage/portworx/portworx.go
```diff
@@ -629,12 +629,15 @@ func setPortworxDefaults(toUpdate *corev1.StorageCluster) {
 		}
 	}
 
-	// Enable CSI flag by default
-	if _, ok := toUpdate.Spec.FeatureGates[string(pxutil.FeatureCSI)]; !ok {
-		if toUpdate.Spec.FeatureGates == nil {
-			toUpdate.Spec.FeatureGates = make(map[string]string)
+	if t.isK3s || toUpdate.Spec.Version == "" {
```
dsupure, line 632: So k3s cannot use in-tree and only uses CSI?

drivers/storage/portworx/util/feature.go
```diff
@@ -15,5 +15,5 @@ const (
 // IsEnabled checks if the feature is enabled in the given feature map
 func (feature Feature) IsEnabled(featureMap map[string]string) bool {
 	enabled, err := strconv.ParseBool(featureMap[string(feature)])
-	return err != nil || enabled
+	return err == nil && enabled
```
dsupure, line 18: Thanks for catching this minor bug.

dsupure (commented):
Generally looking very good. LGTM!

drivers/storage/portworx/portworx.go
piyush-nimbalkar: Correct. In k3s in-tree plugins are disabled to make k8s lighter. CSI is the only way to interact with storage providers.

PR #385, opened 2021-09-01
OP-406 Remove pure cloud drive config map during cleanup

piyush-nimbalkar:

Closes # https://portworx.atlassian.net/browse/OPERATOR-406

### DIFF
--- a/drivers/storage/portworx/portworx_test.go
+++ b/drivers/storage/portworx/portworx_test.go
@@ -5849,7 +5849,13 @@ func TestDeleteClusterWithUninstallWipeStrategyShouldRemoveConfigMaps(t *testing
 			Namespace: bootstrapCloudDriveNamespace,
 		},
 	}
-	k8sClient := testutil.FakeK8sClient(wiperDS, wiperPod, etcdConfigMap, cloudDriveConfigMap)
+	pureConfigMap := &v1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      pureStorageCloudDriveConfigMap,
+			Namespace: bootstrapCloudDriveNamespace,
+		},
+	}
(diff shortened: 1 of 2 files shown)


drivers/storage/portworx/portworx_test.go
```diff
@@ -5849,15 +5849,21 @@ func TestDeleteClusterWithUninstallWipeStrategyShouldRemoveConfigMaps(t *testing
 			Namespace: bootstrapCloudDriveNamespace,
 		},
 	}
-	k8sClient := testutil.FakeK8sClient(wiperDS, wiperPod, etcdConfigMap, cloudDriveConfigMap)
+	pureConfigMap := &v1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      pureStorageCloudDriveConfigMap,
+			Namespace: bootstrapCloudDriveNamespace,
```
Pure-AdamuKaapan, line 5855: Seeing as the configmap is currently always in the `kube-system` namespace, will this work if the namespace is set to something else? Or will it just silently error/ignore it because it's a "not found" on something we're trying to delete (since it's in a different namespace) and it just won't get cleaned up?

(obviously if we change the other PR to use the install namespace, that would also address this)

Pure-AdamuKaapan (approved):
Overall LGTM, just left one question RE the namespace discussion you pointed me to earlier. Thanks for this!

drivers/storage/portworx/portworx_test.go
piyush-nimbalkar: This change would delete the configmap only from the kube-system namespace. If we decide to change it in porx, we will change the operator code too.
For now, if the configmap doesn't exist, it will silently ignore. This is needed even for non Pure cloud drives. If in a different namespace, it will end up not deleting.

Let's first decide on the porx change, then we can revisit the operator change.

Pure-AdamuKaapan: Cool, all sounds good to me. Thanks!

PR #394, opened 2021-09-14
[OPERATOR-440] Add portworx-kvdb-service for internal kvdb monitoring

pure-jliao:
**What this PR does / why we need it**: Add portworx-kvdb-service for internal kvdb monitoring

Closes # [OPERATOR-440](https://portworx.atlassian.net/browse/OPERATOR-440)

### DIFF
--- a/drivers/storage/portworx/component/portworx_basic.go
+++ b/drivers/storage/portworx/component/portworx_basic.go
@@ -78,6 +78,12 @@ func (c *portworxBasic) Reconcile(cluster *corev1.StorageCluster) error {
 	if err := c.createPortworxService(cluster, ownerRef); err != nil {
 		return NewError(ErrCritical, err)
 	}
+	if cluster.Spec.Kvdb != nil && cluster.Spec.Kvdb.Internal {
+		if err := c.createPortworxKVDBService(cluster, ownerRef); err != nil {
+			// This should not block portworx installation
+			return err
+		}
+	}
 	return nil
 }
 
@@ -93,6 +99,9 @@ func (c *portworxBasic) Delete(cluster *corev1.StorageCluster) error {
 		return err
 	}
 	if err := k8sutil.DeleteService(c.k8sClient, pxutil.PortworxServiceName, cluster.Namespace, *ownerRef); err != nil {
+		return err
+	}
+	if err := k8sutil.DeleteService(c.k8sClient, pxutil.PortworxKVDBServiceName, cluster.Namespace, *ownerRef); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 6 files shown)


drivers/storage/portworx/component/portworx_basic.go
```diff
@@ -479,6 +474,68 @@ func getPortworxServiceSpec(
...
+	serviceType := pxutil.ServiceType(cluster)
+	if serviceType != "" {
+		newService.Spec.Type = serviceType
+	}
+
+	return newService
+}
+
+func getTargetPorts(startPort int) (int, int, int) {
+	kvdbTargetPort := 9019
```
dsupure, line 528: Why it is the targetPort?  Shouldn't it be `kvdbDefaultPort` or `kvdbServicePort`?

drivers/storage/portworx/component/portworx_basic.go
```diff
@@ -479,6 +474,68 @@ func getPortworxServiceSpec(
...
+	sdkTargetPort := 9020
```
dsupure, line 529: Same question for restGetway.  Shouldn't it be `restGetwayDefaultPort` or `restGetwayServicePort`?

drivers/storage/portworx/components_test.go
```diff
@@ -389,6 +401,37 @@ func TestBasicInstallWithPortworxDisabled(t *testing.T) {
...
+			Kvdb: &corev1.KvdbSpec{
+				Internal: false,
+			},
+		},
+	}
+
+	err := driver.PreInstall(cluster)
+	require.NoError(t, err)
+
+	// Portworx Services should not be created
```
dsupure, line 429: Why it should not be created?

dsupure (commented):
Please fix the test covergage.

drivers/storage/portworx/component/portworx_basic.go
```diff
@@ -78,6 +78,11 @@ func (c *portworxBasic) Reconcile(cluster *corev1.StorageCluster) error {
 	if err := c.createPortworxService(cluster, ownerRef); err != nil {
 		return NewError(ErrCritical, err)
 	}
+	if cluster.Spec.Kvdb != nil && cluster.Spec.Kvdb.Internal {
+		if err := c.createPortworxKVDBService(cluster, ownerRef); err != nil {
+			return NewError(ErrCritical, err)
```
piyush-nimbalkar, line 83: Actually, let's not return ErrCritical here, but instead just return err directly.
All above are necessary for portworx to function and so we want installation to fail if they fail, but the kvdb service is not critical.

drivers/storage/portworx/component/portworx_basic.go
piyush-nimbalkar: A kubernetes service has two ports - `port` and `targetPort`. Port is what the service exposes and the TargetPort is the port on which the underlying service is accessed. In this case the portworx target ports like kvdb, sdk, rest gateway, etc.
TargetPort is just a k8s terminology.

drivers/storage/portworx/component/portworx_basic.go
piyush-nimbalkar: Same as above.

drivers/storage/portworx/components_test.go
piyush-nimbalkar: .. because internal kvdb is disabled

We can improve the comment to something like - `Portworx kvdb service should not be created`

piyush-nimbalkar (commented):
The changes look good!
Did you try this with px 2.8.1 and a running prometheus?

pure-jliao:
> The changes look good!
> Did you try this with px 2.8.1 and a running prometheus?

Tested manually seems all working fine. some results posted in the jira.  @piyush-nimbalkar

piyush-nimbalkar (approved):

PR #398, opened 2021-09-15
Added FlashArray/FlashBlade matrix tests

Pure-AdamuKaapan:
This adds a functional testing framework that will test various `StorageCluster` specs and validate that Portworx does (or doesn't) come up successfully. It also adds some framework improvements to detect when Portworx fails to start, rather than always starting successfully, enabling for negative testing.

**Notes for your reviewer**:
Most tests are commented out right now, I have just one test enabled. I will uncomment more as I validate they work as expected, either in this PR or in future PRs.

### DIFF
--- a/pkg/util/test/util.go
+++ b/pkg/util/test/util.go
@@ -10,6 +10,7 @@ import (
 	"net/url"
 	"os"
 	"path"
+	"reflect"
 	"regexp"
 	"strconv"
 	"strings"
@@ -349,6 +350,7 @@ func ValidateStorageCluster(
 	pxImageList map[string]string,
 	clusterSpec *corev1.StorageCluster,
 	timeout, interval time.Duration,
+	shouldStartSuccessfully bool,
 	kubeconfig ...string,
 ) error {
 	// Set kubeconfig
@@ -357,8 +359,20 @@ func ValidateStorageCluster(
 	}
 
 	// Validate StorageCluster
-	liveCluster, err := validateStorageClusterIsOnline(clusterSpec, timeout, interval)
-	if err != nil {
+	var liveCluster *corev1.StorageCluster
+	var err error
+	if shouldStartSuccessfully {
+		liveCluster, err = validateStorageClusterIsOnline(clusterSpec, timeout, interval)
+		if err != nil {
+			return err
+		}
+	} else {
+		// If we shouldn't start successfully, this is all we need to check
+		return validateStorageClusterIsFailed(clusterSpec, timeout, interval)
+	}
+
+	// Validate that spec matches live spec
+	if err = validateDeployedSpec(clusterSpec, liveCluster); err != nil {
 		return err
 	}
 
@@ -368,7 +382,7 @@ func ValidateStorageCluster(
 	}
 
 	// Get list of expected Portworx node names
-	expectedPxNodeNameList, err := getExpectedPxNodeNameList(clusterSpec)
+	expectedPxNodeNameList, err := GetExpectedPxNodeNameList(clusterSpec)
 	if err != nil {
 		return err
 	}
@@ -383,7 +397,6 @@ func ValidateStorageCluster(
 		return err
 	}
 
-	// TODO: Add validation to check what is expected (clusterSpec) and what actually got deployed (liveSpec)
 	if err = validateComponents(pxImageList, liveCluster, timeout, interval); err != nil {
 		return err
 	}
@@ -436,6 +449,41 @@ func validateStorageNodes(pxImageList map[string]string, cluster *corev1.Storage
 	if _, err := task.DoRetryWithTimeout(t, timeout, interval); err != nil {
 		return err
 	}
+
+	return nil
+}
+
+func sortNodesByNodeName(nodes []corev1.NodeSpec) func(i, j int) bool {
+	return func(i, j int) bool {
+		return nodes[i].Selector.NodeName < nodes[j].Selector.NodeName
+	}
+}
+
+func nodeSpecsToMaps(nodes []corev1.NodeSpec) map[string]*corev1.CloudStorageNodeSpec {
+	toReturn := map[string]*corev1.CloudStorageNodeSpec{}
+
+	for _, node := range nodes {
+		toReturn[node.Selector.NodeName] = node.CloudStorage
+	}
+
+	return toReturn
+}
+
+func validateDeployedSpec(expected, live *corev1.StorageCluster) error {
+	// Validate cloudStorage
+	if !reflect.DeepEqual(expected.Spec.CloudStorage, live.Spec.CloudStorage) {
+		return fmt.Errorf("deployed CloudStorage spec doesn't match expected")
+	}
+	// Validate kvdb
+	if !reflect.DeepEqual(expected.Spec.Kvdb, live.Spec.Kvdb) {
+		return fmt.Errorf("deployed Kvdb spec doesn't match expected")
+	}
+	// Validate nodes
+	if !reflect.DeepEqual(nodeSpecsToMaps(expected.Spec.Nodes), nodeSpecsToMaps(live.Spec.Nodes)) {
+		return fmt.Errorf("deployed Nodes spec doesn't match expected")
+	}
+
+	// TODO: validate more parts of the spec as we test with them
 
 	return nil
 }
@@ -653,7 +701,10 @@ func validatePortworxNodes(cluster *corev1.StorageCluster, expectedNodes int) er
 	return nil
 }
 
-func getExpectedPxNodeNameList(cluster *corev1.StorageCluster) ([]string, error) {
+// GetExpectedPxNodeNameList will get the list of node names that should be included
+// in the given Portworx cluster, by seeing if each non-master node matches the given
+// node selectors and affinities.
+func GetExpectedPxNodeNameList(cluster *corev1.StorageCluster) ([]string, error) {
 	var nodeNameListWithPxPods []string
 	nodeList, err := coreops.Instance().GetNodes()
 	if err != nil {
@@ -1051,14 +1102,13 @@ func ConstructPxReleaseManifestURL(specGenURL string) (string, error) {
 	return u.String(), nil
 }
 
-func validateStorageClusterIsOnline(cluster *corev1.StorageCluster, timeout, interval time.Duration) (*corev1.StorageCluster, error) {
-	t := func() (interface{}, bool, error) {
+func validateStorageClusterInState(cluster *corev1.StorageCluster, status corev1.ClusterConditionStatus) func() (interface{}, bool, error) {
+	return func() (interface{}, bool, error) {
 		cluster, err := operatorops.Instance().GetStorageCluster(cluster.Name, cluster.Namespace)
 		if err != nil {
 			return nil, true, fmt.Errorf("failed to get StorageCluster %s in %s, Err: %v", cluster.Name, cluster.Namespace, err)
 		}
-
-		if cluster.Status.Phase != "Online" {
+		if cluster.Status.Phase != string(status) {
 			if cluster.Status.Phase == "" {
 				return nil, true, fmt.Errorf("failed to get cluster status")
 			}
@@ -1066,14 +1116,42 @@ func validateStorageClusterIsOnline(cluster *corev1.StorageCluster, timeout, int
 		}
 		return cluster, false, nil
 	}
+}
 
-	out, err := task.DoRetryWithTimeout(t, timeout, interval)
+func validateAllStorageNodesInState(namespace string, status corev1.NodeConditionStatus) func() (interface{}, bool, error) {
+	return func() (interface{}, bool, error) {
+		// Get all StorageNodes
+		storageNodeList, err := operatorops.Instance().ListStorageNodes(namespace)
+		if err != nil {
+			return nil, true, fmt.Errorf("failed to list StorageNodes in %s, Err: %v", namespace, err)
+		}
+
+		for _, node := range storageNodeList.Items {
+			if node.Status.Phase != string(status) {
+				return nil, true, fmt.Errorf("StorageNode %s in %s is in state %v", node.Name, namespace, node.Status.Phase)
+			}
+		}
+
+		return nil, false, nil
+	}
+}
+
+func validateStorageClusterIsOnline(cluster *corev1.StorageCluster, timeout, interval time.Duration) (*corev1.StorageCluster, error) {
+	out, err := task.DoRetryWithTimeout(validateStorageClusterInState(cluster, corev1.ClusterOnline), timeout, interval)
 	if err != nil {
 		return nil, fmt.Errorf("failed to wait for StorageCluster to be ready, Err: %v", err)
 	}
 	cluster = out.(*corev1.StorageCluster)
 
 	return cluster, nil
+}
+
+func validateStorageClusterIsFailed(cluster *corev1.StorageCluster, timeout, interval time.Duration) error {
+	_, err := task.DoRetryWithTimeout(validateAllStorageNodesInState(cluster.Namespace, corev1.NodeFailedStatus), timeout, interval)
+	if err != nil {
+		return fmt.Errorf("failed to wait for StorageNodes to be failed, Err: %v", err)
+	}
+	return nil
 }
 
 // CreateClusterWithTLS is a helper method
--- a/pkg/util/test/util.go
+++ b/pkg/util/test/util.go
@@ -453,6 +453,9 @@ func validateStorageNodes(pxImageList map[string]string, cluster *corev1.Storage
 	return nil
 }
 
+// nodeSpecsToMaps takes the given node spec list and converts it to a map of node names to
+// cloud storage specs. Note that this will not work for label selectors at the moment, only
+// node names.
 func nodeSpecsToMaps(nodes []corev1.NodeSpec) map[string]*corev1.CloudStorageNodeSpec {
 	toReturn := map[string]*corev1.CloudStorageNodeSpec{}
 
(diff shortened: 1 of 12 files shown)


test/integration_test/fafb_matrix_test.go
```diff
@@ -0,0 +1,331 @@
...
+			Kvdb: &internalKVDB,
+			CloudStorage: &v1.CloudStorageSpec{
+				CloudStorageCommon: v1.CloudStorageCommon{
+					KvdbDeviceSpec: &size32,
+					DeviceSpecs:    &disks100,
+				},
+			},
+		},
+	},
+	//// C55130: https://portworx.testrail.net/index.php?/cases/view/55130
```
vilasd, line 82: Why following code is commented out ??? Is it not working or for future coding ??

pkg/util/test/util.go
```diff
@@ -1042,23 +1054,49 @@ func ConstructPxReleaseManifestURL(specGenURL string) (string, error) {
...
 			if cluster.Status.Phase == "" {
 				return nil, true, fmt.Errorf("failed to get cluster status")
 			}
 			return nil, true, fmt.Errorf("cluster state: %s", cluster.Status.Phase)
 		}
 		return cluster, false, nil
 	}
+}
+
+func validateAllStorageNodesInState(namespace string, status corev1.NodeConditionStatus) func() (interface{}, bool, error) {
```
vilasd, line 1074: This validation does not co-relate with the test case you are running. Its not enough just to check if the storage node is online and say its passed. We need to get the configuration of the storage nodes and verify them after state is online on successful test.

vilasd (changes requested):

test/integration_test/fafb_matrix_test.go
Pure-AdamuKaapan: > Most tests are commented out right now, I have just one test enabled. I will uncomment more as I validate they work as expected, either in this PR or in future PRs.

pkg/util/test/util.go
Pure-AdamuKaapan: This is only run for testing cases where portworx should **not** start up successfully, because this is a more consistent way to find that installation failed (the operator `StorageCluster` status doesn't update, but the `StorageNode` status does). The success cases still go through the full validation path that the operator tests use before.

Pure-AdamuKaapan: Talked over with @vilasd in Zoom, we need to add functionality to check that at least the storage sections (or anything else relevant to the test) are set correctly in the storage cluster once it's applied. Should be relatively easy, I'm working on it now and will leave space for future improvements.

test/integration_test/test-deploy.sh
```diff
@@ -112,6 +112,12 @@ for i in $(seq 1 100) ; do
     test_status=$(kubectl -n kube-system get pod operator-test -o json | jq ".status.phase" -r)
     if [ "$test_status" == "Running" ] || [ "$test_status" == "Completed" ]; then
         break
```
dsupure, line 114: [nit], can we add this here?
```
echo "Tests succeeded"
```

test/integration_test/fafb_matrix_test.go
```diff
@@ -0,0 +1,331 @@
...
+var flashArrayPositiveInstallTests = []TestrailCase{
+	// C55129: https://portworx.testrail.net/index.php?/cases/view/55129
+	// Install PX through operator with internal KVDB on FlashArray
+	{
+		ID:                      "C55129",
+		ShouldStartSuccessfully: true,
+		BackendRequirements: BackendRequirements{
+			RequiredArrays: 1,
+		},
+		Spec: v1.StorageClusterSpec{
```
dsupure, line 72: Let's add a TODO here:  Add the `StorageClusterSpecGen` API in the future.

test/integration_test/fafb_matrix_test.go
```diff
@@ -0,0 +1,331 @@
...
+		}
+		// Add extra env variables
+		cluster.Spec.Env = append(cluster.Spec.Env, []k8sv1.EnvVar{
+			{
+				Name:  "DEFAULT_OSB_URL",
+				Value: "https://master-staging.portworx.dev/osb/billing/v1/register",
+			},
+			{
+				Name:  "PX_RELEASE_MANIFEST_URL",
+				Value: "https://install.portworx.com/2.8.0/version",
```
dingyin, line 252: should we use 2.9.0 here now?

test/integration_test/test-deploy.sh
Pure-AdamuKaapan: This isn't for checking if the tests succeeded: this is just to make sure that the pod has started (e.g. the tests are running). The next step in this script is to follow the logs of the pod.

At the [very end](https://github.com/libopenstorage/operator/pull/398/files#diff-812c21dc01db7e65cbf367911b1630f59a7191cd4a36649258c4e373c4a84e21R141) there's already a line printing out "Tests passed".

test/integration_test/fafb_matrix_test.go
Pure-AdamuKaapan: @dsupure what do you mean by that?

test/integration_test/fafb_matrix_test.go
Pure-AdamuKaapan: That's a good question. @piyush-nimbalkar a) should this value instead be taken from some argument to the tests, and b) is there any value that will always keep us up to date with the latest version, or will we have to keep changing the URL over time?

Pure-AdamuKaapan: In hindsight, there's an easily accessible `pxSpecGenURL` variable in the tests that gets passed in from the pod spec: I should probably just base this off of that. Changing...

dingyin: Agreed. It changes every release. One solution could be from an Env variable.

vilasd (approved):
LGTM

test/integration_test/fafb_matrix_test.go
dsupure: nm, please ignore my comment. I was thinking to write the `StorageClusterSpecGen` that can generate a StorageCluster object instead of filling all the fields.

dsupure (approved):
LGTM! Make sure the Jenkins pipeline is work before you merge.

test/integration_test/fafb_matrix_test.go
```diff
@@ -0,0 +1,338 @@
...
+	cluster.Spec.CloudStorage = trc.Spec.CloudStorage
+	cluster.Spec.Kvdb = trc.Spec.Kvdb
+	cluster.Spec.Nodes = trc.Spec.Nodes
+}
+
+var flashArrayPositiveInstallTests = []TestrailCase{
+	// C55129: https://portworx.testrail.net/index.php?/cases/view/55129
+	// Install PX through operator with internal KVDB on FlashArray
+	{
+		ID:                      "C55129",
```
piyush-nimbalkar, line 68: I like this idea of printing the test case ID from Testrail. It would have been even more useful if it was a link or had some description, but this is a great start.
We should explore if we can automate updating Testrail based on the failure/success of these tests using some Testrail API. Just thinking out loud.

test/integration_test/fafb_matrix_test.go
```diff
@@ -0,0 +1,338 @@
+// +build integrationtest
+// +build fafb
```
piyush-nimbalkar, line 2: Why use both the tags and not just fafb?
Is the plan to run the whole operator test suite on fafb too?

pkg/util/k8s/k8s.go
```diff
@@ -1551,6 +1551,23 @@ func removeOwners(current, toBeDeleted []metav1.OwnerReference) []metav1.OwnerRe
 	return newOwners
 }
 
+// GetSecret returns the data from the given secret
+func GetSecret(k8sClient client.Client,
```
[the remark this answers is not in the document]
piyush-nimbalkar: We might not need this if we use sched-ops.

pkg/util/test/util.go
```diff
@@ -1042,23 +1078,49 @@ func ConstructPxReleaseManifestURL(specGenURL string) (string, error) {
...
+	return func() (interface{}, bool, error) {
+		// Get all StorageNodes
+		storageNodeList, err := operatorops.Instance().ListStorageNodes(namespace)
+		if err != nil {
+			return nil, true, fmt.Errorf("failed to list StorageNodes in %s, Err: %v", namespace, err)
+		}
+
+		for _, node := range storageNodeList.Items {
+			for _, cond := range node.Status.Conditions {
+				if cond.Type != corev1.NodeInitCondition {
```
piyush-nimbalkar, line 1108: When the NodeInit condition fails, doesn't the node.Status.Phase change to 'Failed' too? I thought it did. If not, it seems like an issue in operator code. If yes, you can simply use phase instead of going through the conditions.

test/integration_test/fafb_config_util.go
```diff
@@ -0,0 +1,199 @@
...
+		}
+
+		cl, _ := k8sutil.NewK8sClient(scheme)
+		data, err := k8sutil.GetSecret(cl, SourceConfigSecretName, namespace)
+		if err != nil {
+			if !errors.IsNotFound(err) {
+				sourceConfigLoadError = err
+			}
+			return
+		}
```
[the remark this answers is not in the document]
Pure-AdamuKaapan: Heck yes! I'll go ahead and switch over to it, thanks!

test/integration_test/fafb_matrix_test.go
```diff
@@ -0,0 +1,338 @@
...
+		if !strings.HasSuffix(releaseManifestURL, "/") {
+			releaseManifestURL += "/"
+		}
+		releaseManifestURL += "version"
+
+		cluster.Spec.Env = append(cluster.Spec.Env, []k8sv1.EnvVar{
+			{
+				Name:  "DEFAULT_OSB_URL",
+				Value: "https://master-staging.portworx.dev/osb/billing/v1/register",
+			},
```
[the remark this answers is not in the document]
Pure-AdamuKaapan: Honestly wasn't sure if it was, it was in my spec so I copied it over. I'll remove it and if for some reason that does break it I'll look into why

test/integration_test/fafb_matrix_test.go
Pure-AdamuKaapan: Definitely, that was one of our original goals of this project but it didn't quite make it into the P0 tests we wanted to get automated. We would love to add this and will hopefully take a look at it soon, good project for a rainy day

test/integration_test/fafb_matrix_test.go
Pure-AdamuKaapan: No particular reason, was just adding on to the existing tag with an AND in case there was some issue with having no tag or a different one. If it's okay to have a separate one I'll go ahead and change it.

pkg/util/test/util.go
Pure-AdamuKaapan: I'll go ahead and confirm, this may have just been a case of me trying to find the right variable and not noticing "Phase" as being equivalent to "Status". I'll let you know either way.

test/integration_test/fafb_matrix_test.go
Pure-AdamuKaapan: Another thinking out loud idea, I wonder if we could have a `go gen` template that scans through all the test cases and adds flavor comments? e.g. it scans all `TestCase` structs being instantiated and for each test ID it sees, it hits the testrail API and gets the test name and description.

test/integration_test/fafb_matrix_test.go
Pure-AdamuKaapan: @piyush-nimbalkar actually one question is do we want to maybe have this build two different test binaries? We can then have flag in our test run script for which binary to run as well?

pkg/util/test/util.go
Pure-AdamuKaapan: Discussed on Slack, will be assuming StorageNode status will be correct for now and have opened a bug as far as status not updating properly.

test/integration_test/fafb_config_util.go
```diff
@@ -0,0 +1,197 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"sync"
+	"testing"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+
+	coreops "github.com/portworx/sched-ops/k8s/core"
+	"github.com/sirupsen/logrus"
```
pure-jliao, line 14: nit: line 14 and 15 should belong to the second group?

test/integration_test/test-deploy.sh
```diff
@@ -120,6 +126,7 @@ done
 
 kubectl -n kube-system logs -f operator-test
 for i in $(seq 1 100) ; do
+	sleep 5  # Give the test pod a chance to finish first after the logs stop
```
pure-jliao, line 129: nit: seems we have a tab here instead of spaces.

test/integration_test/operator-test-pod.yaml
```diff
@@ -51,6 +51,7 @@ spec:
     - -t
     - /operator.test
     - -test.v
+    - -test.failfast
```
pure-jliao, line 54: do we really need this?

test/integration_test/fafb_matrix_test.go
```diff
@@ -0,0 +1,352 @@
+// +build fafb
+
+package integrationtest
+
+import (
+	"fmt"
+	"k8s.io/apimachinery/pkg/api/errors"
```
pure-jliao, line 7: nit: import group

test/integration_test/fafb_matrix_test.go
```diff
@@ -0,0 +1,352 @@
...
+	}
+
+	// simpleClusterSpec is a StorageCluster with internal KVDB and auto journal drives
+	simpleClusterSpec = v1.StorageClusterSpec{
+		Kvdb:         &internalKVDB,
+		CloudStorage: autoJournalDrive,
+	}
+)
+
+type TestrailCase struct {
```
pure-jliao, line 61: can we move this to a common util package? and make `BackendRequirements` a pointer so it could be nil? this can be used for other tests

pure-jliao (approved):
LGTM!

test/integration_test/operator-test-pod.yaml
Pure-AdamuKaapan: It helps so that if a test fails, we don't keep making more StorageCluster objects (either in case the first one does eventually finish cleaning up, or just to avoid spam). It also helps the tests finish faster if there's some simple setup issue or an early one fails.

Also I _did_ confirm this works unlike that time we added it in PSO tests and it wasn't actually doing anything :joy:

test/integration_test/fafb_matrix_test.go
Pure-AdamuKaapan: Definitely, I was actually thinking about something similar, I have a few ways I can make it more generic (I'll probably make a `TestrailCase` in common test utils, and then a `PureTestrailCase` that will compose that and add the Pure backend requirements).

test/integration_test/fafb_matrix_test.go
Pure-AdamuKaapan: FYI I just converted this to be a list of strings, so that multiple test cases in one run is more viable. It just joins them together with hyphens for the cluster name, but now we can more easily mark each test as passed once we add that API.

test/integration_test/test_common.go
```diff
@@ -51,6 +56,62 @@ const (
 	defaultValidateUninstallRetryInterval = 30 * time.Second
 )
 
+// node* is to be used in the Node section of the StorageCluster spec. node0 will select the
```
Pure-AdamuKaapan, line 59: NB: renamed this file from `common_test` to `test_common` so that non-test files in the test folder can also use methods from here.

pkg/util/test/util.go
```diff
@@ -440,6 +453,35 @@ func validateStorageNodes(pxImageList map[string]string, cluster *corev1.Storage
 	return nil
 }
 
+func nodeSpecsToMaps(nodes []corev1.NodeSpec) map[string]*corev1.CloudStorageNodeSpec {
+	toReturn := map[string]*corev1.CloudStorageNodeSpec{}
+
+	for _, node := range nodes {
+		toReturn[node.Selector.NodeName] = node.CloudStorage
```
piyush-nimbalkar, line 460: A node selector could be either nodeName or a labelSelector. If the nodeName is empty (when using labelSelector), this will not work. 
I guess it is okay for now assuming your tests are only using nodeName, but just something to remember. May be just put a TODO comment here saying this does not handle node specs with label selectors.

test/integration_test/fafb_matrix_test.go
```diff
@@ -0,0 +1,305 @@
...
+		},
+		Data: map[string][]byte{
+			"pure.json": pureJSON,
+		},
+	})
+	return err
+}
+
+func deletePureSecretIfExists(namespace string) error {
+	if err := coreops.Instance().DeleteSecret(OutputSecretName, namespace); !errors.IsNotFound(err) && !errors.IsAlreadyExists(err) {
```
piyush-nimbalkar, line 301: nit: No need to check `errors.IsAlreadyExists`. I don't think k8s apiserver will return that error for a Delete API. It would only happen for a Create API.

test/integration_test/fafb_matrix_test.go
piyush-nimbalkar: I think a flag with single binaries is better.

piyush-nimbalkar (approved):
LGTM
Good work here! 💯

pkg/util/test/util.go
Pure-AdamuKaapan: Yeah, good point, let me document that.

PR #406, opened 2021-09-21
Set max supported openshift version to 4.8

piyush-nimbalkar:
Openshift 4.9 will use Kubernetes 1.22 which does not support
CRD v1beta1. We are adding support for 4.9 in operator 1.6.0.

Closes #

### DIFF
--- a/deploy/olm-catalog/portworx/1.5.2/portworxoperator.v1.5.2.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/portworx/1.5.2/portworxoperator.v1.5.2.clusterserviceversion.yaml
@@ -4,6 +4,7 @@ metadata:
   name: portworx-operator.v1.5.2
   namespace: placeholder
   annotations:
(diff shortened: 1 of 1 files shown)


pure-jliao (approved):

pure-jliao:
does this mean we won't test openshift 4.9 deploy?

piyush-nimbalkar:
> does this mean we won't test openshift 4.9 deploy?

Yes, we won't test openshift 4.9 with operator 1.5.2.
We will be adding support to openshift 4.9 in operator 1.6 because it uses k8s 1.22.

dsupure (approved):
I like it. Can we also add one item to the checklist in our release wiki to make sure we update this in each new release?

piyush-nimbalkar:
> I like it. Can we also add one item to the checklist in our release wiki to make sure we update this in each new release?

Actually, we won't need this starting 1.6.0 as we will support for openshift 4.9 and Kubernetes 1.22 in that release.
But I will add this as a note or a link to openshift docs, so we remember what needs to be done in future.

PR #409, opened 2021-09-29
[OPERATOR-454] Upgrade prometheus operator to a newer version

pure-jliao:

**What this PR does / why we need it**: Upgrade Prometheus operator and related images to newer version and allow operator to create required CRDs for k8s 1.22 and above

Closes # [OPERATOR-454](https://portworx.atlassian.net/browse/OPERATOR-454)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,8 @@ STORAGE_OPERATOR_IMG=$(DOCKER_HUB_REPO)/$(DOCKER_HUB_STORAGE_OPERATOR_IMAGE):$(D
 STORAGE_OPERATOR_TEST_IMG=$(DOCKER_HUB_REPO)/$(DOCKER_HUB_STORAGE_OPERATOR_TEST_IMAGE):$(DOCKER_HUB_STORAGE_OPERATOR_TEST_TAG)
 PX_DOC_HOST ?= https://docs.portworx.com
 PX_INSTALLER_HOST ?= https://install.portworx.com
+PROMETHEUS_OPERATOR_HELM_CHARTS_TAG ?= kube-prometheus-stack-19.0.1
+PROMETHEUS_OPERATOR_CRD_URL_PREFIX ?= https://raw.githubusercontent.com/prometheus-community/helm-charts/$(PROMETHEUS_OPERATOR_HELM_CHARTS_TAG)/charts/kube-prometheus-stack/crds
 
 HAS_GOMODULES := $(shell go help mod why 2> /dev/null)
 
@@ -175,10 +177,18 @@ verify-catalog:
 downloads: getconfigs get-release-manifest
 
 cleanconfigs:
-	rm -f "bin/configs/portworx-prometheus-rule.yaml"
+	rm -rf bin/configs
 
 getconfigs: cleanconfigs
 	wget -q '$(PX_DOC_HOST)/samples/k8s/pxc/portworx-prometheus-rule.yaml' -P bin/configs --no-check-certificate
+	wget -q '$(PROMETHEUS_OPERATOR_CRD_URL_PREFIX)/crd-alertmanagerconfigs.yaml' -O bin/configs/prometheus-crd-alertmanagerconfigs.yaml
+	wget -q '$(PROMETHEUS_OPERATOR_CRD_URL_PREFIX)/crd-alertmanagers.yaml' -O bin/configs/prometheus-crd-alertmanagers.yaml
+	wget -q '$(PROMETHEUS_OPERATOR_CRD_URL_PREFIX)/crd-podmonitors.yaml' -O bin/configs/prometheus-crd-podmonitors.yaml
+	wget -q '$(PROMETHEUS_OPERATOR_CRD_URL_PREFIX)/crd-probes.yaml' -O bin/configs/prometheus-crd-probes.yaml
+	wget -q '$(PROMETHEUS_OPERATOR_CRD_URL_PREFIX)/crd-prometheuses.yaml' -O bin/configs/prometheus-crd-prometheuses.yaml
+	wget -q '$(PROMETHEUS_OPERATOR_CRD_URL_PREFIX)/crd-prometheusrules.yaml' -O bin/configs/prometheus-crd-prometheusrules.yaml
+	wget -q '$(PROMETHEUS_OPERATOR_CRD_URL_PREFIX)/crd-servicemonitors.yaml' -O bin/configs/prometheus-crd-servicemonitors.yaml
+	wget -q '$(PROMETHEUS_OPERATOR_CRD_URL_PREFIX)/crd-thanosrulers.yaml' -O bin/configs/prometheus-crd-thanosrulers.yaml
 
 clean-release-manifest:
 	rm -rf manifests
--- a/drivers/storage/portworx/component/prometheus.go
+++ b/drivers/storage/portworx/component/prometheus.go
@@ -3,6 +3,8 @@ package component
 import (
 	"context"
 	"fmt"
+	"path"
+	"path/filepath"
 
 	"github.com/hashicorp/go-version"
 	pxutil "github.com/libopenstorage/operator/drivers/storage/portworx/util"
@@ -17,6 +19,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metaerrors "k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/api/resource"
@@ -54,6 +57,7 @@ const (
 
 type prometheus struct {
 	k8sClient         client.Client
+	k8sVersion        *version.Version
 	scheme            *runtime.Scheme
 	recorder          record.EventRecorder
 	isOperatorCreated bool
@@ -69,11 +73,12 @@ func (c *prometheus) Priority() int32 {
 
 func (c *prometheus) Initialize(
 	k8sClient client.Client,
-	_ version.Version,
+	k8sVersion version.Version,
 	scheme *runtime.Scheme,
 	recorder record.EventRecorder,
 ) {
 	c.k8sClient = k8sClient
+	c.k8sVersion = &k8sVersion
 	c.scheme = scheme
 	c.recorder = recorder
 }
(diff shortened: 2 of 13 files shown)


drivers/storage/portworx/components_test.go
```diff
@@ -5830,6 +5831,75 @@ func TestCSI_0_3_NodeAffinityChange(t *testing.T) {
...
+			},
+		},
+		&client.CreateOptions{},
+	)
+
+	err := driver.PreInstall(cluster)
+	require.NoError(t, err)
+
+	// Prometheus operator ClusterRole should be updated as well
+	expectedCR := testutil.GetExpectedClusterRole(t, "prometheusNewOperatorClusterRole.yaml")
```
piyush-nimbalkar, line 5894: Can we verify the prometheus-operator deployment too without the config map reload argument and the correct images?

drivers/storage/portworx/portworx.go
```diff
@@ -893,6 +895,12 @@ func hasPrometheusChanged(cluster *corev1.StorageCluster) bool {
 		cluster.Status.DesiredImages.PrometheusOperator == ""
 }
 
+func hasPrometheusVersionChanged(cluster *corev1.StorageCluster, portworx *portworx) bool {
+	return portworx.k8sVersion != nil &&
```
piyush-nimbalkar, line 899: nit: Here we can assume k8sVersion is not nil as we return error from the very beginning if there is an error getting the k8sVersion.

pkg/util/k8s/k8s.go
```diff
@@ -111,6 +116,28 @@ func ParseObjectFromFile(
...
+	} else {
+		crd.ResourceVersion = latestCRD.ResourceVersion
+		if _, err := apiextensionsops.Instance().UpdateCRD(crd); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
```
piyush-nimbalkar, line 140: You could add unit test for this function.

drivers/storage/portworx/components_test.go
piyush-nimbalkar: We should even check that the CRDs have been registered.

drivers/storage/portworx/component/prometheus.go
```diff
@@ -227,15 +274,8 @@ func (c *prometheus) createOperatorClusterRole() error {
...
-					Resources: []string{
-						"alertmanagers",
-						"prometheuses",
-						"prometheuses/finalizers",
-						"servicemonitors",
-						"prometheusrules",
-						"podmonitors",
-					},
-					Verbs: []string{"*"},
+					Resources: monitoringResources,
```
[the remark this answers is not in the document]
pure-jliao: there are some other updates I believe, I'll update operator clusterrole and prometheus clusterrole

drivers/storage/portworx/component/prometheus.go
```diff
@@ -85,6 +90,11 @@ func (c *prometheus) IsEnabled(cluster *corev1.StorageCluster) bool {
 }
 
 func (c *prometheus) Reconcile(cluster *corev1.StorageCluster) error {
+	if c.k8sVersion != nil && c.k8sVersion.GreaterThanOrEqual(k8sutil.K8sVer1_22) {
```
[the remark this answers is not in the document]
pure-jliao: I'll keep it since it could cause problems in unit tests potentially.

Makefile
```diff
@@ -16,6 +16,8 @@ STORAGE_OPERATOR_IMG=$(DOCKER_HUB_REPO)/$(DOCKER_HUB_STORAGE_OPERATOR_IMAGE):$(D
 STORAGE_OPERATOR_TEST_IMG=$(DOCKER_HUB_REPO)/$(DOCKER_HUB_STORAGE_OPERATOR_TEST_IMAGE):$(DOCKER_HUB_STORAGE_OPERATOR_TEST_TAG)
 PX_DOC_HOST ?= https://docs.portworx.com
 PX_INSTALLER_HOST ?= https://install.portworx.com
+PROMETHEUS_OPERATOR_HELM_CHARTS_TAG ?= kube-prometheus-stack-18.1.0
```
dsupure, line 19: Does the helm chart tag change often? Do we need to keep it up-to-date?

drivers/storage/portworx/component/prometheus.go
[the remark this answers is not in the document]
dsupure: Can you add the reference here:
https://github.com/prometheus-operator/prometheus-operator/blob/release-0.50/Documentation/rbac.md
And check the RBAC rule again?

drivers/storage/portworx/manifest/manifest.go
```diff
@@ -243,6 +225,46 @@ func fillCSIDefaults(
 	rel.Components.CSISnapshotter = csiImages.Snapshotter
 }
 
+func fillPrometheusDefaults(
+	rel *Version,
+	k8sVersion *version.Version,
+) {
+	if k8sVersion != nil && k8sVersion.GreaterThanOrEqual(k8sutil.K8sVer1_22) {
+		if rel.Components.Prometheus == "" {
+			rel.Components.Prometheus = "quay.io/prometheus/prometheus:v2.29.1"
```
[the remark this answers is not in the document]
dsupure: Good point!

drivers/storage/portworx/components_test.go
dsupure: Can we rename it to `prometheusNewOperatorClusterRole-v0.50.0.yaml`.
I feel we need to discuss a version management system in the future for the operator. I see many components' version scattering everywhere in the codes.

drivers/storage/portworx/testspec/prometheusNewOperatorClusterRole.yaml
```diff
@@ -0,0 +1,66 @@
+# This OperatorClusterRole is for Prometheus operator 0.50.0
```
dsupure, line 1: Suggest to change the file name to prometheusNewOperatorClusterRole-v0.50.0.yaml

Makefile
pure-jliao: it changes often I think, but as long as the crds are for operator 0.50.0 we should be good.

pkg/util/k8s/k8s_test.go
```diff
@@ -2574,3 +2577,46 @@ func TestWarningEvent(t *testing.T) {
...
+		},
+		Spec: apiextensionsv1.CustomResourceDefinitionSpec{
+			Group: "A",
+		},
+	}
+
+	go func() {
+		err := testutil.ActivateCRDWhenCreated(fakeExtClient, expectedCRD.Name)
+		require.NoError(t, err)
+	}()
```
piyush-nimbalkar, line 2597: nit: I don't think we need this. This is required for other crd registration test because it waits to validate that the CRD has been successfully registered. In our CreateCRD function we are just calling the API and not validating.

drivers/storage/portworx/component/prometheus.go
```diff
@@ -173,6 +183,27 @@ func (c *prometheus) MarkDeleted() {
...
+	files, err := filepath.Glob(filename)
+	if err != nil {
+		return err
+	}
+	for _, file := range files {
+		crd := &apiextensionsv1.CustomResourceDefinition{}
+		if err := k8sutil.ParseObjectFromFile(file, c.scheme, crd); err != nil {
+			return err
+		}
+		if err := k8sutil.CreateCRD(crd); err != nil {
```
piyush-nimbalkar, line 200: nit: Ideally we should optimize this further by first checking if the CRD is already present and then reading+parsing+creating the CRD. Will save some compute and i/o cycles.
You could address this now or just add a TODO and pick it up later.

drivers/storage/portworx/components_test.go
```diff
@@ -5830,6 +5831,130 @@ func TestCSI_0_3_NodeAffinityChange(t *testing.T) {
...
+		"prometheusrules.monitoring.coreos.com",
+		"servicemonitors.monitoring.coreos.com",
+		"thanosrulers.monitoring.coreos.com",
+	}
+	go func() {
+		for _, crdName := range crdNameList {
+			err := testutil.ActivateCRDWhenCreated(fakeExtClient, crdName)
+			require.NoError(t, err)
+		}
+	}()
```
piyush-nimbalkar, line 5943: nit: I guess we might not need this. Just try and decide whether to remove this or not.

drivers/storage/portworx/portworx_test.go
```diff
@@ -7037,6 +7042,12 @@ func (m *fakeManifest) GetVersions(
 			AlertManager:              "quay.io/prometheus/alertmanager:v1.2.3",
 		},
 	}
+	if m.k8sVersion != nil && m.k8sVersion.GreaterThanOrEqual(k8sutil.K8sVer1_22) {
+		version.Components.PrometheusOperator = "quay.io/prometheus-operator/prometheus-operator:v1.2.3"
```
piyush-nimbalkar, line 7046: Why not overwrite the other prometheus images too? That way we will have tested that other prometheus images also changed with the k8s version.

drivers/storage/portworx/components_test.go
```diff
@@ -5830,6 +5831,130 @@ func TestCSI_0_3_NodeAffinityChange(t *testing.T) {
...
+	expectedCR := testutil.GetExpectedClusterRole(t, "prometheusOperatorClusterRole.yaml")
+	actualCR := &rbacv1.ClusterRole{}
+	err = testutil.Get(k8sClient, actualCR, component.PrometheusOperatorClusterRoleName, "")
+	require.NoError(t, err)
+	require.Equal(t, expectedCR.Name, actualCR.Name)
+	require.Empty(t, actualCR.OwnerReferences)
+	require.ElementsMatch(t, expectedCR.Rules, actualCR.Rules)
+
+	// Prometheus operator Deployment
+	expectedDeployment := testutil.GetExpectedDeployment(t, "prometheusOperatorDeployment_v0.50.0.yaml")
```
piyush-nimbalkar, line 5915: Can we check the new prometheus instance and alert manager instance objects as well? The images would have changed based on the k8s version.

piyush-nimbalkar (approved):
LGTM
Minor comments and some more test suggestions.

drivers/storage/portworx/testspec/prometheusRule.yaml
```diff
@@ -148,15 +148,15 @@ spec:
       annotations:
         description: Pool {{$labels.POOL}} from node {{$labels.instance}}, from Portworx cluster {{$labels.clusterid}} successfully expanded.
         summary: Portworx pool {{$labels.POOL}} successfully expanded.
-      expr: px_alerts_poolexpandsuccessful == 1
+      expr: px_alerts_poolexpandsuccessful > 1
```
Pure-AdamuKaapan, line 151: Just for learning/sanity check's sake: my initial reaction would be "shouldn't this be >= 1 instead"? I know the error I got said ">", not ">=" so this is correct to fix the tests, I'm just curious if this is correct (or if the point is that it's a test file and if it makes sense doesn't matter)

pure-jliao: from the new config file in `bin/config` it's `>1` I believe that's the source of truth.

drivers/storage/portworx/component/prometheus.go
pure-jliao: but we get the crd list from files, only after that we fetch the crd names, then check if it exists. so need to parse first anyways.

piyush-nimbalkar: Ah, good point! May be we can do some optimization in the future with the file name being the crd name. Anyway, it is okay for now.

PR #414, opened 2021-10-12
OPERATOR-467 Continue upgrade if all storage nodes are up

piyush-nimbalkar:
- Don't need to check for pods on k8s nodes where the
  storage node is not supposed to run. Only check if the
  SDK is showing the node is up to continue the upgrade.

Closes # https://portworx.atlassian.net/browse/OPERATOR-467


pure-jliao (approved):
LGTM!

PR #428, opened 2021-11-02
Add metrics collector support

haibinxie:
**What this PR does / why we need it**: https://portworx.atlassian.net/browse/OPERATOR-403

Closes #


test/integration_test/basic_test.go
```diff
@@ -16,9 +18,58 @@ import (
 )
 
 func TestStorageClusterBasic(t *testing.T) {
+	testutil.TestSpecPath = ""
```
pure-jliao, line 21: if this is test specific, is it better to put in the test func? or in the common test file

test/integration_test/common_test.go
```diff
@@ -159,6 +159,21 @@ func setup() error {
 	return nil
 }
 
+func constructDefaultStorageCluster() (*corev1.StorageCluster, error) {
```
pure-jliao, line 162: this might be a little confusing, it's more like an empty stc, we populate the default values upon stc creation, see `createStorageCluster`

pure-jliao (commented):
overall looks good to me, some comments regarding unit tests and nits

drivers/storage/portworx/component/telemetry.go
```diff
@@ -59,16 +83,360 @@ func (t *telemetry) Delete(cluster *corev1.StorageCluster) error {
...
+func (t *telemetry) createProxyConfigMap(
+	cluster *corev1.StorageCluster,
+	ownerRef *metav1.OwnerReference,
+) error {
+	arcusLocation, present := cluster.Annotations[pxutil.AnnotationTelemetryArcusLocation]
+	if !present || len(arcusLocation) == 0 {
+		arcusLocation = "rest.cloud-support.purestorage.com"
+	}
+
+	config := fmt.Sprintf(
```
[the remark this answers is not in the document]
haibinxie: discussed with piyush before coding, and we decided to hard code it to be more efficient.

drivers/storage/portworx/manifest/manifest.go
```diff
@@ -31,6 +31,11 @@ const (
 	defaultNodeWiperImage  = "portworx/px-node-wiper:2.5.0"
 	defaultTelemetryImage  = "purestorage/ccm-service:3.0.3"
 
+	// DefaultEnvoyImage is the default Envoy image for metrics collector pod.
+	DefaultEnvoyImage = "envoyproxy/envoy:v1.19.1"
+	// DefaultCollectorImage is the default metrics collector container image.
+	DefaultCollectorImage = "pureezhang/realtime-metrics:latest"
```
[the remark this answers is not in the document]
haibinxie: this will get overwritten via versions in manifect, which is not included in this PR. we have not decided the final image name yet, will send another PR once it is determined.

test/integration_test/basic_test.go
haibinxie: this is in testutil, is that what you mean by common test file?

test/integration_test/common_test.go
haibinxie: createStorageCluster as it is in controller_test.go which is not in the test util to be reused by other tests. slightly different, the function here is to take the manifect URL to generate the storage cluster as it contains the images.

test/integration_test/basic_test.go
```diff
@@ -3,6 +3,8 @@
 package integrationtest
 
 import (
+	"github.com/libopenstorage/operator/drivers/storage/portworx/component"
```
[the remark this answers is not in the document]
haibinxie: what do you mean?

drivers/storage/portworx/component/telemetry.go
```diff
@@ -59,16 +83,360 @@ func (t *telemetry) Delete(cluster *corev1.StorageCluster) error {
...
+	}
+
+	return nil
+}
+
+func (t *telemetry) createCollectorDeployment(
+	cluster *corev1.StorageCluster,
+	ownerRef *metav1.OwnerReference,
+) error {
+	if t.isCollectorDeploymentCreated {
```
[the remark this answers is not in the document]
haibinxie: Good point, I will update it.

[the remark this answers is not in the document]
piyush-nimbalkar: We will have to update the deployment for custom registry, affinity, toleration, etc changes. Look at the deployment creation in other components.

drivers/storage/portworx/component/telemetry.go
```diff
@@ -59,16 +83,360 @@ func (t *telemetry) Delete(cluster *corev1.StorageCluster) error {
...
+		Spec: appsv1.DeploymentSpec{
+			Replicas: &replicas,
+			Selector: &metav1.LabelSelector{MatchLabels: labels},
+			Template: v1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{Labels: labels},
+				Spec: v1.PodSpec{
+					Containers: []v1.Container{
+						{
+							Name:  "collector",
+							Image: manifest.DefaultCollectorImage,
```
piyush-nimbalkar, line 154: Using images from the manifest is fine for this PR, but we need to change this to use `cluster.Status.DesiredImages` to populate the component images. Check other components for that. We can do that as part of [OPERATOR-491](https://portworx.atlassian.net/browse/OPERATOR-491)

drivers/storage/portworx/components_test.go
```diff
@@ -11120,6 +11120,41 @@ func TestTelemetryEnable(t *testing.T) {
 	require.Len(t, telemetryConfigMap.OwnerReferences, 1)
 	require.Equal(t, cluster.Name, telemetryConfigMap.OwnerReferences[0].Name)
 	require.Equal(t, expectedConfigMap.Data, telemetryConfigMap.Data)
+
+	// Proxy config map
+	configMap := &v1.ConfigMap{}
+	err = testutil.Get(k8sClient, configMap, component.CollectorProxyConfigMapName, cluster.Namespace)
+	require.NoError(t, err)
+	require.Len(t, configMap.OwnerReferences, 1)
+	require.Equal(t, cluster.Name, configMap.OwnerReferences[0].Name)
```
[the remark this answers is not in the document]
piyush-nimbalkar: +1

drivers/storage/portworx/components_test.go
```diff
@@ -11120,6 +11120,41 @@ func TestTelemetryEnable(t *testing.T) {
...
+	require.NoError(t, err)
+	require.Len(t, roleBinding.OwnerReferences, 1)
+	require.Equal(t, cluster.Name, roleBinding.OwnerReferences[0].Name)
+
+	// Collector deployment
+	deployment := &appsv1.Deployment{}
+	err = testutil.Get(k8sClient, deployment, component.CollectorDeploymentName, cluster.Namespace)
+	require.NoError(t, err)
+	require.Len(t, deployment.OwnerReferences, 1)
+	require.Equal(t, cluster.Name, deployment.OwnerReferences[0].Name)
```
piyush-nimbalkar, line 11157: Can we add unit tests for disabling metrics collector?
Like we did for repo change, can we add more tests for metrics collector too. For reference - https://github.com/libopenstorage/operator/pull/358#discussion_r683685356

test/integration_test/basic_test.go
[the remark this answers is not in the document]
piyush-nimbalkar: Generally most golang imports are divided in 2 or 3 groups -
- golang internal packages  (eg. strings, log, testing, etc)
- external packages
- internal packages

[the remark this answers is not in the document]
piyush-nimbalkar: Your IDE should have such configuration to automatically sort them on save.

test/integration_test/specs/pure-telemetry-cert.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
```
piyush-nimbalkar, line 1: What is this certificate? It it okay to expose it like this in the code?

test/integration_test/basic_test.go
```diff
@@ -16,9 +18,58 @@ import (
...
+	_, err := coreops.Instance().GetSecret(component.TelemetryCertName, pxNamespace)
+	require.True(t, err == nil || errors.IsNotFound(err))
+
+	if errors.IsNotFound(err) {
+		secret, err = coreops.Instance().CreateSecret(secret)
+		require.NoError(t, err)
+	} else {
+		secret, err = coreops.Instance().UpdateSecret(secret)
+		require.NoError(t, err)
+	}
```
piyush-nimbalkar, line 41: Why bother with a get and then create or update. Let's just create on start and delete on cleanup. The test should just assume the secret is not present when it starts.

test/integration_test/basic_test.go
```diff
@@ -16,9 +18,58 @@ import (
...
+	cluster, err = createStorageCluster(cluster)
+	require.NoError(t, err)
+
+	// Validate metrics collector pod becomes ready
+	err = testutil.ValidatePods(
+		cluster.Namespace,
+		map[string]string{"role": "realtime-metrics-collector"},
+		defaultValidateDeployTimeout,
+		defaultValidateDeployRetryInterval)
+	require.NoError(t, err)
```
piyush-nimbalkar, line 62: We probably don't need this. Should we call `testutil.ValidateStorageCluster()` instead to validate the entire cluster. This function validates all the components as well. We will have to add metrics collector to the validation.

test/integration_test/common_test.go
piyush-nimbalkar: I think he means this one - https://github.com/libopenstorage/operator/blob/master/test/integration_test/common_test.go#L190

test/integration_test/specs/pure-telemetry-cert.yaml
haibinxie: this is just for testing.

test/integration_test/basic_test.go
haibinxie: this is more robust, the create may fail if test exists for any reason and we rerun. I will add deletion in cleanup anyway.

test/integration_test/basic_test.go
haibinxie: ValidateStorageCluster does not verify every component, it only verifies if storagecluster is online. let's let each component verify itself

haibinxie:
@piyush-nimbalkar Somehow this comment does not have a reply box, but the function is different, the function that I added is to construct in memory object, the function you pointed is to create the object to kubenetes (kubectl create)

I think he means this one - https://github.com/libopenstorage/operator/blob/master/test/integration_test/common_test.go#L190

drivers/storage/portworx/component/telemetry.go
```diff
@@ -59,16 +83,360 @@ func (t *telemetry) Delete(cluster *corev1.StorageCluster) error {
...
+	}
+
+	return k8sutil.CreateOrUpdateConfigMap(
+		t.k8sClient,
+		&v1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:            CollectorProxyConfigMapName,
+				Namespace:       cluster.Namespace,
+				OwnerReferences: []metav1.OwnerReference{*ownerRef},
+				Annotations:     map[string]string{constants.AnnotationReconcileObject: "true"},
```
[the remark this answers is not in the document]
haibinxie: Actually during test this fails so i added it, the default value is false.

drivers/storage/portworx/component/telemetry.go
```diff
@@ -59,16 +83,360 @@ func (t *telemetry) Delete(cluster *corev1.StorageCluster) error {
...
+		`
+scrapeConfig:
+  interval: 10
+  k8sConfig:
+    pods:
+    - podSelector:
+        name: portworx
+      namespace: %s
+      endpoint: metrics
+      port: 9001
```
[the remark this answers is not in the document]
haibinxie: I am not sure if i understand, i will sync with you separately.

drivers/storage/portworx/component/telemetry.go
```diff
@@ -59,16 +83,360 @@ func (t *telemetry) Delete(cluster *corev1.StorageCluster) error {
...
+	ownerRef *metav1.OwnerReference,
+) error {
+	config := fmt.Sprintf(
+		`
+scrapeConfig:
+  interval: 10
+  k8sConfig:
+    pods:
+    - podSelector:
+        name: portworx
```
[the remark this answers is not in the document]
haibinxie: will fix

drivers/storage/portworx/component/telemetry.go
```diff
@@ -59,16 +83,360 @@ func (t *telemetry) Delete(cluster *corev1.StorageCluster) error {
...
+      name: envoy.transport_sockets.tls
+      typed_config:
+        "@type": type.googleapis.com/envoy.extensions.transport_sockets.tls.v3.UpstreamTlsContext
+        common_tls_context:
+          tls_certificates:
+          - certificate_chain:
+              filename: /appliance-cert/cert
+            private_key:
+              filename: /appliance-cert/private_key
+`, cluster.UID, pxutil.GetPortworxVersion(cluster), arcusLocation, arcusLocation)
```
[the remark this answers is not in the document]
haibinxie: when do we populate the Status, i am afraid we may run into race before the status is set.

drivers/storage/portworx/component/telemetry.go
```diff
@@ -59,16 +83,360 @@ func (t *telemetry) Delete(cluster *corev1.StorageCluster) error {
...
+	)
+}
+
+func (t *telemetry) createProxyConfigMap(
+	cluster *corev1.StorageCluster,
+	ownerRef *metav1.OwnerReference,
+) error {
+	arcusLocation, present := cluster.Annotations[pxutil.AnnotationTelemetryArcusLocation]
+	if !present || len(arcusLocation) == 0 {
+		arcusLocation = "rest.cloud-support.purestorage.com"
```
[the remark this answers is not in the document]
haibinxie: will fix

drivers/storage/portworx/component/telemetry.go
```diff
@@ -23,10 +26,27 @@ const (
 	TelemetryPropertiesFilename = "ccm.properties"
 	// TelemetryArcusLocationFilename is name of the file storing the location of arcus endpoint to use
 	TelemetryArcusLocationFilename = "location"
+	// TelemetryCertName is name of the telemetry cert.
```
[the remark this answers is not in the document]
haibinxie: I thought about this, but feel it is part of telemetry, hence I have separated the API to deployMetricsCollector

test/integration_test/common_test.go
haibinxie: that's different API, that API creates stc to kuberntes, while this construct*** API just create in memory object.

test/integration_test/basic_test.go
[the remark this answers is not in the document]
haibinxie: cool, looks like gofmt only sorted it

drivers/storage/portworx/component/telemetry.go
```diff
@@ -59,16 +83,360 @@ func (t *telemetry) Delete(cluster *corev1.StorageCluster) error {
...
+		&rbacv1.RoleBinding{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:            CollectorRoleBindingName,
+				Namespace:       cluster.Namespace,
+				OwnerReferences: []metav1.OwnerReference{*ownerRef},
+			},
+			Subjects: []rbacv1.Subject{
+				{
+					Kind:      "ServiceAccount",
+					Name:      "default",
```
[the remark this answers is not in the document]
haibinxie: sounds good

drivers/storage/portworx/components_test.go
haibinxie: That piece of change is a little hard to maintain, i am thinking about using deepEqual to diff the two deployments spec and decide whether to update it. i just posted it in another commit, what do you think about this approach, in this case we don't need to worry what element is changed, any change in the spec would trigger update.

drivers/storage/portworx/component/telemetry.go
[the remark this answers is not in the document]
piyush-nimbalkar: Technically it is not part of telemetry from our side, although it could be seen as one from customer standpoint. 
From our side telemetry/ccm-service is meant to only collect diags from portworx while metrics collector is only sending back metrics to Pure1.
Anyway, it is probably a minor thing right now as these are not big components. I will leave it up to you.

test/integration_test/basic_test.go
pure-jliao: I mean should we put it in `testInstallMetricsCollector` instead of `TestStorageClusterBasic` as it's only used in that subtest only for now.

drivers/storage/portworx/component/telemetry.go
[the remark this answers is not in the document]
piyush-nimbalkar: Your are right. What we can do it not create this component until cluster.Status.ClusterUID is not set. Any way, there is no point of starting the metrics collector until portworx is up.

drivers/storage/portworx/component/telemetry.go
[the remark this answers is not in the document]
piyush-nimbalkar: I am curious, the test should not fail because of that. There are config maps getting created by other components too. We can check it offline if you want.

test/integration_test/common_test.go
pure-jliao: I mean the function name is confusing as the stc created only has basic images in it but not having other default specs populated. we can have a shared utility function like this but better to merge this function with `constructStorageCluster` directly, they serve the same purpose.

drivers/storage/portworx/components_test.go
piyush-nimbalkar: DeepEqual would be ideal, but unfortunately it does not work. When you save a deployment object in k8s it adds a lot of defaults, which causes the DeepEqual to fail. That's why we resorted to this solution. Even I don't like it but comparing complicated objects like deployment, daemonset is quite tricky. You can try it on a cluster and check.

Anyway, the above comment was to add more tests for the deployment changes and checking disable.

test/integration_test/specs/pure-telemetry-cert.yaml
piyush-nimbalkar: Okay, so this is random value and not our pure1 cert?

test/integration_test/basic_test.go
piyush-nimbalkar: ValidateStorageCluster check all the components. It not only checks for the deployment, but also check if they have the right images. You should add the metrics collector component to that function. 
After that if you want, you can call just the ValidateMetricsCollector part out of it in this test.

drivers/storage/portworx/component/telemetry.go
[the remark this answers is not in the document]
haibinxie: @piyush-nimbalkar that port is pxMgmtPort, which is configurable through environment variable, in this case I think I would change it to config.PxMgmtPort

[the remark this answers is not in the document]
piyush-nimbalkar: Use this - https://github.com/libopenstorage/operator/blob/master/drivers/storage/portworx/util/util.go#L324

test/integration_test/basic_test.go
haibinxie: I intended to put here for all functional tests, unit test has a different path as default and without this it would not work for any functional test.

test/integration_test/common_test.go
haibinxie: i will rename it

haibinxie: actually i would set portworx defaults and keep the name in this API so it can be reused

drivers/storage/portworx/components_test.go
haibinxie: I think Kubernetes only adds to other part of the object (such as status) but does not touch the spec, let me test it and see. 

if we do diff of the entire spec then we dont have to worry about each field of change, which is essentially equal change. I would add disabling test first.

drivers/storage/portworx/component/telemetry.go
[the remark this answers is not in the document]
haibinxie: I think you are right, it was my misunderstanding during test.

drivers/storage/portworx/components_test.go
haibinxie: @piyush-nimbalkar this is nice discussion, thanks for the tips. I finally found there is existing API in k8s that can do deep semantics deep equal, which is exactly what we need. 

in 	"k8s.io/apimachinery/pkg/api/equality"

// DeepDerivative is similar to DeepEqual except that unset fields in a1 are
// ignored (not compared). This allows us to focus on the fields that matter to
// the semantic comparison.
//
// The unset fields include a nil pointer and an empty string.
func (e Equalities) DeepDerivative(a1, a2 interface{}) bool {

drivers/storage/portworx/component/telemetry.go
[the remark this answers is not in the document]
haibinxie: I did some test, if we wait for the UID and px does not come online, we will not have metrics for other components too. although we may not have autopilot metrics today but will have that eventually and more components will add metrics, the behavior is not good. do we have to use cluster UID, since it is just an ID maybe just use cluster name which is even more readable than a UUID, what do you think?

drivers/storage/portworx/component/telemetry.go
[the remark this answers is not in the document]
haibinxie: looked at the examples, we need a utility function to change those fields in the deployment according to storageCluster configuration, this function can be reused by all components in the future.

test/integration_test/basic_test.go
haibinxie: @pure-jliao actually i will rename our current folder specs to testspec so that we keep the same name with unit test and dont have to change the value in code.

haibinxie:
@piyush-nimbalkar @pure-jliao I resolved all of the comments except verifying configmaps from files, will add later but having power outage today, please take a look at current changes.

drivers/storage/portworx/component/telemetry.go
```diff
@@ -59,16 +94,496 @@ func (t *telemetry) Delete(cluster *corev1.StorageCluster) error {
...
+	}
+
+	modified := !util.DeploymentDeepEqual(deployment, existingDeployment)
+
+	if !t.isCollectorDeploymentCreated || modified {
+		logrus.WithFields(logrus.Fields{
+			"isCreated": t.isCollectorDeploymentCreated,
+			"modified":  modified,
+		}).Info("will create/update the deployment.")
+		if err = k8sutil.CreateOrUpdateDeployment(t.k8sClient, deployment, ownerRef); err != nil {
```
pure-jliao, line 203: I think `k8sutil.CreateOrUpdateDeployment` already does what you have here so it's duplicated, we can simply call `k8sutil.CreateOrUpdateDeployment`

pkg/util/util.go
```diff
@@ -7,9 +7,13 @@ import (
...
 
 	"github.com/hashicorp/go-version"
+	v1 "k8s.io/api/core/v1"
+
 	corev1 "github.com/libopenstorage/operator/pkg/apis/core/v1"
 	"github.com/libopenstorage/operator/pkg/constants"
-	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
+
+	appsv1 "k8s.io/api/apps/v1"
```
pure-jliao, line 16: nit: group

test/integration_test/basic_test.go
```diff
@@ -3,22 +3,79 @@
...
 	"github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/require"
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"strings"
+	"testing"
+
+	"github.com/libopenstorage/operator/drivers/storage/portworx/component"
+
+	coreops "github.com/portworx/sched-ops/k8s/core"
```
pure-jliao, line 15: nit: group

pure-jliao (approved):
LGTM, I merged the automation test change, so you may need to update the functional test again, we could see whether it will work..

test/integration_test/specs/pure-telemetry-cert.yaml
haibinxie: correct

drivers/storage/portworx/component/telemetry.go
haibinxie: This is to save one update when it is not necessary, personally I don't prefer the optimization as one update in every reconcile loop is trivial. But this is to keep consistent with implementation of all existing components.

pkg/util/util.go
```diff
@@ -149,6 +153,15 @@ func HaveTolerationsChanged(
 	return !reflect.DeepEqual(cluster.Spec.Placement.Tolerations, existingTolerations)
 }
 
+func DeploymentDeepEqual(d1 *appsv1.Deployment, d2 *appsv1.Deployment) bool {
+	return equality.Semantic.DeepDerivative(d1.Spec.Template.Spec.Containers, d2.Spec.Template.Spec.Containers) &&
```
piyush-nimbalkar, line 157: Doesn't this work for the entire deployment spec?

test/integration_test/basic_test.go
```diff
@@ -208,3 +221,70 @@ func BasicUpgrade(tc *TestCase) func(*testing.T) {
...
+		testSpec := tc.TestSpec(t)
+		cluster, ok := testSpec.(*op_corev1.StorageCluster)
+		require.True(t, ok)
+
+		testInstallWithTelemetry(t, cluster)
+	}
+}
+
+func testInstallWithTelemetry(t *testing.T, cluster *op_corev1.StorageCluster) {
+	secret := testutil.GetExpectedSecret(t, "/specs/pure-telemetry-cert.yaml")
```
piyush-nimbalkar, line 240: This should be `/testspec`, right?

drivers/storage/portworx/components_test.go
```diff
@@ -9297,6 +9550,11 @@ func TestCompleteInstallWithNodeAffinityChange(t *testing.T) {
 	err = testutil.Get(k8sClient, pxRepoDeployment, component.PxRepoDeploymentName, cluster.Namespace)
 	require.NoError(t, err)
 	require.Nil(t, pxRepoDeployment.Spec.Template.Spec.Affinity)
+
+	collectorDeployment = &appsv1.Deployment{}
+	err = testutil.Get(k8sClient, collectorDeployment, component.CollectorDeploymentName, cluster.Namespace)
+	require.NoError(t, err)
+	require.Nil(t, collectorDeployment.Spec.Template.Spec.Affinity)
```
piyush-nimbalkar, line 9557: There is test for toleration changes too. Can we modify that too?

drivers/storage/portworx/component/telemetry.go
```diff
@@ -59,16 +94,496 @@ func (t *telemetry) Delete(cluster *corev1.StorageCluster) error {
...
+		return err
+	}
+
+	modified := !util.DeploymentDeepEqual(deployment, existingDeployment)
+
+	if !t.isCollectorDeploymentCreated || modified {
+		logrus.WithFields(logrus.Fields{
+			"isCreated": t.isCollectorDeploymentCreated,
+			"modified":  modified,
+		}).Info("will create/update the deployment.")
```
piyush-nimbalkar, line 202: nit: Let's keep this as trace log. It will create too much unnecessary log. Unfortunately our default log is debug right now, so can't keep this as debug.

drivers/storage/portworx/component/telemetry.go
```diff
@@ -59,16 +94,496 @@ func (t *telemetry) Delete(cluster *corev1.StorageCluster) error {
...
+		if err = k8sutil.CreateOrUpdateDeployment(t.k8sClient, deployment, ownerRef); err != nil {
+			return err
+		}
+	}
+
+	t.isCollectorDeploymentCreated = true
+	logrus.Debug("no change to collector deployment")
```
piyush-nimbalkar, line 209: This is not true.. even when the collector deployment changes, it will come to this line. Let's remove it if was used only for development.

drivers/storage/portworx/component/telemetry.go
```diff
@@ -23,10 +31,33 @@ const (
...
+	// CollectorConfigMapName is name of config map for metrics collector.
+	CollectorConfigMapName = "px-collector-config"
+	// CollectorProxyConfigMapName is name of the config map for envoy proxy.
+	CollectorProxyConfigMapName = "px-collector-proxy-config"
+	// CollectorDeploymentName is name of metrics collector deployment.
+	CollectorDeploymentName = "px-metrics-collector"
+	// DefaultArcusLocation is the default arcus location.
+	DefaultArcusLocation = "rest.cloud-support.purestorage.com"
+	// CollectorServiceAccountName is name of the metrics collector service account
+	CollectorServiceAccountName = "metrics-collector"
```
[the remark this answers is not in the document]
haibinxie: sounds good, will change

drivers/storage/portworx/component/telemetry.go
```diff
@@ -59,16 +94,496 @@ func (t *telemetry) Delete(cluster *corev1.StorageCluster) error {
...
+      name: envoy.transport_sockets.tls
+      typed_config:
+        "@type": type.googleapis.com/envoy.extensions.transport_sockets.tls.v3.UpstreamTlsContext
+        common_tls_context:
+          tls_certificates:
+          - certificate_chain:
+              filename: /appliance-cert/cert
+            private_key:
+              filename: /appliance-cert/private_key
+`, cluster.Status.ClusterUID, pxutil.GetPortworxVersion(cluster), arcusLocation, arcusLocation)
```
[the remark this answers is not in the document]
haibinxie: this is handled before, if clusterUid is empty it already returned err and wont get here.

test/integration_test/basic_test.go
haibinxie: right, have not retested functional test after some changes, but will get to that.

drivers/storage/portworx/component/telemetry.go
haibinxie: This wont print too many, as it is only when it is created or modified. not in every reconcile loop

pkg/util/util.go
haibinxie: Unfortunately no, if a field is empty in the first argument but not empty in the second argument it will return true.

test/integration_test/basic_test.go
```diff
@@ -86,6 +88,17 @@ var testStorageClusterBasicCases = []TestCase{
 		},
 		TestFunc: BasicUpgrade,
 	},
+	{
+		TestName:        "InstallWithTelemetry",
+		TestrailCaseIDs: []string{},
+		TestSpec: func(t *testing.T) interface{} {
+			cluster := &op_corev1.StorageCluster{}
+			cluster.Name = "telemetry-test"
```
pure-jliao, line 96: you can construct the cluster spec here before passing it to the test function
```
			err := constructStorageCluster(cluster, pxSpecGenURL, pxSpecImages)
			require.NoError(t, err)
```

test/integration_test/basic_test.go
```diff
@@ -208,3 +221,70 @@ func BasicUpgrade(tc *TestCase) func(*testing.T) {
...
+		secret, err = coreops.Instance().UpdateSecret(secret)
+		require.NoError(t, err)
+	}
+
+	// Deploy portworx with telemetry set to true
+	imageListMap, err := testutil.GetImagesFromVersionURL(pxSpecGenURL)
+	require.NoError(t, err)
+
+	err = constructStorageCluster(cluster, pxSpecGenURL, imageListMap)
+	require.NoError(t, err)
```
pure-jliao, line 259: those are no longer needed, I initialize `pxSpecImages ` as a global variable during test setup, so we just need to call  `constructStorageCluster` in test spec

test/integration_test/basic_test.go
```diff
@@ -208,3 +221,70 @@ func BasicUpgrade(tc *TestCase) func(*testing.T) {
...
+
+	cluster.Name = "test-sc"
+	cluster.Spec.Monitoring = &op_corev1.MonitoringSpec{
+		Telemetry: &op_corev1.TelemetrySpec{
+			Enabled: true,
+		},
+	}
```
pure-jliao, line 266: and we need to move this to test spec as well

test/integration_test/basic_test.go
```diff
@@ -208,3 +221,70 @@ func BasicUpgrade(tc *TestCase) func(*testing.T) {
...
+
+	cluster, err = createStorageCluster(cluster)
+	require.NoError(t, err)
+
+	err = testutil.ValidateTelemetry(
```
pure-jliao, line 271: we could move this validation to `BasicInstall`, check if telemetry is enabled then validate it, then you should be able to just call Basic install test func and remove this whole function `testInstallWithTelemetry`

pure-jliao (commented):
added a few comments for the new test automation format.

test/integration_test/basic_test.go
```diff
@@ -208,3 +221,70 @@ func BasicUpgrade(tc *TestCase) func(*testing.T) {
...
+func InstallWithTelemetry(tc *TestCase) func(*testing.T) {
+	return func(t *testing.T) {
+		if tc.ShouldSkip() {
+			t.Skip()
+		}
+
+		testSpec := tc.TestSpec(t)
+		cluster, ok := testSpec.(*op_corev1.StorageCluster)
+		require.True(t, ok)
+
```
pure-jliao, line 234: you could parse and create the secret here, then just call `BasicInstall(tc)(t)`

haibinxie: Let's still keep it separate so we can test separately after making changes.

haibinxie:
@piyush-nimbalkar resolved the comments, please take another look

drivers/storage/portworx/component/telemetry.go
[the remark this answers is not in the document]
piyush-nimbalkar: I see the check is before creating the configmap. Couple of points here -
- During first boot, won't this cause the metrics collector deployment to fail for a while as the config map is not present?
- We shouldn't return error if the cluster UID is not present, just log it. If the cluster UID is never getting populated then it is a much bigger issue of px not coming up.

piyush-nimbalkar:
Everything looks good. Just one outstanding comment about the cluster UID check.

test/integration_test/common_test.go
```diff
@@ -44,15 +44,15 @@ const (
 	// defaultValidateDeployTimeout is a default timeout for deployment validation
 	defaultValidateDeployTimeout = 900 * time.Second
 	// defaultValidateDeployRetryInterval is a default retry interval for deployment validation
-	defaultValidateDeployRetryInterval = 30 * time.Second
+	defaultValidateDeployRetryInterval = 5 * time.Second
```
piyush-nimbalkar, line 47: Was this for internal testing or you generally wanted to make the polling faster?

piyush-nimbalkar (approved):
LGTM

PR #430, opened 2021-11-06
[OPERATOR-487] Test framework refactor and upgrade test requirements

pure-jliao:

Closes # OPERATOR-487

**Special notes for your reviewer**: will refactor fa fb test cases as well once everyone agrees on this approach.

### DIFF
--- a/test/integration_test/basic_test.go
+++ b/test/integration_test/basic_test.go
@@ -3,189 +3,208 @@
 package integrationtest
 
 import (
-	"strings"
 	"testing"
 
+	"github.com/hashicorp/go-version"
 	"github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/require"
 	v1 "k8s.io/api/core/v1"
 
 	op_corev1 "github.com/libopenstorage/operator/pkg/apis/core/v1"
+	k8sutil "github.com/libopenstorage/operator/pkg/util/k8s"
 	testutil "github.com/libopenstorage/operator/pkg/util/test"
 	coreops "github.com/portworx/sched-ops/k8s/core"
 )
 
-func TestStorageClusterBasic(t *testing.T) {
-	t.Run("InstallWithAllDefaults", testInstallWithAllDefaults)
-	t.Run("NodeAffinityLabels", testNodeAffinityLabels)
-	t.Run("Upgrade", testUpgrade)
-}
+const (
+	labelKeySkipPX = "skip/px"
+	labelValueTrue = "true"
+)
 
-func testInstallWithAllDefaults(t *testing.T) {
-	// Get versions from URL
-	logrus.Infof("Get component images from versions URL")
-	imageListMap, err := testutil.GetImagesFromVersionURL(pxSpecGenURL)
-	require.NoError(t, err)
+var (
+	pxVer2_9, _ = version.NewVersion("2.9")
+)
 
-	// Construct Portworx StorageCluster object
-	cluster, err := constructStorageCluster(pxSpecGenURL, imageListMap)
-	require.NoError(t, err)
-
-	cluster.Name = "simple-install"
-
-	// Deploy cluster
-	logrus.Infof("Create StorageCluster %s in %s", cluster.Name, cluster.Namespace)
-	cluster, err = createStorageCluster(cluster)
-	require.NoError(t, err)
-
-	// Validate cluster deployment
-	logrus.Infof("Validate StorageCluster %s", cluster.Name)
-	err = testutil.ValidateStorageCluster(imageListMap, cluster, defaultValidateDeployTimeout, defaultValidateDeployRetryInterval, true, "")
-	require.NoError(t, err)
-
-	// Delete cluster
-	logrus.Infof("Delete StorageCluster %s", cluster.Name)
-	err = testutil.UninstallStorageCluster(cluster)
-	require.NoError(t, err)
-
-	// Validate cluster deletion
-	logrus.Infof("Validate StorageCluster %s deletion", cluster.Name)
-	err = testutil.ValidateUninstallStorageCluster(cluster, defaultValidateUninstallTimeout, defaultValidateUninstallRetryInterval)
-	require.NoError(t, err)
-}
-
-func testNodeAffinityLabels(t *testing.T) {
-	var err error
-	labelKey := "skip/px"
-	labelValue := "true"
-
-	// Get versions from URL
-	logrus.Infof("Get component images from versions URL")
-	imageListMap, err := testutil.GetImagesFromVersionURL(pxSpecGenURL)
-	require.NoError(t, err)
-
-	// Construct Portworx StorageCluster object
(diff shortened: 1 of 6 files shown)


test/integration_test/basic_test.go
```diff
@@ -6,186 +6,206 @@ import (
...
-	// Remove Node Affinity label from the node
-	logrus.Infof("Remove label %s from node %s", nodeNameWithLabel, labelKey)
-	if err := coreops.Instance().RemoveLabelOnNode(nodeNameWithLabel, labelKey); err != nil {
+		// Get versions from URL
+		logrus.Infof("Get component images from versions URL")
+		imageListMap, err := testutil.GetImagesFromVersionURL(pxSpecGenURL)
+		require.NoError(t, err)
+
+		// Construct Portworx StorageCluster object
+		err = constructStorageClusterForTestCase(tc, pxSpecGenURL, imageListMap)
```
haibinxie, line 97: The storage cluster should be input, if we create it inside of the test, it's more like internal variable, there is no point to save it back to TC.  

As we discussed in last friday's meeting, I think storagecluster is not a common input for all test cases and it should not be in TestCase struct. For examples:
- for Daemonset to operator migration test cases, the input should be Daemonset and not StorageCluster.
- If we implement autopilot functional tests, not installation of autopilot via operator, autopilot project functional tests. We will have installation of autopilot covered by operator, and functionality of autopilot would be tested via a bunch of test cases that assumes autopilot is already installed, this struct would not fit.

I think a common test case consists of: input, output, setup, execution and teardown, which are common parameters.

test/integration_test/basic_test.go
```diff
@@ -6,186 +6,206 @@ import (
...
+}
 
-	// Set Node Affinity label one of the K8S nodes
-	var nodeNameWithLabel string
-	for _, node := range nodeList.Items {
-		if coreops.Instance().IsNodeMaster(node) {
-			continue // Skip master node, we don't need to label it
+func BasicInstall(tc *TestCase) func(*testing.T) {
+	return func(t *testing.T) {
+		if tc.ShouldSkip {
```
piyush-nimbalkar, line 87: Can ShouldSkip be a function returning bool instead of a bool value itself? That way every test case can have its own skip function or share a common one applicable to different environments.
Also, should we just skip before even calling this test func.. before or inside RunTest method?

test/integration_test/basic_test.go
piyush-nimbalkar: @haibinxie I thought we decided that we can keep the storagecluster in the case struct, but it should be optional. That way, we can use the same test func to test similar test cases by only changing the input storagecluster. 
The tests that don't need or don't want to have an input StorageCluster will keep it nil in their struct and create their own in the test func.

pure-jliao: >The storage cluster should be input, if we create it inside of the test, it's more like internal variable, there is no point to save it back to TC.

we have the storage cluster as test case input, the construct here only populate the env var and images. the constructed stc is still belongs to the current TC so there's actually no difference this is more like a personal preference, I planned to make the function a call belongs to TC object.
>As we discussed in last friday's meeting, I think storagecluster is not a common input for all test cases and it should not be in TestCase struct.

what about we make the field `InputSpec interface{}`? then we could assert the type in different test functions, that way it becomes more scalable for different scenarios?  @haibinxie 
also, the idea of setup a TestFunc field would be applicable for other steps like setup and teardown func as an input spec if we would like to do so, I feel it's feasible.

test/integration_test/basic_test.go
pure-jliao: we could make ShouldSkip a function when defining the test case, but seems there's some issue as mentioned in Upgrade function like the global param was not initialized. but think your suggestion would probably fix that issue so I'll try.
also the t.Skip function only takes args not test name, think about our previous test functions, the skip should be called inside, and the test func is called by t.Run(), so I don't know if this could be called outside of t.Run()

test/integration_test/basic_test.go
haibinxie: Yes, making it a InputSpec sounds good, and different test can have different input as desired.

haibinxie: When we look at the code, the input StorageCluster is empty with just a name, then it gets changed inside of the function, which loses the goal of making StorageCluster as a configurable input and looks confusing. 

`we have the storage cluster as test case input, the construct here only populate the env var and images. the constructed stc is still belongs to the current TC so there's actually no difference this is more like a personal preference, I planned to make the function a call belongs to TC object.

`

pure-jliao: got it, what about this, let's make the InputSpec a `func() interface` similar to ShouldSkip, TestFunc, this way it's more clear, could be defined in testcase list, also fully customizable, and it's consistent with other fields.

test/integration_test/basic_test.go
piyush-nimbalkar: We don't need to call t.Skip(), right? We can just not call the testfunc if the shouldSkip function returns true.

test/integration_test/basic_test.go
haibinxie: sounds good to me. @piyush-nimbalkar what do you think?

test/integration_test/basic_test.go
pure-jliao: yeah we could do that, only thing is the skipped tests will not printed in the console at the end of test run

piyush-nimbalkar: Ah, I see. Then we could call the t.skip inside the test as you have done now.

pure-jliao: yes we could, that way the code is simpler but we don't have a consistent golang test report, but eventually we need to have our own report to make it easier, e.g. print skipped/passed/failed tests and their testrails IDs. I'm fine with change it now or later when we reach there @piyush-nimbalkar what do you think?

piyush-nimbalkar: I am fine with either approach. Would have liked to keep the tests cleaner with only test logic, and have the TestCase struct and loop take care of other things.
If you think the reporting is better this way for now keep it inside the test function or if you are going to keep it outside, just add a log line saying we are skipping xx test.

test/integration_test/basic_test.go
piyush-nimbalkar: Sounds good, but honestly I don't see any other input spec coming up other than StorageCluster in the near term. This could be just a storagecluster spec or if you want to keep it function it could just return the StorageCluster spec. We can extend this in future when we have different kinds of inputs. Also, for migration it won't be just one spec but a bunch of specs. So, it would be better to create those in the test itself.

I agree with Michael, that we should probably not create an input spec if it is not passed. Let the individual test be responsible for that.

piyush-nimbalkar (approved):
This looks good to me organization wise. 
I feel we can get this merged and then work on changes incrementally. Wdyt @haibinxie?

haibinxie (approved):

PR #432, opened 2021-11-11
[OPERATOR-496] Refactor FA FB tests to follow the new format

pure-jliao:
**What this PR does / why we need it**: Refactor fa fb tests

Closes #

### DIFF
--- a/test/integration_test/basic_test.go
+++ b/test/integration_test/basic_test.go
@@ -10,6 +10,7 @@ import (
 	"github.com/stretchr/testify/require"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/libopenstorage/operator/drivers/storage/portworx/component"
 	op_corev1 "github.com/libopenstorage/operator/pkg/apis/core/v1"
@@ -31,23 +32,17 @@ var testStorageClusterBasicCases = []TestCase{
 	{
 		TestName:        "InstallWithAllDefaults",
 		TestrailCaseIDs: []string{},
-		TestSpec: func(t *testing.T) interface{} {
-			cluster := &op_corev1.StorageCluster{}
-			cluster.Name = "simple-install"
-			err := constructStorageCluster(cluster, pxSpecGenURL, pxSpecImages)
-			require.NoError(t, err)
-			return cluster
-		},
-		ShouldSkip: func() bool { return false },
-		TestFunc:   BasicInstall,
+		TestSpec: createStorageClusterTestSpecFunc(&op_corev1.StorageCluster{
+			ObjectMeta: meta.ObjectMeta{Name: "simple-install"},
+		}),
+		TestFunc: BasicInstall,
 	},
 	{
 		TestName:        "NodeAffinityLabels",
 		TestrailCaseIDs: []string{},
-		TestSpec: func(t *testing.T) interface{} {
-			cluster := &op_corev1.StorageCluster{}
-			cluster.Name = "node-affinity-labels"
-			cluster.Spec = op_corev1.StorageClusterSpec{
+		TestSpec: createStorageClusterTestSpecFunc(&op_corev1.StorageCluster{
+			ObjectMeta: meta.ObjectMeta{Name: "node-affinity-labels"},
+			Spec: op_corev1.StorageClusterSpec{
 				Placement: &op_corev1.PlacementSpec{
 					NodeAffinity: &v1.NodeAffinity{
 						RequiredDuringSchedulingIgnoredDuringExecution: &v1.NodeSelector{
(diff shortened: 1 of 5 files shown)


test/integration_test/common_test.go
```diff
@@ -122,10 +122,8 @@ func setup() error {
 	return nil
 }
 
-func (tc *TestCase) PopulateStorageCluster(cluster *corev1.StorageCluster) error {
+func PopulatePureStorageCluster(tc *TestCase, cluster *corev1.StorageCluster) error {
```
pure-jliao, line 125: is this function still needed except the name need to be populated?

Pure-AdamuKaapan: It needs to populate the cluster name, and it has one extra function of searching for node selectors like "node0" and replacing them with the `n`th node name in the cluster (basically allowing us to dynamically test node selectors). That's the for loop further down.

test/integration_test/fafb_matrix_test.go
```diff
@@ -81,21 +88,32 @@ var flashArrayPositiveInstallTests = []PureTestrailCase{
...
-			Spec:                    simpleClusterSpec,
-			ShouldStartSuccessfully: true,
+
+		TestName:        "C55129-C56414-C56411-C55003",
+		TestrailCaseIDs: []string{"C55129", "C56414", "C56411", "C55003"},
+		TestSpec: func(t *testing.T) interface{} {
+			// Construct Portworx StorageCluster object
+			cluster := &v1.StorageCluster{}
+			cluster.Spec = simpleClusterSpec
+			err := constructStorageCluster(cluster, pxSpecGenURL, pxSpecImages)
```
Pure-AdamuKaapan, line 98: You can define this whole thing as one function that will return a function and then reuse it between all the test cases, like this:

```
func createPureTestSpecFunction(cluster *v1.StorageCluster) func(t *testing.T) interface{} {
    return func(t *testing.T) {
        // The same function block as before
    }
}
```
and then invoke as
```
TestSpec: createPureTestSpecFunction(&v1.StorageCluster{
  Spec: simpleClusterSpec
})
```

This will help reduce code duplication and make it a lot more readable

test/integration_test/fafb_matrix_test.go
```diff
@@ -211,21 +247,30 @@ var flashArrayPositiveInstallTests = []PureTestrailCase{
...
 		// Construct Portworx StorageCluster object
-		cluster := &v1.StorageCluster{}
-		err := constructStorageCluster(cluster, pxSpecGenURL, pxSpecImages)
-		require.NoError(t, err)
+		testSpec := tc.TestSpec(t)
+		cluster, ok := testSpec.(*v1.StorageCluster)
+		require.True(t, ok)
 
-		c.PopulateStorageCluster(cluster)
+		logrus.Infof("jliao: tc ids %s", tc.TestrailCaseIDs)
```
Pure-AdamuKaapan, line 270: [nit] Don't forget to remove these if they're temp debugging statements

test/integration_test/types.go
```diff
@@ -31,6 +31,10 @@ type TestCase struct {
 	ShouldStartSuccessfully bool
 	// ShouldSkip check env requirements for the test case
 	ShouldSkip func() bool
+	// BackendRequirements contains additional information to construct a px-pure-secret.
+	BackendRequirements *BackendRequirements
```
Pure-AdamuKaapan, line 35: [nit] Do we want to call this `PureBackendRequirements`? Or keep it as-is and if we have more backends in the future we just expand this struct

Pure-AdamuKaapan (commented):
Overall looks good, had one improvement we can make as far as code reuse for the Pure test functions

test/integration_test/types.go
pure-jliao: I'll change it to pure backend and we could change it in case we have more.

test/integration_test/fafb_matrix_test.go
```diff
@@ -211,21 +216,28 @@ var flashArrayPositiveInstallTests = []PureTestrailCase{
 
 func TestFAFBPositiveInstallation(t *testing.T) {
 	for _, testCase := range flashArrayPositiveInstallTests {
-		t.Run(strings.Join(testCase.TestrailCaseIDs, "-"), matrixTest(testCase))
+		// t.Run(strings.Join(testCase.TestrailCaseIDs, "-"), matrixTest(testCase))
```
piyush-nimbalkar, line 219: Remove this comment

test/integration_test/common_test.go
piyush-nimbalkar: Do we need to pass the testcase to this function? I don't see it getting used.

test/integration_test/fafb_matrix_test.go
```diff
@@ -47,12 +46,14 @@ var (
...
 
-var flashArrayPositiveInstallTests = []PureTestrailCase{
+// TODO: add readable test names
+var flashArrayPositiveInstallTests = []TestCase{
 	//// C55130: https://portworx.testrail.net/index.php?/cases/view/55130
 	// Install PX through operator with internal KVDB and journal drive on FlashArray
 	{
-		TestCase: TestCase{
-			TestrailCaseIDs: []string{"C55130"},
+		TestName:        "C55130",
```
piyush-nimbalkar, line 54: @Pure-AdamuKaapan in future let's add user readable names to existing test cases to better understand the intention.

piyush-nimbalkar (approved):

test/integration_test/fafb_matrix_test.go
Pure-AdamuKaapan: Sounds good to me, I'll also go back and revise these names next time I work on this code

test/integration_test/basic_test.go
```diff
@@ -63,21 +58,17 @@ var testStorageClusterBasicCases = []TestCase{
...
+		TestFunc: BasicInstallWithNodeAffinity,
 	},
 	{
 		TestName:        "Upgrade",
 		TestrailCaseIDs: []string{},
 		TestSpec: func(t *testing.T) interface{} {
-			cluster := &op_corev1.StorageCluster{}
-			cluster.Name = "upgrade-test"
-			return cluster
+			return &op_corev1.StorageCluster{
```
Pure-AdamuKaapan, line 69: Any reason not using the `createStorageClusterTestSpecFunc` from earlier?

Pure-AdamuKaapan (approved):
LGTM!

PR #435, opened 2021-11-13
[OPERATOR-500] Automation code struct refactor

pure-jliao:
**What this PR does / why we need it**: Make the operator automation code more organized and easier to maintain

Closes #

**Special notes for your reviewer**: new structure be like:
```
test/
└── integration_test
    ├── Dockerfile
    ├── basic_test.go
    ├── fafb_test.go
    ├── main_test.go
    ├── operator-test-pod-template.yaml
    ├── test-deploy.sh
    ├── testspec
    │   ├── empty_spec.yaml
    │   └── pure-telemetry-cert.yaml
    ├── types
    │   ├── fafb.go
    │   └── testcase.go
    └── utils
        ├── constants.go
        ├── fafb_config.go
        ├── storagecluster.go
        └── utils.go
```


test/integration_test/pure/types.go
```diff
@@ -0,0 +1,30 @@
+package pure
```
piyush-nimbalkar, line 1: Could we just put all the types in one folder instead of having types for every type of test? Otherwise we will have a lot of folders like these.
May be we can just create a folder called types and have different filenames for each of them - 
```
test/
└── integration_test
    └── types
       ├── pure.go
       └── testcase.go
```

test/integration_test/storagecluster/utils.go
```diff
@@ -0,0 +1,171 @@
+package storagecluster
```
piyush-nimbalkar, line 1: Similar to types, we can just a single folder for utils and just use different file names for different kind of utils. It will also help us share some minor util functions.
```
test/
└── integration_test
    └── utils
       ├── storagecluster.go
       ├── fafb.go
       └── common.go
```

test/integration_test/fafb_matrix_test.go
```diff
@@ -6,15 +6,26 @@ import (
 	"testing"
 	"time"
 
+	"github.com/sirupsen/logrus"
```
piyush-nimbalkar, line 9: Can we rename this file to just `fafb_test.go`? I think it was just named *matrix* because it Adam had a matrix of +ve and -ve test cases which he was looping through.

test/integration_test/pure/types.go
pure-jliao: IMO we need different packages to serve different purposes that's why folders are created. we can have public constants and functions in different packages, so when calling them it's more clear. if we put everything into one package, I feel it's a little bit hard to see what's in the package since there are a lot, and I feel it's a little strange to call `types.TestCase`.. what do you think? but I agree that way from structure wise it's easier to maintain

pure-jliao: I'm not sure which approach is better, can I get more inputs? @haibinxie @Pure-AdamuKaapan

piyush-nimbalkar: The `types` is just a suggestion, you can name it to whatever makes sense. Although `types.TestCase` does not sound that weird to me. May be because kubernetes has some `types.*` objects.

But anyway, I just wanted to group all the testing related structs in one place instead of having those structs all over the place in different folders. That would create a lot of folders.

pure-jliao: ok makes sense, will update the PR.

piyush-nimbalkar (approved):

PR #436, opened 2021-11-13
OPERATOR-459 Migrate portworx daemonset pods

piyush-nimbalkar:

*Does handle:*
- Migration of portworx daemonset pods, assuming daemonset pods were in healthy state already
- Allows skipping nodes if user knows the issue and is okay with it 
- Handles restarts of operator during migration
- Handles daemonset placement constraints
- Logs states and errors

*Does not:*
- Does not raise events on failures
- Does not update status of the StorageCluster
- Does not migrate components
- No provision for rollback

Closes # https://portworx.atlassian.net/browse/OPERATOR-459

### DIFF
--- a/pkg/controller/storagecluster/storagecluster.go
+++ b/pkg/controller/storagecluster/storagecluster.go
@@ -220,15 +220,6 @@ func (c *Controller) Reconcile(ctx context.Context, request reconcile.Request) (
 		return reconcile.Result{}, err
 	}
 
-	if c.waitingForMigrationApproval(cluster) {
-		k8s.InfoEvent(
-			c.recorder, cluster, util.MigrationPendingReason,
-			fmt.Sprintf("To proceed with the migration, set the %s annotation on the "+
-				"StorageCluster to 'true'", constants.AnnotationMigrationApproved),
-		)
-		return reconcile.Result{}, nil
-	}
-
 	if err := c.validate(cluster); err != nil {
 		k8s.WarningEvent(c.recorder, cluster, util.FailedValidationReason, err.Error())
 		cluster.Status.Phase = string(corev1.ClusterOperationFailed)
@@ -236,6 +227,15 @@ func (c *Controller) Reconcile(ctx context.Context, request reconcile.Request) (
 			logrus.Errorf("Failed to update StorageCluster status. %v", updateErr)
 		}
 		return reconcile.Result{}, err
+	}
+
+	if c.waitingForMigrationApproval(cluster) {
+		k8s.InfoEvent(
+			c.recorder, cluster, util.MigrationPendingReason,
+			fmt.Sprintf("To proceed with the migration, set the %s annotation on the "+
+				"StorageCluster to 'true'", constants.AnnotationMigrationApproved),
+		)
+		return reconcile.Result{}, nil
 	}
 
 	if err := c.syncStorageCluster(cluster); err != nil {
@@ -247,8 +247,9 @@ func (c *Controller) Reconcile(ctx context.Context, request reconcile.Request) (
 }
 
 func (c *Controller) waitingForMigrationApproval(cluster *corev1.StorageCluster) bool {
-	approved, err := strconv.ParseBool(cluster.Annotations[constants.AnnotationMigrationApproved])
-	return err == nil && !approved
+	_, migrationLabelPresent := cluster.Annotations[constants.AnnotationMigrationApproved]
+	return (migrationLabelPresent && cluster.Status.Phase == "") ||
+		cluster.Status.Phase == constants.PhaseAwaitingApproval
 }
 
 func (c *Controller) validate(cluster *corev1.StorageCluster) error {
@@ -974,7 +975,7 @@ func (c *Controller) createPodTemplate(
 	newTemplate := v1.PodTemplateSpec{
 		ObjectMeta: metav1.ObjectMeta{
 			Namespace: cluster.Namespace,
-			Labels:    c.storageClusterSelectorLabels(cluster),
+			Labels:    c.StorageClusterSelectorLabels(cluster),
 		},
 		Spec: podSpec,
 	}
@@ -1007,7 +1008,7 @@ func (c *Controller) newSimulationPod(
 	newPod := &v1.Pod{
 		ObjectMeta: metav1.ObjectMeta{
 			Namespace: cluster.Namespace,
-			Labels:    c.storageClusterSelectorLabels(cluster),
+			Labels:    c.StorageClusterSelectorLabels(cluster),
 		},
 		Spec: v1.PodSpec{
 			NodeName: nodeName,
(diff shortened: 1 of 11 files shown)


pkg/migration/migration.go
```diff
@@ -486,6 +757,105 @@ func (h *Handler) findPortworxDaemonSet() (*appsv1.DaemonSet, error) {
...
+	addMigrationConstraints(&ds.Spec.Template.Spec)
+	// Set the update strategy to OnDelete to avoid restart of the daemonset
+	// pods due to the node affinity changes
+	ds.Spec.UpdateStrategy.Type = appsv1.OnDeleteDaemonSetStrategyType
+	return h.client.Update(context.TODO(), ds, &client.UpdateOptions{})
+}
+
+func addMigrationConstraints(podSpec *v1.PodSpec) {
+	if podSpec.Affinity == nil || podSpec.Affinity.NodeAffinity == nil {
+		podSpec.Affinity = &v1.Affinity{
```
pure-jliao, line 790: same here.

pure-jliao (approved):
looks pretty good to me, have some minor questions.

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -247,8 +247,9 @@ func (c *Controller) Reconcile(ctx context.Context, request reconcile.Request) (
 }
 
 func (c *Controller) waitingForMigrationApproval(cluster *corev1.StorageCluster) bool {
-	approved, err := strconv.ParseBool(cluster.Annotations[constants.AnnotationMigrationApproved])
-	return err == nil && !approved
+	_, migrationLabelPresent := cluster.Annotations[constants.AnnotationMigrationApproved]
+	return (migrationLabelPresent && cluster.Status.Phase == "") ||
```
haibinxie, line 251: isn't cluster status waitingForApproval derived from the annotation, why do we need this change?

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -1028,11 +1029,45 @@ func (c *Controller) newSimulationPod(
...
+	addMigrationConstraints(&newPod.Spec)
+
 	// Add default tolerations for StorageCluster pods
 	k8s.AddOrUpdateStoragePodTolerations(&newPod.Spec)
 	return newPod, nil
 }
 
+func addMigrationConstraints(podSpec *v1.PodSpec) {
+	if podSpec.Affinity == nil || podSpec.Affinity.NodeAffinity == nil {
+		podSpec.Affinity = &v1.Affinity{
```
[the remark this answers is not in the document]
haibinxie: we probably only has node affinity now, but would be good to handle it

pkg/migration/migration.go
```diff
@@ -101,15 +113,195 @@ func (h *Handler) createStorageClusterIfAbsent(ds *appsv1.DaemonSet) (*corev1.St
...
+		if err := h.waitForPortworxPod(cluster, node.Name, nodeLog); err != nil {
+			return err
+		}
+
+		if err := h.markMigrationAsDone(node); err != nil {
+			return err
+		}
+
+		value = node.Labels[constants.LabelPortworxDaemonsetMigration]
+		if value == constants.LabelValueMigrationDone {
```
haibinxie, line 181: nit: to simplify replace if/else with one line to log with the label.

pkg/migration/migration.go
```diff
@@ -101,15 +113,195 @@ func (h *Handler) createStorageClusterIfAbsent(ds *appsv1.DaemonSet) (*corev1.St
...
+	nodeLog *logrus.Entry,
+) error {
+	return wait.PollImmediate(podWaitInterval, daemonSetPodTerminationTimeout, func() (bool, error) {
+		node := &v1.Node{}
+		if err := h.client.Get(context.TODO(), types.NamespacedName{Name: nodeName}, node); err != nil {
+			nodeLog.Errorf("Failed to get node. %v", err)
+			return false, nil
+		}
+		value := node.Labels[constants.LabelPortworxDaemonsetMigration]
+		if value == constants.LabelValueMigrationSkip {
```
haibinxie, line 213: when we get here, it is probably too late to skip as it would be scaled down anyway. we may just check skip in the entry and not check here, and if it happens it would migrate anyway, what do you think?

pkg/controller/storagecluster/storagecluster.go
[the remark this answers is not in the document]
piyush-nimbalkar: Did not add it because we just have node affinity right now. But I will change this for future proofing, just in case we add more to placement constraints.

pkg/controller/storagecluster/storagecluster.go
piyush-nimbalkar: There was a race condition where the operator reconcile loop was starting to deploy px pods before the migration handler could add the node labels to prevent the operator from starting px pods right away. Although the pods immediately get terminated but there are a lot of pods created and terminated for few seconds.

This makes sure that the operator reconcile loop waits for the migration to start and only then it is starts allowing the StorageCluster waiting to be migrated.

pkg/migration/migration.go
piyush-nimbalkar: Actually skip does not skip the migration on that node. The daemonset pod is still deleted and the operator pod created. I don't think we need to support a case where user wants to keep some daemonset pods and some operator pods.
I added the skip to allow users to skip a node if it is not coming up for some known issue or may be an existing issue with their cluster. We don't want to get stuck at that node, nor do we want to proceed without user acknowledging the problem is known and we can skip the node. 
Basically the skip will allow the migration to proceed to a new node without worrying about whether pod comes up healthy or not.

Now the reason we want to skip from here is that if the user adds the skip flag later, we end up waiting for the timeout and then on the next restart we honor the skip. To make it react slightly faster, I added the skip check here.

haibinxie (approved):

pkg/migration/migration.go
haibinxie: i was thinking about a scenario that when it is being migrated and user later added the skip annotation, we now have two pods as we dont wait here.

pkg/migration/migration.go
```diff
@@ -101,15 +113,195 @@ func (h *Handler) createStorageClusterIfAbsent(ds *appsv1.DaemonSet) (*corev1.St
...
+			continue
+		}
+
+		if err := h.markMigrationAsStarting(node); err != nil {
+			return err
+		}
+
+		// Wait for daemonset pod to terminate, else it causes conflicts with
+		// the operator managed portworx pod
+		if err := h.waitForDaemonSetPodTermination(ds, node.Name, nodeLog); err != nil {
```
haibinxie, line 163: where do we let operator managed pod wait for portworx pod to be deleted?

pkg/migration/migration.go
piyush-nimbalkar: As I said before, the skip does not stop the daemonset from deleting and operator from creating the pod. That still happens, it just skips the waiting on that node and proceeds with the migration.

PR #439, opened 2021-11-16
Add validations and functional tests for migration

piyush-nimbalkar:

Closes #


test/integration_test/utils/node.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	"testing"
+
+	coreops "github.com/portworx/sched-ops/k8s/core"
+	"github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/require"
+)
+
+// AddLabelToRandomNode adds the given label to a random worker node
+func AddLabelToRandomNode(t *testing.T, key, value string) string {
```
pure-jliao, line 12: so this seems to be the same as node affinity test and could be used there as well. But it doesn't add label to a random node but only the first one.

test/integration_test/migration_test.go
```diff
@@ -30,17 +31,23 @@ const (
...
 	clusterCreationTimeout       = 2 * time.Minute
 	clusterCreationRetryInterval = 5 * time.Second
+	labelKeyPXEnabled            = "px/enabled"
 )
 
 var migrationTestCases = []types.TestCase{
 	{
-		TestName:   "SimpleDaemonSetMigration",
+		TestName:   "SimpleMigration",
 		ShouldSkip: func() bool { return false },
```
pure-jliao, line 40: nit: we could remove this as it will be set to return false by default if missing.

test/integration_test/migration_test.go
```diff
@@ -52,41 +59,76 @@ func BasicMigration(tc *types.TestCase) func(*testing.T) {
...
+	}
+}
+
+func MigrationWithoutNodeAffinity(tc *types.TestCase) func(*testing.T) {
+	return func(t *testing.T) {
+		if tc.ShouldSkip() {
+			t.Skip()
+		}
+
+		ds := testutil.GetExpectedDaemonSet(t, "migration/simple-daemonset.yaml")
```
pure-jliao, line 79: nit: we could move this to TestSpec in test definition

test/integration_test/migration_test.go
```diff
@@ -52,41 +59,76 @@ func BasicMigration(tc *types.TestCase) func(*testing.T) {
...
+	require.NoError(t, err)
 
-		logrus.Infof("Restarting portworx operator to trigger migration")
-		restartPortworxOperator(t)
+	logrus.Infof("Restarting portworx operator to trigger migration")
+	restartPortworxOperator(t)
 
-		pxImageMap, err := getPortworxImagesFromDaemonSet(t, ds)
-		require.NoError(t, err)
+	pxImageMap, err := getPortworxImagesFromDaemonSet(t, ds)
```
pure-jliao, line 99: nit: could create a new `daemonset.go` in utils and move this function there? similar to other utils functions that could be shared.

test/integration_test/migration_test.go
```diff
@@ -52,41 +59,76 @@ func BasicMigration(tc *types.TestCase) func(*testing.T) {
...
-		cluster, err = approveMigration(cluster)
-		require.NoError(t, err)
+	logrus.Infof("Validate original spec is deleted")
+	err = appops.Instance().ValidateDaemonSetIsTerminated(ds.Name, ds.Namespace, 5*time.Minute)
+	require.NoError(t, err)
 
-		logrus.Infof("Validate StorageCluster %s", cluster.Name)
-		err = testutil.ValidateStorageCluster(pxImageMap, cluster, ci_utils.DefaultValidateDeployTimeout, ci_utils.DefaultValidateDeployRetryInterval, true)
-		require.NoError(t, err)
+	logrus.Infof("Delete StorageCluster %s", cluster.Name)
```
pure-jliao, line 118: nit: could call `utils.UninstallAndValidateStorageCluster` instead

test/integration_test/utils/node.go
piyush-nimbalkar: It is already being used there. Calling it random because there is no guarantee which node would be the first one. It will depend on what k8s api returns.

test/integration_test/migration_test.go
piyush-nimbalkar: Actually I changed the logic of this. Instead of having a hard-coded version in the daemonset spec, I am rather creating the DS with the version URL we have passed to the test suite. So relying on PxSpecImages to fill the px version instead of the daemonset version to construct the pxSpecImages map.

piyush-nimbalkar:
Added the missing objects other than daemonset needed for portworx to start.

test/integration_test/utils/k8s.go
```diff
@@ -0,0 +1,146 @@
...
+			_, err = coreops.Instance().CreateServiceAccount(sa)
+		} else if role, ok := obj.(*rbacv1.Role); ok {
+			_, err = rbacops.Instance().CreateRole(role)
+		} else if rb, ok := obj.(*rbacv1.RoleBinding); ok {
+			_, err = rbacops.Instance().CreateRoleBinding(rb)
+		} else if clusterRole, ok := obj.(*rbacv1.ClusterRole); ok {
+			_, err = rbacops.Instance().CreateClusterRole(clusterRole)
+		} else if crb, ok := obj.(*rbacv1.ClusterRoleBinding); ok {
+			_, err = rbacops.Instance().CreateClusterRoleBinding(crb)
+		}
```
pure-jliao, line 82: can we return type not supported error here?

test/integration_test/utils/k8s.go
```diff
@@ -0,0 +1,146 @@
...
+			err = coreops.Instance().DeleteServiceAccount(sa.Name, sa.Namespace)
+		} else if role, ok := obj.(*rbacv1.Role); ok {
+			err = rbacops.Instance().DeleteRole(role.Name, role.Namespace)
+		} else if rb, ok := obj.(*rbacv1.RoleBinding); ok {
+			err = rbacops.Instance().DeleteRoleBinding(rb.Name, rb.Namespace)
+		} else if clusterRole, ok := obj.(*rbacv1.ClusterRole); ok {
+			err = rbacops.Instance().DeleteClusterRole(clusterRole.Name)
+		} else if crb, ok := obj.(*rbacv1.ClusterRoleBinding); ok {
+			err = rbacops.Instance().DeleteClusterRoleBinding(crb.Name)
+		}
```
pure-jliao, line 108: same here

pure-jliao (approved):
LGTM if we could pass the functional tests

PR #448, opened 2021-11-19
[OPERATOR-502] Add support to check portworx-operator version

pure-jliao:
**What this PR does / why we need it**: Check the px-operator version and skip tests

Closes #

### DIFF
--- a/test/integration_test/basic_test.go
+++ b/test/integration_test/basic_test.go
@@ -100,6 +100,9 @@ var testStorageClusterBasicCases = []types.TestCase{
 			return cluster
 		},
 		TestFunc: InstallWithTelemetry,
+		ShouldSkip: func(tc *types.TestCase) bool {
+			return ci_utils.PxOperatorVersion.LessThan(ci_utils.PxOperatorVer1_7)
+		},
 	},
 	{
 		TestName:        "InstallWithCSI",
--- a/test/integration_test/main_test.go
+++ b/test/integration_test/main_test.go
@@ -4,6 +4,7 @@ package integrationtest
 
 import (
 	"flag"
+	"fmt"
 	"os"
 	"strings"
 	"testing"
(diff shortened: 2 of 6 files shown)


test/integration_test/basic_test.go
```diff
@@ -100,6 +100,9 @@ var testStorageClusterBasicCases = []types.TestCase{
 			return cluster
 		},
 		TestFunc: InstallWithTelemetry,
+		ShouldSkip: func(tc *types.TestCase) bool {
+			return ci_utils.PxOperatorVersion.LessThanOrEqual(ci_utils.PxOperatorVer1_6)
```
haibinxie, line 104: 1.6.1 will also not have it.

haibinxie (approved):
good to merge after resolving the one comment.

test/integration_test/basic_test.go
pure-jliao: I planned to use 1.7 for checking that, but we may have different tags like 1.7.0-dev, 1.7.0-rc1. did some experiments seems `1.7-` should do the job,

test/integration_test/main_test.go
```diff
@@ -63,6 +64,11 @@ func setup() error {
 
 	ci_utils.PxUpgradeHopsURLList = strings.Split(pxUpgradeHopsURLs, ",")
 
+	ci_utils.PxOperatorVersion, err = ci_utils.GetPXOperatorVersion()
+	if err != nil {
+		return fmt.Errorf("failed to discover installed portworx operator version: %v", err)
+	}
+
```
piyush-nimbalkar, line 71: nit: If we are setting the variable `PxOperatorVersion` using a function also in the same package, should we just call `ci_utils.SetPxOperatorVersion` or `ci_utils.InitPxOperatorVersion` and let it set the value inside the function itself?

piyush-nimbalkar (approved):
LGTM
Minor comment. But feel free to merge without it.

test/integration_test/main_test.go
pure-jliao: the operator right now is installed out of the automation, so Get may work better for now. in the future if we install operator in the test, probably will change it to set.

PR #452, opened 2021-11-29
PTX-4560 Fix Upgrade hops test

nikolaypopov:
Pass MANIFEST_URL env var in case of edge-install and add check for empty expected PX version

### DIFF
--- a/pkg/util/test/util.go
+++ b/pkg/util/test/util.go
@@ -64,6 +64,9 @@ const (
 	PxRegistryPasswordEnvVarName = "REGISTRY_PASS"
 	// PxImageEnvVarName is the env variable to specify a specific Portworx image to install
 	PxImageEnvVarName = "PX_IMAGE"
+
+	// PxMasterVersion is a tag for Portworx master version
+	PxMasterVersion = "3.0.0.0"
 )
 
 // TestSpecPath is the path for all test specs. Due to currently functional test and
@@ -428,7 +431,13 @@ func validateStorageNodes(pxImageList map[string]string, cluster *corev1.Storage
 		if len(cluster.Spec.Env) > 0 {
 			for _, env := range cluster.Spec.Env {
 				if env.Name == PxReleaseManifestURLEnvVarName {
-					expectedPxVersion = strings.TrimSpace(regexp.MustCompile(`\S+\/(\S+)\/version`).FindStringSubmatch(env.Value)[1])
+					// Looking for clear PX version before /version in the URL
+					ver := regexp.MustCompile(`\S+\/(\d.\S+)\/version`).FindStringSubmatch(env.Value)
+					if ver != nil {
+						expectedPxVersion = ver[1]
+					} else { // If the above regex found nothing, assuming it was a master version URL
+						expectedPxVersion = PxMasterVersion
+					}
 				}
 			}
 		}
--- a/pkg/util/test/util.go
+++ b/pkg/util/test/util.go
@@ -428,16 +428,18 @@ func validateStorageNodes(pxImageList map[string]string, cluster *corev1.Storage
 
 	// Construct PX Version string used to match to deployed expected PX version
 	if strings.Contains(pxImageList["version"], "_") {
-		if len(cluster.Spec.Env) > 0 {
+		if cluster.Spec.Env != nil {
 			for _, env := range cluster.Spec.Env {
 				if env.Name == PxReleaseManifestURLEnvVarName {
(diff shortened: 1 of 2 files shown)


pkg/util/test/util.go
```diff
@@ -428,15 +428,16 @@ func validateStorageNodes(pxImageList map[string]string, cluster *corev1.Storage
 		if len(cluster.Spec.Env) > 0 {
 			for _, env := range cluster.Spec.Env {
 				if env.Name == PxReleaseManifestURLEnvVarName {
-					pxVersion = strings.TrimSpace(regexp.MustCompile(`\S+\/(\S+)\/version`).FindStringSubmatch(env.Value)[1])
-					if pxVersion == "" {
-						return fmt.Errorf("failed to extract version from value of %s", PxReleaseManifestURLEnvVarName)
-					}
+					expectedPxVersion = strings.TrimSpace(regexp.MustCompile(`\S+\/(\S+)\/version`).FindStringSubmatch(env.Value)[1])
```
piyush-nimbalkar, line 431: How do we handle master where the url does not have the version?

nikolaypopov: We haven't before, added support for handling master builds

pkg/util/test/util.go
```diff
@@ -428,15 +431,22 @@ func validateStorageNodes(pxImageList map[string]string, cluster *corev1.Storage
 		if len(cluster.Spec.Env) > 0 {
 			for _, env := range cluster.Spec.Env {
 				if env.Name == PxReleaseManifestURLEnvVarName {
-					pxVersion = strings.TrimSpace(regexp.MustCompile(`\S+\/(\S+)\/version`).FindStringSubmatch(env.Value)[1])
-					if pxVersion == "" {
-						return fmt.Errorf("failed to extract version from value of %s", PxReleaseManifestURLEnvVarName)
+					// Looking for clear PX version before /version in the URL
+					ver := regexp.MustCompile(`\S+\/(\d.\S+)\/version`).FindStringSubmatch(env.Value)
+					if ver != nil {
+						expectedPxVersion = ver[1]
```
pure-jliao, line 437: `break` here if found?

pkg/util/test/util.go
```diff
@@ -428,15 +431,22 @@ func validateStorageNodes(pxImageList map[string]string, cluster *corev1.Storage
...
+					} else { // If the above regex found nothing, assuming it was a master version URL
+						expectedPxVersion = PxMasterVersion
```
pure-jliao, line 439: we should set the master version outside of this env iteration loop if we didn't find anything matches

test/integration_test/basic_test.go
```diff
@@ -187,6 +187,9 @@ func BasicUpgrade(tc *types.TestCase) func(*testing.T) {
 				// Set defaults
 				portworx.SetPortworxDefaults(cluster)
 
+				err = ci_utils.ConstructStorageCluster(cluster, hopURL, specImages)
```
pure-jliao, line 190: are we calling this to update the env variables? then we could replace `cluster.Spec.Image = specImages["version"]` with `err = ci_utils.ConstructStorageCluster(cluster, hopURL, specImages)`, it's safer to call `SetPortworxDefaults` afterwards.

pkg/util/test/util.go
nikolaypopov: added break

pkg/util/test/util.go
```diff
@@ -425,18 +428,26 @@ func validateStorageNodes(pxImageList map[string]string, cluster *corev1.Storage
...
 			for _, env := range cluster.Spec.Env {
 				if env.Name == PxReleaseManifestURLEnvVarName {
-					pxVersion = strings.TrimSpace(regexp.MustCompile(`\S+\/(\S+)\/version`).FindStringSubmatch(env.Value)[1])
-					if pxVersion == "" {
-						return fmt.Errorf("failed to extract version from value of %s", PxReleaseManifestURLEnvVarName)
+					// Looking for clear PX version before /version in the URL
+					ver := regexp.MustCompile(`\S+\/(\d.\S+)\/version`).FindStringSubmatch(env.Value)
+					if ver != nil {
+						expectedPxVersion = ver[1]
+						break
```
pure-jliao, line 438: nit: better to move break under `expectedPxVersion = PxMasterVersion`

pure-jliao (approved):
LGTM as long as it's tested. one code suggestion left.

piyush-nimbalkar (approved):

PR #453, opened 2021-11-30
OPERATOR-505 Add spec fields for components during migration

piyush-nimbalkar:

- Created a separate file for StorageCluster spec generation code as the file was getting big. Have added only the new code to it to make it easier for review. Will move the rest of the StorageCluster creation code in a separate PR later.

Closes # https://portworx.atlassian.net/browse/OPERATOR-505

### DIFF
--- a/pkg/migration/generate.go
+++ b/pkg/migration/generate.go
@@ -1,0 +1,138 @@
+package migration
+
+import (
+	"context"
+
+	"github.com/libopenstorage/operator/drivers/storage/portworx/component"
+	pxutil "github.com/libopenstorage/operator/drivers/storage/portworx/util"
+	corev1 "github.com/libopenstorage/operator/pkg/apis/core/v1"
+	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+const (
+	storkDeploymentName              = "stork"
+	prometheusOperatorDeploymentName = "prometheus-operator"
+	serviceMonitorName               = "portworx-prometheus-sm"
+)
+
+func (h *Handler) addStorkSpec(cluster *corev1.StorageCluster) error {
+	dep, err := h.getDeployment(storkDeploymentName, cluster.Namespace)
+	if err != nil {
+		return err
+	}
+	cluster.Spec.Stork = &corev1.StorkSpec{
+		Enabled: dep != nil,
+	}
+	return nil
+}
+
+func (h *Handler) addAutopilotSpec(cluster *corev1.StorageCluster) error {
+	dep, err := h.getDeployment(component.AutopilotDeploymentName, cluster.Namespace)
+	if err != nil {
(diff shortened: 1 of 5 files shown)


pure-jliao (approved):
Overall looks good to me. some minor suggestions but ok to merge without changes

pkg/migration/generate.go
```diff
@@ -0,0 +1,165 @@
...
+		Enabled: alertManagerFound,
+	}
+
+	if !cluster.Spec.Monitoring.Prometheus.ExportMetrics &&
+		!cluster.Spec.Monitoring.Prometheus.AlertManager.Enabled {
+		return nil
+	}
+
+	// Check for prometheus operator
+	prometheusOpDeployment := &appsv1.Deployment{}
```
haibinxie, line 144: this code block is used in several places, it can be put in a function doesDeploymentExist() (bool, error)

pkg/migration/migration.go
```diff
@@ -613,14 +622,24 @@ func (h *Handler) constructStorageCluster(ds *appsv1.DaemonSet) *corev1.StorageC
...
+	csiEnabled := false
+	telemetryEnabled := false
 	for _, c := range ds.Spec.Template.Spec.Containers {
-		if c.Name == "csi-node-driver-registrar" {
-			cluster.Spec.FeatureGates = map[string]string{
-				"CSI": "true",
-			}
-			break
+		switch c.Name {
+		case "csi-node-driver-registrar":
```
haibinxie, line 629: can we add two const strings for the container names, use here and in portworx code creates the container.

pkg/migration/generate.go
```diff
@@ -0,0 +1,165 @@
...
+	corev1 "github.com/libopenstorage/operator/pkg/apis/core/v1"
+	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+const (
+	storkDeploymentName              = "stork"
+	autopilotDeploymentName          = "autopilot"
```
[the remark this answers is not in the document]
piyush-nimbalkar: Done. Used the matching variables from the driver package.
Some of the components have different names in operator and the spec generator. Those I have kept in the migration package itself.

pkg/migration/generate.go
```diff
@@ -0,0 +1,165 @@
...
+			Namespace: cluster.Namespace,
+		},
+		svcMonitor,
+	)
+	if errors.IsNotFound(err) {
+		svcMonitorFound = false
+	} else if err != nil {
+		return err
+	}
+	cluster.Spec.Monitoring.Prometheus = &corev1.PrometheusSpec{
```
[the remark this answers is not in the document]
piyush-nimbalkar: Makes sense. Will initialize the monitoring spec here too if empty.

pkg/migration/generate.go
piyush-nimbalkar: Done.
Instead of a boolean, just returning the actual deployment as in one place we need the deployment object.

haibinxie (approved):

PR #454, opened 2021-11-30
OPERATOR-518: PKS install fix

zoxpx:
Fixes containerd-based PKS installs:
* mounting `-v /run/containerd:/run/containerd`
* adding new `-v /var/vcap/sys/run/containerd/containerd.sock:/run/containerd/containerd.sock` mount  (PKS only)

Additional:
* renamed `containerdsock` to `containerddir` label for all "/run/containerd" mounts  (all specs)
* removed `--keep-px-up` option -- this is a default PX behavior since px-2.1  (PKS only)

Closes # OPERATOR-518

Similar changes implemented for PKS-installs via https://install.portworx.com


zoxpx:
Note also -- did a test-deployment on a new PKS deploy, confirmed the install works OK  (build with this change, and associated PWX-22105 oci-monitor change)

drivers/storage/portworx/deployment.go
```diff
@@ -1464,6 +1460,20 @@ func (t *template) getK3sVolumeInfoList() []volumeInfo {
...
 
+func (t *template) getPKSVolumeInfoList() []volumeInfo {
+	if !t.isPKS {
+		return []volumeInfo{}
+	}
+
+	return []volumeInfo{
+		{
+			name:      "containerd-pks",
+			hostPath:  "/var/vcap/sys/run/containerd/containerd.sock",
```
piyush-nimbalkar, line 1471: If the containerd directory is under `/var/vcap/sys/run/containerd`, then why do we need to mount `/run/containerd` in PKS? Rather is `/run/containerd` even present on the PKS cluster?

zoxpx: We'll actually need both --
* the `containerd` APIs are a little bit "weird" in a way that just the access to the socket-file is not sufficient, and we also need the access to the `/run/containerd` directory
   - (i.e. unlike DBus or Docker, where the only thing required for APIs to work is the socket-endpoint)
* as for the PKS specifically, those guys "broken up" their install, so they keep majority of the data in `/run/containerd`, but the socket is at `/var/vcap/sys/run/containerd/containerd.sock`

zoxpx:
Rebased off of latest `master` -- re-requesting the review.

piyush-nimbalkar (approved):

zoxpx:
Thanks Piyush -- merging the PR, cherry-picking for v1.7.0

PR #457, opened 2021-12-02
Migrate spec based install components to operator managed components

piyush-nimbalkar:
- Pausing any component from getting installed until all the Portworx pods have been migrated. This is to ensure if there is any failure during Portworx pod migration, then there are less pieces to revert. Not all components are paused though. Components like PSP, auth, PX rbac and service and installed to not hinder Portworx functionality.

- Once the Portworx pod migration is completed, just deleting the old components and letting operator take care of creating the new ones.

- Still working on the unit tests and functional tests. Will add them to this PR as separate commits.

Closes # https://portworx.atlassian.net/browse/OPERATOR-458


pure-jliao (commented):
so far looks pretty good to me.

pkg/migration/cleanup.go
```diff
@@ -0,0 +1,241 @@
...
+		return err
+	}
+	if err := h.deletePortworxRbac(cluster); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (h *Handler) deletePortworxRbac(cluster *corev1.StorageCluster) error {
+	if err := h.deleteObject("node-role-binding", "", &rbacv1.ClusterRoleBinding{}, cluster); err != nil {
```
haibinxie, line 44: nit: please use const for the object names in the function.

pkg/migration/cleanup.go
```diff
@@ -0,0 +1,241 @@
...
+	if err := h.deleteObject("node-get-put-list-role", "", &rbacv1.ClusterRole{}, cluster); err != nil {
+		return err
+	}
+	if err := h.deleteObject("px-role-binding", cluster.Namespace, &rbacv1.RoleBinding{}, cluster); err != nil {
+		return err
+	}
+	if err := h.deleteObject("px-role", cluster.Namespace, &rbacv1.Role{}, cluster); err != nil {
+		return err
+	}
+	if err := h.deleteObject("px-role-binding", "portworx", &rbacv1.RoleBinding{}, cluster); err != nil {
```
haibinxie, line 56: namespace is always "portworx"?

pkg/migration/generate.go
```diff
@@ -2,23 +2,390 @@ package migration
...
+			envMap["PORTWORX_AUTH_SYSTEM_KEY"] = &v1.EnvVar{
+				Name:  "PORTWORX_AUTH_SYSTEM_KEY",
+				Value: c.Args[i+1],
+			}
+			i++
+		} else if arg == "--log" {
+			cluster.Annotations[pxutil.AnnotationLogFile] = c.Args[i+1]
+			i++
+		} else {
+			miscArgs += arg + " "
```
haibinxie, line 293: Just curious, due to the big list above, if we miss any parameter above, is there any pattern/way to cross check if this arg should belong to miscArgs?

haibinxie (approved):
Looks good

pkg/migration/cleanup.go
piyush-nimbalkar: Yes, this is the default namespace we had for storing volume encryption secrets. Our spec generator still uses that namespace by default.

pkg/migration/generate.go
piyush-nimbalkar: There is no pattern as such. If an argument does not have corresponding field in StorageCluster spec, we are adding it to the misc args. Misc args will take anything and pass it verbatim to oci-mon. At least, we won't change or break the existing PX deployment.
Saying that when we add new corresponding fields, we will have to keep this convert function up to date for better migration.

pure-jliao (approved):
LGTM!

PR #460, opened 2021-12-02
[OPERATOR-438] Add portworx proxy support for telemetry ccm container

pure-jliao:

**What this PR does / why we need it**: mount PX_HTTP(S)_PROXY as a configmap to ccm container

Closes # [OPERATOR-438](https://portworx.atlassian.net/browse/OPERATOR-438)

**Special notes for your reviewer**: will add unit tests once verified the functionality manually.

### DIFF
--- a/drivers/storage/portworx/component/telemetry.go
+++ b/drivers/storage/portworx/component/telemetry.go
@@ -28,6 +28,12 @@ const (
 	TelemetryComponentName = "Portworx Telemetry"
 	// TelemetryConfigMapName is the name of the config map that stores the telemetry configuration for CCM
 	TelemetryConfigMapName = "px-telemetry-config"
+	// TelemetryCCMProxyConfigMapName is the name of the config map that stores the PX_HTTP(S)_PROXY env var
+	TelemetryCCMProxyConfigMapName = "px-ccm-service-proxy-config"
+	// TelemetryCCMProxyFilePath path of the http proxy file
+	TelemetryCCMProxyFilePath = "/cache/network/http_proxy"
+	// TelemetryCCMProxyFileName name of the http proxy file
+	TelemetryCCMProxyFileName = "http_proxy"
 	// TelemetryPropertiesFilename is the name of the CCM properties file
 	TelemetryPropertiesFilename = "ccm.properties"
 	// TelemetryArcusLocationFilename is name of the file storing the location of arcus endpoint to use
@@ -85,6 +91,9 @@ func (t *telemetry) Reconcile(cluster *corev1.StorageCluster) error {
 	if err := t.createConfigMap(cluster, ownerRef); err != nil {
 		return err
 	}
+	if err := t.reconcileCCMProxyConfigMap(cluster, ownerRef); err != nil {
+		return err
+	}
 
 	if err := t.deployMetricsCollector(cluster, ownerRef); err != nil {
 		return err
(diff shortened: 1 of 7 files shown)


drivers/storage/portworx/component/telemetry.go
```diff
@@ -99,6 +108,11 @@ func (t *telemetry) Delete(cluster *corev1.StorageCluster) error {
 		return err
 	}
 
+	err := k8sutil.DeleteConfigMap(t.k8sClient, TelemetryCCMProxyConfigMapName, cluster.Namespace, *ownerRef)
+	if err != nil && !errors.IsNotFound(err) {
```
piyush-nimbalkar, line 112: nit: `k8sutil.DeleteConfigMap` handles the `IsNotFound` error. We don't have to check it again here.

drivers/storage/portworx/util/util.go
```diff
@@ -496,6 +500,16 @@ func GetClusterEnvVarValue(ctx context.Context, cluster *corev1.StorageCluster,
 	return ""
 }
 
+// GetPxProxyEnvVarValue returns the PX_HTTP(S)_PROXY environment variable value for a cluster.
+func GetPxProxyEnvVarValue(ctc context.Context, cluster *corev1.StorageCluster) string {
```
piyush-nimbalkar, line 504: No need to pass the context if we are not using it.

drivers/storage/portworx/component/telemetry.go
```diff
@@ -697,6 +711,46 @@ func (t *telemetry) createConfigMap(
...
+				Name:      TelemetryCCMProxyConfigMapName,
+				Namespace: cluster.Namespace,
+			},
+			existingConfigMap,
+		)
+		if errors.IsNotFound(err) {
+			return nil
+		} else if err != nil {
+			return err
+		}
```
piyush-nimbalkar, line 734: We don't need this. The DeleteConfigMap does the same thing. We can directly call `DeleteConfigMap` if proxy is nil.

drivers/storage/portworx/util/util.go
```diff
@@ -496,6 +500,17 @@ func GetClusterEnvVarValue(ctx context.Context, cluster *corev1.StorageCluster,
 	return ""
 }
 
+// GetPxProxyEnvVarValue returns the PX_HTTP(S)_PROXY environment variable value for a cluster.
+// Note: we only expect one proxy for the telemetry CCM container
+func GetPxProxyEnvVarValue(cluster *corev1.StorageCluster) string {
+	for _, env := range cluster.Spec.Env {
+		if env.Name == EnvKeyPortworxHTTPProxy || env.Name == EnvKeyPortworxHTTPSProxy {
```
piyush-nimbalkar, line 507: Did we confirm, if there is a precedence between https and http in case both are passed?

drivers/storage/portworx/component/telemetry.go
```diff
@@ -697,6 +710,32 @@ func (t *telemetry) createConfigMap(
...
+	return k8sutil.CreateOrUpdateConfigMap(
+		t.k8sClient,
+		&v1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:            TelemetryCCMProxyConfigMapName,
+				Namespace:       cluster.Namespace,
+				OwnerReferences: []metav1.OwnerReference{*ownerRef},
+			},
+			Data: map[string]string{
+				"http_proxy": proxy,
```
piyush-nimbalkar, line 732: nit: Let's use the constant `TelemetryCCMProxyFileName` here.

drivers/storage/portworx/component/telemetry.go
```diff
@@ -668,7 +681,7 @@ func (t *telemetry) createConfigMap(
         "use_appliance_id": "true"
       },
       "proxy": {
-        "path": "/dev/null"
+        "path": "/cache/network/http_proxy"
```
piyush-nimbalkar, line 684: nit: We should use the constant `TelemetryCCMProxyFilePath` here too.

piyush-nimbalkar (approved):
LGTM
Few minor comments.

drivers/storage/portworx/util/util.go
pure-jliao: who should I ask again? for the http_proxy file, it only expect one, so it's more like a question for portworx.

drivers/storage/portworx/component/telemetry.go
pure-jliao: the whole ccm property config is a const string, I feel there's no point to just change this field only..

PR #461, opened 2021-12-03
PTX-4514 PVC Controller validation and basic regression test

nikolaypopov:
PVC Controller basic regression test and various improvements for validation of PVC Controller components and images

* Added PVC Controller basic regression test that covers multiple PVC Controller test cases
             Validate PVC Controller annotations are missing by default
             Enable portworx-pvc-controller in StorageCluster annotations and validate PVC Controller pods are created
             Disable portworx-pvc-controller in StorageCluster annotations and validate PVC Controller pods are deleted 
             Validate portworx-pvc-controller deployment and pods 
             Delete PVC Controller pods and validate pods get re-deployed 
             Set custom PVC Controller secure-port in StorageCluster annotations and validate
             Delete custom PVC Controller secure-port in StorageCluster annotations and validate ports are changed to default
* Added more in-depth validation for PVC Controller components and images

### DIFF
--- a/pkg/util/test/util.go
+++ b/pkg/util/test/util.go
@@ -788,17 +788,9 @@ func validateComponents(pxImageList map[string]string, cluster *corev1.StorageCl
 		return err
 	}
 
-	if isPVCControllerEnabled(cluster) {
-		pvcControllerDp := &appsv1.Deployment{}
-		pvcControllerDp.Name = "portworx-pvc-controller"
-		pvcControllerDp.Namespace = cluster.Namespace
-		if err = appops.Instance().ValidateDeployment(pvcControllerDp, timeout, interval); err != nil {
-			return err
-		}
-
-		if err = validateImageTag(k8sVersion, cluster.Namespace, map[string]string{"name": "portworx-pvc-controller"}); err != nil {
-			return err
-		}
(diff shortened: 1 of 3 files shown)


pkg/util/test/util.go
```diff
@@ -1159,6 +1219,76 @@ func validateCsiContainerInPxPods(namespace string, csi bool, timeout, interval
...
 }
 
+func validatePvcControllerPorts(annotations map[string]string, pvcControllerDeployment *appsv1.Deployment, timeout, interval time.Duration) error {
+	logrus.Debug("Validate PVC Controller custom ports")
+
+	if annotations == nil {
+		return nil
+	}
+
+	pvcPort := annotations["portworx.io/pvc-controller-port"]
```
piyush-nimbalkar, line 1229: Starting k8s 1.22 we don't honor the `pvc-controller-port` annotation. Only `secure-port` annotation is honored. That is because the new pvc controller doesn't take the port parameter now.
Let's remove this port validation and keep just the secure-port, else this will start failing soon on k8s 1.22 and above.

nikolaypopov: Removed `pvc-controller-port` from test and validation

piyush-nimbalkar (approved):

PR #468, opened 2021-12-07
add more functional tests for metrics collector

haibinxie:
**What this PR does / why we need it**: https://portworx.atlassian.net/browse/OPERATOR-520

Closes #


pkg/util/test/util.go
```diff
@@ -1478,78 +1479,153 @@ func validateMonitoring(pxImageList map[string]string, cluster *corev1.StorageCl
...
+		return err
+	}
+
+	logrus.Infof("Telemetry successfully uninstalled.")
+	return nil
+}
+
+// ValidateTelemetry validates telemetry component is installed/uninstalled as expected
+func ValidateTelemetry(pxImageList map[string]string, cluster *corev1.StorageCluster, timeout, interval time.Duration) error {
+	if cluster.Spec.Monitoring != nil && cluster.Spec.Monitoring.Telemetry.Enabled {
```
piyush-nimbalkar, line 1533: It should be -
```
if cluster.Spec.Monitoring != nil && 
  cluster.Spec.Monitoring.Telemetry != nil &&
  cluster.Spec.Monitoring.Telemetry.Enabled {
```

pkg/util/util.go
```diff
@@ -158,16 +158,35 @@ func HaveTolerationsChanged(
...
 	// The reason we don't use deepEqual for volumes is k8s API server may add defaultMode to it.
-	return equality.Semantic.DeepDerivative(d1.Spec.Template.Spec.Containers, d2.Spec.Template.Spec.Containers) &&
-		len(d1.Spec.Template.Spec.Volumes) == len(d2.Spec.Template.Spec.Volumes) &&
-		equality.Semantic.DeepDerivative(d1.Spec.Template.Spec.Volumes, d2.Spec.Template.Spec.Volumes) &&
-		equality.Semantic.DeepEqual(d1.Spec.Template.Spec.ImagePullSecrets, d2.Spec.Template.Spec.ImagePullSecrets) &&
-		equality.Semantic.DeepEqual(d1.Spec.Template.Spec.Affinity, d2.Spec.Template.Spec.Affinity) &&
-		equality.Semantic.DeepEqual(d1.Spec.Template.Spec.Tolerations, d2.Spec.Template.Spec.Tolerations) &&
-		equality.Semantic.DeepEqual(d1.Spec.Template.Spec.ServiceAccountName, d2.Spec.Template.Spec.ServiceAccountName)
+	if !equality.Semantic.DeepDerivative(d1.Spec.Template.Spec.Containers, d2.Spec.Template.Spec.Containers) {
+		return false, fmt.Errorf("vontainers not equal, first: %+v, second: %+v", d1.Spec.Template.Spec.Containers, d2.Spec.Template.Spec.Containers)
```
piyush-nimbalkar, line 165: nit: *containers

test/integration_test/basic_test.go
```diff
@@ -10,19 +10,18 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/require"
 	v1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+	appsops "github.com/portworx/sched-ops/k8s/apps"
+	coreops "github.com/portworx/sched-ops/k8s/core"
+	"github.com/portworx/sched-ops/k8s/operator"
```
piyush-nimbalkar, line 17: nit: format the imports

test/integration_test/testspec/metricsCollectorDeployment.yaml
```diff
@@ -0,0 +1,85 @@
...
+          securityContext:
+            runAsUser: 1111
+          volumeMounts:
+            - mountPath: /config
+              name: px-collector-proxy-config
+              readOnly: true
+            - mountPath: /appliance-cert
+              name: pure-telemetry-certs
+              readOnly: true
+      schedulerName: default-scheduler
```
piyush-nimbalkar, line 73: Where does this come scheduler name come from? Let's remove it.
I check the unit test and seems like it is present there too. Let's fix that and in the unit test do a direct spec comparison (`require.Equal`) instead of DeploymentDeepEqual.

test/integration_test/basic_test.go
```diff
@@ -228,25 +241,41 @@ func InstallWithTelemetry(tc *types.TestCase) func(*testing.T) {
...
+	require.NoError(t, err)
 
-	// Deploy portworx with telemetry set to true
-	cluster, err = ci_utils.CreateStorageCluster(cluster)
+	cluster.Spec.Monitoring.Telemetry.Enabled = false
+	cluster, err = ci_utils.UpdateStorageCluster(cluster)
 	require.NoError(t, err)
 
-	err = testutil.ValidateTelemetry(
+	err = testutil.ValidateTelemetryUninstalled(
```
piyush-nimbalkar, line 263: Instead of calling the ValidateInstall and ValidateUninstall, let's just call ValidateTelemetry. Check how it is done for other components. Let's follow a consistent pattern.

pkg/util/test/util.go
```diff
@@ -1478,78 +1479,153 @@ func validateMonitoring(pxImageList map[string]string, cluster *corev1.StorageCl
...
+	if !ok {
+		return fmt.Errorf("failed to find image for metrics collector")
+	}
+
+	imageName = util.GetImageURN(cluster, imageName)
+
+	deployment, err := appops.Instance().GetDeployment("px-metrics-collector", cluster.Namespace)
+	if err != nil {
+		return err
+	}
```
piyush-nimbalkar, line 1607: We already have the deployment from line 1554 in `actualDeployment`.

test/integration_test/basic_test.go
haibinxie: Actually in the last step after it's uninstalled you can't do ValiateTelemetry, so i explicitly used TelemetryUninstalled and Installed APIs.

test/integration_test/basic_test.go
haibinxie: isnt this already formatted? this group is for portworx

test/integration_test/basic_test.go
piyush-nimbalkar: That makes me think if we want to call ValidateUninstallTelemetry on the last step. Ideally ValidateUninstallStorageCluster should check that px and all components are completely removed. This doesn't apply to just telemetry, but rather to all other components.

test/integration_test/basic_test.go
piyush-nimbalkar: The practice is not to format the imports for every group. Generally there are 2 or 3 groups - 
- in-built packages
- external package
- packages internal to this repository. In our case - `github.com/libopenstorage/operator/...`

haibinxie: this section was actually auto-generated by intellij, I think it detected the code and figured out the import and alias.

haibinxie: actually there were there before my PR, so somehow intellij decided to move them here. let's compare our intellij settings in todays standup meeting.

piyush-nimbalkar: If you remove the blank lines between those imports, intellij should automatically group them and sort it.

haibinxie: sounds good, done

drivers/storage/portworx/components_test.go
```diff
@@ -11564,7 +11564,10 @@ func TestTelemetryEnableAndDisable(t *testing.T) {
 	require.NoError(t, err)
 
 	expectedDeployment := testutil.GetExpectedDeployment(t, "metricsCollectorDeployment.yaml")
-	require.True(t, util.DeploymentDeepEqual(expectedDeployment, deployment))
+	// Revise unrelated fields so we can compare.
+	expectedDeployment.Spec.Template.Spec.DeprecatedServiceAccount = ""
```
piyush-nimbalkar, line 11568: Instead of setting the deprecated service account empty here, let's just remove it from the test spec.

test/integration_test/basic_test.go
```diff
@@ -100,6 +98,20 @@ var testStorageClusterBasicCases = []types.TestCase{
...
+			trueVal := true
+			cluster.Spec.Storage = &corev1.StorageSpec{
+				UseAll:               nil,
+				UseAllWithPartitions: &trueVal,
+				ForceUseDisks:        &trueVal,
+				Devices:              nil,
+				CacheDevices:         nil,
+				JournalDevice:        nil,
+				SystemMdDevice:       nil,
+				KvdbDevice:           nil,
```
piyush-nimbalkar, line 110: Just curious. Why do we need to set the rest of them to nil?

test/integration_test/basic_test.go
```diff
@@ -100,6 +98,20 @@ var testStorageClusterBasicCases = []types.TestCase{
...
+			}
+			cluster.Spec.DeleteStrategy = &corev1.StorageClusterDeleteStrategy{
+				Type: corev1.UninstallAndWipeStorageClusterStrategyType,
+			}
```
piyush-nimbalkar, line 114: The UninstallStorageCluster function will automatically add the UninstallAndWipe delete strategy. We don't need to add it here.

piyush-nimbalkar (approved):
LGTM after addressing comments

test/integration_test/basic_test.go
haibinxie: it's auto generated.

test/integration_test/basic_test.go
haibinxie: It is useful to run on dev env. without wiping after some runs it would fail due to no more available drives

piyush-nimbalkar: You need to get a cluster with backing drives.
Also, if the wipe fails on your cluster so often and if it is a valid issue, we should be filing a PWX ticket.

test/integration_test/basic_test.go
piyush-nimbalkar: But we can remove it from the code, right?

drivers/storage/portworx/components_test.go
piyush-nimbalkar: This comment doesn't seem to be addressed.

test/integration_test/basic_test.go
haibinxie: the wipe did not fail

drivers/storage/portworx/components_test.go
haibinxie: missed this one, it is not in test spec. it is deprecated but gets set even if it is not in test spec.

PR #469, opened 2021-12-08
Add CSI spec, snapshot controller, and CRDs

ggriffiths:

* Adds CSI spec to storagecluster and option to enable snapshot controller / CRDs

Closes #


ggriffiths:
fixing/adding tests now. Ready for review soon.

ggriffiths:
Spec addition will look like the following:
```
  csi:
    enabled: true
    installSnapshotController: true
    installSnapshotCRDs: true
```

drivers/storage/portworx/component/csi.go
```diff
@@ -127,6 +139,13 @@ func (c *csi) Reconcile(cluster *corev1.StorageCluster) error {
 		}
 		c.csiNodeInfoCRDCreated = true
 	}
+	if !c.csiSnapshotCRDsCreated && cluster.Spec.CSI.InstallSnapshotCRDs != nil && *cluster.Spec.CSI.InstallSnapshotCRDs {
+		if err := c.createCSISnapshotCRDs(&c.k8sVersion); err != nil {
```
piyush-nimbalkar, line 143: We don't need to pass the `k8sVersion`. It is already part of the receiver `c`.

drivers/storage/portworx/util/csi_generator.go
```diff
@@ -181,6 +187,19 @@ func (g *CSIGenerator) GetCSIConfiguration() *CSIConfiguration {
...
+	if g.kubeVersion.GreaterThanOrEqual(k8sVer1_17) {
+		cv.IncludeSnapshotController = g.includeSnapshotController
+	} else {
+		cv.IncludeSnapshotController = false
+	}
+
+	pxVer2_5, _ := version.NewVersion("2.5")
+	if g.pxVersion.GreaterThanOrEqual(pxVer2_5) {
+		cv.IncludeEphemeralSupport = true
+	}
```
piyush-nimbalkar, line 201: This is already present on line 186

pkg/apis/core/v1/storagecluster.go
```diff
@@ -176,6 +178,13 @@ type NodeSpec struct {
 	CommonConfig
 }
 
+// CSISpec is used to define the CSI configurations
+type CSISpec struct {
+	Enabled                   bool  `json:"enabled,omitempty"`
+	InstallSnapshotController *bool `json:"installSnapshotController,omitempty"`
+	InstallSnapshotCRDs       *bool `json:"installSnapshotCRDs,omitempty"`
```
piyush-nimbalkar, line 185: Would it happen that users have the CRDs but not the controller? Should we just have one flag to enable snapshot controller? We can try to create the CRDs and ignore if they are already present.

pkg/util/util_test.go
```diff
@@ -3,10 +3,11 @@ package util
 import (
 	"testing"
 
+	"github.com/libopenstorage/operator/pkg/constants"
```
piyush-nimbalkar, line 6: nit: Can we format the imports?

pkg/util/util_test.go
```diff
@@ -15,7 +16,7 @@ func TestImageURN(t *testing.T) {
 	out := getImageURN("", "registry.io", "")
 	require.Equal(t, "", out)
 
-	// TestCase: Empty repo and registry
+	// TestCase: Empty repo and registrypkg/util/util_test.go
```
piyush-nimbalkar, line 19: I guess this was done unintentionally

drivers/storage/portworx/component/csi.go
```diff
@@ -989,15 +1036,58 @@ func createCSINodeInfoCRD() error {
...
 ) *pxutil.CSIConfiguration {
 	deprecatedCSIDriverName := pxutil.UseDeprecatedCSIDriverName(cluster)
 	disableCSIAlpha := pxutil.DisableCSIAlpha(cluster)
 	kubeletPath := pxutil.KubeletPath(cluster)
+	if cluster.Spec.CSI == nil {
+		cluster.Spec.CSI = &corev1.CSISpec{
+			Enabled:                   false,
+			InstallSnapshotController: boolPtr(false),
+		}
+	}
```
piyush-nimbalkar, line 1088: Why do we need this?

drivers/storage/portworx/portworx.go
```diff
@@ -736,6 +742,23 @@ func setNodeSpecDefaults(toUpdate *corev1.StorageCluster) {
...
+			// Enabled by feature gate, but not spec
+			Enabled:                   pxutil.FeatureCSI.IsEnabled(toUpdate.Spec.FeatureGates),
+			InstallSnapshotController: boolPtr(false),
+			InstallSnapshotCRDs:       boolPtr(false),
+		}
+	}
+
+	if toUpdate.Spec.CSI.Enabled && toUpdate.Spec.CSI.InstallSnapshotController == nil {
+		// Do not install snapshot controller by default
+		toUpdate.Spec.CSI.InstallSnapshotController = boolPtr(false)
```
piyush-nimbalkar, line 757: I think we can leave it to empty. Anyway we consider empty as false by default. Wdyt?

drivers/storage/portworx/portworx.go
```diff
@@ -299,6 +302,9 @@ func (p *portworx) SetDefaultsOnStorageCluster(toUpdate *corev1.StorageCluster)
 	}
 
 	setDefaultAutopilotProviders(toUpdate)
+	if pxutil.IsCSIEnabled(toUpdate) {
```
piyush-nimbalkar, line 305: Should we reset the CSI feature gate and use `cluster.Spec.CSI` in this function?
- If we don't reset then users will have to toggle two flags to disable CSI
- Once we have populated `cluster.Spec.CSI.Enabled`, then rest of the code, should be able to just use the above IsCSIEnabled function which checks the new flag instead of the featureGates map.

drivers/storage/portworx/components_test.go
```diff
@@ -4849,6 +4849,41 @@ func TestCSIInstallEphemeralWithK8s1_20VersionAndPX2_5(t *testing.T) {
...
+	// Should have snapshot controller
+	expectedDeployment = testutil.GetExpectedDeployment(t, "csiDeployment_1.0_k8s120_with_snapcontroller.yaml")
+	err = testutil.Get(k8sClient, deployment, component.CSIApplicationName, cluster.Namespace)
+	require.NoError(t, err)
+	fmt.Println("actual", deployment.Spec.Template.Spec.Containers)
+	require.Equal(t, len(expectedDeployment.Spec.Template.Spec.Containers), len(deployment.Spec.Template.Spec.Containers))
+	require.Equal(t, expectedDeployment.Spec, deployment.Spec)
+
+	// Should have volumesnapshot CRDs
+	_, err = apiextensionsops.Instance().GetCRD("volumesnapshots.snapshot.storage.k8s.io", metav1.GetOptions{})
```
piyush-nimbalkar, line 4884: Can we check all 3 crds?
Also, can we test for v3 and v4 separately?

piyush-nimbalkar (commented):
Mostly looks good!

pkg/apis/core/v1/storagecluster.go
ggriffiths: Sure

drivers/storage/portworx/component/csi.go
ggriffiths: There were some code paths breaking when the csi spec wasn't initialized. That code path also didn't fill the defaults.

drivers/storage/portworx/portworx.go
ggriffiths: sure, will leave empty

drivers/storage/portworx/portworx.go
ggriffiths: lets do that change in a separate PR, feel free to open a bug to track that.

drivers/storage/portworx/components_test.go
ggriffiths: Yes, and I added a test for k8s 1.17 that checks for v3.

piyush-nimbalkar (approved):
LGTM
Created a ticket to address the duplicate CSI flags - https://portworx.atlassian.net/browse/OPERATOR-551. Let's do this as part of 1.8.0 too.

drivers/storage/portworx/components_test.go
```diff
@@ -6018,8 +6098,10 @@ func TestPrometheusUpgradeDefaultDesiredImages(t *testing.T) {
 		"prometheusrules.monitoring.coreos.com",
 		"servicemonitors.monitoring.coreos.com",
 		"thanosrulers.monitoring.coreos.com",
+		"volumesnapshots.snapshot.storage.k8s.io",
+		"volumesnapshotclasses.snapshot.storage.k8s.io",
+		"volumesnapshotcontents.snapshot.storage.k8s.io",
```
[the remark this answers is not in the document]
ggriffiths: This was needed to make this test pass, since CSI is enabled by default.

PR #470, opened 2021-12-08
[PTX-4229] Test deploy StorageCluster in custom namespace and custom start port

pure-jliao:

Closes #


test/integration_test/basic_test.go
```diff
@@ -42,6 +42,18 @@ var testStorageClusterBasicCases = []types.TestCase{
...
+		TestName:        "InstallInCustomNamespaceWithShiftedPort",
+		TestrailCaseIDs: []string{"C52411", "C52430", "C53572"},
+		TestSpec: ci_utils.CreateStorageClusterTestSpecFunc(&corev1.StorageCluster{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "custom-ns-shifted-port",
+				Namespace: "custom-namespace",
+			},
+			Spec: corev1.StorageClusterSpec{StartPort: func(val uint32) *uint32 { return &val }(17001)},
+		}),
+		TestFunc: BasicInstall,
```
piyush-nimbalkar, line 55: We need to validate that the [portworx proxy component](https://github.com/libopenstorage/operator/blob/master/drivers/storage/portworx/component/portworx_proxy.go) is created when custom port and namespaces are used. We can add that validation in the ValidateStorageCluster function itself.

test/integration_test/utils/k8s.go
```diff
@@ -79,6 +79,9 @@ func CreateObjects(objects []runtime.Object) error {
 		var err error
 		if ns, ok := obj.(*v1.Namespace); ok {
 			_, err = coreops.Instance().CreateNamespace(ns)
+			if errors.IsAlreadyExists(err) {
+				err = nil
+			}
```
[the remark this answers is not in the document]
pure-jliao: this is mainly for running test manually on the same cluster if the previous one failed, in which case the ns may not be deleted, so we don't need to do more manual work to delete the ns. once allowing to run test cases even previous one failed, we will have a good way to clean everything up before running new cases.

test/integration_test/utils/storagecluster.go
```diff
@@ -165,6 +179,18 @@ func UninstallAndValidateStorageCluster(cluster *corev1.StorageCluster, t *testi
...
+	// Delete namespace if not kube-system
+	if cluster.Namespace != PxNamespace {
+		ns := &v1.Namespace{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: cluster.Namespace,
+			},
+		}
+		err = DeleteObjects([]runtime.Object{ns})
+		require.NoError(t, err)
+		ValidateObjectsAreTerminated([]runtime.Object{ns}, false)
```
piyush-nimbalkar, line 192: Check the error here too.

pkg/util/test/util.go
```diff
@@ -412,6 +414,17 @@ func ValidateStorageCluster(
 		return err
 	}
 
+	// If portworx proxy is enabled, validate the portworx-proxy DaemonSet
+	if isPortworxProxyEnabled(liveCluster) {
```
piyush-nimbalkar, line 418: - We can move this to validateComponents just like the other components. 
- Call it ValidatePortworxProxy and validate both it is present and absent if enabled or disabled respectively. 
- Check for portworx-service, clusterrolebinding, service account, etc.

Look at stork, autopilot, csi validations for reference. Let's keep it consistent.

test/integration_test/utils/k8s.go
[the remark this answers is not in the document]
piyush-nimbalkar: If this is about manual failures, we can just put a small script to clean up things. We shouldn't change the test code to deal with manual failure runs. The migration test cases are dependent on the namespace to be completely fresh, so would some other test cases in future.

pkg/util/test/util.go
```diff
@@ -1043,6 +1069,76 @@ func ValidateAutopilot(pxImageList map[string]string, cluster *corev1.StorageClu
...
+		}
+		_, err = coreops.Instance().GetService(pxService.Name, pxService.Namespace)
+		if errors.IsNotFound(err) {
+			return fmt.Errorf("failed to validate Service %s, Err: %v", pxService.Name, err)
+		}
+
+		// Validate Portworx proxy DaemonSet
+		if err := validateDaemonSet(proxyDs, timeout); err != nil {
+			return err
+		}
```
piyush-nimbalkar, line 1109: Let's move this validation before the ServiceAccount validation. In case we call ValidatePortworxProxy directly, it might fail because the operator might not have started creating the proxy components. If we call validateDaemonSet before, it will wait for the daemonset to be ready. At that point the service account and other components would be created for sure.

piyush-nimbalkar (approved):

PR #471, opened 2021-12-08
OPERATOR-519 Mount /var/lib/osd directly only for px > 2.9.1

piyush-nimbalkar:
oci-monitor 2.9.1 has a fix to allow /var/lib/osd mounted as a shared directory,
when the container runtime is cri-o.
Without the oci monitor change, mounting /var/lib/osd is not very useful as it
is not getting mounted as a shared mount. So keeping the old behavior for older
px releases, while mounting /var/lib/osd for 2.9.1 and above.

Closes #

### DIFF
--- a/drivers/storage/portworx/deployment.go
+++ b/drivers/storage/portworx/deployment.go
@@ -59,48 +59,6 @@ type pksVolumeInfo struct {
 }
 
 var (
-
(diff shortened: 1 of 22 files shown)


zoxpx (approved):
Thanks Piyush

PR #478, opened 2021-12-14
OPERATOR-519 Using /var/lib/osd mount even for PKS

piyush-nimbalkar:
- For backward compatibility using the `/var/lib/osd` host path even for PKS.
- Continuing to mount `/var/lib/osd/log` from the `/var/vcap/store/lib/osd/log` directory like before.
- As Zoran suggested, mounting `/var/lib/osd` before `/var/lib/osd/log` to keep the old behavior.

Closes # https://portworx.atlassian.net/browse/OPERATOR-519


zoxpx (approved):
I think this looks OK -- thanks Piyush

PR #479, opened 2021-12-15
OPERATOR-540 Run metrics collector only for px 2.9.1 and above

piyush-nimbalkar:
This ensures that existing customers don't start seeing metrics
collector on their px upgrade.

Closes # https://portworx.atlassian.net/browse/OPERATOR-540

### DIFF
--- a/drivers/storage/portworx/components_test.go
+++ b/drivers/storage/portworx/components_test.go
@@ -6470,7 +6470,7 @@ func TestCompleteInstallWithCustomRegistryChange(t *testing.T) {
 			},
 		},
 		Spec: corev1.StorageClusterSpec{
-			Image:               "portworx/image:2.8",
+			Image:               "portworx/image:2.9.1",
 			CustomImageRegistry: customRegistry,
 			StartPort:           &startPort,
 			UserInterface: &corev1.UserInterfaceSpec{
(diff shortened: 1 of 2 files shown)


drivers/storage/portworx/components_test.go
```diff
@@ -11616,6 +11617,71 @@ func TestTelemetryEnableAndDisable(t *testing.T) {
 	require.True(t, errors.IsNotFound(err))
 }
 
+func TestTelemetryIsDisabledForOldPxVersions(t *testing.T) {
```
pure-jliao, line 11620: nit: TestMetricsCollectorIsDisabled?

drivers/storage/portworx/components_test.go
```diff
@@ -11616,6 +11617,71 @@ func TestTelemetryEnableAndDisable(t *testing.T) {
...
+			ClusterUID: "test-clusteruid",
+		},
+	}
+
+	driver.SetDefaultsOnStorageCluster(cluster)
+
+	err := driver.PreInstall(cluster)
+	require.NoError(t, err)
+
+	// Now disable telemetry
```
pure-jliao, line 11654: why do we need to disable first? should we just check those metrics collector components don't exist directly?

piyush-nimbalkar: Good catch. Both of the above are just copy paste mistakes done in hurry.
Thanks for looking thoroughly!

pure-jliao (approved):
LGTM!

PR #480, opened 2021-12-21
[OPERATOR-533] pass PX_SHARED_SECRET to the autopilot when security i…

haibinxie:
…s enabled

Closes #

### DIFF
--- a/drivers/storage/portworx/portworx.go
+++ b/drivers/storage/portworx/portworx.go
@@ -114,8 +114,8 @@ func (p *portworx) GetStorkEnvMap(cluster *corev1.StorageCluster) map[string]*v1
 	}
 
 	if pxutil.AuthEnabled(&cluster.Spec) {
-		envMap[pxutil.EnvKeyStorkPXSharedSecret] = &v1.EnvVar{
-			Name: pxutil.EnvKeyStorkPXSharedSecret,
+		envMap[pxutil.EnvKeyPXSharedSecret] = &v1.EnvVar{
+			Name: pxutil.EnvKeyPXSharedSecret,
 			ValueFrom: &v1.EnvVarSource{
 				SecretKeyRef: &v1.SecretKeySelector{
 					LocalObjectReference: v1.LocalObjectReference{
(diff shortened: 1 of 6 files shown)


drivers/storage/portworx/util/util.go
```diff
@@ -148,6 +148,8 @@ const (
 	EnvKeyPortworxAuthJwtIssuer = "PORTWORX_AUTH_JWT_ISSUER"
 	// EnvKeyPortworxAuthSystemAppsKey is an environment variable defining the PX Security shared secret for Portworx Apps
 	EnvKeyPortworxAuthSystemAppsKey = "PORTWORX_AUTH_SYSTEM_APPS_KEY"
+	// EnvKeyAutopilotPXSharedSecret is an environment variable defining the shared secret for Autopilot
+	EnvKeyAutopilotPXSharedSecret = "PX_SHARED_SECRET"
```
piyush-nimbalkar, line 152: Can we rename the constant name for stork to something common, instead of having two different variables for stork and autopilot?

drivers/storage/portworx/portworx.go
```diff
@@ -800,6 +801,36 @@ func setDefaultAutopilotProviders(
 	}
 }
 
+func setDefaultAutopilotSecret(cluster *corev1.StorageCluster) {
```
piyush-nimbalkar, line 804: We don't need to add the env to the storagecluster. We should directly add it to the autopilot deployment. This should be in the autopilot component when creating the deployment.

haibinxie: i think either way works, but this becomes storagecluster defaults so we can track from storagecluster, it would be confusing if we see storagecluster env is different from deployment env.

piyush-nimbalkar: There are lot of other defaults args, envs, etc that we pass to different components but we don't always reflect them back to the StorageCluster. Ideally StorageCluster should just reflect what the user has added/modified. We should not be populating all our defaults in it if we can avoid it.

haibinxie: there are already many properties set to default, looks like they are mixed, let's discuss what should, what should not and why.

piyush-nimbalkar: Yes, unfortunately it is mixed. I think we need a clear definition on what should be exposed to user and what we should not. Ideally we should not populate defaults as much as possible. For the existing fields, there were reasons for some of them to add back to storagecluster. 
At least in this case we should not expose it to the user in the storagecluster because it is internal. When security is enabled, the user should just see the security section and should have to worry about envs in other components.

haibinxie: resolved

piyush-nimbalkar (approved):

PR #482, opened 2022-01-04
[OPERATOR-425] storage node label validation

haibinxie:

Closes #

### DIFF
--- a/pkg/controller/storagecluster/controller_test.go
+++ b/pkg/controller/storagecluster/controller_test.go
@@ -433,6 +433,9 @@ func TestCloudStorageLabelSelector(t *testing.T) {
 			CloudStorage: getCloudStorageSpec("type2"),
 		},
 	}
+
+	driver.EXPECT().GetSelectorLabels().Return(nil).AnyTimes()
+	driver.EXPECT().String().Return("mockDriverName").AnyTimes()
 	err := controller.validateCloudStorageLabelKey(cluster)
 	require.Error(t, err)
 
@@ -825,6 +828,8 @@ func TestReconcileWithDaemonSet(t *testing.T) {
 		Driver:            driver,
 	}
 
+	driver.EXPECT().GetSelectorLabels().Return(nil).AnyTimes()
+	driver.EXPECT().String().Return("mockDriverName").AnyTimes()
 	driver.EXPECT().Validate().Return(fmt.Errorf("daemonset is present"))
 
 	request := reconcile.Request{
--- a/pkg/controller/storagecluster/controller_test.go
+++ b/pkg/controller/storagecluster/controller_test.go
@@ -416,6 +416,21 @@ func TestCloudStorageLabelSelector(t *testing.T) {
 		}
 	}
 
+	// NodeName selector
+	cluster.Spec.Nodes = []corev1.NodeSpec{
+		{
+			Selector: corev1.NodeSelector{
+				NodeName: "Test",
+			},
+			CloudStorage: getCloudStorageSpec("type1"),
+		},
+	}
+
+	driver.EXPECT().GetSelectorLabels().Return(nil).AnyTimes()
+	driver.EXPECT().String().Return("mockDriverName").AnyTimes()
+	err := controller.validateCloudStorageLabelKey(cluster)
+	require.Error(t, err)
+
 	// Empty selector
 	cluster.Spec.Nodes = []corev1.NodeSpec{
 		{
@@ -436,7 +451,7 @@ func TestCloudStorageLabelSelector(t *testing.T) {
 
 	driver.EXPECT().GetSelectorLabels().Return(nil).AnyTimes()
 	driver.EXPECT().String().Return("mockDriverName").AnyTimes()
-	err := controller.validateCloudStorageLabelKey(cluster)
+	err = controller.validateCloudStorageLabelKey(cluster)
 	require.Error(t, err)
 
 	// Two keys in one selector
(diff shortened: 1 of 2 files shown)


pkg/controller/storagecluster/storagecluster.go
```diff
@@ -334,6 +339,78 @@ func (c *Controller) validateCustomAnnotations(current *corev1.StorageCluster) e
...
+	first := false
+	key := ""
+	for _, node := range cluster.Spec.Nodes {
+		if node.CloudStorage == nil {
+			continue
+		}
+
+		keys := getKeysFromNodeSelector(&node.Selector)
+		if len(keys) != 1 {
+			return "", fmt.Errorf("it's required to have 1 key in cloud storage node label selector, key(s) %v", keys)
```
piyush-nimbalkar, line 395: * to have __only__ 1 key
This makes it clear that only 1 key is allowed

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -334,6 +339,78 @@ func (c *Controller) validateCustomAnnotations(current *corev1.StorageCluster) e
...
+		}
+
+		keyTemp := ""
+		for k, _ := range keys {
+			keyTemp = k
+		}
+
+		if !first {
```
piyush-nimbalkar, line 403: nit: We could just check `key == ""` instead of the `first` boolean, right?

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -334,6 +339,78 @@ func (c *Controller) validateCustomAnnotations(current *corev1.StorageCluster) e
...
+	key, err := c.getCloudStorageLabelKey(cluster)
+	if err != nil {
+		return err
+	}
+
+	if cluster.Spec.CloudStorage != nil &&
+		cluster.Spec.CloudStorage.NodePoolLabel != "" &&
+		key != "" &&
+		cluster.Spec.CloudStorage.NodePoolLabel != key {
+		return fmt.Errorf("node pool label key incorrect, expected %s, actual %s", cluster.Spec.CloudStorage.NodePoolLabel, key)
```
piyush-nimbalkar, line 375: Shouldn't `key` be the expected and `Spec.CloudStorage.NodePoolLabel` be the actual?

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -334,6 +339,78 @@ func (c *Controller) validateCustomAnnotations(current *corev1.StorageCluster) e
...
 
+
+
+func getKeysFromNodeSelector(ns *corev1.NodeSelector) map[string]bool {
+	keys := make(map[string]bool)
+
+	// Use node name as selector.
+	if ns.NodeName != "" {
+		keys["NodeName"] = true
+	}
```
piyush-nimbalkar, line 350: If cloud storage is not empty, then we should not even allow `nodeName` selector. The problem would be that if different nodes define different cloud storage configurations, then we won't be able to honor them on restarts. Same as different label selector key problem we are trying to solve/validate.

pkg/controller/storagecluster/controller_test.go
```diff
@@ -395,6 +395,205 @@ func TestSingleClusterValidation(t *testing.T) {
...
+				},
+			},
+			CloudStorage: getCloudStorageSpec("type2"),
+		},
+	}
+	err = controller.validateCloudStorageLabelKey(cluster)
+	require.Error(t, err)
+
+	// Different keys in selectors
+	cluster.Spec.Nodes = []corev1.NodeSpec {
```
piyush-nimbalkar, line 507: Let's add a case where one of the nodes section does not have cloud storage.

piyush-nimbalkar (commented):
Overall looks good. 
Just need to handle the case of node name selector.

piyush-nimbalkar:
Take a look at the travis build as well. I think formatting is broken in few places.

pkg/controller/storagecluster/storagecluster.go
haibinxie: correct

pkg/controller/storagecluster/storagecluster.go
haibinxie: if nodePoolLabel is configured, then it should be used and expected. well depends on how we understand it, either one can be source of truth

pkg/controller/storagecluster/storagecluster.go
haibinxie: "different nodes define different cloud storage configurations, then we won't be able to honor them on restarts", the node name does not change upon restarts and the cloud storage configuration should be honored right?

piyush-nimbalkar: The issue here is how clouddrive is implemented. Currently the entire logic is in the px layer. The operator sends the params to the px container, but when failover of clouddrive happens in the background, the original spec may not be honored. This could be confusing to the user. That is why we are adding these validations. We can talk offline if you need more context.

pkg/controller/storagecluster/storagecluster.go
piyush-nimbalkar: hmm.. that is true
Still I feel we should call `key` as expected so it tells the user that the `cloudStorage.NodePoolLabel` is the user defined key and it should match the expected value in `key`. Users can only modify that and not the 'expected'. Wdyt?

haibinxie:
@piyush-nimbalkar I have resolved the comment, this may cause existing storagecluster to fail validation after upgrading operator, this could be bad to users, and it could be very tricky to update storagecluster if users run into this. let's discuss the impact before merging.

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -334,6 +339,69 @@ func (c *Controller) validateCustomAnnotations(current *corev1.StorageCluster) e
...
+		return "", nil
+	}
+
+	key := ""
+	for _, node := range cluster.Spec.Nodes {
+		if node.CloudStorage == nil {
+			continue
+		}
+
+		keys := getKeysFromNodeSelector(&node.Selector)
```
piyush-nimbalkar, line 385: So, in case of nodeName selector, we are going to not enforce the validation?

haibinxie: i will add it back later. thinking of only enforcing the validation on fresh installation, the problem is the first time we know status is empty it is fresh installation we can fail the validation then update the status in storagecluster, but in second reconcile loop the status is updated so we no longer know it is fresh installation.

haibinxie: resolved. also add check to skip validation on existing cluster.

haibinxie:
@piyush-nimbalkar please take a final look, i added a change to skip validation is storage pod exists.

piyush-nimbalkar (approved):
Can you add tests for the two cases you added? 
- node name used with cloudstorage
- storage pods present and see that the validation is skipped

haibinxie:
> Can you add tests for the two cases you added?
> 
> * node name used with cloudstorage
> * storage pods present and see that the validation is skipped

@piyush-nimbalkar added in a new commit. will merge once the build passes.

PR #483, opened 2022-01-06
OPERATOR-398: Adding px-repo integration test

zoxpx:
* adding negative (failed install w/o px-repo) + positive (fixed install w/ px-repo) tests
* adding `ValidateStorageClusterFailedEvents` that can validate "stuck" installation and check K8s events
* modified `validateStorageClusterPods` to take external "POD readyness" eval function

* this is an integration test for `px-repo` component  (will be productized w/ px-2.10.x)

Closes # OPERATOR-398

### DIFF
--- a/pkg/util/test/util.go
+++ b/pkg/util/test/util.go
@@ -403,7 +403,10 @@ func ValidateStorageCluster(
 	}
 
 	// Validate Portworx pods
-	if err = validateStorageClusterPods(clusterSpec, expectedPxNodeNameList, timeout, interval); err != nil {
+	podTestFn := func(pod v1.Pod) bool {
+		return coreops.Instance().IsPodReady(pod)
+	}
+	if err = validateStorageClusterPods(clusterSpec, expectedPxNodeNameList, timeout, interval, podTestFn); err != nil {
 		return err
 	}
 
@@ -651,10 +654,13 @@ func ValidateUninstallStorageCluster(
 	return nil
 }
 
+type podTestFnType func(pod v1.Pod) bool
+
 func validateStorageClusterPods(
 	clusterSpec *corev1.StorageCluster,
 	expectedPxNodeNameList []string,
 	timeout, interval time.Duration,
+	podTestFn podTestFnType,
 ) error {
 	t := func() (interface{}, bool, error) {
 		cluster, err := operatorops.Instance().GetStorageCluster(clusterSpec.Name, clusterSpec.Namespace)
@@ -676,7 +682,8 @@ func validateStorageClusterPods(
 		var podsNotReady []string
 		var podsReady []string
 		for _, pod := range pods {
-			if !coreops.Instance().IsPodReady(pod) {
+			// if test-function fails, POD is considered as "not ready"
+			if !podTestFn(pod) {
 				podsNotReady = append(podsNotReady, pod.Name)
 			}
 			pxNodeNameList = append(pxNodeNameList, pod.Spec.NodeName)
@@ -738,6 +745,7 @@ func defaultPxNodeAffinityRules(runOnMaster bool) *v1.NodeAffinity {
 func validatePortworxNodes(cluster *corev1.StorageCluster, expectedNodes int) error {
 	conn, err := getSdkConnection(cluster)
 	if err != nil {
+		// CHECKME -- shouldn't we return err ?
 		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


zoxpx:
Note, we're digging into ...
```
# kubectl -n kube-system get events --field-selector reason=FileSystemDependency
kube-system  29m    Warning  FileSystemDependency  pod/px-repo-test-negative-bpqqc  Failed to load PX filesystem dependencies for kernel 4.15.0-162-generic
kube-system  8m20s  Warning  FileSystemDependency  pod/px-repo-test-negative-dpxgz  Failed to load PX filesystem dependencies for kernel 4.15.0-162-generic
kube-system  29m    Warning  FileSystemDependency  pod/px-repo-test-negative-kxk7t  Failed to load PX filesystem dependencies for kernel 4.15.0-162-generic
```

pkg/util/test/util.go
```diff
@@ -674,7 +680,8 @@ func validateStorageClusterPods(
 		var podsNotReady []string
 		var podsReady []string
 		for _, pod := range pods {
-			if !coreops.Instance().IsPodReady(pod) {
+			// if test-function fails, POD is considered as "not ready"
+			if !podTestFn(pod) {
```
haibinxie, line 684: nit: why not call coreops.Instance().IsPodReady(pod) here.

test/integration_test/pxrepo_test.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/require"
+	k8sv1 "k8s.io/api/core/v1"
+)
+
+var (
+	pxVer2_10, _ = version.NewVersion("2.10")
+)
+
+var zoxTestCases = []types.TestCase{
```
haibinxie, line 25: zoxTestCases --> pxRepoTestCases

test/integration_test/pxrepo_test.go
```diff
@@ -0,0 +1,129 @@
...
+			cluster.Spec.PxRepo = &corev1.PxRepoSpec{
+				Enabled: true,
+				Image:   "docker.io/portworx/px-repo:latest",
+			}
+			cluster.Spec.DeleteStrategy = &corev1.StorageClusterDeleteStrategy{
+				Type: corev1.UninstallAndWipeStorageClusterStrategyType,
+			}
+			return cluster
+		},
+		TestFunc: BasicInstall,
```
haibinxie, line 117: In jenkins the test case would run on non airgapped env, could you add validation that modules are installed through px-repo not public?

pkg/util/test/util.go
```diff
@@ -728,6 +735,7 @@ func defaultPxNodeAffinityRules() *v1.NodeAffinity {
 func validatePortworxNodes(cluster *corev1.StorageCluster, expectedNodes int) error {
 	conn, err := getSdkConnection(cluster)
 	if err != nil {
+		// CHECKME -- shouldn't we return err ?
```
haibinxie, line 738: There is probably a lower version of px that does not have the support. Even in that case, the error should be checked before return nil

test/integration_test/pxrepo_test.go
zoxpx: Thanks -- updated in 6905a2d

test/integration_test/pxrepo_test.go
zoxpx: This is OK -- the tests are using internal args to "emulate" the `air-gapped` env:
* `env TEST_SKIP_EXTRACT_PX_MODULE=true` -- this skips unpacking the internal KO-archive embedded in `px-enterprise` container
* `-air-gapped` flag added -- this prevents any downloads from mirrors.portworx.com or yum.portworx.com  (e.g. remote download and install of KO-modules and kernel-headers)

test/integration_test/pxrepo_test.go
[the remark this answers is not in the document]
zoxpx: Done in 6905a2d
(note, it used to be worse -- the test-filename itself was called `zox_test.go`  ;)  -- forgot to clean up the test-names before the commit [sorry])

zoxpx:
Re-requesting the review

test/integration_test/pxrepo_test.go
haibinxie: how do we know if px-repo is used? my theory for a happy path validation is running with a px version that does not work, with a wrong px-repo image px wont run, with a correct px-repo image px runs correctly.

zoxpx: The first (negative) test must fail with `Failed to load PX filesystem dependencies for kernel XXX` error -- the second test should pass, after we added the px-repo into the install.

haibinxie (approved):

zoxpx:
Thanks Michael -- merging the PR

PR #486, opened 2022-01-10
OPERATOR-517: fetch manifest from proxy

haibinxie:

Closes #

### DIFF
--- a/drivers/storage/portworx/manifest/common.go
+++ b/drivers/storage/portworx/manifest/common.go
@@ -3,8 +3,14 @@ package manifest
 import (
 	"io/ioutil"
 	"net/http"
+	"net/url"
+	"strings"
 
-	yaml "gopkg.in/yaml.v2"
+	"gopkg.in/yaml.v2"
+)
+
+const (
+	httpPrefix = "http://"
 )
 
 // Methods to override for testing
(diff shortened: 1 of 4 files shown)


drivers/storage/portworx/manifest/common.go
```diff
@@ -3,20 +3,54 @@ package manifest
...
+	if proxy == "" {
+		resp, err = httpGet(manifestURL)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		// Add the http prefix so we can parse the URL:
+		// Wrong format: url.Parse("127.0.0.1:3213")
+		// Correct format: url.Parse("http://127.0.0.1:3213")
+		if !strings.HasPrefix(strings.ToLower(proxy), httpPrefix) {
```
piyush-nimbalkar, line 33: Shouldn't we look for https proxy as well?

drivers/storage/portworx/manifest/remote_test.go
```diff
@@ -8,11 +8,41 @@ import (
...
 
 	version "github.com/hashicorp/go-version"
+	"github.com/libopenstorage/operator/drivers/storage/portworx/util"
 	corev1 "github.com/libopenstorage/operator/pkg/apis/core/v1"
 	"github.com/stretchr/testify/require"
 	"k8s.io/api/core/v1"
 )
 
+func TestRemoteManifestWithProxy(t *testing.T) {
+	// Do not run the test by default. Change httpProxy below to a working proxy.
```
piyush-nimbalkar, line 18: For unit tests, we want it to run on any environment. Let's test the proxy only in functional tests. 
For this unit test, see if you can test the behavior. May be check if http client is called with the proxy. 
If it is getting too complicated, we can skip the unit test.

piyush-nimbalkar:
@haibinxie, sorry for the delay in review.

drivers/storage/portworx/manifest/remote.go
```diff
@@ -65,7 +65,7 @@ func (m *remote) downloadVersionManifest(
 	params.Add("kbver", m.k8sVersion.String())
 	u.RawQuery = params.Encode()
 
-	body, err := getManifestFromURL(u.String())
+	body, err := getManifestFromURL(u.String(), pxutil.GetPxProxyEnvVarValue(m.cluster))
```
pure-jliao, line 68: this function can return https first if exist, we should handle https when parsing the url as well

pure-jliao (commented):
one comments regarding https proxy

drivers/storage/portworx/manifest/remote_test.go
haibinxie: this unit test works and talks to real proxy, but I just dont enable it by default as the proxy is setup manually.

piyush-nimbalkar: I see, so you are keeping it to test this manually only. We won't be running it any environment, correct?

haibinxie: Yes

drivers/storage/portworx/manifest/remote.go
haibinxie: this string is later used by url.Parse which works for https. if that's where you meant to handle

drivers/storage/portworx/manifest/common.go
haibinxie: good point, resolved in a new commit. pls take another look

pure-jliao (approved):
LGTM

piyush-nimbalkar (approved):

PR #487, opened 2022-01-11
PTX-4833 Temporary comment out DeploymentDeepEqual validation for Telemetry

nikolaypopov:
Temporary comment out DeploymentDeepEqual validation for Telemetry. I have tried using it with just object, but it didn't work for me.

NOTE: This is not working for spawn, also it is not P1. Let's get P1 working in Operator Integration Tests and Spawn as well, because most of our testing happens in Spawn/Torpedo, we still need that to work until we implement Torpedo and all the Clouds/Environment in the Operator Integration Test suite.

P.S. I left all that code to work on it in future when I have more time available.


pkg/util/test/util.go
```diff
@@ -1829,6 +1838,197 @@ func ValidateTelemetryInstalled(pxImageList map[string]string, cluster *corev1.S
 	return nil
 }
 
+/* TODO: We need to make this work for spawn
+func createMetrixCollectorDeploymentObject(cluster *corev1.StorageCluster, pxImageList map[string]string) (*appsv1.Deployment, error) {
```
piyush-nimbalkar, line 1842: I believe we don't need this. Ideally we should be reading a spec from the file and changing the individual components as needed by the test in future. Should we just remove this?

piyush-nimbalkar (approved):
LGTM.
Let's unblock the spawn tests for now.

pkg/util/test/util.go
```diff
@@ -1755,14 +1755,23 @@ func ValidateTelemetryInstalled(pxImageList map[string]string, cluster *corev1.S
 		return err
 	}
 
-	expectedDeployment := GetExpectedDeployment(&testing.T{}, "metricsCollectorDeployment.yaml")
+	// TODO: We need to make this work for spawn
+	//expectedDeployment := GetExpectedDeployment(&testing.T{}, "metricsCollectorDeployment.yaml") // Original yaml to object way
+	//expectedDeployment, err := createMetrixCollectorDeploymentObject(cluster, pxImageList) // Object way, just creating object straight away
```
haibinxie, line 1760: i am ok with unblocking first, but i would rather keep the code  and for validation just validate the name and namespace etc, whatever that passes validation.

nikolaypopov:
Guys I removed my commented out code and I just commented out `GetExpectedDeployment()` and `DeploymentDeepEqual()` until we can think of a way that can work for Spawn as well. Please take a look and I'm going to merge this so I can pull this change in Spawn

pkg/util/test/util.go
nikolaypopov: All the other validations still exist, i just commented out `GetExpectedDeployment()` and `DeploymentDeepEqual()`

pkg/util/test/util.go
nikolaypopov: Done, removed

piyush-nimbalkar (approved):

PR #491, opened 2022-01-21
OPERATOR-301 Do not remove portworx-service

piyush-nimbalkar:
- When running in storage disabled mode, which is mostly used by
PX Central, the operator used to delete the portworx-service
created by the daemonset spec. We specially handle this by ignoring
the deletion of the service when storage is disabled.
- Ignore the GetStorageNodes call when storage is disabled.

Closes # https://portworx.atlassian.net/browse/OPERATOR-301


pure-jliao (approved):
LGTM, one question is do we only have this portworx service deletion in portworx_proxy?

piyush-nimbalkar:
> LGTM, one question is do we only have this portworx service deletion in portworx_proxy?

That's a good question. We do have the deletion in the portworx_basic component as well. The only difference there is that we only delete it from the StorageCluster's namespace and if it is owned by the StorageCluster. Even in the case where px-central deploys the StorageCluster in the same namespace as the old portworx daemonset, as we don't own the old portworx-service object it won't get deleted.

PR #495, opened 2022-01-27
OPERATOR-559 Migrate only if the annotation is set to true

piyush-nimbalkar:

Closes # https://portworx.atlassian.net/browse/OPERATOR-559

### DIFF
--- a/pkg/controller/storagecluster/storagecluster.go
+++ b/pkg/controller/storagecluster/storagecluster.go
@@ -248,8 +248,21 @@ func (c *Controller) Reconcile(ctx context.Context, request reconcile.Request) (
 }
 
 func (c *Controller) waitingForMigrationApproval(cluster *corev1.StorageCluster) bool {
-	_, migrationLabelPresent := cluster.Annotations[constants.AnnotationMigrationApproved]
-	return (migrationLabelPresent && cluster.Status.Phase == "") ||
+	label, migrationLabelPresent := cluster.Annotations[constants.AnnotationMigrationApproved]
+	if !migrationLabelPresent {
+		return false
+	}
+
+	approved, err := strconv.ParseBool(label)
+	if err != nil {
+		// Wait until valid boolean value is provided
+		return true
+	}
+
+	// Even if the migration is approved, wait for the status to come out of initial
+	// or waiting migration approval state. This gives time to the migration routine,
+	// to prepare the cluster for migration after the approval.
+	return !approved || cluster.Status.Phase == "" ||
 		cluster.Status.Phase == constants.PhaseAwaitingApproval
 }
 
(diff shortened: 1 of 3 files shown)


pkg/controller/storagecluster/storagecluster.go
```diff
@@ -248,8 +248,21 @@ func (c *Controller) Reconcile(ctx context.Context, request reconcile.Request) (
 }
 
 func (c *Controller) waitingForMigrationApproval(cluster *corev1.StorageCluster) bool {
```
haibinxie, line 250: use different event message when annotation is approved but status is still waitingForApproval. the time window for it to get into migrationInProgress may be long. you can return two bool to indicate the status.

piyush-nimbalkar: The time window is not long actually. We just label the node and then start the migration immediately. Will seem like an overkill in this case.

haibinxie: Backup should also be added prior to it later, and other stuff like validating if migration is possible or not. 

if user deploys op 1.7.0 will migration be triggered?

piyush-nimbalkar: The backup can happen after the status has been updated to InProgress and the node labels have been added. I guess we won't have to block the operator reconcile loop as long as the node labels have been updated. 
The validation happens when the StorageCluster is first created, which is when op 1.7.0 is installed.

haibinxie: Regarding op 1.7, the scenario i want to confirm is if user accident deploy op 1.7 with Daemonset, it should not start migration, is it the current behavior?

piyush-nimbalkar: The current behavior is that it will create the StorageCluster object but will wait in pending migration state. The user has to change the annotation to 'true' to trigger the actual migration.

piyush-nimbalkar: Now I am feeling if we should add a flag to the operator to disable the migration feature by default in 1.7.0. Wdyt?

dsupure: I incline to disable it by default.

haibinxie: a flag would be nice, or we can just disable by default, and when the feature is ready to release enable by default

piyush-nimbalkar: Done. Added a flag `--migration` which is set to false by default.

pure-jliao (approved):
LGTM! thanks!

haibinxie (approved):

PR #496, opened 2022-02-01
OPERATOR-554: change portworx pod dns policy to DNSClusterFirstWithHostNet bcz host…

haibinxie:
…network is true

Closes #

### DIFF
--- a/drivers/storage/portworx/deployment.go
+++ b/drivers/storage/portworx/deployment.go
@@ -261,6 +261,8 @@ func (p *portworx) GetStoragePodSpec(
 
 	containers := t.portworxContainer(cluster)
 	podSpec := v1.PodSpec{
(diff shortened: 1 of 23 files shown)


zoxpx (approved):
It would have been better to make the `dnsPolicy` configurable via the `StorageCluster` YAML in case we would need to modify it at customer's site -- but I'll take this change.

Thanks Michael

piyush-nimbalkar (approved):

piyush-nimbalkar:
@haibinxie I am surprised that no unit test failed. Can you check why existing tests do not catch this change and modify or add a new test for this?

haibinxie:
> @haibinxie I am surprised that no unit test failed. Can you check why existing tests do not catch this change and modify or add a new test for this?

dnspolicy was not compared in unit tests, i have added a commit to add it.

piyush-nimbalkar (approved):

PR #497, opened 2022-02-01
OPERATOR-499: add flag to set hostpid

haibinxie:

Closes #


zoxpx (approved):
Thanks Michael

piyush-nimbalkar (approved):
Can we add unit test for this too?

Note that changing this will not automatically bounce the px pods. The users will have to bounce the px pods one at at time. I guess it is okay for now as it is not a common flag.

haibinxie:
> Can we add unit test for this too?
> 
> Note that changing this will not automatically bounce the px pods. The users will have to bounce the px pods one at at time. I guess it is okay for now as it is not a common flag.

added

PR #502, opened 2022-02-07
Set image registry for ones that don't have it

nashant:
Adding the default docker.io registry to the start of the default of all
the images fixes a problem where GetImageURN swallows the image
namespacing

Remediates Pure portal ticket #01194532


nashant:
This one would have changed current behaviour

PR #505, opened 2022-02-08
Add preserveFullCustomRegistry to not swallow part of image

nashant:
This is to stop GetImageURN from swallowing part of the image tag when
setting a customImageRegistry with a / in it. Original behaviour is,
when a registry of `example.io/public` is used with an image of
`portworx/oci-monitor`, for the image to become
`emaple.io/public/oci-monitor`. Setting this to true will preserve the
full path and the image will be `example.io/public/portworx/oci-monitor`

Remediates Pure portal ticket #01194532

### DIFF
--- a/deploy/crds/core_v1_storagecluster_crd.yaml
+++ b/deploy/crds/core_v1_storagecluster_crd.yaml
@@ -88,6 +88,14 @@ spec:
                   Custom container image registry server that will be used instead of
                   index.docker.io to download Docker images. This may include the repository as well.
                   (Example: myregistry.net:5443 or myregistry.com/myrepository)
+              preserveFullCustomImageRegistry:
+                type: boolean
+                description: >-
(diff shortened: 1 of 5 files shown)


piyush-nimbalkar (approved):

piyush-nimbalkar:
Thank you for making the changes, @nashant!

nashant:
No problem whatsoever, that's the advantage of open source! When do you think you would be able to cut a release so we can crack on with our daemonset->operator migration?

piyush-nimbalkar:
We are planning to release 1.8.0 (with this change) around end of the next month. We will keep you posted through the internal channels.
Thanks again!

nashant:
Could you please cut a 1.7.1 release with just this commit? End of next month is no good to us, we're already well behind with migrating from daemonset to operator.

PR #511, opened 2022-02-12
[OPERATOR-543] Allow adding custom labels to portworx-api service

pure-jliao:

**What this PR does / why we need it**: To pass custom labels to portworx-api service

Closes # [OPERATOR-543](https://portworx.atlassian.net/browse/OPERATOR-543)

### DIFF
--- a/pkg/util/test/util.go
+++ b/pkg/util/test/util.go
@@ -420,6 +420,11 @@ func ValidateStorageCluster(
 		return err
 	}
 
+	// Validate Portworx API Service
+	if err = validatePortworxAPIService(liveCluster, timeout, interval); err != nil {
+		return err
+	}
+
 	if err = validateComponents(pxImageList, liveCluster, timeout, interval); err != nil {
 		return err
 	}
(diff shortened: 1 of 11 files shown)


pkg/util/test/util.go
```diff
@@ -784,6 +789,30 @@ func validatePortworxService(namespace string) error {
...
+		for k, expectedVal := range cluster.Spec.Metadata.Labels["service/portworx-api"] {
+			if actualVal, ok := service.Labels[k]; !ok || actualVal != expectedVal {
+				return fmt.Errorf("failed to validate Service %s/%s custom labels, label %s doesn't exist or value doesn't match", cluster.Namespace, pxAPIServiceName, k)
+			}
+		}
+		for k := range service.Labels {
+			if k == "name" {
+				continue
+			}
+			if _, ok := cluster.Spec.Metadata.Labels["service/portworx-api"]; !ok {
```
piyush-nimbalkar, line 808: Here we are only checking of the portworx-api labels exist or not in the metadata. We should also check if the actual label `k` is present or not in the metadata labels of portworx-api. Something like this - 
```
if metadataLabels, ok := cluster.Spec.Metadata.Labels["service/portworx-api"]; ok {
  if _, ok := metadataLabels[k]; ok {
    return fmt.Errorf("failed to validate Service %s/%s custom labels, found unexpected label %s", cluster.Namespace, pxAPIServiceName, k)
  }
} else {
  return fmt.Errorf("failed to validate Service %s/%s custom labels, found unexpected label %s", cluster.Namespace, pxAPIServiceName, k)
}
```

piyush-nimbalkar (approved):
This looks great! Feel free to merge after addressing the comment.

pkg/util/test/util.go
pure-jliao: dou you mean the validation from line 798 to 803? I have 2 loops of validations here.

piyush-nimbalkar: No, I do not mean that. I see the first loop is checking that all labels in storagecluster are added correctly to the service. While the second loop is checking that all the service labels are coming from the stc and stc only. Basically, we are validating the removal of labels.
In this `if` condition, we should be checking if the `key` from the service labels exists in the stc. What you are checking right now is just whether their are labels present in the stc for `service/portworx-api`.

PR #515, opened 2022-02-22
[OPERATOR-510] [OPERATOR-565] handle custom registry and component images during migration

haibinxie:


pkg/migration/migration_test.go
```diff
@@ -23,8 +17,211 @@ import (
...
+
+	testImageMigration(t, dsImages, expectedStcImages)
+
+	// Test with custom image registry. Component images do not match spec gen.
+	// StorageCluster should configure component image.
+	dsImages = ImageConfig{
+		PortworxImage: "test/oci-monitor:2.9.1",
+		Components: manifest.Release{
+			Stork:     "test/a/stork:2.7.0",
+			Autopilot: "test/b/autopilot:1.3.1",
```
piyush-nimbalkar, line 112: Can we also add a test where the versions don't match but the path is fine? This would be most common scenario with customer's spec.

pkg/migration/migration_test.go
```diff
@@ -23,8 +17,211 @@ import (
...
+	}
+
+	testImageMigration(t, dsImages, expectedStcImages)
+
+	// Test with custom image registry. Component images match spec gen.
+	// StorageCluster should not configure any component image.
+	dsImages = ImageConfig{
+		PortworxImage: "test/oci-monitor:2.9.1",
+		Components: manifest.Release{
+			Stork:     "test/openstorage/stork:2.7.0",
```
piyush-nimbalkar, line 88: Can we test for a longer path for custom image registry? Eg. `test/foo/portworx/oci-monitor:2.9.1`

pkg/migration/migration_test.go
```diff
@@ -23,8 +17,211 @@ import (
...
+	testutil "github.com/libopenstorage/operator/pkg/util/test"
 )
 
+type ImageConfig struct {
+	CustomImageRegistry string
+	PortworxImage       string
+	Components          manifest.Release
+}
+
+func TestImageMigration(t *testing.T) {
```
piyush-nimbalkar, line 36: Very nicely written test! 👌

pkg/migration/generate.go
```diff
@@ -392,6 +427,10 @@ func (h *Handler) addStorkSpec(cluster *corev1.StorageCluster) error {
 	cluster.Spec.Stork = &corev1.StorkSpec{
 		Enabled: dep != nil,
 	}
+
+	if dep != nil && cluster.Status.DesiredImages.Stork != dep.Spec.Template.Spec.Containers[0].Image {
+		cluster.Spec.Stork.Image = h.removeCustomImageRegistry(dep.Spec.Template.Spec.Containers[0].Image, cluster.Spec.CustomImageRegistry)
```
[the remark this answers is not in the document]
haibinxie: Thought about this during implementation, if customers use custom image for component which does not match px version, then they will likely still use custom image for component even after upgrading px version. but if customers use component images that match px version, we wont set this and they will get auto-updated when px version changes.

pkg/mock/storagedriver.mock.go
```diff
@@ -188,6 +190,24 @@ func (mr *MockDriverMockRecorder) PreInstall(arg0 interface{}) *gomock.Call {
 // SetDefaultsOnStorageCluster mocks base method.
 func (m *MockDriver) SetDefaultsOnStorageCluster(arg0 *v1.StorageCluster) {
 	m.ctrl.T.Helper()
+	arg0.Status.DesiredImages = &v1.ComponentImages{
```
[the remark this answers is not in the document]
haibinxie: we could change the behavior, if caller wants to use different versions, they can generate different component images based on px version in storage cluster.

pkg/migration/generate.go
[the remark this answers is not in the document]
piyush-nimbalkar: Most customers don't intentionally use custom images. Some just update the oci-mon image and will update the components whenever they need to. That creates the difference between what is recommended and what they have in the cluster.

piyush-nimbalkar:
@haibinxie btw we need to handle air-gapped customers as well. We need to think how and when we can create the config map for version map.

haibinxie:
@piyush-nimbalkar handled manifest configmap creation for airgapped env and resolved other comments as well. take another look when you get a chance.

pkg/migration/generate.go
```diff
@@ -358,7 +394,7 @@ func (h *Handler) constructStorageCluster(ds *appsv1.DaemonSet) *corev1.StorageC
 	}
 	cluster.Spec.Monitoring = &corev1.MonitoringSpec{
 		Telemetry: &corev1.TelemetrySpec{
-			Enabled: telemetryEnabled,
+		Enabled: telemetryEnabled,
```
piyush-nimbalkar, line 397: nit: Indentation missing

pkg/migration/generate.go
```diff
@@ -384,6 +420,19 @@ func (h *Handler) constructStorageCluster(ds *appsv1.DaemonSet) *corev1.StorageC
...
+	if image == "" || customImageRegistry == "" {
+		return image
+	}
+
+	if !strings.HasPrefix(image, customImageRegistry) {
+		logrus.Warningf("image %s does not have custom image registry prefix %s", image, customImageRegistry)
+		return image
+	}
+
+	return strings.TrimPrefix(image, customImageRegistry+"/")
```
piyush-nimbalkar, line 433: What would happen if the custom registry is `registry.io/custom-namespace`? 
The image in this case would be `registry.io/custom-namespace/stork:x.x.x`. If we trim it, then would just `stork:x.x.x` remain? It might still work, but just verify that it works with just the image portion.

pkg/migration/generate.go
```diff
@@ -447,6 +491,24 @@ func (h *Handler) addMonitoringSpec(cluster *corev1.StorageCluster) error {
...
+	// Check for telemetry
+	telemetryImage := ""
+	for _, container := range ds.Spec.Template.Spec.Containers {
+		if strings.Contains(container.Image, "ccm-service") {
+			telemetryImage = container.Image
+			break
+		}
+	}
+	if telemetryImage != "" {
+		cluster.Spec.Monitoring.Telemetry = &corev1.TelemetrySpec{
```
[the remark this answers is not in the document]
piyush-nimbalkar: Let's remove it from the other place now that we have added the telemetry spec here?

pkg/migration/generate.go
```diff
@@ -34,7 +37,11 @@ func (h *Handler) createStorageCluster(
 	if err := h.addPVCControllerSpec(stc); err != nil {
 		return nil, err
 	}
-	if err := h.addMonitoringSpec(stc); err != nil {
+	if err := h.addMonitoringSpec(stc, ds); err != nil {
```
piyush-nimbalkar, line 40: We need to populate the CSI, prometheus and alert manager images too.

pkg/migration/migration_test.go
```diff
@@ -23,8 +17,240 @@ import (
...
+	cluster := &corev1.StorageCluster{}
+	err := testutil.Get(k8sClient, cluster, clusterName, ds.Namespace)
+	require.NoError(t, err)
+	require.Equal(t, expectedStcImages.PortworxImage, cluster.Spec.Image)
+	require.Equal(t, expectedStcImages.CustomImageRegistry, cluster.Spec.CustomImageRegistry)
+	require.Equal(t, expectedStcImages.Components.Stork, cluster.Status.DesiredImages.Stork)
+	require.Equal(t, expectedStcImages.Components.Autopilot, cluster.Status.DesiredImages.Autopilot)
+	require.Equal(t, expectedStcImages.Components.Telemetry, cluster.Status.DesiredImages.Telemetry)
+
+	cm := &v1.ConfigMap{}
```
piyush-nimbalkar, line 245: We should check for non-airgapped envs as well to verify that we don't create a configmap.

pkg/migration/migration_test.go
```diff
@@ -23,8 +17,240 @@ import (
...
+
+	mockController := gomock.NewController(t)
+	driver := testutil.MockDriver(mockController)
+	ctrl := &storagecluster.Controller{
+		Driver: driver,
+	}
+	ctrl.SetKubernetesClient(k8sClient)
+	mockManifest := mock.NewMockManifest(mockController)
+	manifest.SetInstance(mockManifest)
+	mockManifest.EXPECT().CanAccessRemoteManifest(gomock.Any()).AnyTimes()
```
piyush-nimbalkar, line 226: Can we explicitly return false here and in all the following tests?

pkg/migration/migration_test.go
piyush-nimbalkar: This test is still missing.

pkg/migration/generate.go
```diff
@@ -47,6 +50,25 @@ func (h *Handler) createStorageCluster(
...
+
+	// default format, such as portworx/oci-monitor:2.9.0
+	if len(imageParts) == 2 && imageParts[0] == "portworx" {
+		return "", image
+	}
+
+	imageOnly := imageParts[len(imageParts)-1]
+	paths := imageParts[:len(imageParts)-1]
+
+	return path.Join(paths...), imageOnly
```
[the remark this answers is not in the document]
piyush-nimbalkar: This is a good change. 
But, now I am thinking what if the custom registry path is actually `registry.io/portworx`. In that case the oci-mon images would look like `registry.io/portworx.oci-monitor` but the stork image would be `registry.io/portworx/stork`. Should we look at stork image as well to decide the custom image registry?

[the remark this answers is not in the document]
haibinxie: let's leave it as is. once we implement dry run, we will output the glitches so users can review and update StorageCluster before approval.

pkg/migration/migration_test.go
haibinxie: because it is set in desiredImage entirely now, we no longer check version.

pkg/migration/generate.go
haibinxie: yes, that is expected. as we would add it back in getImageURN

piyush-nimbalkar: I mean let's verify that we don't end up using default and that the component deployments come up fine even with just name in the `status.desiredImages`.

pkg/migration/generate.go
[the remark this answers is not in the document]
piyush-nimbalkar: A lot of users I have seen use `portworx` as the namespace. This would cause a problem for all such customers.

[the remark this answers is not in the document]
haibinxie: even with that, portworx could be prefix of image instead of suffix of registry. correct?

pkg/migration/generate.go
haibinxie: I will add automated functional test for it after FC.

piyush-nimbalkar: I just wanted to know if we verified this case manually at least, where the image in desired images is just the name (eg. `stork:2.7.1`) and no path

pkg/migration/generate.go
```diff
@@ -493,6 +571,7 @@ func (h *Handler) addMonitoringSpec(cluster *corev1.StorageCluster) error {
 		prometheus.Spec.ServiceMonitorSelector != nil &&
 		prometheus.Spec.ServiceMonitorSelector.MatchLabels["name"] == serviceMonitorName {
 		cluster.Spec.Monitoring.Prometheus.Enabled = true
+		cluster.Status.DesiredImages.Prometheus = *prometheus.Spec.Image
```
piyush-nimbalkar, line 574: Other prometheus images are still missing - 
```
PrometheusOperator:         cluster.Status.DesiredImages.PrometheusOperator,
PrometheusConfigMapReload:  cluster.Status.DesiredImages.PrometheusConfigMapReload,
PrometheusConfigReloader:   cluster.Status.DesiredImages.PrometheusConfigReloader,
```

pkg/migration/generate.go
```diff
@@ -339,13 +382,11 @@ func (h *Handler) constructStorageCluster(ds *appsv1.DaemonSet) *corev1.StorageC
...
 	// Enable CSI
 	csiEnabled := false
-	telemetryEnabled := false
 	for _, c := range ds.Spec.Template.Spec.Containers {
 		switch c.Name {
 		case pxutil.CSIRegistrarContainerName:
 			csiEnabled = true
-		case pxutil.TelemetryContainerName:
-			telemetryEnabled = true
+			cluster.Status.DesiredImages.CSINodeDriverRegistrar = c.Image
```
piyush-nimbalkar, line 389: Other CSI images are still missing - 
```
			CSINodeDriverRegistrar:     cluster.Status.DesiredImages.CSINodeDriverRegistrar,
			CSIProvisioner:             cluster.Status.DesiredImages.CSIProvisioner,
			CSIAttacher:                cluster.Status.DesiredImages.CSIAttacher,
			CSIResizer:                 cluster.Status.DesiredImages.CSIResizer,
			CSISnapshotter:             cluster.Status.DesiredImages.CSISnapshotter,
			CSISnapshotController:      cluster.Status.DesiredImages.CSISnapshotController,
			CSIHealthMonitorController: cluster.Status.DesiredImages.CSIHealthMonitorController,
```
I would recommend adding a new method like `addCSISpec` just like the other component methods we have and populating the CSI desired images there, just like how you have done it for telemetry and prometheus.

piyush-nimbalkar (approved):
Looks good after adding the missing CSI and prometheus images to the StorageCluster status.

pkg/migration/generate.go
haibinxie: the images anyway will be regenerated based on px version when they upgrade px, why not leave all of them empty and let them be generated based on current px version? 
could you point me to the code/script customer uses to generate the manifest configmap for airgapped env, I want to take a look at it and see if we can reuse that to generate the configmap, i feel there might be glitches in the manifest configmap we generated during migration.

piyush-nimbalkar: Yes, when they upgrade the images will sync, but we in case of air-gapped env we don't want anything to fail. 
If we don't add the other images from the existing CSI deployment, then we will end up using the default as we cannot reach the version manifest. So better to use whatever they are already running instead of using the defaults.

https://docs.portworx.com/portworx-install-with-kubernetes/on-premise/airgapped/

haibinxie: which means before upgrading px version, customer needs to run the airgapped script to update the manifest configmap on airgapped env. 

what if we let customer run the airgapped script before migration on airgapped env. they need to do this when doing fresh install or when upgrading, so it does not hurt to run before migration with operator. in this case we dont need to carry those desiredImages in storagecluster -- although this is mostly done now, but still need more comprehensive tests. we just need to output a warning tell them manifest configmap is not present please run airgapped script.

piyush-nimbalkar: Sure, asking them to run the script before migration won't hurt. But, now that this is already handled, we don't need to ask them to use the script anymore, right? 
It is not a lot of change to just read the images from the CSI and prometheus deployments and populate the desired images and configmap anyway.

haibinxie: in general if airgapped script is production ready we shall use it, the way we implement here in DesiredImage still have uncertainties and will require more efforts in testing and maintaining going forward. such as

1. migrate while upgrade would not work
2. looks like there are more images to handle, all images in the airgapped list needs to be handled. 

IMAGES="$IMAGES docker.io/portworx/px-enterprise:2.8.1.5"
IMAGES="$IMAGES docker.io/portworx/oci-monitor:2.8.1.5"
IMAGES="$IMAGES docker.io/openstorage/stork:2.7.0"
IMAGES="$IMAGES docker.io/portworx/autopilot:1.3.1"
IMAGES="$IMAGES docker.io/portworx/px-node-wiper:2.5.0"
IMAGES="$IMAGES docker.io/purestorage/ccm-service:3.1.1"
IMAGES="$IMAGES docker.io/portworx/px-lighthouse:2.0.7"
IMAGES="$IMAGES docker.io/portworx/lh-config-sync:2.0.7"
IMAGES="$IMAGES docker.io/portworx/lh-stork-connector:2.0.7"
IMAGES="$IMAGES docker.io/portworx/talisman:1.1.0"
IMAGES="$IMAGES docker.io/portworx/px-operator:1.7.0"
IMAGES="$IMAGES k8s.gcr.io/pause:3.1"
IMAGES="$IMAGES gcr.io/google_containers/kube-controller-manager-amd64:v1.14.10"
IMAGES="$IMAGES gcr.io/google_containers/kube-scheduler-amd64:v1.14.10"
IMAGES="$IMAGES k8s.gcr.io/sig-storage/csi-node-driver-registrar:v2.1.0"
IMAGES="$IMAGES docker.io/openstorage/csi-provisioner:v1.6.1-1"
IMAGES="$IMAGES docker.io/openstorage/csi-attacher:v1.2.1-1"
IMAGES="$IMAGES k8s.gcr.io/sig-storage/csi-resizer:v1.1.0"
IMAGES="$IMAGES docker.io/openstorage/csi-snapshotter:v1.2.2-1"
IMAGES="$IMAGES k8s.gcr.io/sig-storage/snapshot-controller:v3.0.3"
IMAGES="$IMAGES quay.io/prometheus/prometheus:v2.7.1"
IMAGES="$IMAGES quay.io/coreos/prometheus-operator:v0.34.0"
IMAGES="$IMAGES quay.io/coreos/configmap-reload:v0.0.1"
IMAGES="$IMAGES quay.io/coreos/prometheus-config-reloader:v0.34.0"
IMAGES="$IMAGES quay.io/prometheus/alertmanager:v0.17.0"

piyush-nimbalkar: - I went through the list already. The only images that are not accounted for are the pvc controller and stork scheduler images. Rest all the images will be covered by the desired images.
- We are not going to recommend upgrade during the migration at least for now. 

Basically I am saying there is no harm in handling the rest of the images if we can. It will definitely make the experience better. We can definitely recommend running the air-gapped script before the migration, but even without it we can handle the migration. Additionally we also, take care of creating the px-version configmap for them.

haibinxie: with the issues above and potentially overlooked issues, I would prefer just stop migration and output a warning let user run airgapped script first. we dont want to run into various issues with this implementation and maintaining it. 

1. how about pause image, k8s controller images etc
2. it's not blocked, chances are still there

what do you think?

piyush-nimbalkar: Pause image has never changed. K8s controller image == pvc controller image. Version for this image is taken from the k8s version. But it is possible that customers updated the k8s version but did not change the k8s controller and k8s scheduler images (stork-scheduler) in their deployments.

I agree that we should throw a warning and not proceed with the migration. From what comes to my mind, we should stop the migration and ask users to run the air-gapped script when - 
- stork-scheduler image version != current k8s version
- pvc controller image version != current k8s version 
- all images don't have the same custom image registry prefix 

If none of the above issues are encountered we should be able to proceed with the migration.

pkg/migration/generate.go
```diff
@@ -516,6 +636,34 @@ func (h *Handler) addMonitoringSpec(cluster *corev1.StorageCluster) error {
 		prometheus.Spec.ServiceMonitorSelector != nil &&
 		prometheus.Spec.ServiceMonitorSelector.MatchLabels["name"] == serviceMonitorName {
 		cluster.Spec.Monitoring.Prometheus.Enabled = true
+		cluster.Status.DesiredImages.Prometheus = *prometheus.Spec.Image
+
+		dep := &appsv1.Deployment{}
+		err := h.client.Get(
+			context.TODO(),
+			types.NamespacedName{
+				Name:      component.PrometheusOperatorDeploymentName,
```
piyush-nimbalkar, line 645: This should be `prometheusOpDeploymentName` as we should read the old prometheus deployment created by the spec generator.

pkg/migration/generate.go
```diff
@@ -516,6 +636,34 @@ func (h *Handler) addMonitoringSpec(cluster *corev1.StorageCluster) error {
...
+		} else if err != nil {
+			return err
+		}
+
+		container := dep.Spec.Template.Spec.Containers[0]
+		cluster.Status.DesiredImages.PrometheusOperator = container.Image
+
+		for _, arg := range container.Args {
+			if strings.HasPrefix(arg, prometheusConfigReloaderArg) {
+				cluster.Status.DesiredImages.PrometheusConfigReloader = strings.TrimPrefix(arg, prometheusConfigReloaderArg)
```
piyush-nimbalkar, line 661: The existing customer have installed the prometheus operator using this  - https://docs.portworx.com/samples/k8s/portworx-pxc-operator.yaml. You can find this in the spec generator with DaemonSet on the final page.
This does not have the `--prometheus-config-reload` argument. If the arg is not passed prometheus operator uses the image with its own version. So after this for loop, let's check if the `DesiredImage.PrometheusConfigReloader` is empty. If empty, then let's add `quay.io/coreos/prometheus-config-reloader:<prometheus-image-version>` to the desired image.

pkg/migration/migration_test.go
```diff
@@ -23,8 +17,244 @@ import (
...
+	time.Sleep(2 * time.Second)
+
+	cluster := &corev1.StorageCluster{}
+	err := testutil.Get(k8sClient, cluster, clusterName, ds.Namespace)
+	require.NoError(t, err)
+	require.Equal(t, expectedStcImages.PortworxImage, cluster.Spec.Image)
+	require.Equal(t, expectedStcImages.CustomImageRegistry, cluster.Spec.CustomImageRegistry)
+	require.Equal(t, expectedStcImages.Components.Stork, cluster.Status.DesiredImages.Stork)
+	require.Equal(t, expectedStcImages.Components.Autopilot, cluster.Status.DesiredImages.Autopilot)
+	require.Equal(t, expectedStcImages.Components.Telemetry, cluster.Status.DesiredImages.Telemetry)
```
piyush-nimbalkar, line 243: Can we extend this test for all other images?

pkg/migration/migration_test.go
```diff
@@ -23,8 +17,244 @@ import (
...
+		require.NoError(t, err)
+	} else {
+		require.True(t, errors.IsNotFound(err))
+	}
+
+	// Stop the migration process by removing the daemonset
+	err = testutil.Delete(k8sClient, ds)
+	require.NoError(t, err)
+}
+
```
piyush-nimbalkar, line 257: Can we write a quick test to check that the desired images are reset when the config map is already present?

haibinxie:
@piyush-nimbalkar all resolved.

PR #516, opened 2022-02-22
[OPERATOR-560] Add pod topology spread constraints to stork, pvc and csi pods

pure-jliao:
**What this PR does / why we need it**: Add pod topology spread constraints to csi, pvc controller, stork and stork scheduler pods.

Closes # [OPERATOR-560](https://portworx.atlassian.net/browse/OPERATOR-560)

### DIFF
--- a/pkg/controller/storagecluster/stork.go
+++ b/pkg/controller/storagecluster/stork.go
@@ -60,6 +60,24 @@ const (
 	userVolumeNamePrefix    = "user-"
 )
 
+var (
+	storkServiceLabels = map[string]string{
+		"name": "stork",
+	}
+	storkDeploymentLabels = map[string]string{
+		"tier": "control-plane",
+	}
+	storkTemplateLabels = map[string]string{
+		"name": "stork",
+		"tier": "control-plane",
+	}
+	storkSchedulerDeploymentLabels = map[string]string{
+		"tier":      "control-plane",
+		"component": "scheduler",
+		"name":      storkSchedDeploymentName,
+	}
+)
+
 func (c *Controller) syncStork(
 	cluster *corev1.StorageCluster,
 ) error {
@@ -477,7 +495,7 @@ func (c *Controller) createStorkService(
 				OwnerReferences: []metav1.OwnerReference{*ownerRef},
 			},
 			Spec: v1.ServiceSpec{
-				Selector: getStorkServiceLabels(),
+				Selector: storkServiceLabels,
 				Ports: []v1.ServicePort{
 					{
 						Name:       "extender",
@@ -597,6 +615,11 @@ func (c *Controller) createStorkDeployment(
 	existingVolumes := append([]v1.Volume{}, existingDeployment.Spec.Template.Spec.Volumes...)
 	sort.Sort(k8sutil.VolumeByName(existingVolumes))
 
+	updatedTopologySpreadConstraints, err := util.GetTopologySpreadConstraints(c.client, storkTemplateLabels)
+	if err != nil {
+		return err
+	}
+
 	// Check if image, envs, cpu or args are modified
 	modified := existingImage != imageName ||
 		!reflect.DeepEqual(existingCommand, command) ||
@@ -607,11 +630,13 @@ func (c *Controller) createStorkDeployment(
 		existingDeployment.Spec.Template.Spec.HostNetwork != hostNetwork ||
 		util.HasPullSecretChanged(cluster, existingDeployment.Spec.Template.Spec.ImagePullSecrets) ||
 		util.HasNodeAffinityChanged(cluster, existingDeployment.Spec.Template.Spec.Affinity) ||
-		util.HaveTolerationsChanged(cluster, existingDeployment.Spec.Template.Spec.Tolerations)
+		util.HaveTolerationsChanged(cluster, existingDeployment.Spec.Template.Spec.Tolerations) ||
+		util.HaveTopologySpreadConstraintsChanged(updatedTopologySpreadConstraints,
+			existingDeployment.Spec.Template.Spec.TopologySpreadConstraints)
 
 	if !c.isStorkDeploymentCreated || modified {
 		deployment := c.getStorkDeploymentSpec(cluster, ownerRef, imageName,
-			command, envVars, volumes, volumeMounts, targetCPUQuantity)
+			command, envVars, volumes, volumeMounts, targetCPUQuantity, updatedTopologySpreadConstraints)
 		if err = k8sutil.CreateOrUpdateDeployment(c.client, deployment, ownerRef); err != nil {
 			return err
 		}
@@ -629,16 +654,9 @@ func (c *Controller) getStorkDeploymentSpec(
 	volumes []v1.Volume,
 	volumeMounts []v1.VolumeMount,
 	cpuQuantity resource.Quantity,
+	topologySpreadConstraints []v1.TopologySpreadConstraint,
 ) *apps.Deployment {
 	pullPolicy := imagePullPolicy(cluster)
-	deploymentLabels := map[string]string{
-		"tier": "control-plane",
-	}
-	templateLabels := getStorkServiceLabels()
-	for k, v := range deploymentLabels {
-		templateLabels[k] = v
-	}
-
 	replicas := int32(3)
 	maxUnavailable := intstr.FromInt(1)
 	maxSurge := intstr.FromInt(1)
@@ -650,7 +668,7 @@ func (c *Controller) getStorkDeploymentSpec(
 			Annotations: map[string]string{
 				"scheduler.alpha.kubernetes.io/critical-pod": "",
 			},
-			Labels:          deploymentLabels,
+			Labels:          storkDeploymentLabels,
 			OwnerReferences: []metav1.OwnerReference{*ownerRef},
 		},
 		Spec: apps.DeploymentSpec{
@@ -662,7 +680,7 @@ func (c *Controller) getStorkDeploymentSpec(
 				},
 			},
 			Selector: &metav1.LabelSelector{
-				MatchLabels: templateLabels,
+				MatchLabels: storkTemplateLabels,
 			},
 			Replicas: &replicas,
 			Template: v1.PodTemplateSpec{
@@ -670,7 +688,7 @@ func (c *Controller) getStorkDeploymentSpec(
 					Annotations: map[string]string{
 						"scheduler.alpha.kubernetes.io/critical-pod": "",
 					},
-					Labels: templateLabels,
+					Labels: storkTemplateLabels,
 				},
 				Spec: v1.PodSpec{
 					ServiceAccountName: storkServiceAccountName,
@@ -750,6 +768,10 @@ func (c *Controller) getStorkDeploymentSpec(
 		deployment.Spec.Template.Spec.Containers[0].VolumeMounts = volumeMounts
 	}
 
+	if len(topologySpreadConstraints) != 0 {
+		deployment.Spec.Template.Spec.TopologySpreadConstraints = topologySpreadConstraints
+	}
+
 	return deployment
 }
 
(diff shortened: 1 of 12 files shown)


pkg/util/k8s/node.go
```diff
@@ -91,3 +101,24 @@ func IsNodeCordoned(node *v1.Node) (bool, time.Time) {
 	}
 	return false, time.Time{}
 }
+
+// GetNodeTopologyLabelKeys returns a list of sorted node label keys that exist in a candidate list
+func GetNodeTopologyLabelKeys() ([]string, error) {
+	nodeList, err := coreops.Instance().GetNodes()
```
piyush-nimbalkar, line 107: Let's pass the sigs controller k8s client from the caller here instead of using the sched-ops GetNodes method. The controller client will get the nodes from the cache, while the GetNodes will hit the k8s API server everytime.

pkg/util/util.go
```diff
@@ -298,3 +297,38 @@ func ComponentsPausedForMigration(cluster *corev1.StorageCluster) bool {
...
+) bool {
+	nodeTopologyLabelKeySet := make(map[string]bool)
+	constraintTopologyKeySet := make(map[string]bool)
+	for _, key := range nodeTopologyLabelKeys {
+		nodeTopologyLabelKeySet[key] = true
+	}
+	for _, constraint := range topologySpreadConstraints {
+		constraintTopologyKeySet[constraint.TopologyKey] = true
+	}
+	return !reflect.DeepEqual(nodeTopologyLabelKeySet, constraintTopologyKeySet)
```
piyush-nimbalkar, line 314: Why check just the keys and not the entire spread constraints? What if the labels have changed?

pkg/util/util.go
```diff
@@ -298,3 +297,38 @@ func ComponentsPausedForMigration(cluster *corev1.StorageCluster) bool {
 	componentsPaused, err := strconv.ParseBool(cluster.Annotations[constants.AnnotationPauseComponentMigration])
 	return migrating && err == nil && componentsPaused
 }
+
+// HasTopologySpreadConstraintsChanged checks if the deployment has pod topology spread constraints changed
+func HasTopologySpreadConstraintsChanged(
```
piyush-nimbalkar, line 302: nit: *Have

pkg/controller/storagecluster/stork.go
```diff
@@ -821,15 +832,21 @@ func (c *Controller) createStorkSchedDeployment(
 		}
 	}
 
+	nodeTopologyLabelKeys, err := k8sutil.GetNodeTopologyLabelKeys()
```
piyush-nimbalkar, line 835: We could call the `util.GetTopologySpreadConstraints` here and then pass that to `util.HasTopologySpreadConstraintsChanged` to compare the entire object. Wdyt?

test/integration_test/basic_test.go
```diff
@@ -771,3 +788,59 @@ func AlertManagerRegression(tc *types.TestCase) func(*testing.T) {
...
+// 5. remove all topology labels, then do step 3 again to validate all topology keys should be removed
+// 6. uninstall the cluster
+func InstallWithNodeTopologyLabels(tc *types.TestCase) func(*testing.T) {
+	return func(t *testing.T) {
+		// Add topology label to all nodes
+		nodeList, err := coreops.Instance().GetNodes()
+		require.NoError(t, err)
+		regionKey := "topology.kubernetes.io/region"
+		topologyKeys := []string{regionKey}
+		for i, node := range nodeList.Items {
```
piyush-nimbalkar, line 806: I would say copy the original `topology.kuberenetes.io/*` values and remove them before we overwrite with our own values. Once we start deploying on clouds, it would have some values already set, which make this test behave unexpectedly. Reset the original values back when the test is successful.

pkg/util/test/util.go
```diff
@@ -2002,6 +2002,49 @@ func ValidateTelemetryInstalled(pxImageList map[string]string, cluster *corev1.S
...
 
+// ValidatePodTopologySpreadConstraints validates pod topology spread constraints for following component deployments:
+// csi, pvc controller, stork, stork controller
+func ValidatePodTopologySpreadConstraints(t *testing.T, namespace string, topologyKeys []string, timeout, interval time.Duration) {
+	topologyKeySet := make(map[string]bool)
+	for _, key := range topologyKeys {
+		topologyKeySet[key] = true
+	}
+
+	validateDeploymentTopology := func(deploymentName string) func() (interface{}, bool, error) {
```
piyush-nimbalkar, line 2013: Instead of iterating over all the different deployments, we should call this validate function from individual component's Validate function. So this gets executed every time we call ValidateStorageCluster (or the ValidateComponents, ValidateStork, ValidateCSI, etc). On clouds it will be checked for other tests as well automatically.

We can always generate the list of topologyKeys on the fly looking at the current nodes, instead of the test supplying it.

pkg/util/test/util.go
```diff
@@ -2002,6 +2002,49 @@ func ValidateTelemetryInstalled(pxImageList map[string]string, cluster *corev1.S
...
+		return func() (interface{}, bool, error) {
+			deployment, err := appops.Instance().GetDeployment(deploymentName, namespace)
+			if err != nil {
+				return nil, true, fmt.Errorf("failed to get deployment %s/%s", namespace, deploymentName)
+			}
+			constraintKeySet := make(map[string]bool)
+			for _, constraint := range deployment.Spec.Template.Spec.TopologySpreadConstraints {
+				constraintKeySet[constraint.TopologyKey] = true
+			}
+			if !reflect.DeepEqual(topologyKeySet, constraintKeySet) {
```
piyush-nimbalkar, line 2023: Let's check all the fields of the spread constraint with the key. Although those fields are static, it would be a good validation in case we change anything in future.

test/integration_test/basic_test.go
```diff
@@ -771,3 +788,59 @@ func AlertManagerRegression(tc *types.TestCase) func(*testing.T) {
...
+
+		// Install the cluster and do basic validations
+		testSpec := tc.TestSpec(t)
+		cluster, ok := testSpec.(*corev1.StorageCluster)
+		require.True(t, ok)
+
+		cluster = ci_utils.DeployAndValidateStorageCluster(cluster, ci_utils.PxSpecImages, t)
+
+		// Validate pod topology spread constraints
+		testutil.ValidatePodTopologySpreadConstraints(t, cluster.Namespace, topologyKeys,
```
piyush-nimbalkar, line 821: We could call ValidateComponents, if we address above comment.

pkg/util/test/util.go
```diff
@@ -2002,6 +2002,49 @@ func ValidateTelemetryInstalled(pxImageList map[string]string, cluster *corev1.S
...
+				return nil, true, fmt.Errorf("failed to validate deployment pod topology spread constraints, expected keys: %s, actual keys: %v", topologyKeys, constraintKeySet)
+			}
+			return nil, false, nil
```
piyush-nimbalkar, line 2026: Can we additionally validate that the pods are actually spread across the nodes with different fault domains?

test/integration_test/basic_test.go
```diff
@@ -771,3 +788,59 @@ func AlertManagerRegression(tc *types.TestCase) func(*testing.T) {
...
+	return func(t *testing.T) {
+		// Add topology label to all nodes
+		nodeList, err := coreops.Instance().GetNodes()
+		require.NoError(t, err)
+		regionKey := "topology.kubernetes.io/region"
+		topologyKeys := []string{regionKey}
+		for i, node := range nodeList.Items {
+			val := fmt.Sprintf("region%v", i)
+			logrus.Infof("Label node %s with %s=%s", node.Name, regionKey, val)
+			err := coreops.Instance().AddLabelOnNode(node.Name, regionKey, val)
```
piyush-nimbalkar, line 809: Instead of adding a different region for all nodes, can we assume 2 or 3 regions and distribute them evenly across the nodes? So, if there are 6 nodes multiple nodes, some are in the same region.

Rather, I would suggest first add different zone labels on every node, then validate. After that add region labels like I suggested above and then validate. This will ensure that if some pods are in different zones, they might get redistributed because now they are in the same region.

piyush-nimbalkar (commented):
Overall the change looks good.

pkg/util/test/util.go
```diff
@@ -2331,3 +2387,28 @@ func CreateClusterWithTLS(caCertFileName, serverCertFileName, serverKeyFileName
 func BoolPtr(val bool) *bool {
 	return &val
 }
+
+// NewK8sClient returns a new controller runtime Kubernetes client
+func NewK8sClient(scheme *runtime.Scheme) (client.Client, error) {
```
pure-jliao, line 2392: due to loop import issue I dup the whole function here, don't know if there are better solutions, please let me know @piyush-nimbalkar

pkg/controller/storagecluster/stork.go
```diff
@@ -60,6 +60,28 @@ const (
...
+	}
+	storkSchedulerDeploymentLabels = map[string]string{
+		"name":      storkSchedDeploymentName,
+		"tier":      "control-plane",
+		"component": "scheduler",
+	}
+	storkSchedulerTemplateLabels = map[string]string{
+		"tier":      "control-plane",
+		"component": "scheduler",
+	}
```
piyush-nimbalkar, line 82: This will conflict because of https://github.com/libopenstorage/operator/pull/524. I have merged the deployment and template labels for the scheduler. You can just use one set of labels here.

pkg/util/util_test.go
```diff
@@ -281,3 +289,49 @@ func TestComponentsPausedForMigration(t *testing.T) {
 	cluster.Annotations[constants.AnnotationPauseComponentMigration] = "true"
 	require.True(t, ComponentsPausedForMigration(cluster))
 }
+
+func TestGetTopologySpreadConstraints(t *testing.T) {
```
piyush-nimbalkar, line 293: Can we add a quick test for HaveTopologySpreadConstraintsChanged as well just to keep the unit test coverage up? I know this is tested from the individual components, but go test doesn't count it in the coverage.

pkg/util/test/util.go
```diff
@@ -2002,6 +2024,40 @@ func ValidateTelemetryInstalled(pxImageList map[string]string, cluster *corev1.S
...
+	k8sClient, err := NewK8sClient(s)
+	if err != nil {
+		return err
+	}
+	t := func() (interface{}, bool, error) {
+		existingDeployment, err := appops.Instance().GetDeployment(deployment.Name, deployment.Namespace)
+		if err != nil {
+			return nil, true, fmt.Errorf("failed to get deployment %s/%s", deployment.Namespace, deployment.Name)
+		}
+		expectedConstraints, err := util.GetTopologySpreadConstraints(k8sClient, existingDeployment.Spec.Template.Labels)
```
piyush-nimbalkar, line 2041: This could be outside the retry function. We don't need to calculate the spread constraints on every retry.

pkg/util/test/util.go
piyush-nimbalkar: One way we can avoid this is by using the [sched-ops function - GetNodes](https://github.com/portworx/sched-ops/blob/master/k8s/core/nodes.go#L22). It is fine to directly talk to the k8s API directly in tests.

If we do that then we can split `GetTopologySpreadConstraints` and extract the portion of getting nodes. The test could use GetNodes and pass the node list to the function calculating the spread constraints.

```
// Actual code will call this
func GetTopologySpreadConstraints(k8sClient, labels) {
  nodeList := ...
  k8sClient.Get(... nodeList ...)
  GetTopologySpreadConstraintsForNodes(nodeList, labels)
}

// Above test will call this
func GetTopologySpreadConstraintsForNodes(nodeList, labels) {}
```

pkg/util/test/util.go
```diff
@@ -2002,6 +2024,40 @@ func ValidateTelemetryInstalled(pxImageList map[string]string, cluster *corev1.S
...
+		if err != nil {
+			return nil, true, fmt.Errorf("failed to get expected pod topology spread constraints from %s/%s deployment template",
+				deployment.Namespace, deployment.Name)
+		}
+		existingConstraints := existingDeployment.Spec.Template.Spec.TopologySpreadConstraints
+		if !reflect.DeepEqual(expectedConstraints, existingConstraints) {
+			return nil, true, fmt.Errorf("failed to validate deployment pod topology spread constraints for %s/%s, expected constraints: %v, actual constraints: %v",
```
piyush-nimbalkar, line 2048: We can use `%+v` or `%#v` to get well formatted constraint objects.

test/integration_test/basic_test.go
```diff
@@ -771,3 +788,86 @@ func AlertManagerRegression(tc *types.TestCase) func(*testing.T) {
...
+		logrus.Infof("Validate StorageCluster %s", cluster.Name)
+		err = testutil.ValidateStorageCluster(ci_utils.PxSpecImages, cluster,
+			ci_utils.DefaultValidateDeployTimeout, ci_utils.DefaultValidateDeployRetryInterval, true, "")
+		require.NoError(t, err)
+
+		// Uninstall storage cluster and recover labels
+		ci_utils.UninstallAndValidateStorageCluster(cluster, t)
+		logrus.Infof("Recovering cluster topology labels")
+		for _, node := range nodeList.Items {
+			if val, ok := node.Labels[regionKey]; ok {
```
piyush-nimbalkar, line 861: If the nodes don't have any topology labels, then we still want to remove them after the test, right? This won't handle that part.

pkg/util/test/util.go
pure-jliao: we need to get template labels from deployment template to construct the constraints, deployment labels can be changed, so it's better to get the labels in each retry.

test/integration_test/basic_test.go
pure-jliao: removing label is done as a test step before uninstalling the cluster, so at this step all topology labels are already removed.

piyush-nimbalkar (approved):

PR #520, opened 2022-03-01
[CSI] Deprecate CSI feature flag in favor of spec.CSI.Enabled

ggriffiths:

Closes #

### DIFF
--- a/drivers/storage/portworx/portworx.go
+++ b/drivers/storage/portworx/portworx.go
@@ -326,9 +326,6 @@ func (p *portworx) SetDefaultsOnStorageCluster(toUpdate *corev1.StorageCluster) 
 	}
 
 	setDefaultAutopilotProviders(toUpdate)
-	if pxutil.IsCSIEnabled(toUpdate) {
-		setCSISpecDefaults(toUpdate)
-	}
 }
 
 func (p *portworx) PreInstall(cluster *corev1.StorageCluster) error {
@@ -683,8 +680,11 @@ func SetPortworxDefaults(toUpdate *corev1.StorageCluster) {
 			toUpdate.Spec.CSI.Enabled = toUpdate.Spec.CSI.Enabled || csiEnabled
 		}
 
-		// Always remove the feature flag.
+		// Always remove the feature flag. Clear the feature gates map if none are set.
 		delete(toUpdate.Spec.FeatureGates, string(pxutil.FeatureCSI))
+		if len(toUpdate.Spec.FeatureGates) == 0 {
+			toUpdate.Spec.FeatureGates = nil
+		}
 	}
 
 	// Enable CSI if running in k3s environment or
@@ -786,18 +786,6 @@ func setNodeSpecDefaults(toUpdate *corev1.StorageCluster) {
 		updatedNodeSpecs = append(updatedNodeSpecs, *nodeSpecCopy)
 	}
 	toUpdate.Spec.Nodes = updatedNodeSpecs
-}
-
-func setCSISpecDefaults(toUpdate *corev1.StorageCluster) {
-	if toUpdate.Spec.CSI == nil {
-		toUpdate.Spec.CSI = &corev1.CSISpec{
-			// Enabled by feature gate, but not spec
-			Enabled:                   pxutil.FeatureCSI.IsEnabled(toUpdate.Spec.FeatureGates),
-			InstallSnapshotController: boolPtr(false),
-		}
-	} else if toUpdate.Spec.CSI.InstallSnapshotController == nil {
-		toUpdate.Spec.CSI.InstallSnapshotController = boolPtr(false)
-	}
 }
 
 func setSecuritySpecDefaults(toUpdate *corev1.StorageCluster) {
--- a/drivers/storage/portworx/portworx.go
+++ b/drivers/storage/portworx/portworx.go
@@ -3,6 +3,7 @@ package portworx
 import (
 	"context"
 	"fmt"
+	"strconv"
 	"strings"
 
 	version "github.com/hashicorp/go-version"
@@ -663,14 +664,36 @@ func SetPortworxDefaults(toUpdate *corev1.StorageCluster) {
 		}
 	}
 
-	if t.isK3s || toUpdate.Spec.Version == "" {
-		// Enable CSI if running in k3s environment or
-		// if this is a new Portworx installation
-		if _, ok := toUpdate.Spec.FeatureGates[string(pxutil.FeatureCSI)]; !ok {
-			if toUpdate.Spec.FeatureGates == nil {
-				toUpdate.Spec.FeatureGates = make(map[string]string)
+	// Check if feature gate is set. If it is, honor the flag here and remove it.
+	csiFeatureFlag, featureGateSet := toUpdate.Spec.FeatureGates[string(pxutil.FeatureCSI)]
+	if featureGateSet {
+		csiEnabled, err := strconv.ParseBool(csiFeatureFlag)
+		if err != nil {
+			csiEnabled = true
+		}
+
+		// Feature flag set, but CSI spec is not defined. In this case,
+		// we want to use the feature flag value in the CSI spec.
+		if toUpdate.Spec.CSI == nil {
+			toUpdate.Spec.CSI = &corev1.CSISpec{
+				Enabled: csiEnabled,
 			}
-			toUpdate.Spec.FeatureGates[string(pxutil.FeatureCSI)] = "true"
+		} else {
+			// Set CSI spec as enabled if it's already enabled OR the flag is true.
+			toUpdate.Spec.CSI.Enabled = toUpdate.Spec.CSI.Enabled || csiEnabled
+		}
+
+		// Always remove the feature flag.
+		delete(toUpdate.Spec.FeatureGates, string(pxutil.FeatureCSI))
+	}
+
+	// Enable CSI if running in k3s environment or
+	// if this is a new Portworx installation
+	if t.isK3s || toUpdate.Spec.Version == "" {
+		if toUpdate.Spec.CSI == nil {
+			toUpdate.Spec.CSI = &corev1.CSISpec{
+				Enabled: true,
+			}
 		}
 	}
 
@@ -772,6 +795,8 @@ func setCSISpecDefaults(toUpdate *corev1.StorageCluster) {
 			Enabled:                   pxutil.FeatureCSI.IsEnabled(toUpdate.Spec.FeatureGates),
 			InstallSnapshotController: boolPtr(false),
 		}
+	} else if toUpdate.Spec.CSI.InstallSnapshotController == nil {
+		toUpdate.Spec.CSI.InstallSnapshotController = boolPtr(false)
 	}
 }
 
--- a/drivers/storage/portworx/portworx_test.go
+++ b/drivers/storage/portworx/portworx_test.go
@@ -990,8 +990,12 @@ func TestStorageClusterDefaultsForCSI(t *testing.T) {
 	require.True(t, pxutil.IsCSIEnabled(cluster))
 	require.NotEmpty(t, cluster.Status.DesiredImages.CSIProvisioner)
 
+	// Feature gate should be removed when CSI is enabled
+	require.NotContains(t, cluster.Spec.FeatureGates, string(pxutil.FeatureCSI))
+
 	// Don't enable CSI by default for existing cluster
 	cluster.Spec.FeatureGates = nil
+	cluster.Spec.CSI = nil
 	driver.SetDefaultsOnStorageCluster(cluster)
 	require.False(t, pxutil.IsCSIEnabled(cluster))
 	require.Empty(t, cluster.Status.DesiredImages.CSIProvisioner)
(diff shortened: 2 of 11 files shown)


drivers/storage/portworx/portworx.go
```diff
@@ -772,6 +788,8 @@ func setCSISpecDefaults(toUpdate *corev1.StorageCluster) {
 			Enabled:                   pxutil.FeatureCSI.IsEnabled(toUpdate.Spec.FeatureGates),
```
lpitstickpstg, line 788: Still looking at the feature gate here?

ggriffiths: Yeah- before, we were enabling CSI if one of the two were set:
1. feature flag for CSI is enabled
2. CSI Spec has enabled flag as true

So, it was a logical `or` with the above flags. If one was set, enable CSI. To keep backwards compatibility, we need to read the feature flag, set CSI spec as enabled, and delete the feature flag.

drivers/storage/portworx/portworx.go
```diff
@@ -663,14 +664,29 @@ func SetPortworxDefaults(toUpdate *corev1.StorageCluster) {
 		}
 	}
 
+	// Enable CSI if running in k3s environment or
+	// if this is a new Portworx installation
 	if t.isK3s || toUpdate.Spec.Version == "" {
```
haibinxie, line 669: what's the reason to only do this for new installation? IMO this should be done on all installations.

drivers/storage/portworx/util/util.go
```diff
@@ -236,7 +236,7 @@ func IsPortworxEnabled(cluster *corev1.StorageCluster) bool {
 
 // IsCSIEnabled returns true if CSI is not disabled by the feature flag
 func IsCSIEnabled(cluster *corev1.StorageCluster) bool {
-	return IsPortworxEnabled(cluster) && FeatureCSI.IsEnabled(cluster.Spec.FeatureGates)
+	return IsPortworxEnabled(cluster) && cluster.Spec.CSI != nil && cluster.Spec.CSI.Enabled
```
haibinxie, line 239: if we only remove feature gate on new installations then we shall check the csi flag here as well for existing installations.

drivers/storage/portworx/portworx.go
ggriffiths: Simplified logic and existing customers might be using the feature gate w/ Gitops. If a customer is using GitOps and we keep removing the feature flag, gitops will keep adding it back in.

Also keeps the behavior of enabling CSI by default only for new installations.

drivers/storage/portworx/util/util.go
ggriffiths: ah good point.. maybe the tradeoff is better to always move from feature gate -> CSI spec.

drivers/storage/portworx/portworx.go
ggriffiths: after reading your next comment, I agree - will make this change.

drivers/storage/portworx/portworx_test.go
```diff
@@ -990,8 +990,12 @@ func TestStorageClusterDefaultsForCSI(t *testing.T) {
 	require.True(t, pxutil.IsCSIEnabled(cluster))
 	require.NotEmpty(t, cluster.Status.DesiredImages.CSIProvisioner)
 
+	// Feature gate should be removed when CSI is enabled
```
piyush-nimbalkar, line 993: nit: Did you mean feature gate should not be set?

drivers/storage/portworx/portworx_test.go
```diff
@@ -1075,6 +1078,26 @@ func TestStorageClusterDefaultsForCSI(t *testing.T) {
 	require.Empty(t, cluster.Status.DesiredImages.CSINodeDriverRegistrar)
 	require.Empty(t, cluster.Status.DesiredImages.CSIResizer)
 	require.Empty(t, cluster.Status.DesiredImages.CSISnapshotter)
+
+	// If CSI feature flag is true and CSI Spec is false, honor feature flag true
+	cluster.Spec.CSI.Enabled = false
```
piyush-nimbalkar, line 1083: Can we add a simple test with CSI spec as empty and feature gate is set to true value? This will test the most common scenario where existing customers are moved to CSI spec.

drivers/storage/portworx/portworx.go
ggriffiths: true, will remove this.

drivers/storage/portworx/portworx.go
```diff
@@ -772,6 +788,8 @@ func setCSISpecDefaults(toUpdate *corev1.StorageCluster) {
 			Enabled:                   pxutil.FeatureCSI.IsEnabled(toUpdate.Spec.FeatureGates),
 			InstallSnapshotController: boolPtr(false),
 		}
+	} else if toUpdate.Spec.CSI.InstallSnapshotController == nil {
```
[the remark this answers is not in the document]
ggriffiths: No, was just set to false to be more verbose instead of having a nil value. We can remove it though.

drivers/storage/portworx/portworx_test.go
ggriffiths: yeah, will fix

ggriffiths:
manually tested a few cases:
1. CSI disabled, spec is nil, add CSI feature gate as true. Result - featureGate set to nil, CSI enabled in spec:
```
grant@ip-70-0-0-129:~/workspace/go/src/github.com/libopenstorage/operator$ kn get stc -o yaml | grep -A 1 'csi:' 
          f:csi:
            .: {}
--
    csi:
      enabled: true
grant@ip-70-0-0-129:~/workspace/go/src/github.com/libopenstorage/operator$ kn get stc -o yaml | grep featureGates
        {"apiVersion":"core.libopenstorage.org/v1","kind":"StorageCluster","metadata":{"annotations":{"portworx.io/install-source":"https://install.portworx.com/?mc=false\u0026kbver=1.20.4\u0026b=true\u0026f=true\u0026c=px-cluster-72d387a1-6e69-46e8-8799-7e9b08ee8fe1\u0026stork=true\u0026csi=true\u0026st=k8s\u0026stork=true\u0026operator=true"},"name":"px-cluster-72d387a1-6e69-46e8-8799-7e9b08ee8fe1","namespace":"kube-system"},"spec":{"autopilot":{"enabled":true},"deleteStrategy":{"type":"UninstallAndWipe"},"env":[{"name":"REGISTRY_USER","value":"pwxbuild"},{"name":"REGISTRY_PASS","value":"fridaydemos"}],"featureGates":{"CSI":"true"},"image":"portworx/oci-monitor:2.9.0","imagePullPolicy":"Always","kvdb":{"internal":true},"secretsProvider":"k8s","storage":{"forceUseDisks":true,"useAllWithPartitions":true},"stork":{"args":{"webhook-controller":"false"},"enabled":true}}}
```
2. pods are bouncing with CSI spec enabled -> disabled and vice versa

ggriffiths:
fixing a few tests now actually... will update when it's ready for review again

lpitstickpstg (approved):

haibinxie (approved):

drivers/storage/portworx/portworx.go
```diff
@@ -663,14 +661,39 @@ func SetPortworxDefaults(toUpdate *corev1.StorageCluster) {
...
+
+	// Enable CSI if running in k3s environment or
+	// if this is a new Portworx installation
 	if t.isK3s || toUpdate.Spec.Version == "" {
-		// Enable CSI if running in k3s environment or
-		// if this is a new Portworx installation
-		if _, ok := toUpdate.Spec.FeatureGates[string(pxutil.FeatureCSI)]; !ok {
-			if toUpdate.Spec.FeatureGates == nil {
-				toUpdate.Spec.FeatureGates = make(map[string]string)
+		if toUpdate.Spec.CSI == nil {
```
piyush-nimbalkar, line 693: Just to be cautious, let's set the flag outside. In case someone has created the csi spec but not enabled the flag.
```
if toUpdate.Spec.CSI == nil {
  toUpdate.Spec.CSI = &corev1.CSISpec{}
}
toUpdate.Spec.CSI.Enabled = true
```

pkg/controller/storagecluster/update.go
```diff
@@ -732,7 +732,28 @@ func doesTelemetryMatch(oldSpec, currentSpec *corev1.StorageClusterSpec) bool {
 
 // isBounceRequired handles miscellaneous fields that requrie a pod bounce
 func isBounceRequired(oldSpec, currentSpec *corev1.StorageClusterSpec) bool {
-	return isSecurityBounceRequired(oldSpec, currentSpec)
+	return isSecurityBounceRequired(oldSpec, currentSpec) || isCSIBoundRequired(oldSpec, currentSpec)
+}
+
+// isCSIBoundRequired handles CSI fields that requrie a pod bounce
```
piyush-nimbalkar, line 738: nit: *Bounce?

pkg/controller/storagecluster/update.go
```diff
@@ -704,6 +702,8 @@ func matchSelectedFields(
 		return false, nil
 	} else if !reflect.DeepEqual(oldSpec.Resources, currentSpec.Resources) {
 		return false, nil
+	} else if !reflect.DeepEqual(oldSpec.CSI, currentSpec.CSI) {
```
piyush-nimbalkar, line 705: Let's remove this now as we are already checking if the portworx pod needs to be bounced below based on the CSI spec.
Also, this will cause the portworx pod to bounce even if some one enables the snapshot controller.

pkg/migration/generate.go
```diff
@@ -348,8 +348,13 @@ func (h *Handler) constructStorageCluster(ds *appsv1.DaemonSet) *corev1.StorageC
 			telemetryEnabled = true
 		}
 	}
-	cluster.Spec.FeatureGates = map[string]string{
-		"CSI": strconv.FormatBool(csiEnabled),
+	// Install snapshot controller, as Daemonset spec generator
```
piyush-nimbalkar, line 351: Thank you for adding it here too!

pkg/controller/storagecluster/controller_test.go
```diff
@@ -5685,9 +5685,9 @@ func TestUpdateStorageClusterFeatureGates(t *testing.T) {
 	}
 	k8sClient.Create(context.TODO(), oldPod)
 
-	// TestCase: Add spec.featureGates
-	cluster.Spec.FeatureGates = map[string]string{
-		"feature1": "enabled",
+	// TestCase: Add spec.CSI.Enbled
```
piyush-nimbalkar, line 5688: nit: *Enabled

pkg/controller/storagecluster/controller_test.go
```diff
@@ -5705,8 +5705,45 @@ func TestUpdateStorageClusterFeatureGates(t *testing.T) {
...
+
+	podControl.DeletePodName = nil
+
+	result, err = controller.Reconcile(context.TODO(), request)
+	require.NoError(t, err)
+	require.Empty(t, result)
+	require.Equal(t, []string{oldPod.Name}, podControl.DeletePodName)
+
+	// TestCase: No spec.CSI changes
+	cluster.Spec.CSI = nil
```
piyush-nimbalkar, line 5746: Can we add a test where we toggle the snapshot controller and verify that portworx pod doesn't bounce?

piyush-nimbalkar (commented):
Some minor comments and a fix in the upgrade logic, other wise this looks good.
Thank you @ggriffiths for introducing the CSI spec!

piyush-nimbalkar:
@ggriffiths can you also change the integration test to use the new spec instead of the feature flags - https://github.com/libopenstorage/operator/blob/master/test/integration_test/basic_test.go#L346?

pkg/controller/storagecluster/update.go
ggriffiths: good catch, left this in by accident - was originally testing here before adding isCSIBounceRequired...

pkg/controller/storagecluster/controller_test.go
```diff
@@ -5705,8 +5705,57 @@ func TestUpdateStorageClusterFeatureGates(t *testing.T) {
...
+	trueBool := true
+	cluster.Spec.CSI.InstallSnapshotController = &trueBool
+	k8sClient.Update(context.TODO(), cluster)
+
+	podControl.DeletePodName = nil
+
+	result, err = controller.Reconcile(context.TODO(), request)
+	require.NoError(t, err)
+	require.Empty(t, result)
+	require.Equal(t, []string{oldPod.Name}, podControl.DeletePodName)
```
piyush-nimbalkar, line 5744: Ideally this should be nil. If the test is passing it means the pod is bouncing when it should not.

ggriffiths: Tried this case and yeah you're right. Found an issue in isCSIBounceRequired. Testing a few other cases now.

ggriffiths: I was missing `replaceOldPod` and a few other required things for these tests. Fixed this up and did some manually testing, looking good now.

piyush-nimbalkar (approved):

PR #521, opened 2022-03-02
change PreserveFullCustomRegistryPath in CRD yaml to be consistent wi…

haibinxie:
…th spec in code

Closes #

### DIFF
--- a/deploy/crds/core_v1_storagecluster_crd.yaml
+++ b/deploy/crds/core_v1_storagecluster_crd.yaml
@@ -96,7 +96,7 @@ spec:
                   Custom container image registry server that will be used instead of
                   index.docker.io to download Docker images. This may include the repository as well.
                   (Example: myregistry.net:5443 or myregistry.com/myrepository)
-              preserveFullCustomRegistryPath:
+              preserveFullCustomImageRegistry:
                 type: boolean
                 description: >-
                   Setting this to true this stops part of the image tag being swallowed when setting a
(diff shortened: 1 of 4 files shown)


deploy/crds/core_v1_storagecluster_crd.yaml
```diff
@@ -96,7 +96,7 @@ spec:
                   Custom container image registry server that will be used instead of
                   index.docker.io to download Docker images. This may include the repository as well.
                   (Example: myregistry.net:5443 or myregistry.com/myrepository)
-              preserveFullCustomImageRegistry:
+              preserveFullCustomRegistryPath:
```
piyush-nimbalkar, line 99: To be consistent with `customImageRegistry` can we call this `preserveFullCustomImageRegistry` or `preserveFullCustomImageRegistryPath`? Not sure if we will have a different registry than image registry in future, but it is better to be safe as this is a user facing field.
I know this involves more changes in the code, but better get this done once.

haibinxie: Sounds good, changed.

pkg/apis/core/v1/storagecluster.go
```diff
@@ -77,7 +77,7 @@ type StorageClusterSpec struct {
 	// This is a hack to stop GetImageURN from swallowing part of the image tag.
 	// Without it having a customImageRegistry with a / in it means that an image
 	// `portworx/oci-monitor` becomes just `oci-monitor`
-	PreserveFullCustomRegistryPath bool `json:"preserveFullCustomRegistryPath"`
+	PreserveFullCustomImageRegistry bool `json:"preserveFullCustomImageRegistry"`
```
piyush-nimbalkar, line 80: Let's also add `omitempty` here.

piyush-nimbalkar (approved):

pkg/apis/core/v1/storagecluster.go
haibinxie: great catch, updated.

PR #535, opened 2022-03-09
OPERATOR-603: Fix CSI case where op 1.7.0 to 1.8.0 bounced PX pods

ggriffiths:

Fixes case where PX pods were bouncing due the following conditions:
1. Operator was on version 1.7.0
2. STC had CSI enabled w/ featureGate
3. Operator was upgraded to 1.8.0 with CSI.Enabled flag
4. PX pods bounced we noticed a new spec.CSI.Enabled flag, but did NOT check what the feature gate was set to. Unfortunately, we'll always still need check for the previous feature gate value.

OPERATOR-603

I was able to repro this by upgrading from 1.7.0 -> 1.8.0.

Did a manual test w/ upgrading operator 1.7.0 to an image with this change, and the pods were not bounced.

A new unit test also covers the case that I missed before.

### DIFF
--- a/pkg/controller/storagecluster/update.go
+++ b/pkg/controller/storagecluster/update.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"reflect"
 	"sort"
+	"strconv"
 
 	storageapi "github.com/libopenstorage/openstorage/api"
 	"github.com/libopenstorage/operator/drivers/storage/portworx/util"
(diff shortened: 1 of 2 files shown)


ggriffiths:
Did a manual test w/ upgrading operator 1.7.0 to an image with this change, and the pods were not bounced.

A new unit test also covers the case that I missed before.

david-enli (approved):
makes sense, LGTM, I'll give it another try later.

pkg/controller/storagecluster/update.go
```diff
@@ -740,9 +741,9 @@ func isCSIBounceRequired(oldSpec, currentSpec *corev1.StorageClusterSpec) bool {
 		return oldSpec.CSI.Enabled != currentSpec.CSI.Enabled
 	}
 
-	// New CSI spec added to cluster. If enabled, pod will bounce
-	if oldSpec.CSI == nil && currentSpec.CSI != nil {
-		return currentSpec.CSI.Enabled
+	// Spec changed from enabled -> disabled or vice versa
+	if isCSIEnabled(oldSpec) != isCSIEnabled(currentSpec) {
```
pure-jliao, line 745: nit: can we assume for currentSpec we just need to check the new field?
```
if isCSIEnabled(oldSpec) != currentSpec.CSI
```

pure-jliao (approved):
LGTM, left 2 comments.

pkg/controller/storagecluster/update.go
piyush-nimbalkar, line 745: I think the entire function can now be simplified to just these lines, right? Wdyt?
```
func isCSIBounceRequired(..) bool {
    return isCSIEnabled(oldSpec) != isCSIEnabled(currentSpec)
}
```

ggriffiths: We could, but there's always the case where a user re-adds the feature gate even though it's deprecated. That'll be reconciled to the CSI.Enabled field. Just wanted to be safe and cover our bases here.

pkg/controller/storagecluster/update.go
```diff
@@ -754,6 +755,29 @@ func isCSIBounceRequired(oldSpec, currentSpec *corev1.StorageClusterSpec) bool {
...
+		if featureGateSet {
+			csiFeatureFlagEnabled, _ := strconv.ParseBool(csiFeatureFlag)
+			if csiFeatureFlagEnabled {
+				return true
+			}
+		}
+	}
+
+	// Check if enabled in spec
+	if spec.CSI != nil && spec.CSI.Enabled {
```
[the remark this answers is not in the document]
ggriffiths: good point, will do

piyush-nimbalkar (approved):
LGTM

pkg/controller/storagecluster/update.go
ggriffiths: My unit tests cases fail when I switch to just this logic. Let me work through why..

ggriffiths: This was a test issue where the feature gate wasn't getting set back to nil. Fixed, and simplified the code as suggested.

piyush-nimbalkar (approved):

pkg/controller/storagecluster/update.go
piyush-nimbalkar: The feature gate is set to nil by the portworx driver, but it is mocked here. So, we will have to set it to nil in the test.

PR #538, opened 2022-03-10
OPERATOR-538 Add default secrets namespace if not provided

piyush-nimbalkar:

-  The 'portworx' secrets namesapce needs to be explicitly added to the storage
    cluster during migration. As most daemonset clusters have been using portworx
    namespace to store secrets by default. If a secrets namespace is provided
    explicitly then we honor that.
-  Some env variables are added by the operator to the portworx pod directly. We
    don't need to copy them to the storage cluster.

Closes # https://portworx.atlassian.net/browse/OPERATOR-538


pure-jliao (approved):
LGTM

PR #540, opened 2022-03-10
[CSI] Add desired snapshot controller image when empty

ggriffiths:
Add desired snapshot controller image when empty

OPERATOR-613

### DIFF
--- a/drivers/storage/portworx/portworx.go
+++ b/drivers/storage/portworx/portworx.go
@@ -234,18 +234,23 @@ func (p *portworx) SetDefaultsOnStorageCluster(toUpdate *corev1.StorageCluster) 
 			toUpdate.Status.DesiredImages.PxRepo = release.Components.PxRepo
 		}
 
-		if pxutil.IsCSIEnabled(toUpdate) &&
-			(toUpdate.Status.DesiredImages.CSIProvisioner == "" ||
+		if pxutil.IsCSIEnabled(toUpdate) {
+			if toUpdate.Status.DesiredImages.CSIProvisioner == "" ||
 				pxVersionChanged ||
-				autoUpdateComponents(toUpdate)) {
-			toUpdate.Status.DesiredImages.CSIProvisioner = release.Components.CSIProvisioner
-			toUpdate.Status.DesiredImages.CSINodeDriverRegistrar = release.Components.CSINodeDriverRegistrar
-			toUpdate.Status.DesiredImages.CSIDriverRegistrar = release.Components.CSIDriverRegistrar
-			toUpdate.Status.DesiredImages.CSIAttacher = release.Components.CSIAttacher
-			toUpdate.Status.DesiredImages.CSIResizer = release.Components.CSIResizer
-			toUpdate.Status.DesiredImages.CSISnapshotter = release.Components.CSISnapshotter
-			toUpdate.Status.DesiredImages.CSISnapshotController = release.Components.CSISnapshotController
-			toUpdate.Status.DesiredImages.CSIHealthMonitorController = release.Components.CSIHealthMonitorController
+				autoUpdateComponents(toUpdate) {
+				toUpdate.Status.DesiredImages.CSIProvisioner = release.Components.CSIProvisioner
+				toUpdate.Status.DesiredImages.CSINodeDriverRegistrar = release.Components.CSINodeDriverRegistrar
+				toUpdate.Status.DesiredImages.CSIDriverRegistrar = release.Components.CSIDriverRegistrar
+				toUpdate.Status.DesiredImages.CSIAttacher = release.Components.CSIAttacher
+				toUpdate.Status.DesiredImages.CSIResizer = release.Components.CSIResizer
+				toUpdate.Status.DesiredImages.CSISnapshotter = release.Components.CSISnapshotter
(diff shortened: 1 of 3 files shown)


drivers/storage/portworx/portworx.go
```diff
@@ -937,7 +942,7 @@ func hasLighthouseChanged(cluster *corev1.StorageCluster) bool {
 
 func hasCSIChanged(cluster *corev1.StorageCluster) bool {
 	return pxutil.IsCSIEnabled(cluster) &&
-		cluster.Status.DesiredImages.CSIProvisioner == ""
+		(cluster.Status.DesiredImages.CSIProvisioner == "" || cluster.Status.DesiredImages.CSISnapshotController == "")
```
piyush-nimbalkar, line 945: I just realized that we need to break this down into two again, because the controller can be independently enabled later after CSI is enabled.
Let's introduce a new method called `autoUpdateCSISnapshotController()` - 
```
func hasCSIChanged(cluster *corev1.StorageCluster) bool {
	return pxutil.IsCSIEnabled(cluster) &&
		(cluster.Status.DesiredImages.CSIProvisioner == "" || autoUpdateCSISnapshotController(cluster))
}

func autoUpdateCSISnapshotController(cluster *corev1.StorageCluster) bool {
	return cluster.Spec.CSI.InstallSnapshotController &&
                 cluster.Status.DesiredImages.CSISnapshotController == ""
}
```
and change the condition on line 249 to - 
```
			if autoUpdateCSISnapshotController(toUpdate) ||
				pxVersionChanged ||
				autoUpdateComponents(toUpdate) {
```

The reason for all of the above change is that we will populate the snapshot controller image in the desired images as soon as CSI is enabled. We can rather delay that to whenever the controller is enabled so we can pick up the latest recommended version instead of using the one we got at CSI install time.

drivers/storage/portworx/portworx_test.go
```diff
@@ -981,14 +981,18 @@ func TestStorageClusterDefaultsForCSI(t *testing.T) {
 			Namespace: "kube-test",
 		},
 		Spec: corev1.StorageClusterSpec{
-			Image: "px/image:2.1.5.1",
+			Image: "px/image:2.9.0.1",
 		},
 	}
+	// Simulate DesiredImages.CSISnapshotController being empty for old operator version w/o this image
+	driver.SetDefaultsOnStorageCluster(cluster)
+	cluster.Status.DesiredImages.CSISnapshotController = ""
```
piyush-nimbalkar, line 989: After the above change we will have to split this test into two steps. 
First, just enable CSI - DesiredImages.CSISnapshotController should be empty
Second, enable snapshot controller - DesiredImages.CSISnapshotController should not be empty

ggriffiths: Made the above changes.

SetDefaultsOnStorageCluster seems to still always set the `desiredImages.SnapshotController`. I had to keep this line to clear it:
```
cluster.Status.DesiredImages.CSISnapshotController = ""
```

However, when SetDefaultsOnStorageCluster is called again with InstallSnapshotController=true set, the image is set properly. This is expected, right?

piyush-nimbalkar: Actually, it is not expected, but I guess it does not have a bad effect. So far we have populated the desired images of only components that are enabled. Let's try to be consistent just to be safe.
I see what the issue is after comparing to other components. Our `autoUpdateCSISnapshotController` is a bit incorrect.
```
func autoUpdateCSISnapshotController(cluster *corev1.StorageCluster) bool {
	return cluster.Spec.CSI.InstallSnapshotController != nil && *cluster.Spec.CSI.InstallSnapshotController
}

func hasCSIChanged(cluster *corev1.StorageCluster) bool {
	return pxutil.IsCSIEnabled(cluster) &&
		(cluster.Status.DesiredImages.CSIProvisioner == "" || hasCSISnapshotControllerChanged(cluster))
}

func hasCSISnapshotControllerChanged(cluster *corev1.StorageCluster) bool {
         return autoUpdateCSISnapshotController(cluster) && 
                      cluster.Status.DesiredImages.CSISnapshotController == ""

}
```
and line 249 should be - 
```
			if autoUpdateCSISnapshotController(toUpdate) &&
                                 (cluster.Status.DesiredImages.CSISnapshotController == "" ||
				 pxVersionChanged ||
				 autoUpdateComponents(toUpdate) {
```

These changes should be inline with other components and should work as expected.

---
Another issue I observed is that we are not resetting the snapshot controller desired image when it is disabled. We should add this after line 315 - 
```
        if !pxutil.IsCSIEnabled(toUpdate) {
              ...
	else if !autoUpdateCSISnapshotController(toUpdate) {
		toUpdate.Status.DesiredImages.CSISnapshotController = ""
	}
```

ggriffiths: Makes sense. Just made those changes. Test passes now without the image reset line:
```
cluster.Status.DesiredImages.CSISnapshotController = ""
```

piyush-nimbalkar (approved):

drivers/storage/portworx/portworx_test.go
piyush-nimbalkar: Could we add a test to just disable csi snapshot controller and check that the desired image for controller is reset?

PR #545, opened 2022-03-17
handle prometheus and alertmanager image migration

haibinxie:

Closes #


pkg/migration/generate.go
```diff
@@ -546,6 +547,18 @@ func (h *Handler) addPVCControllerSpec(cluster *corev1.StorageCluster) error {
 	return nil
 }
 
+func (h *Handler) getContaienrImage(statefulSetList *appsv1.StatefulSetList, containerName string) string {
```
piyush-nimbalkar, line 550: nit: *Container

pkg/migration/generate.go
```diff
@@ -636,7 +665,12 @@ func (h *Handler) addMonitoringSpec(cluster *corev1.StorageCluster, ds *appsv1.D
 		prometheus.Spec.ServiceMonitorSelector != nil &&
 		prometheus.Spec.ServiceMonitorSelector.MatchLabels["name"] == serviceMonitorName {
 		cluster.Spec.Monitoring.Prometheus.Enabled = true
-		cluster.Status.DesiredImages.Prometheus = *prometheus.Spec.Image
+
+		if prometheus.Spec.Image != nil {
+			cluster.Status.DesiredImages.Prometheus = *prometheus.Spec.Image
+		} else {
+			cluster.Status.DesiredImages.Prometheus = h.getContaienrImage(statefulSetList, "prometheus")
```
piyush-nimbalkar, line 672: This seems like a very crudeway of finding the prometheus/alertmanager image. We already have the prometheus and alertmanager objects. We can easily get the corresponding statefulsets for them and then the image from it. Something like this - 
```
for _, ss := range statefulSetList.Items {
  if metav1.IsControlledBy(ss, alertManager) {
    alertManagerImg = h.getContainerImage(ss, "alertmanager")
  } else if metav1.IsControlledBy(ss, prometheus) {
    prometheusImg = h.getContainerImage(ss, "prometheus")
  }
}
```
We won't iterate over the statefulsets in `getContainerImage` then.

haibinxie: there wont be a container with same name. 

the ownerref check is based on the assumption that there wont be another statefulset owned by it, and the statefulset always has a ownerref with it (not deployed by statefulset directly). although I think this is always true, there is no need to do the extra check

piyush-nimbalkar: Why won't there be a container with the same name? There is nothing unique about these names.
Also, most stateful sets will never have an owner reference. They will have it only if they are owned by an operator. Additionally, we are not just checking if the sts has a owner ref or not, but we are getting the exact stateful set that is owned by our prometheus and alert manager CRs. 
We can never be sure. What if there is another prometheus/alert manager running in the same namespace?

piyush-nimbalkar (approved):
Looks great!

Can we modify/add unit test to test the image is retrieved from statefulset? You can do in a later PR if you want, but let's keep the coverage up.

PR #550, opened 2022-03-20
backup portworx namespace

haibinxie:

Closes #

### DIFF
--- a/pkg/migration/backup.go
+++ b/pkg/migration/backup.go
@@ -184,6 +184,10 @@ func (h *Handler) getPortworxRbac(namespace string, objs *[]client.Object) error
 	if err := h.addObject(pxRoleName, namespace, &rbacv1.Role{}, objs); err != nil {
 		return err
 	}
+
+	if err := h.addObject(secretsNamespace, "", &v1.Namespace{}, objs); err != nil {
+		return err
+	}
 	if err := h.addObject(pxRoleBindingName, secretsNamespace, &rbacv1.RoleBinding{}, objs); err != nil {
 		return err
 	}


pure-jliao (approved):

pkg/migration/backup.go
```diff
@@ -184,6 +184,10 @@ func (h *Handler) getPortworxRbac(namespace string, objs *[]client.Object) error
 	if err := h.addObject(pxRoleName, namespace, &rbacv1.Role{}, objs); err != nil {
 		return err
 	}
+
+	if err := h.addObject(secretsNamespace, "", &v1.Namespace{}, objs); err != nil {
```
piyush-nimbalkar, line 188: We should not have to backup the portworx namespace as we don't clean it up. Rather we should not even be deleting it during migration.
This change should not be needed.

haibinxie: I knew that, this is just to match spec gen as we generated it. just in case it is deleted by accident

piyush-nimbalkar: It is not our responsibility to create a namespace for the user. Additionally, restoring it does not help anything.
The namespace is used to store the pvc encryption secrets. So if that namespace is deleted by accident, then they have bigger problems, restoring must the namespace obj is not going to help.
Why do we need to match the spec gen?

piyush-nimbalkar: Also, if the user is using a different namespace to store secrets using PX_SECRETS_NAMESPACE, then `portworx` namespace is not in picture. We use that other namespace for storing encryption secrets.

haibinxie: without the namespace, restoring of px-role and px-role-binding in the namespace would fail

haibinxie: we dont delete during migration, but we backup objects in the namespace. backup is for accident i would not predict why it is gone:)

PR #551, opened 2022-03-21
set px version in status during migration

haibinxie:

Closes #

### DIFF
--- a/pkg/migration/generate.go
+++ b/pkg/migration/generate.go
@@ -835,6 +835,7 @@ func (h *Handler) handleCustomImageRegistry(cluster *corev1.StorageCluster) erro
 		MetricsCollectorProxy:      h.removeCustomImageRegistry(cluster.Spec.CustomImageRegistry, cluster.Status.DesiredImages.MetricsCollectorProxy),
 		PxRepo:                     h.removeCustomImageRegistry(cluster.Spec.CustomImageRegistry, cluster.Status.DesiredImages.PxRepo),
 	}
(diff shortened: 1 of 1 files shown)


pkg/migration/generate.go
```diff
@@ -835,6 +835,7 @@ func (h *Handler) handleCustomImageRegistry(cluster *corev1.StorageCluster) erro
 		MetricsCollectorProxy:      h.removeCustomImageRegistry(cluster.Spec.CustomImageRegistry, cluster.Status.DesiredImages.MetricsCollectorProxy),
 		PxRepo:                     h.removeCustomImageRegistry(cluster.Spec.CustomImageRegistry, cluster.Status.DesiredImages.PxRepo),
 	}
+	cluster.Status.Version = pxutil.GetImageTag(cluster.Spec.Image)
```
piyush-nimbalkar, line 838: Why do we need this? This will be automatically set by the operator when the migration starts.

haibinxie: If we dont have this, the desiredImages will be overwritten after stc reconcile.

piyush-nimbalkar: Oh yes, that would happen.

haibinxie: Good point, changed

piyush-nimbalkar (approved):
Let's add this check in the existing unit test as well.

haibinxie:
> 

I was trying to change unit test but turned out complicated to mock controller reconcile in migration unit tests. I will add functional tests in next sprint which would cover this as well, in next sprint I will list out the plan for all functional tests.

PR #557, opened 2022-03-25
[OPERATOR-634] Fix operator migration test component image validation

pure-jliao:
**What this PR does / why we need it**: We should update hardcoded daemonset component images from spec gen url

Closes #

### DIFF
--- a/test/integration_test/migration_test.go
+++ b/test/integration_test/migration_test.go
@@ -147,7 +147,7 @@ func testMigration(t *testing.T, objects []runtime.Object, ds *appsv1.DaemonSet)
 	err := ci_utils.AddPortworxImageToDaemonSet(ds)
 	require.NoError(t, err)
 
-	err = updateImagesWithKubernetesVersion(objects)
+	err = updateComponentDeploymentImages(objects)
 	require.NoError(t, err)
 
 	logrus.Infof("Creating portworx objects")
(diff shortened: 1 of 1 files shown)


test/integration_test/migration_test.go
```diff
@@ -268,16 +268,37 @@ func extractPortworxDaemonSetFromObjects(objects []runtime.Object) (*appsv1.Daem
...
 	k8sVersion, err := k8sutil.GetVersion()
 	if err != nil {
 		return err
 	}
 	for _, obj := range objects {
-		if dep, ok := obj.(*appsv1.Deployment); ok &&
-			(dep.Name == "stork-scheduler" || dep.Name == "portworx-pvc-controller") {
-			image := strings.Split(dep.Spec.Template.Spec.Containers[0].Image, ":")[0]
-			dep.Spec.Template.Spec.Containers[0].Image = image + ":v" + k8sVersion.String()
+		if dep, ok := obj.(*appsv1.Deployment); ok {
```
piyush-nimbalkar, line 277: This is good. We also need to handle images for - 
- ccm-service (px daemonset)
- csi-node-driver-registrar (px daemonset)
- alertmanager (alert manager object)

pkg/util/test/util.go
```diff
@@ -425,6 +425,11 @@ func ValidateStorageCluster(
 		return err
 	}
 
+	if _, ok := liveCluster.Annotations["portworx.io/migration-approved"]; ok {
+		logrus.Debug("Sleeping 10 seconds to ensure all components are bounced before validation")
+		time.Sleep(10 * time.Second)
```
haibinxie, line 430: change to sleep 1s

haibinxie: oops, i thought it has retry, we have too many sleep in code, the perf is very slow now. this should be a PollImmediate with timeout, even 10 seconds may not be enough occasionally. can you verify by retry with timeout, the retry interval should be 1 second or less.

pure-jliao: that makes the change complicated, this is just to make sure we don't get the old objects from daemonset, as after px pods are ready we'll create new component deployments, and the validations already have retries, also it's for migration tests only, IMO it should be good.

test/integration_test/migration_test.go
pure-jliao: right now we don't have validations for those images, maybe we could add them when we have the validation?

pkg/util/test/util.go
haibinxie: it wont be a complicated change, just verify if ds is gone with pollImmediate. if the migration is not finished, the validation would still pass as the deployment and images are the same?

pure-jliao: sounds good, that case an easier way to do this is wait for ds is gone then validate the stc, we did this in a reverse way before. updated

haibinxie (approved):

PR #561, opened 2022-04-04
[OPERATOR-641] Add application workload validation for migration test

pure-jliao:

**What this PR does / why we need it**: Add app validation before and after migration

Closes # OPERATOR-641

**Special notes for your reviewer**: App only added to `AllComponentsEnabled` test case


test/integration_test/utils/k8s.go
```diff
@@ -124,6 +145,31 @@ func CreateObjects(objects []runtime.Object) error {
...
+				} else if livePVC.Status.Phase != v1.ClaimBound {
+					return nil, true, fmt.Errorf("failed to validate PVC %s/%s, phase %s", pvc.Namespace, pvc.Name, pvc.Status.Phase)
+				}
+				return nil, false, nil
+			}
+			_, err = task.DoRetryWithTimeout(t, timeout, interval)
+		}
+		if err != nil {
+			return err
+		}
```
piyush-nimbalkar, line 168: We can use ValidatePersistentVolumeClaim from sched-ops here - https://github.com/portworx/sched-ops/blob/master/k8s/core/persistentvolumeclaims.go#L94.

piyush-nimbalkar (approved):

test/integration_test/utils/k8s.go
pure-jliao: thanks for the hint. I was surprised I didn't find any PVC validation utilities.

PR #565, opened 2022-04-08
OPERATOR-666 Set maxSurge and maxUnavailable in stork scheduler

piyush-nimbalkar:
In a 3 node cluster, the stork scheduler upgrades were getting stuck because of
the pod anti-affinity policy. The new 4th pod was going in a Pending state and
waiting for one of the existing 3 pods to go down. This was happening because
the default maxUnavailable is 25%. This was preventing the existing pod from
going down, and the pod anti-affinity stopped the new pod from coming up.

Closes # https://portworx.atlassian.net/browse/OPERATOR-666
 
### DIFF
--- a/pkg/controller/storagecluster/stork.go
+++ b/pkg/controller/storagecluster/stork.go
@@ -883,6 +883,8 @@ func getStorkSchedDeploymentSpec(
 ) *apps.Deployment {
 	pullPolicy := imagePullPolicy(cluster)
 	replicas := int32(3)
+	maxUnavailable := intstr.FromInt(1)
+	maxSurge := intstr.FromInt(1)
 
 	deployment := &apps.Deployment{
 		ObjectMeta: metav1.ObjectMeta{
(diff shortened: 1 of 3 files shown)


pkg/controller/storagecluster/stork.go
```diff
@@ -883,6 +883,7 @@ func getStorkSchedDeploymentSpec(
 ) *apps.Deployment {
 	pullPolicy := imagePullPolicy(cluster)
 	replicas := int32(3)
+	maxSurge := intstr.FromInt(0)
```
pure-jliao, line 886: don't know why we have a 0 here? 
also this
```
maxUnavailable declares the maximum number (or percentage) of unavailable pods during the update. If maxSurge is set to 0, this field cannot be 0.
```

pure-jliao (commented):
we can hit this issue in pvc controller right? but probably not for stork or csi.

pkg/controller/storagecluster/stork.go
piyush-nimbalkar: The default value of maxUnavailable is 25%. So we can set maxSurge to 0 here.

piyush-nimbalkar: Max surge = 0 means that without adding an extra pod, the deployment can start terminating pods during an upgrade.

Anyway, I code to make it similar to pvc controller. The problem was because of maxUnavailable. The default value of 25% was preventing even 1/3 pods to go down.

pure-jliao (approved):

pkg/controller/storagecluster/stork.go
pure-jliao: oh we already have this update strategy in pvc controller, just saw that.

PR #576, opened 2022-04-26
[OPERATOR-667] Node migration labels should be removed when deleting a StorageCluster.

pure-jliao:

**What this PR does / why we need it**: When recovering to DaemonSet in case of migration failures, migration labels added by operator should be removed.

Closes # OPERATOR-667

### DIFF
--- a/pkg/controller/storagecluster/storagecluster.go
+++ b/pkg/controller/storagecluster/storagecluster.go
@@ -631,6 +631,9 @@ func (c *Controller) deleteStorageCluster(
 		}
 
 		if toDelete.Status.Conditions[foundIndex].Status == corev1.ClusterOperationCompleted {
+			if err := c.removeMigrationLabels(); err != nil {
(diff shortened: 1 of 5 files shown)


haibinxie (approved):
LGTM. In theory we should also remove migration related node anti-affinity rules on the daemonset, however I dont see there is issue related to it, we can leave it as is

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -643,6 +643,28 @@ func (c *Controller) deleteStorageCluster(
 		msg := fmt.Sprintf("Failed to cleanup Stork. %v", err)
 		k8s.WarningEvent(c.recorder, cluster, util.FailedComponentReason, msg)
 	}
+
+	if err := c.removeMigrationLabels(); err != nil {
```
piyush-nimbalkar, line 647: I think we should do this before the finalizer is removed. That way if the node update fails, we retry again.

piyush-nimbalkar (approved):

PR #580, opened 2022-04-28
[OPERATOR-690] change SCC from privileged to custom scc

haibinxie:


drivers/storage/portworx/component/securitycontextconstraints.go
```diff
@@ -0,0 +1,147 @@
...
+			},
+			Volumes: []ocp_secv1.FSType{"*"},
+			Groups: []string{
+				"system:cluster-admins",
+				"system:nodes",
+				"system:masters",
+			},
+			Users: []string{
+				"system:admin",
+				"system:serviceaccount:openshift-infra:build-controller",
```
piyush-nimbalkar, line 133: Why do we need to add these users?

drivers/storage/portworx/component/securitycontextconstraints.go
```diff
@@ -0,0 +1,147 @@
...
+				fmt.Sprintf("system:serviceaccount:%s:%s", namespace, PxSCCName),
```
piyush-nimbalkar, line 134: We should be using `pxutil.PortworxServiceAccountName(cluster)` instead of PxSCCName here.

drivers/storage/portworx/component/securitycontextconstraints.go
```diff
@@ -0,0 +1,147 @@
...
+	} else {
+		logrus.Debugf("security context constraints is disabled, %v", err)
+	}
+
+	return err == nil
+}
+
+func (s *scc) Reconcile(cluster *opcorev1.StorageCluster) error {
+	for _, scc := range s.getSCCs(cluster.Namespace) {
+		ownerRef := metav1.NewControllerRef(cluster, pxutil.StorageClusterKind())
```
piyush-nimbalkar, line 69: Let's move this line outside the `for` loop. The ownerRef is not going to change per scc.

drivers/storage/portworx/component/securitycontextconstraints.go
```diff
@@ -0,0 +1,147 @@
...
+		scc.SetOwnerReferences([]metav1.OwnerReference{*ownerRef})
+
+		err := s.k8sClient.Create(context.TODO(), &scc)
+		if err != nil && !errors.IsAlreadyExists(err) {
+			return fmt.Errorf("failed to create %s security context constraints: %s", scc.Name, err)
+		}
```
piyush-nimbalkar, line 75: We should do a Get and then Create to avoid unnecessary API calls the k8s api server. The Get comes from the cache.

drivers/storage/portworx/component/securitycontextconstraints.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+func (s *scc) Delete(cluster *opcorev1.StorageCluster) error {
+	if cluster.DeletionTimestamp != nil &&
+		cluster.Spec.DeleteStrategy != nil &&
+		cluster.Spec.DeleteStrategy.Type != "" {
+		return nil
+	}
+
+	return nil
```
piyush-nimbalkar, line 87: Who is responsible for deleting the SCCs after StorageCluster is removed?

drivers/storage/portworx/component/securitycontextconstraints.go
```diff
@@ -0,0 +1,147 @@
...
+func (s *scc) Initialize(k8sClient client.Client, k8sVersion version.Version, scheme *runtime.Scheme, recorder record.EventRecorder) {
+	s.k8sClient = k8sClient
+}
+
+func (s *scc) IsPausedForMigration(cluster *opcorev1.StorageCluster) bool {
+	return false
+}
+
+func (s *scc) IsEnabled(cluster *opcorev1.StorageCluster) bool {
+	ocp_secv1.Install(s.k8sClient.Scheme())
```
piyush-nimbalkar, line 51: Q: What does this Install do?

drivers/storage/portworx/component/securitycontextconstraints.go
```diff
@@ -0,0 +1,147 @@
...
+	apiextensionsv1.AddToScheme(s.k8sClient.Scheme())
+	list := &ocp_secv1.SecurityContextConstraintsList{}
+	err := s.k8sClient.List(
+		context.TODO(),
+		list,
+	)
+	if err == nil {
+		logrus.Debugf("security context constraints is enabled")
+	} else {
+		logrus.Debugf("security context constraints is disabled, %v", err)
```
piyush-nimbalkar, line 61: nit: *are enabled
*are disabled

piyush-nimbalkar, line 61: We should log these as trace as this will get called every reconcile loop.

drivers/storage/portworx/component/securitycontextconstraints.go
haibinxie: this is to be consistent with original scc,  i doubt it might be needed by ocp controller/upgrades etc, just to be cautious of regressions

drivers/storage/portworx/component/securitycontextconstraints.go
haibinxie: client-go does not recognize the CRD type without it.

haibinxie: install is a replacement of AddToScheme, AddToScheme is deprecated by ocp

piyush-nimbalkar: Oh got it. It is adding the SecurityContextConstraint object to scheme. Makes sense.
The next line is adding the CRD type.

piyush-nimbalkar (commented):
LGTM, just one comment.

drivers/storage/portworx/component/securitycontextconstraints.go
```diff
@@ -0,0 +1,181 @@
...
+				Type: ocp_secv1.SupplementalGroupsStrategyRunAsAny,
+			},
+			Volumes: []ocp_secv1.FSType{"*"},
+			Groups:  nil,
+			Users: []string{
+				fmt.Sprintf("system:serviceaccount:%s:%s", cluster.Namespace, pxutil.PortworxServiceAccountName(cluster)),
+				fmt.Sprintf("system:serviceaccount:%s:%s", cluster.Namespace, CSIServiceAccountName),
+				fmt.Sprintf("system:serviceaccount:%s:%s", cluster.Namespace, LhServiceAccountName),
+				fmt.Sprintf("system:serviceaccount:%s:%s", cluster.Namespace, PVCServiceAccountName),
+				fmt.Sprintf("system:serviceaccount:%s:%s", cluster.Namespace, CollectorServiceAccountName),
```
piyush-nimbalkar, line 167: Do we use privileged SCC for metrics collector?

piyush-nimbalkar (approved):

drivers/storage/portworx/component/securitycontextconstraints.go
haibinxie: Yes.

PR #586, opened 2022-05-02
filter more volumes in dryrun

haibinxie:


pkg/migration/dryrun.go
```diff
@@ -15,6 +15,17 @@ import (
 	v1 "k8s.io/api/core/v1"
 )
 
+var (
+	// "crioconf" only exists with operator install.
+	// "dev" only exists with daemonset install.
+	// varlibosd only exists with operator install
+	skipVolumes = map[string]bool {
```
haibinxie, line 22: @pure-jliao not sure how did you get the list of volumes to skip in your PR, i am thinking whether I should copy your list as it's more, we can discuss tomorrow after the standup meeting.

pure-jliao: just for your reference, here is a sample I'm using https://portworx.atlassian.net/browse/OPERATOR-509?focusedCommentId=202070
I added `builtinPxVolumeNames` list in my PR, you could add another `builtinOperatorVolumeNames` list and put all volumes from the `getVolumes()` function. then in `deepEqualVolumes` if the volume exists in either set, we remove it from comparing. how does that sound?

haibinxie: the question is are those built-in volumes always stay the same or they may change in different environment, if they change then we should also compare them.

pure-jliao: that case we could do this, it's easier to maintain 2 name lists, if the volume name is in both list, or not in either list, we check the volume content. I think the user could modify the default volume and mount, or they should be the same, @piyush-nimbalkar suggestions?

PR #596, opened 2022-05-05
[OPERATOR-691] Add StorageCluster CSI topology spec to toggle the feature gate

pure-jliao:
* default: spec.csi.topology.enabled=false

**What this PR does / why we need it**: Add stc spec.csi.topology.enabled flag to toggle csi topology feature gate. disabled by default.

Closes # OPERATOR-691


piyush-nimbalkar (approved):
Can you copy the crd schema changes to 1.9.0 and 2.0.0 folders too? You can do it in a separate PR if you want.

drivers/storage/portworx/testspec/csiDeployment_1.0_k8s120_with_topology.yaml
```diff
@@ -0,0 +1,84 @@
...
+            - name: ADDRESS
+              value: /csi/csi.sock
+          volumeMounts:
+            - name: socket-dir
+              mountPath: /csi
+      volumes:
+        - name: socket-dir
+          hostPath:
+            path: /var/lib/kubelet/plugins/pxd.portworx.com
+            type: DirectoryOrCreate
```
haibinxie, line 84: missing line at end

drivers/storage/portworx/component/csi.go
```diff
@@ -1167,6 +1196,12 @@ func (c *csi) getCSIConfiguration(
 			InstallSnapshotController: boolPtr(false),
 		}
 	}
+	if cluster.Spec.CSI.Enabled && cluster.Spec.CSI.Topology == nil {
+		// CSI topology feature gate is disabled by default
+		cluster.Spec.CSI.Topology = &corev1.CSITopologySpec{
```
haibinxie, line 1201: @piyush-nimbalkar @ggriffiths do you think we should enable it by default?

pure-jliao: as it's only for direct access I think it's better to leave it disabled by default. also here the cluster object is just used for populating the csi configuration I don't think the actual object will be updated, as by default the whole topology field should be nil.

haibinxie (approved):

PR #599, opened 2022-05-08
dryrun tool

haibinxie:
1. add reconcile API call to storageCluster to populate default values and generate other objects
2. write one object to one file instead of all objects to a file
3. add validations to migration
4. restructure code


pkg/dryrun/dryrun.go
```diff
@@ -0,0 +1,810 @@
...
+	if err = d.writeFiles(objs); err != nil {
+		return err
+	}
+	logrus.Infof("Operator will deploy %d objects, saved to folder %s", len(objs), d.outputFolder)
+
+	if d.realClient != nil {
+		controller := &storagecluster.Controller{}
+		controller.SetKubernetesClient(d.realClient)
+		h := migration.New(controller)
+		dsObjs, err := h.GetAllDaemonSetObjects(d.cluster.Namespace)
```
dsupure, line 228: Are you assuming we only have portworx daemonset in the namespace. Is it possible you have multiple DaemonSets ?

pkg/dryrun/dryrun.go
```diff
@@ -0,0 +1,810 @@
...
+					kind = "StatefulSet"
+					logrus.Info("px-csi-ext is deployed as StatefulSet instead of Deployment by operator")
+				}
+			}
+		} else if kind == "Prometheus" && name == "prometheus" {
+			logrus.Warning("Prometheus service name will change from prometheus to px-prometheus after migration")
+			name = "px-prometheus"
+		} else if kind == "Service" && name == "prometheus" {
+			name = "px-prometheus"
+		} else if kind == "Service" && name == "autopilot" {
```
dsupure, line 464: Do we need to consider ccm here?

pkg/dryrun/dryrun.go
```diff
@@ -0,0 +1,810 @@
...
+	}
+
+	start := func() {
+		if err := d.realControllerMgr.Start(ctrl.SetupSignalHandler()); err != nil {
+			logrus.Fatalf("Manager exited non-zero error: %v", err)
+		}
+	}
+	go start()
+	// Wait for the cache to be started.
+	time.Sleep(time.Second)
```
dsupure, line 639: Is 1 sec enough?

pkg/dryrun/validation.go
```diff
@@ -0,0 +1,272 @@
...
+	"fmt"
+	"reflect"
+	"sort"
+
+	v1 "k8s.io/api/core/v1"
+
+	"github.com/libopenstorage/operator/pkg/util"
+)
+
+var (
```
dsupure, line 13: Why we need to define twice in seperate files? Can we have one place instead?

cmd/dryrun/dryrun.go
```diff
@@ -0,0 +1,73 @@
...
+
+func main() {
+	app := cli.NewApp()
+	app.Name = "portworx-operator-dryrun"
+	app.Usage = "Portworx operator dry run tool"
+	app.Version = version.Version
+	app.Action = execute
+
+	app.Flags = []cli.Flag{
+		cli.BoolFlag{
```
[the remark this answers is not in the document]
haibinxie: good idea, resolved, there is no aliases param, i used a different way.

cmd/dryrun/dryrun.go
```diff
@@ -0,0 +1,73 @@
...
+			Name:  flagVerbose,
+			Usage: "Enable verbose logging",
+		},
+		cli.StringFlag{
+			Name:  flagStorageCluster,
+			Usage: "File for storage cluster spec, retrieve from k8s if it's not configured",
+		},
+		cli.StringFlag{
+			Name:  flagKubeConfig,
```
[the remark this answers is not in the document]
haibinxie: what do you mean?

pkg/dryrun/dryrun.go
```diff
@@ -0,0 +1,810 @@
...
+	mockDriver   storage.Driver
+	cluster      *corev1.StorageCluster
+	outputFolder string
+}
+
+// Init performs required initialization
+func (d *DryRun) Init(kubeconfig, outputFolder, storageClusterFile string) error {
+	var err error
+
+	k8sVersion := "v1.22.0"
```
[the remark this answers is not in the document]
haibinxie: if kubeconfig is not provided, 1.22.0 will be the default version to setup simulator.

pkg/dryrun/dryrun.go
```diff
@@ -0,0 +1,810 @@
...
+	} else {
+		ver, err := testutil.GetK8SVersion()
+		if err != nil {
+			logrus.WithError(err).Error("failed to get k8s version")
+		} else {
+			k8sVersion = ver
+			logrus.Infof("k8s version is %s", k8sVersion)
+			minVer, err := goversion.NewVersion("1.16")
+			if err != nil {
+				return fmt.Errorf("error parsing version '1.16': %v", err)
```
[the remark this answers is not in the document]
haibinxie: yes, right now it's actually blocked during migration too

pkg/dryrun/dryrun.go
haibinxie: yes, there is a ticket to support two daemonset installs, it's not supported yet. we will document this prior to release

pkg/dryrun/dryrun.go
haibinxie: ccm is part of telemetry. we dont support in daemonset spec however

pkg/dryrun/dryrun.go
haibinxie: yes, this should be good. it just needs to start without the cache to be fully populated.

pkg/dryrun/validation.go
haibinxie: var is only used by the current file

haibinxie:
@dsupure resolved comments in a new commit, also updated the confluence page with the two docs about min k8s version and two daemonset installs.

cmd/dryrun/dryrun.go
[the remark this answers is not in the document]
dsupure: I mean keep the name simple.  can we use `kubeconfig` instead of `flagKubeConfig`.

dsupure (approved):
LGTM!

PR #601, opened 2022-05-09
Add 1.8.1 olm catalog

pure-jliao:

Closes #

### DIFF
--- a/deploy/olm-catalog/portworx/1.8.1/portworx-certified.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/portworx/1.8.1/portworx-certified.clusterserviceversion.yaml
@@ -1,0 +1,540 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: ClusterServiceVersion
+metadata:
+  name: portworx-operator.v1.8.1
+  namespace: placeholder
+  annotations:
+    capabilities: Auto Pilot
+    categories: "Storage"
+    description: Cloud native storage solution for production workloads
+    containerImage: portworx/px-operator:1.8.1-dev
+    repository: https://github.com/libopenstorage/operator
+    createdAt: 2022-05-09T08:00:00Z
+    support: Portworx, Inc
+    certified: "true"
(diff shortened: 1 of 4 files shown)


deploy/olm-catalog/portworx/1.8.1/portworx-certified.clusterserviceversion.yaml
```diff
@@ -0,0 +1,540 @@
...
+          },
+          "spec": {}
+        }
+      ]
+spec:
+  displayName: Portworx Enterprise
+  version: 1.8.1
+  minKubeVersion: "1.12.0"
+  maturity: stable
+  replaces: portworx-operator.v1.7.1
```
piyush-nimbalkar, line 55: This should be 1.8.0

test/integration_test/utils/px_operator.go
```diff
@@ -10,7 +10,7 @@ import (
 )
 
 const (
-	nextReleaseTag = "1.9.0-dev"
+	nextReleaseTag = "1.8.1-dev"
```
piyush-nimbalkar, line 13: Should we leave this to 1.9.0 if 1.8.1 is going to be a quick hot fix release?

piyush-nimbalkar (approved):

test/integration_test/utils/px_operator.go
pure-jliao: this doesn't matter, the pipeline will use the default jenkins env parameter, I can update this again once 1.8.1 is released.

deploy/olm-catalog/portworx/1.8.1/portworx-certified.clusterserviceversion.yaml
pure-jliao: ok so the rule applies between patch versions as well not only the minor versions?

piyush-nimbalkar: Yes, all versions

PR #607, opened 2022-05-11
[OPERATOR-633] Add exponential backoff to pod restart when nodes are cordoned

pure-jliao:
…

Closes #

### DIFF
--- a/pkg/util/k8s/node.go
+++ b/pkg/util/k8s/node.go
@@ -19,6 +19,13 @@ import (
 file contains node level k8s utility functions
 */
 
+// NodeInfo contains information of a k8s node, consumed by controllers
+type NodeInfo struct {
+	NodeName             string
+	LastPodCreationTime  time.Time
+	CordonedRestartDelay time.Duration
+}
+
 // IsNodeBeingDeleted returns true if the underlying machine for the Kubernetes node is being deleted.
 // This method is only supported on platforms that use the cluster-api (https://github.com/kubernetes-sigs/cluster-api)
 func IsNodeBeingDeleted(node *v1.Node, cl client.Client) (bool, error) {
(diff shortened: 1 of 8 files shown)


pkg/util/k8s/node.go
```diff
@@ -65,14 +82,17 @@ func IsPodRecentlyCreatedAfterNodeCordoned(
 		return true
 	}
 
-	lastCreationTime, ok := lastCreationTimeOnNode[node.Name]
-	// If the pod has never been created, no timestamp.
-	if !ok {
-		return false
+	// If pod has been recently created, return true
+	// Otherwise the pod will get deleted, increase the restart delay
```
haibinxie, line 86: The pod won't get deleted, it will return shouldRun false, shouldContinueRun true by caller. 
"Otherwise the pod will get deleted"

pkg/util/k8s/node.go
```diff
@@ -65,14 +82,17 @@ func IsPodRecentlyCreatedAfterNodeCordoned(
...
+	recentlyCreated := cutOffTime.Before(nodeInfo.LastPodCreationTime)
+	if !recentlyCreated && !overwriteRestartDelay {
+		nodeInfo.CordonedRestartDelay = waitDuration * 2
+		if nodeInfo.CordonedRestartDelay > constants.MaxCordonedRestartDelay {
+			nodeInfo.CordonedRestartDelay = constants.MaxCordonedRestartDelay
+		}
+		logrus.Infof("node %s is cordoned, next pod restart is scheduled in %s", nodeInfo.NodeName, nodeInfo.CordonedRestartDelay)
```
haibinxie, line 93: "in %s" -> "in %d seconds"

haibinxie (approved):

pkg/util/k8s/node.go
pure-jliao: it's already a time.duration object, so it will be parsed correctly, with %d, it will become nanosecond

pkg/util/k8s/node.go
pure-jliao: it has different callers, will add comments at the caller.

piyush-nimbalkar (approved):

PR #609, opened 2022-05-11
[CSI] Update images to latest

ggriffiths:

Closes #


haibinxie (commented):
Is there any functional test automation? what's the process to verify if the new images wont introduce regression

ggriffiths:
We have automated functional testing with the csi-e2e-test suite. Also torpedo CSI tests cover this path.

drivers/storage/portworx/component/csi.go
```diff
@@ -563,6 +563,7 @@ func getCSIDeploymentSpec(
 			"--csi-address=$(ADDRESS)",
 			"--leader-election=true",
 			"--default-fstype=ext4",
+			"--extra-create-metadata=true",
```
pure-jliao, line 566: provisioner v2+ is not covered in unit tests, we may want to add that as the default in the future

test/integration_test/testspec/migration/daemonset-with-all-components.yaml
```diff
@@ -132,7 +132,7 @@ spec:
             - name: dbusmount
               mountPath: /var/run/dbus
         - name: csi-node-driver-registrar
-          image: k8s.gcr.io/sig-storage/csi-node-driver-registrar:v2.1.0
+          image: k8s.gcr.io/sig-storage/csi-node-driver-registrar:v2.5.0
```
pure-jliao, line 135: do we update the default versions for daemonset installation? if not probably we don't need to change here.

pure-jliao (approved):
LGTM

test/integration_test/testspec/migration/daemonset-with-all-components.yaml
ggriffiths: Yes we updated the version for the daemonset version

drivers/storage/portworx/component/csi.go
ggriffiths: Ok yes we can add that in the future

PR #613, opened 2022-05-12
PWX-23693 Add update pvc permission to stork

lpitstickpstg:
Signed-off-by: Luke Pitstick <lpitstick@purestorage.com>

Add update pvc permission to stork so the stork-scheduler can modify the pvc in WaitForFirstConsumer scenario (PWX-23693)

Closes # PWX-23693


piyush-nimbalkar (approved):
The test will fail. You need to update the test spec https://github.com/libopenstorage/operator/blob/master/pkg/controller/storagecluster/testspec/storkSchedClusterRole.yaml

LGTM after fixing the test specs.

haibinxie (approved):

PR #615, opened 2022-05-12
PTX-6585 Add method to get credentials from the PX vSphere secret

nikolaypopov:
* Add method to get credentials from the PX vSphere secret, due to some special characters aren't being allowed
* Add method to find and copy PX vSphere secret to custom namespaces, if we deploy PX into any other namespace but kube-system
* Fix how we treat upgrade string, there was a bug there that I missed


pkg/util/test/util.go
```diff
@@ -391,6 +394,95 @@ func UninstallStorageCluster(cluster *corev1.StorageCluster, kubeconfig ...strin
...
+
+// CreateVsphereCredentialEnvVarsFromSecret check if px-vsphere-secret exists and returns vSphere crendentials Env vars
+func CreateVsphereCredentialEnvVarsFromSecret(namespace string) ([]v1.EnvVar, error) {
+	var envVars []v1.EnvVar
+
+	// Get PX vSphere secret
+	_, err := coreops.Instance().GetSecret(DefaultPxVsphereSecretName, namespace)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			logrus.Infof("PX vSphere secret %s in not found in %s namespace, unable to get credentials from secret, please make sure you have specified them in the Env vars", DefaultPxVsphereSecretName, namespace)
```
pure-jliao, line 452: we should use warn or error I think.

test/integration_test/main_test.go
```diff
@@ -96,9 +96,13 @@ func setup() error {
 		return err
 	}
 
-	ci_utils.PxUpgradeHopsURLList = strings.Split(pxUpgradeHopsURLs, ",")
+	if len(pxUpgradeHopsURLs) != 0 {
```
pure-jliao, line 99: `pxUpgradeHopsURLs ` is a string right? do you mean to check the url list length after split? or just if `pxUpgradeHopsURLs == ""`?

test/integration_test/main_test.go
```diff
@@ -96,9 +96,13 @@ func setup() error {
...
+		ci_utils.PxUpgradeHopsURLList = strings.Split(pxUpgradeHopsURLs, ",")
+	}
 
-	ci_utils.OperatorUpgradeHopsImageList = strings.Split(operatorUpgradeHopsImages, ",")
+	if len(operatorUpgradeHopsImages) != 0 {
```
pure-jliao, line 103: same here.

test/integration_test/utils/storagecluster.go
```diff
@@ -34,17 +35,58 @@ func CreateStorageClusterTestSpecFunc(cluster *corev1.StorageCluster) func(t *te
...
 func ConstructStorageCluster(cluster *corev1.StorageCluster, specGenURL string, specImages map[string]string) error {
+	var envVarList []v1.EnvVar
+
 	cluster.Spec.Image = specImages["version"]
 	if cluster.Namespace == "" {
 		cluster.Namespace = PxNamespace
 	}
 
+	if cluster.Namespace != PxNamespace {
+		// Create custom namespace
```
pure-jliao, line 46: IMO we should avoid creating this ns here, it should be created in the test itself, otherwise how can we make sure it's deleted after the test? if not deleted properly it could cause other test failures.

pkg/util/test/util.go
```diff
@@ -391,6 +394,95 @@ func UninstallStorageCluster(cluster *corev1.StorageCluster, kubeconfig ...strin
...
+			return nil, nil
+		}
+		return nil, fmt.Errorf("failed to get secret %s in %s namespace, err %v", DefaultPxVsphereSecretName, namespace, err)
+	}
+
+	envVars = []v1.EnvVar{
```
pure-jliao, line 458: do we need to check if the actual values are empty or not? if they are we should not add the item.

test/integration_test/main_test.go
nikolaypopov: Done

test/integration_test/main_test.go
nikolaypopov: Done

test/integration_test/utils/storagecluster.go
nikolaypopov: Done, moved it to test

pkg/util/test/util.go
nikolaypopov: I don't want to rely on this being populated or not existing at all . The secret is created by automation, so if its created it should always be populated and also we don't really need secret, the values can also be passed as Env vars.

pure-jliao (approved):
LGTM

PR #618, opened 2022-05-13
reduce log

haibinxie:
here is the new output of the tool:

```
operator# ksys exec -it portworx-operator-6977cdcbd4-5nzmx -- /dryrun
INFO[0001] k8s version is v1.20.15                      
INFO[0001] Reconciling StorageCluster                    Request.Name=px-cluster-ad6dc338-d754-49a8-ab42-a98ce7488f67 Request.Namespace=kube-system
WARN[0001] Found mountPath conflict for volume containerdsock at /run/containerd, volume will be ignored 
INFO[0001] Operator will deploy 60 objects, saved to folder /tmp/portworxSpecs 
INFO[0003] Got 41 objects from k8s cluster, will compare with objects installed by operator 
INFO[0003] Portworx will be deployed via Pod scheduled by operator (instead of DaemonSet) 
WARN[0003] failed to compare Deployment stork-scheduler, Containers are different: container stork-scheduler is different: image is different: before-migration k8s.gcr.io/kube-scheduler-amd64:v1.19.10, after-migration k8s.gcr.io/kube-scheduler-amd64:v1.20.15, please make sure the image exists if running on air-gapped env.

 
WARN[0003] ServiceMonitor name will change from portworx-prometheus-sm to portworx after migration, ok to proceed with migration 
WARN[0003] Prometheus service name will change from prometheus to px-prometheus after migration, ok to proceed with migration 
```

### DIFF
--- a/pkg/dryrun/dryrun.go
+++ b/pkg/dryrun/dryrun.go
@@ -97,7 +97,7 @@ type DryRun struct {
 
 // Init performs required initialization
 func (d *DryRun) Init(kubeconfig, outputFolder, storageClusterFile string) error {
-	logrus.Infof("command line args, kubeconfig: %s, outputFolder: %s, storageClusterFile: %s", kubeconfig, outputFolder, storageClusterFile)
+	logrus.Debugf("command line args, kubeconfig: %s, outputFolder: %s, storageClusterFile: %s", kubeconfig, outputFolder, storageClusterFile)
 	var err error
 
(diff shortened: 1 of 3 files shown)


pure-jliao (approved):
LGTM, btw do we plan to add the splitter in the log for different section or errors?

dsupure (approved):
LGTM!

PR #620, opened 2022-05-16
update log when image changes during migration, make it info when not…

haibinxie:
… running on airgaped env.

Updates to dryrun log:
1. split log into multiple lines
2. if image is different, only warn if running on air gapped env.
3. filter some more expected env var and volume differences found during test

### DIFF
--- a/pkg/dryrun/dryrun.go
+++ b/pkg/dryrun/dryrun.go
@@ -41,6 +41,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/yaml"
 
+	"github.com/google/uuid"
 	"github.com/libopenstorage/operator/drivers/storage"
 	"github.com/libopenstorage/operator/drivers/storage/portworx/component"
 	pxutil "github.com/libopenstorage/operator/drivers/storage/portworx/util"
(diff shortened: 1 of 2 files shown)


pkg/dryrun/dryrun.go
```diff
@@ -41,6 +41,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/yaml"
 
+	"github.com/google/uuid"
```
pure-jliao, line 44: nit: external import in second group

pure-jliao (approved):
LGTM

pkg/dryrun/validation.go
```diff
@@ -227,6 +243,7 @@ func (d *DryRun) deepEqualEnvVars(env1, env2 []v1.EnvVar) error {
 			"NODE_NAME":            true,
 			"PX_NAMESPACE":         true,
 			"PX_SECRETS_NAMESPACE": true,
+			"controller_sn":        true,
```
dsupure, line 246: What is controller_sn? Why it's all lowercase?

haibinxie: it's an env variable required by ccm container, daemonset spec gen just pass through

piyush-nimbalkar (approved):

PR #624, opened 2022-05-19
Do not set px pod dns policy by default and add annotation to configu…

haibinxie:
…re it

### DIFF
--- a/drivers/storage/portworx/deployment.go
+++ b/drivers/storage/portworx/deployment.go
@@ -263,14 +263,15 @@ func (p *portworx) GetStoragePodSpec(
 
 	containers := t.portworxContainer(cluster)
 	podSpec := v1.PodSpec{
-		// For Pods running with hostNetwork, you should explicitly set its DNS policy to DNSClusterFirstWithHostNet
-		DNSPolicy:          v1.DNSClusterFirstWithHostNet,
 		HostPID:            pxutil.IsHostPidEnabled(cluster),
 		HostNetwork:        true,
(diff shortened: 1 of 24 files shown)


drivers/storage/portworx/deployment.go
```diff
@@ -263,15 +263,16 @@ func (p *portworx) GetStoragePodSpec(
...
-		DNSPolicy:          v1.DNSClusterFirstWithHostNet,
 		HostPID:            pxutil.IsHostPidEnabled(cluster),
 		HostNetwork:        true,
 		RestartPolicy:      v1.RestartPolicyAlways,
 		ServiceAccountName: pxutil.PortworxServiceAccountName(cluster),
 		Containers:         []v1.Container{containers},
 		Volumes:            t.getVolumes(),
 	}
+	if cluster.Annotations[pxutil.AnnotationDNSPolicy] != "" {
+		podSpec.DNSPolicy = v1.DNSPolicy(cluster.Annotations[pxutil.AnnotationDNSPolicy])
```
pure-jliao, line 274: we have consts for DNS policies, think we can check if the user provided string can be parsed to those consts? if cannot we can set it to none, this could add a layer of validation for the annotation input.

haibinxie: I intended to leave it to k8s to decide, if it is not provided actually it is ClusterFirst instead of none, if it is invalid it might be clusterFirst although I did not test but let's leave it to k8s

pure-jliao (approved):

zoxpx (approved):
Looks good -- thanks Michael

PR #626, opened 2022-05-20
[OPERATOR-696] Handle service type annotation migration

pure-jliao:

Closes #

### DIFF
--- a/pkg/migration/generate.go
+++ b/pkg/migration/generate.go
@@ -56,7 +56,10 @@ var (
 func (h *Handler) createStorageCluster(
 	ds *appsv1.DaemonSet,
 ) (*corev1.StorageCluster, error) {
-	stc := h.constructStorageCluster(ds)
+	stc, err := h.constructStorageCluster(ds)
+	if err != nil {
+		return nil, err
+	}
 
 	if err := h.addCSISpec(stc, ds); err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


pkg/migration/generate.go
```diff
@@ -467,6 +467,9 @@ func (h *Handler) constructStorageCluster(ds *appsv1.DaemonSet) *corev1.StorageC
 	// Populate StorageCluster spec.volumes from extra volumes mounted to portworx container
 	cluster.Spec.Volumes = getVolumeSpecs(c.VolumeMounts, ds.Spec.Template.Spec.Volumes, builtinPxVolumeNames)
 
+	// Add service type annotation based on portworx and portworx-api service if it's not ClusterIP by default
+	h.addServiceTypeAnnotation(cluster)
```
haibinxie, line 471: the error should be handled, if there is network error or k8s api server returns some temporary error such as "server busy", this should be returned so we can retry. inside of addServiceTypeAnnotation if service is not found write a log and return nil error, otherwise return error.

pkg/migration/migration_test.go
```diff
@@ -2838,6 +2838,133 @@ func TestStorageClusterWithExtraListOfVolumes(t *testing.T) {
...
+	versionClient := fakek8sclient.NewSimpleClientset()
+	coreops.SetInstance(coreops.New(versionClient))
+	versionClient.Discovery().(*fakediscovery.FakeDiscovery).FakedServerVersion = &k8sversion.Info{
+		GitVersion: "v1.16.0",
+	}
+
+	migrator := New(ctrl)
+
+	go migrator.Start()
+	time.Sleep(2 * time.Second)
```
haibinxie, line 2906: nit: we should avoid use sleep, which may be unstable and also slows test performance.
instead use PollImmediate 
 such as https://github.com/libopenstorage/operator/blob/master/pkg/migration/backup_test.go#L174

PR #629, opened 2022-05-23
[OPERATOR-511] Migrate portworx pod custom annotations

pure-jliao:

Closes #

### DIFF
--- a/pkg/migration/generate.go
+++ b/pkg/migration/generate.go
@@ -19,6 +19,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	meta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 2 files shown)


pkg/migration/generate.go
```diff
@@ -830,6 +835,45 @@ func (h *Handler) addMonitoringSpec(cluster *corev1.StorageCluster, ds *appsv1.D
...
+	)
+	if err != nil {
+		return err
+	}
+
+	customAnnotations := make(map[string]string)
+	for _, pod := range podList.Items {
+		owner := metav1.GetControllerOf(&pod)
+		if owner != nil && owner.UID == ds.UID && pod.Annotations != nil {
+			for k, v := range pod.Annotations {
```
haibinxie, line 856: there is no need to get pod, annotation is in daemonset.

pure-jliao: I thought user could just annotate the pod manually. so do we look at the daemonset only? or look at both the pods and daemonset? @haibinxie

haibinxie: the user wont have access to pod when first deploying it. if users change the pod later directly, then when they upgrade portworx (px version or other stuff), the annotation will be changed back during daemonset rolling upgrade. In any case, I think it's correct to refer to pod as source of truth. so let's keep it unchanged.

haibinxie (approved):

pkg/migration/generate.go
piyush-nimbalkar: I agree with Michael about just checking daemonset. If customers are adding annotation directly to pods, then it is not our responsibility to copy them as they are anyway ephemeral. When the DS is updated, the annotations are going to go away. We should only pick up the persisted annotations from the DS.

pkg/migration/generate.go
```diff
@@ -830,6 +835,45 @@ func (h *Handler) addMonitoringSpec(cluster *corev1.StorageCluster, ds *appsv1.D
...
+				if k != constants.AnnotationPodSafeToEvict {
```
piyush-nimbalkar, line 857: Almost all spec generator daemonsets have this annotation - `portworx.com/install-source`. Let's skip that too.

PR #642, opened 2022-06-07
[OPERATOR-719] Enable CSI snapshot controller by default

pure-jliao:

* Enable csi snapshot controller by default
* Scan all pods once on operator start to find if there's a snapshot controller preinstalled
* Skip installing the snapshot controller if found preinstalled controller
* User need to delete existing controller and restart operator to use operator managed controller

Closes #OPERATOR-719

### DIFF
--- a/drivers/storage/portworx/component/csi.go
+++ b/drivers/storage/portworx/component/csi.go
@@ -65,12 +65,13 @@ var (
 )
 
 type csi struct {
-	isCreated              bool
-	csiNodeInfoCRDCreated  bool
-	csiSnapshotCRDsCreated bool
-	k8sClient              client.Client
-	k8sVersion             version.Version
-	scheme                 *runtime.Scheme
+	isCreated                         bool
+	csiNodeInfoCRDCreated             bool
+	csiSnapshotCRDsCreated            bool
+	csiSnapshotControllerPreInstalled *bool
+	k8sClient                         client.Client
+	k8sVersion                        version.Version
+	scheme                            *runtime.Scheme
 }
 
 func (c *csi) Name() string {
@@ -471,10 +472,19 @@ func (c *csi) createDeployment(
 	if cluster.Spec.CSI.InstallSnapshotController != nil &&
 		*cluster.Spec.CSI.InstallSnapshotController &&
 		cluster.Status.DesiredImages.CSISnapshotController != "" {
-		snapshotControllerImage = util.GetImageURN(
-			cluster,
-			cluster.Status.DesiredImages.CSISnapshotController,
-		)
+		// Check if a snapshot controller was installed already using image "snapshot-controller",
+		// Only do this once to avoid scanning all pods frequently.
+		if c.csiSnapshotControllerPreInstalled == nil {
+			if err := c.findPreinstalledCSISnapshotController(); err != nil {
+				return err
+			}
+		}
+		if !*c.csiSnapshotControllerPreInstalled {
+			snapshotControllerImage = util.GetImageURN(
+				cluster,
+				cluster.Status.DesiredImages.CSISnapshotController,
+			)
+		}
 	}
 
 	if csiConfig.IncludeHealthMonitorController && cluster.Status.DesiredImages.CSIHealthMonitorController != "" {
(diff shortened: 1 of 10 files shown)


drivers/storage/portworx/component/csi.go
```diff
@@ -513,6 +523,24 @@ func (c *csi) createDeployment(
 	return nil
 }
 
+func (c *csi) findPreinstalledCSISnapshotController() error {
+	podList := &v1.PodList{}
+	err := c.k8sClient.List(context.TODO(), podList)
```
haibinxie, line 528: this would list the default namespace I think. @ggriffiths shall we search all namespaces?

ggriffiths: I would search in all namespaces only once as that is expensive

pure-jliao: from unit test this already list pods from all namespaces, and we only list once in the first reconcile loop.

ggriffiths: Sounds good :+1:

drivers/storage/portworx/components_test.go
```diff
@@ -12970,6 +13000,214 @@ func TestServiceTypeAnnotation(t *testing.T) {
...
+	require.True(t, *cluster.Spec.CSI.InstallSnapshotController)
+	require.NotEmpty(t, cluster.Status.DesiredImages.CSISnapshotController)
+
+	expectedDeployment = testutil.GetExpectedDeployment(t, "csiDeployment_1.0_k8s120.yaml")
+	deployment = &appsv1.Deployment{}
+	err = testutil.Get(driver.k8sClient, deployment, component.CSIApplicationName, cluster.Namespace)
+	require.NoError(t, err)
+	require.Equal(t, expectedDeployment.Spec, deployment.Spec)
+
+	// TestCase: If snapshot controller exists before creating the StorageCluster, don't create the controller
```
haibinxie, line 13144: please have two test cases for snapshot controller exists in same namespace and different namespace

drivers/storage/portworx/components_test.go
```diff
@@ -12970,6 +13000,214 @@ func TestServiceTypeAnnotation(t *testing.T) {
...
+			Image: "portworx/image:2.6.1",
+		},
+	}
+	driver.SetDefaultsOnStorageCluster(cluster)
+	err = driver.PreInstall(cluster)
+	require.NoError(t, err)
+	require.NotNil(t, cluster.Spec.CSI)
+	require.True(t, cluster.Spec.CSI.Enabled)
+	require.NotNil(t, cluster.Spec.CSI.InstallSnapshotController)
+	require.True(t, *cluster.Spec.CSI.InstallSnapshotController)
```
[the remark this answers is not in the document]
pure-jliao: we discussed yesterday, that requires updating the stc in component which is not good. This case the flag is true by default as it should be, just the snapshot controller is not installed as there's one already. the user need to restart the operator anyway to use the operator managed snapshot controller.

drivers/storage/portworx/portworx.go
```diff
@@ -235,6 +235,12 @@ func (p *portworx) SetDefaultsOnStorageCluster(toUpdate *corev1.StorageCluster)
 		}
 
 		if pxutil.IsCSIEnabled(toUpdate) {
+			// Enable CSI snapshot controller by default if it's not configured on k8s 1.17+
+			if toUpdate.Spec.CSI.InstallSnapshotController == nil &&
+				p.k8sVersion != nil && p.k8sVersion.GreaterThanOrEqual(k8sutil.K8sVer1_17) {
+				toUpdate.Spec.CSI.InstallSnapshotController = boolPtr(true)
+			}
+
```
piyush-nimbalkar, line 243: Let's move this code to https://github.com/libopenstorage/operator/blob/4ad3571c9d6d93ac68c5dcb567a101eb01407ffb/drivers/storage/portworx/portworx.go#L721
This is not the right place to set the CSI default parameters.

piyush-nimbalkar (approved):
LGTM after addressing the comment

drivers/storage/portworx/portworx.go
pure-jliao: @piyush-nimbalkar we need k8s version to set it only for k8s 1.17+, but I don't see anything else in SetPortworxDefaults is k8s version related, so I put it here, but I could pass the k8s version to `SetPortworxDefaults` as well. wdyt?

pure-jliao: so there's another function I can call in SetPortworxDefaults is `GetVersion()` which uses `coreops.Instance()`, so this will work. I agree it's better to move the function and I originally did this, and it shouldn't hurt to call the GetVersion function, I'll change it.

haibinxie (approved):

drivers/storage/portworx/portworx.go
pure-jliao: I chose to pass the k8s version to SetPortworxDefaults as the final solution.

PR #645, opened 2022-06-13
OPERATOR-752: cleanup px created configmaps

haibinxie:

Closes #

### DIFF
--- a/pkg/controller/storagecluster/controller_test.go
+++ b/pkg/controller/storagecluster/controller_test.go
@@ -2736,6 +2736,154 @@ func TestUpdateDriverWithInstanceInformation(t *testing.T) {
 	require.Empty(t, recorder.Events)
 }
 
+func TestGarbageCollection(t *testing.T) {
+	mockCtrl := gomock.NewController(t)
+	defer mockCtrl.Finish()
+
+	cluster := createStorageCluster()
+
+	// Configmaps created by px are expected to be deleted https://portworx.atlassian.net/browse/OPERATOR-752
+	configMaps := []v1.ConfigMap{
+		{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "px-attach-driveset-lock",
+				Namespace: cluster.Namespace,
+			},
+			Data: map[string]string{},
+		},
+		{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "px-attach-driveset-lock",
+				Namespace: "kube-system",
+			},
+			Data: map[string]string{},
+		},
+		{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "px-bringup-queue-lockb",
+				Namespace: cluster.Namespace,
+			},
+			Data: map[string]string{},
+		},
+		{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "px-bringup-queue-locka",
+				Namespace: "kube-system",
+			},
+			Data: map[string]string{},
+		},
+		{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "cm1",
+				Namespace: cluster.Namespace,
+				Annotations: map[string]string{
+					"operator.libopenstorage.org/garbage-collection": "true",
+				},
+			},
+			Data: map[string]string{},
+		},
+		{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "cm2",
+				Namespace: "kube-system",
+				Annotations: map[string]string{
+					"operator.libopenstorage.org/garbage-collection": "true",
+				},
+			},
+			Data: map[string]string{},
+		},
+	}
+	nonGCConfigMaps := []v1.ConfigMap{
+		{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "a",
+				Namespace: cluster.Namespace,
+			},
+			Data: map[string]string{},
+		},
+	}
+
+	k8sVersion, _ := version.NewVersion(minSupportedK8sVersion)
+	driver := testutil.MockDriver(mockCtrl)
+	k8sClient := testutil.FakeK8sClient(cluster)
+	for _, cm := range configMaps {
+		err := k8sClient.Create(context.TODO(), &cm)
+		require.NoError(t, err)
+	}
+	for _, cm := range nonGCConfigMaps {
+		err := k8sClient.Create(context.TODO(), &cm)
+		require.NoError(t, err)
+	}
+
+	controller := Controller{
+		client:            k8sClient,
+		Driver:            driver,
+		kubernetesVersion: k8sVersion,
+		nodeInfoMap:       make(map[string]*k8s.NodeInfo),
+	}
+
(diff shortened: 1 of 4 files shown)


piyush-nimbalkar (commented):
Not sure if we need to do all of this to just delete the config maps. We already have some logic to delete configmaps - https://github.com/libopenstorage/operator/blob/master/drivers/storage/portworx/uninstall.go#L133
We can just change that code to list and delete the known configmaps.

haibinxie:
> Not sure if we need to do all of this to just delete the config maps. We already have some logic to delete configmaps - https://github.com/libopenstorage/operator/blob/master/drivers/storage/portworx/uninstall.go#L133 We can just change that code to list and delete the known configmaps.

@piyush-nimbalkar i moved this to operator bcz this is general, after this PR we won't need to handle deletion in the future anymore, anyone (px, telemetry) wants to let operator delete something just add the annotation in their object.

piyush-nimbalkar:
> i moved this to operator bcz this is general, after this PR we won't need to handle deletion in the future anymore, anyone (px, telemetry) wants to let operator delete something just add the annotation in their object.

Sounds good to me and will work well in this situation. 
In general and for future, we should be careful when we add operator awareness in px. It could cause dependency from both sides and make changing operator code very difficult to maintain backward compatibility.

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -649,6 +658,45 @@ func (c *Controller) deleteStorageCluster(
...
+	if obj.GetObjectKind().GroupVersionKind().Kind == "ConfigMap" {
+		if obj.GetName() == "px-attach-driveset-lock" ||
+			strings.HasPrefix(obj.GetName(), "px-bringup-queue-lock") {
+			return true
+		}
+	}
+
+	return false
+}
+
```
piyush-nimbalkar, line 699: This is very portworx specific. Can we move at least this or the entire removeResources function to the portworx driver? Let's not spill px specific code in the controller.

pkg/util/k8s/k8s.go
```diff
@@ -1836,3 +1836,173 @@ func GetV1beta1CRDFromFile(
 	}
 	return crd, nil
 }
+
+// GetAllObjects gets all objects from k8s
+func GetAllObjects(k8sClient client.Client, namespace string) ([]client.Object, error) {
```
piyush-nimbalkar, line 1841: nit: Do we really need to go overboard and include all the objects?
I guess just secrets, configmaps, services, endpoints and storageclasses should do it.

pkg/util/k8s/k8s.go
```diff
@@ -1836,3 +1836,173 @@ func GetV1beta1CRDFromFile(
...
+	} else if l, ok := list.(*monitoringv1.PrometheusList); ok {
+		for _, o := range l.Items {
+			o.Kind = "Prometheus"
+			*objs = append(*objs, o.DeepCopy())
+		}
+	} else if l, ok := list.(*monitoringv1.AlertmanagerList); ok {
+		for _, o := range l.Items {
+			o.Kind = "Alertmanager"
+			*objs = append(*objs, o.DeepCopy())
+		}
```
piyush-nimbalkar, line 2000: Do we need to check all these objects?

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -649,6 +658,45 @@ func (c *Controller) deleteStorageCluster(
...
+	for _, obj := range objs {
+		gcAnnotated := false
+		if obj.GetAnnotations() != nil {
+			gcAnnotated, _ = strconv.ParseBool(obj.GetAnnotations()[constants.AnnotationGarbageCollection])
+		}
+
+		if gcAnnotated || c.gcNeeded(obj) {
+			logrus.Infof("Garbage collect object %s %s", obj.GetObjectKind().GroupVersionKind().Kind, obj.GetName())
+			if err := c.client.Delete(context.TODO(), obj); err != nil {
```
pure-jliao, line 675: need to ignore `NotFound` errors

pkg/util/k8s/k8s.go
pure-jliao: doesn't hurt IMO probably need those somedays.

pure-jliao (commented):
So moving forward resources added by operator would be managed by operator of course, external added resources like other config maps generated by Portworx, the creator should be responsible to add this annotation to the object right?

pkg/controller/storagecluster/controller_test.go
```diff
@@ -2736,6 +2736,154 @@ func TestUpdateDriverWithInstanceInformation(t *testing.T) {
...
 }
 
+func TestGarbageCollection(t *testing.T) {
+	mockCtrl := gomock.NewController(t)
+	defer mockCtrl.Finish()
+
+	cluster := createStorageCluster()
+
+	// Configmaps created by px are expected to be deleted https://portworx.atlassian.net/browse/OPERATOR-752
+	configMaps := []v1.ConfigMap{
```
dsupure, line 2746: Can we have a list and append 1 by 1? Instead hard-coded inside of the test?

pkg/util/k8s/k8s.go
```diff
@@ -1836,3 +1836,173 @@ func GetV1beta1CRDFromFile(
 	}
 	return crd, nil
 }
+
+// GetAllObjects gets all objects from k8s
+func GetAllObjects(k8sClient client.Client, namespace string) ([]client.Object, error) {
+	var objs []client.Object
+
+	if err := AppendObjectList(k8sClient, namespace, &v1.ServiceList{}, &objs); err != nil {
```
dsupure, line 1844: Can we change this to be a loop?

pkg/util/k8s/k8s.go
```diff
@@ -1836,3 +1836,173 @@ func GetV1beta1CRDFromFile(
...
+		list,
+		&client.ListOptions{
+			Namespace: namespace,
+		})
+	if err != nil {
+		logrus.WithError(err).Errorf("failed to list objects %v", list.GetObjectKind())
+		return err
+	}
+
+	if l, ok := list.(*v1.ServiceList); ok {
```
dsupure, line 1906: Can we change this to be a loop?

pkg/controller/storagecluster/storagecluster.go
haibinxie: It should not happen as the object is found before deleting it.

pkg/util/k8s/k8s.go
haibinxie: good idea, i will make a change

pkg/util/k8s/k8s.go
haibinxie: looks like this type conversion is hard to do in list, but i changed the other one into loop

pkg/controller/storagecluster/controller_test.go
dsupure: I mean do a for loop.

dsupure (approved):
LGTM!

pure-jliao (approved):

PR #649, opened 2022-06-14
OPERATOR-773: Update to new base image for security scan

haibinxie:


pure-jliao (approved):
LGTM! would be nice if this would fix the issue. but how can we confirm?

dsupure (commented):
Do we always want to pull the latest? It's better to use a fix version. Or you might hit a regression without any notification and it's hard to debug.

piyush-nimbalkar:
> Do we always want to pull the latest? It's better to use a fix version. Or you might hit a regression without any notification and it's hard to debug.

I agree. Using latest may not always be good. It is better if we manually update it every release and test it explicitly during that release.

haibinxie:
> > Do we always want to pull the latest? It's better to use a fix version. Or you might hit a regression without any notification and it's hard to debug.
> 
> I agree. Using latest may not always be good. It is better if we manually update it every release and test it explicitly during that release.

@dsupure @piyush-nimbalkar the new base image is required in image scan anyway. using the tag explicitly would make we integrate and test it later. and we don't always monitor it so always get last minute failure.

haibinxie:
@dsupure @piyush-nimbalkar in the last release and this release, we both found this in the last minute, no functional tests were rerun due to this, so the sooner the better using latest.

PR #674, opened 2022-06-24
OPERATOR-785: stop creating more kvdb pods when hitting outOfPods status

haibinxie:
Fixing the customer escalation. When kvdb runs into outofpods status, operator will keep creating more pods, as the controller client API does not return the kvdb in outOfPods status, so operator does not know the pod exists. switch to sched-ops API and delete the outofpods pods first.
https://portworx.atlassian.net/browse/CEE-400

### DIFF
--- a/pkg/controller/storagenode/storagenode.go
+++ b/pkg/controller/storagenode/storagenode.go
@@ -3,6 +3,7 @@ package storagenode
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/hashicorp/go-version"
@@ -12,6 +13,7 @@ import (
 	"github.com/libopenstorage/operator/pkg/util"
 	"github.com/libopenstorage/operator/pkg/util/k8s"
 	apiextensionsops "github.com/portworx/sched-ops/k8s/apiextensions"
+	coreops "github.com/portworx/sched-ops/k8s/core"
 	"github.com/sirupsen/logrus"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


pkg/controller/storagenode/storagenode.go
```diff
@@ -234,18 +236,27 @@ func (c *Controller) syncKVDB(
...
+	podList, err := coreops.Instance().GetPods(cluster.Namespace, c.kvdbPodLabels(cluster))
 	if err != nil {
 		return err
 	}
+	var kvdbPods []v1.Pod
+	for _, p := range podList.Items {
+		if p.Spec.NodeName == storageNode.Name {
+			// Let's delete the pod if it's in outOfPods status, so that a new one will be created.
+			if strings.ToLower(p.Status.Reason) == "outofpods" {
+				logrus.Warningf("Found pod with outOfPods status, will delete it: %v", p)
```
pure-jliao, line 250: nit: `OutOfpods` status. Also could we use `%#v` to print formatted Pod, or just print the pod namespace/name?

pkg/controller/storagenode/storagenode.go
```diff
@@ -234,18 +236,27 @@ func (c *Controller) syncKVDB(
...
+	// operator would keep creating kvdb pods in this case. https://portworx.atlassian.net/browse/CEE-400
+	podList, err := coreops.Instance().GetPods(cluster.Namespace, c.kvdbPodLabels(cluster))
 	if err != nil {
 		return err
 	}
+	var kvdbPods []v1.Pod
+	for _, p := range podList.Items {
+		if p.Spec.NodeName == storageNode.Name {
+			// Let's delete the pod if it's in outOfPods status, so that a new one will be created.
+			if strings.ToLower(p.Status.Reason) == "outofpods" {
```
pure-jliao, line 249: nit: safer to use strings.Contains ?

pkg/controller/storagenode/storagenode.go
```diff
@@ -267,14 +278,14 @@ func (c *Controller) syncKVDB(
...
-		if !isBeingDeleted && !isRecentlyCreatedAfterNodeCordoned && len(kvdbPodList.Items) == 0 {
+		if !isBeingDeleted && !isRecentlyCreatedAfterNodeCordoned && len(kvdbPods) == 0 {
 			pod, err := c.createKVDBPod(cluster, storageNode)
 			if err != nil {
 				return err
 			}
 
 			c.log(storageNode).Infof("creating kvdb pod: %s/%s", pod.Namespace, pod.Name)
-			err = c.client.Create(context.TODO(), pod)
+			_, err = coreops.Instance().CreatePod(pod)
```
pure-jliao, line 288: I think it's safe to use the controller client API here and below as well? do we need to replace all of them?

pkg/controller/storagenode/storagenode.go
haibinxie: Namespace/name may not have enough information for debugging, as this runs into failed code path we should be ok with more info in the log. %#v requires go formatter, not sure if it's implemented, any reason in favor of it?

PR #688, opened 2022-07-11
OPERATOR-809: remove evicted kvdb pods during reconcile loop

haibinxie:

Closes #


david-enli (approved):
LGTM, can put the string on a const but I'm okay with the way it is especially given the test already passed.

PR #692, opened 2022-07-15
PWX-24421: Adding azure disk encryption integration test.

pp511:

disEncryptionSetID can be passed in the StorageCluster deviceSpecs to use the specific key associated with this disk encryption set for encryption in an AKS cluster. With this PR we are adding two test cases to test PX installation with this param.

Closes #PWX-24421

Test [**InstallWithInvalidDiskEncryptionSetID**] : Pass "invalid" diskEncryptionSetID in the device spec and check for NodeStartFailure event from K8.
Test[**InstallWithInvalidDiskEncryptionSetID**] : Pass a valid diskEncryptionSetID using --portworx-device-specs in the integration test param and PX installation should succeed.

### DIFF
--- a/test/integration_test/azure_disk_test.go
+++ b/test/integration_test/azure_disk_test.go
@@ -4,15 +4,17 @@
 package integrationtest
 
 import (
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/require"
+
 	corev1 "github.com/libopenstorage/operator/pkg/apis/core/v1"
 	"github.com/libopenstorage/operator/pkg/util/test"
 	"github.com/libopenstorage/operator/test/integration_test/types"
 	ci_utils "github.com/libopenstorage/operator/test/integration_test/utils"
-	"github.com/sirupsen/logrus"
-	"github.com/stretchr/testify/require"
-	"strings"
-	"testing"
-	"time"
 )
 
 var (
@@ -34,12 +36,13 @@ var TestAzureDiskEncryptionCases = []types.TestCase{
 					DeviceSpecs: &disksEnc,
 				},
 			}
-			cluster.Spec.DeleteStrategy = &corev1.StorageClusterDeleteStrategy{
-				Type: corev1.UninstallAndWipeStorageClusterStrategyType,
-			}
 			return cluster
 		},
 		TestFunc: EncryptedDiskInstallFail,
+		ShouldSkip: func(tc *types.TestCase) bool {
+			return !ci_utils.IsAks
+
+		},
 	},
 	{
 		TestName:        "InstallWithDiskEncryptionSetID",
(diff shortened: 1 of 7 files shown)


test/integration_test/azure_disk_test.go
```diff
@@ -0,0 +1,123 @@
...
+			require.NoError(t, err)
+
+			cluster.Spec.CloudStorage = &corev1.CloudStorageSpec{
+				CloudStorageCommon: corev1.CloudStorageCommon{
+					DeviceSpecs: &disksEnc,
+				},
+			}
+			cluster.Spec.DeleteStrategy = &corev1.StorageClusterDeleteStrategy{
+				Type: corev1.UninstallAndWipeStorageClusterStrategyType,
+			}
```
piyush-nimbalkar, line 39: I believe we don't need this. When we call `UninstallAndValidateStorageCluster()`, it will always set the UninstallAndWipe strategy by default, if empty.

test/integration_test/azure_disk_test.go
```diff
@@ -0,0 +1,123 @@
...
+		TestrailCaseIDs: []string{"C82915"},
+		TestSpec: func(t *testing.T) interface{} {
+			cluster := &corev1.StorageCluster{}
+			cluster.Name = "px-cluster-azure-enc"
+			err := ci_utils.ConstructStorageCluster(cluster, ci_utils.PxSpecGenURL, ci_utils.PxSpecImages)
+			require.NoError(t, err)
+
+			cluster.Spec.DeleteStrategy = &corev1.StorageClusterDeleteStrategy{
+				Type: corev1.UninstallAndWipeStorageClusterStrategyType,
+			}
```
piyush-nimbalkar, line 55: Same as above.

test/integration_test/azure_disk_test.go
```diff
@@ -0,0 +1,123 @@
...
+				for _, s := range deviceSpecs {
+					if strings.Contains(s, "diskEncryptionSetID=") {
+						encryptedDiskParam = true
+						logrus.Infof("Param diskEncryptionSetID is present in deviceSpec")
+						break
+					}
+				}
+			}
+		}
+		require.True(t, encryptedDiskParam, "failed to validate the presence of diskEncryptionSetID in deviceSpec")
```
piyush-nimbalkar, line 107: May be it is okay to fail the test now, but I was thinking may be we should just log a warning and skip the test. If we use this test file for all azure tests, it may have a spec where encryption set is not specified.

piyush-nimbalkar (commented):
We should introduced a new parameter called `is-aks` ([refer here](https://github.com/libopenstorage/operator/blob/master/test/integration_test/main_test.go#L80)) and set that to true when running tests in Azure environment.

Once we have that, we should skip these tests if the flag is not set ([refer here](https://github.com/libopenstorage/operator/blob/master/test/integration_test/basic_test.go#L155)). This will ensure that these tests run only on azure env and are skipped for other environments.

test/integration_test/azure_disk_test.go
```diff
@@ -0,0 +1,123 @@
+//go:build integrationtest
```
pure-jliao, line 1: is this line needed?

test/integration_test/azure_disk_test.go
```diff
@@ -0,0 +1,123 @@
...
+// +build integrationtest
+
+package integrationtest
+
+import (
```
pure-jliao, line 6: nit: group the import into 3 groups: golang internal, external github, portworx repo. see other test files for reference.

test/integration_test/azure_disk_test.go
pp511: It is being auto generated in VS code

pure-jliao (approved):
LGTM

piyush-nimbalkar (approved):
As discussed offline, let's skip the test if drive specs don't have an encryption key.

PR #693, opened 2022-07-21
[OPERATOR-816] Handle clusterID with special characters during migration

pure-jliao:
**What this PR does / why we need it**: During migration if the cluster ID in DaemonSet cannot be a valid k8s object name, format the ID and save the original ID to STC annotation, so kvdb path and existing configmap names are not messed up.

Closes #

### DIFF
--- a/drivers/storage/portworx/deployment.go
+++ b/drivers/storage/portworx/deployment.go
@@ -772,7 +772,7 @@ func (t *template) getCloudProvider() string {
 
 func (t *template) getArguments() []string {
 	args := []string{
-		"-c", t.cluster.Name,
+		"-c", pxutil.GetClusterID(t.cluster),
 		"-x", "kubernetes",
 	}
 
--- a/drivers/storage/portworx/status.go
+++ b/drivers/storage/portworx/status.go
@@ -109,7 +109,8 @@ func (p *portworx) updateStorageNodes(
 	// If cluster is running internal kvdb, get current bootstrap nodes
 	kvdbNodeMap := make(map[string]*kvdb_api.BootstrapEntry)
 	if cluster.Spec.Kvdb != nil && cluster.Spec.Kvdb.Internal {
-		strippedClusterName := strings.ToLower(configMapNameRegex.ReplaceAllString(cluster.Name, ""))
+		clusterID := pxutil.GetClusterID(cluster)
+		strippedClusterName := strings.ToLower(configMapNameRegex.ReplaceAllString(clusterID, ""))
 		cmName := fmt.Sprintf("%s%s", internalEtcdConfigMapPrefix, strippedClusterName)
 
 		cm := &v1.ConfigMap{}
--- a/drivers/storage/portworx/util/util.go
+++ b/drivers/storage/portworx/util/util.go
@@ -123,6 +123,8 @@ const (
 	AnnotationHostPid = pxAnnotationPrefix + "/host-pid"
 	// AnnotationDNSPolicy configures dns policy for portworx pod.
 	AnnotationDNSPolicy = pxAnnotationPrefix + "/dns-policy"
+	// AnnotationClusterID overwrites portworx cluster ID, which is the storage cluster name by default
+	AnnotationClusterID = pxAnnotationPrefix + "/cluster-id"
 
 	// EnvKeyPXImage key for the environment variable that specifies Portworx image
 	EnvKeyPXImage = "PX_IMAGE"
@@ -1012,4 +1014,12 @@ func ApplyStorageClusterSettings(cluster *corev1.StorageCluster, deployment *app
 			}
 		}
 	}
+}
+
+// GetClusterID returns portworx instance cluster ID
+func GetClusterID(cluster *corev1.StorageCluster) string {
+	if cluster.Annotations[AnnotationClusterID] != "" {
+		return cluster.Annotations[AnnotationClusterID]
+	}
+	return cluster.Name
 }
--- a/pkg/migration/generate.go
+++ b/pkg/migration/generate.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"path"
+	"regexp"
 	"strconv"
 	"strings"
 
@@ -184,7 +185,12 @@ func (h *Handler) constructStorageCluster(ds *appsv1.DaemonSet) (*corev1.Storage
 	for i := 0; i < len(c.Args); i++ {
 		arg := c.Args[i]
 		if arg == "-c" {
-			cluster.Name = c.Args[i+1]
+			clusterID := c.Args[i+1]
+			cluster.Name = getStorageClusterNameFromClusterID(clusterID)
+			// If the cluster name got formatted, save the original cluster ID to storage cluster annotation
+			if cluster.Name != clusterID {
+				cluster.Annotations[pxutil.AnnotationClusterID] = clusterID
+			}
 			i++
 		} else if arg == "-x" {
 			i++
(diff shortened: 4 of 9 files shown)


drivers/storage/portworx/deployment.go
```diff
@@ -771,8 +771,13 @@ func (t *template) getCloudProvider() string {
 }
 
 func (t *template) getArguments() []string {
+	clusterID := t.cluster.Name
+	if t.cluster.Annotations != nil && t.cluster.Annotations[pxutil.AnnotationClusterID] != "" {
```
piyush-nimbalkar, line 775: nit: In golang, you don't need to check if a map is nil. You can directly try to check the key.
```
if t.cluster.Annotations[pxutil.AnnotationClusterID] != "" {
```

drivers/storage/portworx/status.go
```diff
@@ -109,7 +109,11 @@ func (p *portworx) updateStorageNodes(
 	// If cluster is running internal kvdb, get current bootstrap nodes
 	kvdbNodeMap := make(map[string]*kvdb_api.BootstrapEntry)
 	if cluster.Spec.Kvdb != nil && cluster.Spec.Kvdb.Internal {
-		strippedClusterName := strings.ToLower(configMapNameRegex.ReplaceAllString(cluster.Name, ""))
+		clusterID := cluster.Name
+		if cluster.Annotations != nil && cluster.Annotations[pxutil.AnnotationClusterID] != "" {
+			clusterID = cluster.Annotations[pxutil.AnnotationClusterID]
+		}
```
piyush-nimbalkar, line 115: Can we just extract this in a util function like this - `pxutil.GetClusterName(cluster)`?

pkg/migration/generate.go
```diff
@@ -1152,6 +1158,17 @@ func initRuntimeOptions(cluster *corev1.StorageCluster) {
 	}
 }
 
+func getStorageClusterNameFromClusterID(clusterID string) string {
+	// Must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character
```
piyush-nimbalkar, line 1162: Does the regex work for `.`? I don't see how.

pkg/migration/generate.go
```diff
@@ -1152,6 +1158,17 @@ func initRuntimeOptions(cluster *corev1.StorageCluster) {
...
+	// If the cluster ID is a valid k8s object name, just set it as the storage cluster name,
+	// otherwise format it to be k8s compatible
+	validNameRegex := regexp.MustCompile(`^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`)
+	if validNameRegex.MatchString(clusterID) {
+		return clusterID
+	}
+	return pxutil.GetK8sCompatibleObjectName(clusterID)
```
piyush-nimbalkar, line 1169: Just for migration, should we leave the hyphens? Otherwise the stc name could become unreadable.
Most people might have underscores in their names as separators. So we can replace the `_` with `-` first and then remove the rest of the unknown characters, lower case it, etc.
For instance, `System_Test_Cluster` can become `system-test-cluster` instead of `systemtestcluster`. 
Wdyt?

piyush-nimbalkar (commented):
Minor comments, but the PR looks good.
Great job on covering all the scenarios with UTs! 👏

drivers/storage/portworx/util/util.go
```diff
@@ -1013,3 +1016,9 @@ func ApplyStorageClusterSettings(cluster *corev1.StorageCluster, deployment *app
 		}
 	}
 }
+
+// GetK8sCompatibleObjectName formats an object name to be k8s compatible
+func GetK8sCompatibleObjectName(name string) string {
+	invalidCharRegex := regexp.MustCompile("[^a-zA-Z0-9]+")
+	return strings.ToLower(invalidCharRegex.ReplaceAllString(name, ""))
```
haibinxie, line 1023: Why convert to lower case?

drivers/storage/portworx/status.go
```diff
@@ -109,7 +109,11 @@ func (p *portworx) updateStorageNodes(
 	// If cluster is running internal kvdb, get current bootstrap nodes
 	kvdbNodeMap := make(map[string]*kvdb_api.BootstrapEntry)
 	if cluster.Spec.Kvdb != nil && cluster.Spec.Kvdb.Internal {
-		strippedClusterName := strings.ToLower(configMapNameRegex.ReplaceAllString(cluster.Name, ""))
+		clusterID := cluster.Name
+		if cluster.Annotations != nil && cluster.Annotations[pxutil.AnnotationClusterID] != "" {
```
haibinxie, line 113: we don't need to change here, the configMap name should always be valid here as long as cluster.Name is valid. I dont understand why we need configMapNameRegex.ReplaceAllString in the old code.

pkg/migration/generate.go
```diff
@@ -1152,6 +1158,17 @@ func initRuntimeOptions(cluster *corev1.StorageCluster) {
 	}
 }
 
+func getStorageClusterNameFromClusterID(clusterID string) string {
+	// Must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character
+	// If the cluster ID is a valid k8s object name, just set it as the storage cluster name,
+	// otherwise format it to be k8s compatible
+	validNameRegex := regexp.MustCompile(`^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`)
```
haibinxie, line 1165: we don't need to check here, why not directly use pxutil.GetK8sCompatibleObjectName?

pkg/migration/generate.go
pure-jliao: yes it works with `\\.`. Get the regex from the log, and verified the regex in a tester.

pkg/migration/generate.go
pure-jliao: I thought about it and did this originally, but I found how we process the configmap name and thought it's better to be consistent and share the same util function. removing all the special chars is more safe, and don't need to worry about other corner cases like tailing `-` or other invalid inputs. but if we don't see those corner cases I think it's good to just replace invalid chars with `-`

drivers/storage/portworx/status.go
pure-jliao: yes you're right, cm name should be valid if cluster name is valid. but we originally have this logic here, to strip the cluster name, remove all special chars. I see we have this vendor with the same logic https://github.com/libopenstorage/operator/blob/master/vendor/github.com/portworx/kvdb/api/bootstrap/k8s/k8s.go probably it's just for consistency, the kvdb path uses the same format. So I retain this logic to avoid any regressions caused by changing the logic here.

pkg/migration/generate.go
pure-jliao: my understand is, if the original cluster ID passed to the px container args is valid for an object name then just use it, so we don't need to set the annotation. the util function here is more strict and trims more chars from the name, so don't use it unless necessary. if we call the util function every time, we need to set the annotation always.

haibinxie: GetK8sCompatibleObjectName should not remove more chars than needed. above regex can be put in GetK8sCompatibleObjectName for validation and replacement (if the regex does not work for replacing entire string, we can still validate each char and remove)

if we call the util function every time, we need to set the annotation always.
you only add the annotation when the object is not the same as -c arg, that's after the util function is called anyway. correct?

drivers/storage/portworx/status.go
haibinxie: sounds good

pkg/migration/generate.go
piyush-nimbalkar: The configmap case is actually different. We do it that way because portworx creates the configmaps that way. Operator just does the work of deleting them on uninstall. So we had to generate the name accordingly.

If it is not very complicated, can we replace the `_` with `-`? Asking because it would be the most common scenario.
```
name := strings.Trim(strings.ReplaceAll(clusterID, "_", "-"), "-")
// after this do what you are doing already except that just keep the hyphen in the string/regex
```

drivers/storage/portworx/status.go
piyush-nimbalkar: Yes, let's not change the names here because portworx creates the name. We just use the same regex as Portworx to match the configmap name.

piyush-nimbalkar: I mean what @pure-jliao has done is correct.

pkg/migration/generate.go
piyush-nimbalkar: @haibinxie, he is using a much relaxed regex for validation, which will avoid unnecessary name replacement. 
`GetK8sCompatibleObjectName` function removes more chars because it is used by the configmap names. Unfortunately, that is how we will have to keep it to be consistent with the px generated names.

piyush-nimbalkar (approved):

PR #698, opened 2022-07-25
ccm-go integration

pure-jliao:
**What this PR does / why we need it**: telemetry ccm-go integration
Notes:
* ccm go is supported on PX 2.12+ only, older version will install ccm java
* when px upgrade to 2.12, if old ccm java is enabled, operator will upgrade to ccm go automatically , so telemetryProxy image must be available from manifest, otherwise should continue running old ccm java
* when px upgrade to 2.12, if old ccm java is enabled, but image is set explicitly, operator will reset the image and upgrade ccm
* during ccm upgrade, once old ccm components are deleted, operator will start creating new components. However it doesn't wait for all telemetry containers teardown in px pods.
* when disable telemetry, it will try to do a cleanup for both ccm java and ccm go

Closes #

### DIFF
--- a/deploy/ccm/phonehome-cluster.yaml
+++ b/deploy/ccm/phonehome-cluster.yaml
@@ -14,7 +14,7 @@ spec:
     spec:
       containers:
         - name: log-upload-service
-          image: pureezhang/log-upload:0.0.11
+          image: purestorage/log-upload:1.0.0
           volumeMounts:
           - name: etcpwx
             mountPath: /etc/pwx
--- a/deploy/ccm/phonehome-cluster.yaml
+++ b/deploy/ccm/phonehome-cluster.yaml
@@ -1,0 +1,98 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: phonehome-cluster
+spec:
+  replicas: 3
+  selector:
+    matchLabels:
+      role: phonehome-cluster
+  template:
+    metadata:
+      labels:
+        role: phonehome-cluster
+    spec:
+      containers:
+        - name: log-upload-service
+          image: pureezhang/log-upload:0.0.11
+          volumeMounts:
+          - name: etcpwx
+            mountPath: /etc/pwx
+          - mountPath: /var/cache
+            name: varcache
+          - mountPath: /var/log
+            name: journalmount2
+            readOnly: true
+          - mountPath: /var/cores
+            name: varcores
+          - readOnly: true
+            mountPath: /config
+            name: ccm-config
+          env:
+            - name: K8S_NODE_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: spec.nodeName
+          ports:
+            - name: loguploader
+              containerPort: 9090
+              hostPort: 9090
+              protocol: TCP
+          imagePullPolicy: Always
+        - name: envoy
+          image: envoyproxy/envoy:v1.22.2
+          securityContext:
+            runAsUser: 1111
+          volumeMounts:
+          - readOnly: true
+            mountPath: /config
+            name: proxy-config
+          - readOnly: true
+            mountPath: /etc/envoy/
+            name: tls-certificate
+          - readOnly: true
+            mountPath: /appliance-cert
+            name: pure-telemetry-certs
+          args:
+          - "envoy"
+          - "--config-path"
+          - "/config/envoy-config-rest.yaml"
+          ports:
+            - name: envoy
+              containerPort: 12002
+              hostPort: 12002
+              protocol: TCP
+      initContainers:
+        - name: init-cont
+          image: bitnami/kubectl:1.24.3
+          command: ['/bin/bash', '-c', "cert=$(kubectl get secrets -n kube-system --field-selector=metadata.name=pure-telemetry-certs -ojson |jq .items[0].data.cert); until [[ ${#cert} -gt 4 ]]; do echo waiting for cert generation; sleep 4; cert=$(kubectl get secrets -n kube-system --field-selector=metadata.name=pure-telemetry-certs -ojson |jq .items[0].data.cert); done;"]
+      volumes:
+        - name: etcpwx
+          hostPath:
+            path: /etc/pwx
+        - hostPath:
+            path: /var/cache
+          name: varcache
+        - name: varcores
+          hostPath:
+            path: /var/cores
+        - name: journalmount2
+          hostPath:
+            path: /var/log
+        - configMap:
+            items:
+              - key: ccm.properties
+                path: ccm.properties
+              - key: location
+                path: location
+            name: px-telemetry-config
+          name: ccm-config
+        - name: proxy-config
+          configMap:
+            name: proxy-config-rest
+        - name: tls-certificate
+          configMap:
+            name: tls-certificate
+        - name: pure-telemetry-certs
+          secret:
+            secretName: pure-telemetry-certs
--- a/deploy/crds/core_v1_storagecluster_crd.yaml
+++ b/deploy/crds/core_v1_storagecluster_crd.yaml
@@ -926,7 +926,7 @@ spec:
                       image:
                         type: string
                         description: Docker image of the telemetry container.
-                      imageLogUpload:
+                      logUploaderImage:
                         type: string
                         description: Docker image of the log-upload-service container.
               security:
@@ -1472,7 +1472,7 @@ spec:
                     description: Desired image for metrics collector proxy.
                   telemetryProxy:
                     type: string
-                    description: Desired image for telemetry.
+                    description: Desired image for telemetry proxy.
                   logUploader:
                     type: string
                     description: Desired image for log uploader.
--- a/deploy/crds/core_v1_storagecluster_crd.yaml
+++ b/deploy/crds/core_v1_storagecluster_crd.yaml
@@ -926,6 +926,9 @@ spec:
                       image:
                         type: string
                         description: Docker image of the telemetry container.
+                      imageLogUpload:
+                        type: string
+                        description: Docker image of the log-upload-service container.
               security:
                 type: object
                 description: Contains security configuration for the storage cluster.
@@ -1467,6 +1470,12 @@ spec:
                   metricsCollectorProxy:
                     type: string
                     description: Desired image for metrics collector proxy.
+                  telemetryProxy:
+                    type: string
+                    description: Desired image for telemetry.
+                  logUploader:
+                    type: string
+                    description: Desired image for log uploader.
                   pxRepo:
                     type: string
                     description: Desired image for pxRepo.
--- a/drivers/storage/portworx/component/telemetry.go
+++ b/drivers/storage/portworx/component/telemetry.go
@@ -3,14 +3,11 @@ package component
 import (
 	"context"
 	"fmt"
+	"io/ioutil"
+	"path"
 	"strings"
 
 	"github.com/hashicorp/go-version"
-	pxutil "github.com/libopenstorage/operator/drivers/storage/portworx/util"
-	corev1 "github.com/libopenstorage/operator/pkg/apis/core/v1"
-	"github.com/libopenstorage/operator/pkg/constants"
-	"github.com/libopenstorage/operator/pkg/util"
-	k8sutil "github.com/libopenstorage/operator/pkg/util/k8s"
 	"github.com/sirupsen/logrus"
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
@@ -22,25 +19,24 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	pxutil "github.com/libopenstorage/operator/drivers/storage/portworx/util"
+	corev1 "github.com/libopenstorage/operator/pkg/apis/core/v1"
+	"github.com/libopenstorage/operator/pkg/constants"
+	"github.com/libopenstorage/operator/pkg/util"
+	k8sutil "github.com/libopenstorage/operator/pkg/util/k8s"
+	testutil "github.com/libopenstorage/operator/pkg/util/test"
 )
 
 const (
 	// TelemetryComponentName name of the telemetry component
 	TelemetryComponentName = "Portworx Telemetry"
-	// TelemetryConfigMapName is the name of the config map that stores the telemetry configuration for CCM
-	TelemetryConfigMapName = "px-telemetry-config"
 	// TelemetryCCMProxyConfigMapName is the name of the config map that stores the PX_HTTP(S)_PROXY env var
 	TelemetryCCMProxyConfigMapName = "px-ccm-service-proxy-config"
 	// TelemetryCCMProxyFilePath path of the http proxy file
 	TelemetryCCMProxyFilePath = "/cache/network/http_proxy"
 	// TelemetryCCMProxyFileName name of the http proxy file
 	TelemetryCCMProxyFileName = "http_proxy"
-	// TelemetryPropertiesFilename is the name of the CCM properties file
-	TelemetryPropertiesFilename = "ccm.properties"
-	// TelemetryArcusLocationFilename is name of the file storing the location of arcus endpoint to use
-	TelemetryArcusLocationFilename = "location"
-	// TelemetryCertName is name of the telemetry cert.
-	TelemetryCertName = "pure-telemetry-certs"
 	// CollectorRoleName is name of the Role for metrics collector.
 	CollectorRoleName = "px-metrics-collector"
 	// CollectorRoleBindingName is name of the role binding for metrics collector.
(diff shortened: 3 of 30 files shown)


harsh-px:
@pure-jliao travis is failing static check

```
# staticcheck check ...
drivers/storage/portworx/component/telemetry.go:1114:9: empty branch (SA9003)
drivers/storage/portworx/component/telemetry.go:1155:9: empty branch (SA9003)
```

pkg/controller/storagecluster/update.go
```diff
@@ -716,6 +716,7 @@ func matchSelectedFields(
 }
 
 func doesTelemetryMatch(oldSpec, currentSpec *corev1.StorageClusterSpec) bool {
+	// TODO: check match for old CCM Java only, should not bound pod when using new ccm
```
harsh-px, line 719: Create a PWX for this and track it with the epic.

drivers/storage/portworx/util/util.go
```diff
@@ -962,6 +967,31 @@ func IsTelemetryEnabled(spec corev1.StorageClusterSpec) bool {
 		spec.Monitoring.Telemetry.Enabled
 }
 
+// IsCCMGoSupported returns true if px version is higher than 2.12
+func IsCCMGoSupported(pxVersion *version.Version) bool {
+	// TODO: remove after testing
```
harsh-px, line 972: Is this comment referring to removal of the ` || true` ? Considering we are prepping the PR to merge, let's remove this now.

drivers/storage/portworx/portworx.go
```diff
@@ -227,6 +229,22 @@ func (p *portworx) SetDefaultsOnStorageCluster(toUpdate *corev1.StorageCluster)
 			toUpdate.Status.DesiredImages.MetricsCollectorProxy = release.Components.MetricsCollectorProxy
 		}
 
+		logrus.Infof("JLIAO: log uploader version: %s", release.Components.LogUploader)
```
harsh-px, line 232: remove debugging logs. Same for other places too.

drivers/storage/portworx/manifest/manifest.go
```diff
@@ -29,9 +29,12 @@ const (
 	defaultAutopilotImage      = "portworx/autopilot:1.3.2.1"
 	defaultLighthouseImage     = "portworx/px-lighthouse:2.0.7"
 	defaultNodeWiperImage      = "portworx/px-node-wiper:2.10.0"
-	defaultTelemetryImage      = "purestorage/ccm-service:3.2.11"
+	defaultCCMJavaImage        = "purestorage/ccm-service:3.2.11"
 	defaultCollectorProxyImage = "envoyproxy/envoy:v1.21.4"
 	defaultCollectorImage      = "purestorage/realtime-metrics:1.0.1"
+	defaultCCMGoImage          = "pureezhang/ccm-go:0.0.8"
+	defaultLogUploaderImage    = "pureezhang/log-upload:0.0.11"
```
harsh-px, line 36: Why are these still referencing Eugene's private docker repo? This should now come from our official docker hub. Please work with Eugene to upload and use that.

drivers/storage/portworx/component/telemetry.go
```diff
@@ -59,12 +55,62 @@ const (
...
+	configParameterApplianceID                     = "APPLIANCE_ID"
+	configParameterComponentSN                     = "COMPONENT_SN"
+	configParameterProductVersion                  = "PRODUCT_VERSION"
+	configParameterRegisterProxyURL                = "REGISTER_PROXY_URL"
+	configParameterRestProxyURL                    = "REST_PROXY_URL"
+	configParameterCertSecretNamespace             = "CERT_SECRET_NAMESPACE"
+	configParameterCustomProxyPort                 = "CUSTOM_PROXY_PORT"
+	configParameterPortworxPort                    = "PORTWORX_PORT"
+
+	logUploaderPort = 9090 // TODO: avoid 1970 as used by old ccm
```
harsh-px, line 113: What is the TODO here? I thought we finalized using 9090.

drivers/storage/portworx/component/telemetry.go
```diff
@@ -133,6 +269,7 @@ func (t *telemetry) setTelemetryCertOwnerRef(
 
 	// The cert is created after ccm container starts, so we may not have it for a while.
 	if errors.IsNotFound(err) {
+		logrus.Infof("JLIAO: not found")
```
haibinxie, line 272: let's keep the log always for debugging

drivers/storage/portworx/component/telemetry.go
```diff
@@ -855,6 +993,476 @@ func (t *telemetry) reconcileCCMProxyConfigMap(
 	)
 }
 
+func (t *telemetry) reconcileCCMGoRolesAndRoleBindings(
```
haibinxie, line 996: nit: you may put old telemetry into telemetry-old.go, the new telemetry.go can just call the APIs of the old telemetry object. in that case we dont need to rename the APIs. just something to consider, dont want to add overhead as this is already done up to you evaluate

drivers/storage/portworx/deployment.go
```diff
@@ -287,7 +287,7 @@ func (p *portworx) GetStoragePodSpec(
 		}
 	}
 
-	if pxutil.IsTelemetryEnabled(cluster.Spec) {
+	if pxutil.IsTelemetryEnabled(cluster.Spec) && !pxutil.RunCCMGo(cluster) {
```
haibinxie, line 290: nit: use new and old ccm instead of go and java, easier to understand as other people may not know what language it's written

deploy/crds/core_v1_storagecluster_crd.yaml
```diff
@@ -1467,6 +1470,12 @@ spec:
                   metricsCollectorProxy:
                     type: string
                     description: Desired image for metrics collector proxy.
+                  telemetryProxy:
+                    type: string
+                    description: Desired image for telemetry.
```
haibinxie, line 1475: Desired image for telemetry proxy

pkg/apis/core/v1/storagecluster.go
```diff
@@ -545,6 +545,8 @@ type TelemetrySpec struct {
 	Enabled bool `json:"enabled,omitempty"`
 	// Image is docker image of the telemetry container
 	Image string `json:"image,omitempty"`
+	// ImageLogUpload is docker image of the log-upload-service container.
+	ImageLogUpload string `json:"imageLogUpload,omitempty"`
```
haibinxie, line 549: Use ImageLogUploader to be consistent

drivers/storage/portworx/manifest/manifest.go
```diff
@@ -312,6 +307,35 @@ func fillPrometheusDefaults(
...
+			rel.Components.Telemetry = defaultCCMJavaImage
+		}
+		if rel.Components.MetricsCollector == "" {
+			rel.Components.MetricsCollector = defaultCollectorImage
+		}
+		if rel.Components.MetricsCollectorProxy == "" {
+			rel.Components.MetricsCollectorProxy = defaultCollectorProxyImage
+		}
+	} else {
+		// New CCM Go telemetry
```
haibinxie, line 326: let's reset the images of the old ccm.

drivers/storage/portworx/manifest/manifest.go
```diff
@@ -312,6 +307,35 @@ func fillPrometheusDefaults(
...
+	pxVersion, err := version.NewSemver(rel.PortworxVersion)
+	if err != nil || !pxutil.IsCCMGoSupported(pxVersion) {
+		// Old CCM Java telemetry
+		if rel.Components.Telemetry == "" {
+			rel.Components.Telemetry = defaultCCMJavaImage
+		}
+		if rel.Components.MetricsCollector == "" {
+			rel.Components.MetricsCollector = defaultCollectorImage
+		}
+		if rel.Components.MetricsCollectorProxy == "" {
```
haibinxie, line 322: let's reset images of the new ccm to empty, just in case customer downgrade. add a unit test to upgrade then downgrade and add test case in testrail too

drivers/storage/portworx/deployment.go
piyush-nimbalkar: IMO `CCMGo` is better than calling old/new because if we later have another version of ccm, it will be get confusing to name it. Have some logical name is better.

drivers/storage/portworx/component/telemetry.go
piyush-nimbalkar: Yes, I like moving the old code to a different file and keeping the new components clean.

drivers/storage/portworx/util/util.go
```diff
@@ -962,6 +970,16 @@ func IsTelemetryEnabled(spec corev1.StorageClusterSpec) bool {
...
+// IsCCMGoSupported returns true if px version is higher than 2.12
+func IsCCMGoSupported(pxVersion *version.Version) bool {
+	return pxVersion.GreaterThanOrEqual(MinimumPxVersionCCMGO)
+}
+
+// RunCCMGo returns whether to run new ccm go or old ccm java based on if the telemetry proxy image is set
+func RunCCMGo(cluster *corev1.StorageCluster) bool {
+	return cluster.Status.DesiredImages.TelemetryProxy != ""
+}
+
```
piyush-nimbalkar, line 982: Don't think this util function correctly conveys the intent. Just checking the status.desiredImages does not tell if CCMGo should be run or not. Does this even need to be a util that anyone can consume? Can it be local to the caller?

drivers/storage/portworx/manifest/manifest.go
```diff
@@ -312,6 +307,35 @@ func fillPrometheusDefaults(
 	}
 }
 
+func fillTelemetryDefaults(
+	rel *Version,
+) {
+	pxVersion, err := version.NewSemver(rel.PortworxVersion)
+	if err != nil || !pxutil.IsCCMGoSupported(pxVersion) {
```
piyush-nimbalkar, line 314: Shouldn't this be `err == nil && !pxutil.IsCCMGoSupported(pxVersion)`?
For unreleased builds the PortworxVersion would be `sha_sha`, so the err will be non-nil. We can assume that the unreleased builds are newer versions, while older released versions will have a proper semantic version.

drivers/storage/portworx/manifest/manifest.go
```diff
@@ -312,6 +307,35 @@ func fillPrometheusDefaults(
...
+		// Old CCM Java telemetry
+		if rel.Components.Telemetry == "" {
+			rel.Components.Telemetry = defaultCCMJavaImage
+		}
+		if rel.Components.MetricsCollector == "" {
```
piyush-nimbalkar, line 319: We don't need to put all the images in if-else statements, just the `Telemetry` image.
```
if rel.Components.Telemetry == "" {
  pxVersion, err := version.NewSemver(rel.PortworxVersion)
  if err != nil || !pxutil.IsCCMGoSupported(pxVersion) {		
    rel.Components.Telemetry = defaultCCMJavaImage
  } else {
    rel.Components.Telemetry = defaultCCMGoImage
  }
}

if rel.Components.MetricsCollector == "" {
  ...
}

...

File: drivers/storage/portworx/manifest/manifest.go (hunk 12 above)
Comment (piyush-nimbalkar) [reply drivers/storage/portworx/manifest/manifest.go:322]: We don't need to reset the images here. The `Version` object can send all the images from the manifest. We handle what needs to be set from this in the desired images in `portworx.go`.

File: drivers/storage/portworx/portworx.go [hunk 16]
```diff
@@ -181,13 +181,21 @@ func (p *portworx) SetDefaultsOnStorageCluster(toUpdate *corev1.StorageCluster)
 			*toUpdate.Spec.AutoUpdateComponents == corev1.OnceAutoUpdate)
 		release := manifest.Instance().GetVersions(toUpdate, force)
 
+		logrus.Infof("JLIAO: versions in set px default: %s", release.Components)
+		logrus.Infof("JLIAO: has component changed %v", p.hasComponentChanged(toUpdate))
+		logrus.Infof("JLIAO: telemetry , auto update: %v, has changed: %v", autoUpdateTelemetry(toUpdate), hasTelemetryChanged(toUpdate))
+		logrus.Infof("JLIAO: telemetry proxy, auto update: %v, has changed: %v", autoUpdateTelemetryProxy(toUpdate), hasTelemetryProxyChanged(toUpdate))
+		logrus.Infof("JLIAO: metrics, auto update: %v, has changed: %v", autoUpdateMetricsCollector(toUpdate), hasMetricsCollectorChanged(toUpdate))
+		logrus.Infof("JLIAO: log uploader, auto update: %v, has changed: %v", autoUpdateLogUploader(toUpdate), hasLogUploaderChanged(toUpdate))
+
```
piyush-nimbalkar, line 190: Remove this and all other debug logs from this file.

drivers/storage/portworx/portworx.go
```diff
@@ -1016,8 +1072,21 @@ func autoUpdateTelemetry(cluster *corev1.StorageCluster) bool {
...
 func autoUpdateMetricsCollector(cluster *corev1.StorageCluster) bool {
-	return pxutil.IsTelemetryEnabled(cluster.Spec)
+	return pxutil.IsTelemetryEnabled(cluster.Spec) &&
+		!pxutil.RunCCMGo(cluster)
+}
+
+func autoUpdateLogUploader(cluster *corev1.StorageCluster) bool {
+	return pxutil.IsTelemetryEnabled(cluster.Spec) &&
+		pxutil.IsCCMGoSupported(pxutil.GetPortworxVersion(cluster)) &&
+		pxutil.RunCCMGo(cluster) &&
```
piyush-nimbalkar, line 1088: Why do we need to check this? Feels like the rest of the 3 conditions should suffice.

drivers/storage/portworx/portworx.go
```diff
@@ -1016,8 +1072,21 @@ func autoUpdateTelemetry(cluster *corev1.StorageCluster) bool {
...
 
+func autoUpdateTelemetryProxy(cluster *corev1.StorageCluster) bool {
+	return pxutil.IsTelemetryEnabled(cluster.Spec) &&
+		pxutil.IsCCMGoSupported(pxutil.GetPortworxVersion(cluster))
+}
+
 func autoUpdateMetricsCollector(cluster *corev1.StorageCluster) bool {
-	return pxutil.IsTelemetryEnabled(cluster.Spec)
+	return pxutil.IsTelemetryEnabled(cluster.Spec) &&
+		!pxutil.RunCCMGo(cluster)
```
piyush-nimbalkar, line 1082: Why can't we just use `pxutil.IsCCMGoSupported` here?

drivers/storage/portworx/portworx_test.go
```diff
@@ -254,9 +254,9 @@ func TestSetDefaultsOnStorageCluster(t *testing.T) {
 	driver.SetDefaultsOnStorageCluster(cluster)
 
 	// Use default image from release manifest when spec.image is not set
-	require.Equal(t, defaultPortworxImage+":3.0.0", cluster.Spec.Image)
-	require.Equal(t, "3.0.0", cluster.Spec.Version)
-	require.Equal(t, "3.0.0", cluster.Status.Version)
+	require.Equal(t, defaultPortworxImage+":2.10.0", cluster.Spec.Image)
```
piyush-nimbalkar, line 257: Can we add some tests in this function - https://github.com/libopenstorage/operator/blob/894e0117f6f130e9a623b77c4d9c55369a8724da/drivers/storage/portworx/portworx_test.go#L7264?

drivers/storage/portworx/deployment.go
piyush-nimbalkar, line 290: Can we not use `pxutil.IsCCMGoSupported` here?

pure-jliao: addressed offline.

drivers/storage/portworx/portworx.go
pure-jliao: addressed offline.

drivers/storage/portworx/portworx.go
pure-jliao: addressed offline.

deploy/ccm/phonehome-cluster.yaml
```diff
@@ -0,0 +1,98 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: phonehome-cluster
```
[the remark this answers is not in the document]
pure-jliao: created https://portworx.atlassian.net/browse/PWX-26142 to track

drivers/storage/portworx/deployment.go
pure-jliao: keeping go and java

drivers/storage/portworx/manifest/manifest.go
pure-jliao: we don't support downgrade, but on safe side I'll reset those to empty. as when px is supported it should only run new ccm

deploy/crds/core_v1_storagecluster_crd.yaml
```diff
@@ -926,6 +926,9 @@ spec:
                       image:
                         type: string
                         description: Docker image of the telemetry container.
+                      imageLogUpload:
```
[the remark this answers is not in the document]
pure-jliao: I'll call it imageLogUploader, I feel better to have the image prefix, so it could sorted correctly.

drivers/storage/portworx/manifest/manifest.go
pure-jliao: better not to set the rest images if old/new ccm is not enabled to be safe, though we expect everything from manifest is correct. for now it should be ok.

drivers/storage/portworx/component/telemetry.go
```diff
@@ -149,712 +224,516 @@ func (t *telemetry) setTelemetryCertOwnerRef(
...
-	return k8sutil.CreateOrUpdateConfigMap(
-		t.k8sClient,
-		&v1.ConfigMap{
-			ObjectMeta: metav1.ObjectMeta{
-				Name:            TelemetryConfigMapName,
-				Namespace:       cluster.Namespace,
-				OwnerReferences: []metav1.OwnerReference{*ownerRef},
-			},
-			Data: data,
+func (t *telemetry) createDeploymentPhonehomeCluster(
```
piyush-nimbalkar, line 533: In none of the deployments, we are setting the tolerations, node affinity, image pull secret, etc. Please check other deployments for comparison.

drivers/storage/portworx/component/telemetry.go
```diff
@@ -149,712 +224,516 @@ func (t *telemetry) setTelemetryCertOwnerRef(
...
+		} else if c.Name == containerNameTelemetryProxy {
+			c.Image = proxyImage
+		}
+		containers = append(containers, *c.DeepCopy())
+	}
+	deployment.Name = DeploymentNameRegistrationService
+	deployment.Namespace = cluster.Namespace
+	deployment.OwnerReferences = []metav1.OwnerReference{*ownerRef}
+	deployment.Spec.Template.Spec.Containers = containers
+	return k8sutil.CreateOrUpdateDeployment(t.k8sClient, deployment, ownerRef)
```
piyush-nimbalkar, line 530: This is not a good method to call. We avoid calling this is all the components, because it will call update deployment k8s API on every reconcile loop. We don't want to add load to the k8s api server unnecessarily.

piyush-nimbalkar (commented):
Overall, I don't think it is a good idea read multiple files on every reconcile loop. Especially because these files are not going to change until the operator upgrades. We should either cache them or just embed them in the code like we do for other components.

drivers/storage/portworx/util/util.go
```diff
@@ -450,6 +464,11 @@ func GetPortworxVersion(cluster *corev1.StorageCluster) *version.Version {
 				pxVersion, err = version.NewSemver(pxVersionStr)
 				if err != nil {
 					logrus.Warnf("Invalid PX version %s extracted from annotation: %v", pxVersionStr, err)
+					pxVersionStr = getPortworxVersionFromManifestURL(manifestURL)
```
piyush-nimbalkar, line 467: Just curious, why/where exactly do we need to get the px version from the manifest?

pure-jliao: because when we doing the upgrade we just set the env var `PX_RELEASE_MANIFEST_URL` to `"https://edge-install.portworx.com/2.12/version"`, the oci monitor image will be sha_sha, currently operator cannot get the version correctly without this.

drivers/storage/portworx/component/telemetry.go
```diff
@@ -149,712 +224,516 @@ func (t *telemetry) setTelemetryCertOwnerRef(
...
+	deployment.Spec.Template.Spec.Affinity = &v1.Affinity{
+		NodeAffinity: cluster.Spec.Placement.NodeAffinity,
+	}
+	deployment.Spec.Template.Spec.Containers = containers
+	deployment.Spec.Template.Spec.InitContainers = []v1.Container{{
+		Name:  "init-cont",
+		Image: "bitnami/kubectl:1.24.3",
+		Command: []string{
+			"/bin/bash", "-c",
+			fmt.Sprintf("cert=$(kubectl get secrets -n %s --field-selector=metadata.name=pure-telemetry-certs -ojson |jq .items[0].data.cert); until [[ ${#cert} -gt 4 ]]; do echo waiting for cert generation; sleep 4; cert=$(kubectl get secrets -n %s --field-selector=metadata.name=pure-telemetry-certs -ojson |jq .items[0].data.cert); done;",
```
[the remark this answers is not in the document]
pure-jliao: the original one doesn't work for custom namespace install, I'll create a ticket for this.

drivers/storage/portworx/component/telemetry.go
```diff
@@ -595,8 +614,26 @@ func (t *telemetry) createDeploymentPhonehomeCluster(
...
 	deployment.Spec.Replicas = &num
+	pxutil.ApplyStorageClusterSettings(cluster, deployment)
+
+	existingDeployment, err := k8sutil.GetDeployment(t.k8sClient, deployment)
+	if err != nil {
+		return err
+	}
 
-	return k8sutil.CreateOrUpdateDeployment(t.k8sClient, deployment, ownerRef)
+	updated, _ := util.DeepEqualDeployment(deployment, existingDeployment)
```
haibinxie, line 624: updated --> equal

drivers/storage/portworx/util/util.go
piyush-nimbalkar: If the version is sha_sha, normally this function returns a very high valid sem version. So any check against 2.12 should have passed and we should have enabled the ccm-go.  So, ideally we should not have needed this. Anyway, if it unblocks you for now, it is okay.

haibinxie:
> Overall, I don't think it is a good idea read multiple files on every reconcile loop. Especially because these files are not going to change until the operator upgrades. We should either cache them or just embed them in the code like we do for other components.

read files should be lightweight, even px is writing lots of logs

drivers/storage/portworx/component/telemetry.go
```diff
@@ -524,10 +529,28 @@ func (t *telemetry) createDeploymentRegistrationService(
...
 	deployment.Spec.Template.Spec.Containers = containers
-	return k8sutil.CreateOrUpdateDeployment(t.k8sClient, deployment, ownerRef)
+	pxutil.ApplyStorageClusterSettings(cluster, deployment)
+
+	existingDeployment, err := k8sutil.GetDeployment(t.k8sClient, deployment)
+	if err != nil {
+		return err
+	}
+
+	updated, _ := util.DeepEqualDeployment(deployment, existingDeployment)
```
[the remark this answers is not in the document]
pure-jliao: oh the same thing, if they are equal, then it's updated, I'll change this to make it clear.

piyush-nimbalkar:
> > Overall, I don't think it is a good idea read multiple files on every reconcile loop. Especially because these files are not going to change until the operator upgrades. We should either cache them or just embed them in the code like we do for other components.
> 
> read files should be lightweight, even px is writing lots of logs

Logs are buffered and then persisted to disk. Reading files is not going to be lightweight as it is finally an IO operation. It especially doesn't make sense as the file never changes.

pure-jliao:
> e we write logs, does not mean we should do more unnecessary d

I agree we should only read it once at the beginning. but allow reading it we can offer the capability to update the configs in realtime to resolve issues

drivers/storage/portworx/util/util.go
pure-jliao: seems yes, alright i got it, previously I didn't set it to run ccm-go if the version is invalid, for now without this it should work as well, but adding this shouldn't hurt.

haibinxie (approved):

piyush-nimbalkar:
> > e we write logs, does not mean we should do more unnecessary d
> 
> I agree we should only read it once at the beginning. but allow reading it we can offer the capability to update the configs in realtime to resolve issues

These are configs sitting inside the operator container. The user can always modify the configmap directly by setting the reconcile annotation to false. Reference - https://github.com/libopenstorage/operator/blob/54e8500e909454fcc62bd25e8a1eba77685946a2/pkg/util/k8s/k8s.go#L522

haibinxie:
> > > Overall, I don't think it is a good idea read multiple files on every reconcile loop. Especially because these files are not going to change until the operator upgrades. We should either cache them or just embed them in the code like we do for other components.
> > 
> > 
> > read files should be lightweight, even px is writing lots of logs
> 
> Logs are buffered and then persisted to disk. Reading files is not going to be lightweight as it is finally an IO operation. It especially doesn't make sense as the file never changes. Also, just because we write logs, does not mean we should do more unnecessary disk operations.

if there is a quick way to do the caching, we should do it, we shall leverage simplicity vs optimization

piyush-nimbalkar (approved):
LGTM. Good job on getting the ccm migration done.

Please ensure that we change the deployment, config map, etc names before we release the new ccm.
Also, think about reading files every time and see how we can cache and keep the operator lightweight.

pure-jliao:
> These are configs sitting inside the operator container. The user can always modify the configmap directly by setting the reconcile annotation to false. Reference -

Thanks @piyush-nimbalkar and @haibinxie . I've created tickets to track all the improvements.

PR #707, opened 2022-08-03
PWX-24825: Check for a specific secure port in pvc controller for AKS.

pp511:

Operator integration tests were failing on AKS in the function validatePvcControllerPorts with error `14:33:08 2022/08/02 21:28:27 failed to validate secure-port, secure-port is missing from annotations in the StorageCluster, but is found in the PVC Controler pod portworx-pvc-controller-99f55f488-5q79p Next retry in: 20s`
Azure --secure-port get updated to value  AksPVCControllerSecurePort = "10261" in pvc controller irrespective of the StorageCluster annotation.  Updating this function to compare  secure-port against this value in case of AKS.

### DIFF
--- a/pkg/util/test/util.go
+++ b/pkg/util/test/util.go
@@ -1716,16 +1716,16 @@ func validatePvcControllerPorts(cluster *corev1.StorageCluster, pvcControllerDep
 					if len(container.Command) > 0 {
 						for _, containerCommand := range container.Command {
 							if strings.Contains(containerCommand, "--secure-port") {
-								if isAKS(cluster) {
-									if strings.Split(containerCommand, "=")[1] != AksPVCControllerSecurePort {
-										return nil, true, fmt.Errorf("failed to validate secure-port, secure-port is missing in the PVC Controler pod %s", pod.Name)
-									}
-								} else {
-									if len(pvcSecurePort) == 0 {
+								if len(pvcSecurePort) == 0 {
+									if isAKS(cluster) {
+										if strings.Split(containerCommand, "=")[1] != AksPVCControllerSecurePort {
+											return nil, true, fmt.Errorf("failed to validate secure-port, secure-port is missing in the PVC Controler pod %s", pod.Name)
+										}
+									} else {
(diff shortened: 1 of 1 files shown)


pkg/util/test/util.go
```diff
@@ -1713,10 +1716,16 @@ func validatePvcControllerPorts(annotations map[string]string, pvcControllerDepl
...
+								if isAKS(cluster) {
+									if strings.Split(containerCommand, "=")[1] != AksPVCControllerSecurePort {
+										return nil, true, fmt.Errorf("failed to validate secure-port, secure-port is missing in the PVC Controler pod %s", pod.Name)
+									}
+								} else {
+									if len(pvcSecurePort) == 0 {
+										return nil, true, fmt.Errorf("failed to validate secure-port, secure-port is missing from annotations in the StorageCluster, but is found in the PVC Controler pod %s", pod.Name)
+									} else if pvcSecurePort != strings.Split(containerCommand, "=")[1] {
+										return nil, true, fmt.Errorf("failed to validate secure-port, wrong --secure-port value in the command in PVC Controller pod [%s]: expected: %s, got: %s", pod.Name, pvcSecurePort, strings.Split(containerCommand, "=")[1])
+									}
```
piyush-nimbalkar, line 1728: If the `portworx.io/pvc-controller-secure-port` is set, we need to overwrite the port even if it is AKS. So the correct testing order should be like below - 
```
if (pvcSecurePort) == 0 {
  if isAKS(cluster) && if strings.Split(containerCommand, "=")[1] != AksPVCControllerSecurePort {
      return nil, true, fmt.Errorf("failed to validate secure-port, secure-port on the PVC Controler pod %s is invalid", pod.Name)
  }
  return nil, true, fmt.Errorf("failed to validate secure-port, secure-port is missing from annotations in the StorageCluster, but is found in the PVC Controler pod %s", pod.Name) 
} else if pvcSecurePort != strings.Split(containerCommand, "=")[1] {
  ...
}
```

piyush-nimbalkar (approved):

PR #710, opened 2022-08-09
[PWX-26146][PWX-26177] Add ccm-go missing scc configuration and metrics collector

pure-jliao:
1. Add missing scc configuration for ocp environment to allow pod creation in custom namespace.
* service account px-telemetry
* cluster role px-telemetry
* cluster role binding px-telemetry
2. Add metrics collector back to ccm-go components

Closes #

### DIFF
--- a/drivers/storage/portworx/component/telemetry.go
+++ b/drivers/storage/portworx/component/telemetry.go
@@ -21,6 +21,7 @@ import (
 
 	pxutil "github.com/libopenstorage/operator/drivers/storage/portworx/util"
 	corev1 "github.com/libopenstorage/operator/pkg/apis/core/v1"
+	"github.com/libopenstorage/operator/pkg/constants"
 	"github.com/libopenstorage/operator/pkg/util"
 	k8sutil "github.com/libopenstorage/operator/pkg/util/k8s"
 )
@@ -48,6 +49,12 @@ const (
 	DeploymentNameRegistrationService = "registration-service"
 	// DeploymentNamePhonehomeCluster is name of deployment phonehome-cluster
 	DeploymentNamePhonehomeCluster = "phonehome-cluster"
+	// ServiceAccountNamePxTelemetry is name of service account px-telemetry
+	ServiceAccountNamePxTelemetry = "px-telemetry"
+	// ClusterRoleNamePxTelemetry is name of cluster role px-telemetry
+	ClusterRoleNamePxTelemetry = "px-telemetry"
+	// ClusterRoleBindingNamePxTelemetry is name of cluster role binding px-telemetry
+	ClusterRoleBindingNamePxTelemetry = "px-telemetry"
 
 	roleFileNameSecretManager             = "secret-manager-role.yaml"
 	roleBindingFileNameSecretManager      = "secret-manager-role-binding.yaml"
@@ -178,6 +185,11 @@ func (t *telemetry) reconcileCCMGo(
 	cluster *corev1.StorageCluster,
 	ownerRef *metav1.OwnerReference,
 ) error {
+	// TODO: add metrics collector back
+	// TODO: let metrics collector use new serviceaccount, clusterrole, clusterrolebining
+	if err := t.createCCMGoSCCComponents(cluster, ownerRef); err != nil {
+		return err
+	}
 	if err := t.reconcileCCMGoRolesAndRoleBindings(cluster, ownerRef); err != nil {
 		return err
 	}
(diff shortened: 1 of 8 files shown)


drivers/storage/portworx/component/telemetry_java.go
```diff
@@ -144,16 +147,30 @@ func (t *telemetry) deployMetricsCollector(
...
+			return err
+		}
+	} else {
+		if err := t.createCollectorServiceAccount(cluster.Namespace, ownerRef); err != nil {
+			return err
+		}
+		if err := t.createCollectorClusterRole(); err != nil {
+			return err
+		}
+		if err := t.createCollectorClusterRoleBinding(cluster.Namespace); err != nil {
```
haibinxie, line 168: how about role and roleBinding?

pure-jliao: They will not be changed, still same as before but service account will point to the new one.

drivers/storage/portworx/component/telemetry.go
```diff
@@ -178,12 +183,18 @@ func (t *telemetry) reconcileCCMGo(
...
+	if err := k8sutil.DeleteConfigMap(t.k8sClient, TelemetryCCMProxyConfigMapName, cluster.Namespace, *ownerRef); err != nil {
 		return err
 	}
-	if cluster.Status.ClusterUID == "" {
-		logrus.Warn("clusterUID is empty, wait for it to fill telemetry proxy config")
-		return nil
+	if err := t.createCCMGoSCCComponents(cluster, ownerRef); err != nil {
+		return err
+	}
+	if err := t.deployMetricsCollector(cluster, ownerRef); err != nil {
```
haibinxie, line 193: metrics collector uses ccm roles and rolebindings, so this should be moved to after those are created.

drivers/storage/portworx/component/telemetry.go
```diff
@@ -178,12 +183,18 @@ func (t *telemetry) reconcileCCMGo(
...
 ) error {
-	if err := t.reconcileCCMGoRolesAndRoleBindings(cluster, ownerRef); err != nil {
+	// Remove old ccm components first if exist
+	if err := k8sutil.DeleteConfigMap(t.k8sClient, TelemetryCCMProxyConfigMapName, cluster.Namespace, *ownerRef); err != nil {
 		return err
 	}
-	if cluster.Status.ClusterUID == "" {
-		logrus.Warn("clusterUID is empty, wait for it to fill telemetry proxy config")
-		return nil
+	if err := t.createCCMGoSCCComponents(cluster, ownerRef); err != nil {
```
haibinxie, line 190: rename this to reconcileCCMGoClusterRolesAndClusterRoleBindings

drivers/storage/portworx/component/telemetry_java.go
haibinxie: clusterRole and role should coexist. they should be reused/created/deleted at the same places.

pure-jliao: After upgrade the cluster role is used by registration, log uploaded and metrics collector roles, I’ll group the object creation based on components like metrics collector later. Here is when upgrading, we delete the old service account, cluster role and binding, reconcile it to use the new one which should be created already at this point

drivers/storage/portworx/component/telemetry.go
```diff
@@ -178,12 +183,18 @@ func (t *telemetry) reconcileCCMGo(
 	cluster *corev1.StorageCluster,
 	ownerRef *metav1.OwnerReference,
 ) error {
-	if err := t.reconcileCCMGoRolesAndRoleBindings(cluster, ownerRef); err != nil {
+	// Remove old ccm components first if exist
+	if err := k8sutil.DeleteConfigMap(t.k8sClient, TelemetryCCMProxyConfigMapName, cluster.Namespace, *ownerRef); err != nil {
```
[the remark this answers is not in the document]
pure-jliao: Ccm go migration currently is not supported and not in the scope, have a ticket to fix it later.

drivers/storage/portworx/component/telemetry.go
pure-jliao: The deploy function creates all objects needed except the scc related ones

drivers/storage/portworx/component/telemetry.go
pure-jliao: We have a service account here inside as well, all required to support scc

haibinxie: Understood, but it's not for SCC only, i was thinking about matching this with role and roleBinding, serviceAccount can be created separately

drivers/storage/portworx/component/telemetry.go
haibinxie: what's the reason to share clusterRole but not share role.

drivers/storage/portworx/component/telemetry.go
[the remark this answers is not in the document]
haibinxie: I am referring to old ccm to new ccm migration.

drivers/storage/portworx/component/telemetry.go
haibinxie: looks like the new ccm does not have a role, perhaps it's correct, we shall test it on various environment to see if any permission is missing.

drivers/storage/portworx/component/telemetry.go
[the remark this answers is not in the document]
pure-jliao: I changed the function names to v1 and v2 for metrics collector deployments only to make it clear, but the actual object names are not changed.

drivers/storage/portworx/component/telemetry.go
```diff
@@ -130,33 +138,30 @@ func (t *telemetry) Reconcile(cluster *corev1.StorageCluster) error {
...
-		// Remove old ccm components first if exist
-		if err := t.deleteCCMJava(cluster, ownerRef); err != nil {
-			return err
-		}
+	if cluster.Status.ClusterUID == "" {
+		logrus.Warn("clusterUID is empty, wait for it to reconcile telemetry components")
+		return nil
+	}
+	t.isCCMGoSupported = pxutil.IsCCMGoSupported(pxutil.GetPortworxVersion(cluster))
+	if t.isCCMGoSupported {
```
piyush-nimbalkar, line 146: nit: This can be a local variable. No need to add it to the receiver.

pure-jliao: it's used by other functions in ccm java code to determine which service account to use.

haibinxie (approved):

PR #711, opened 2022-08-09
PWX-24024: Default value for max_storage_nodes_per_zone

nrevanna:
A default value for max_storage_nodes_per_zone will be set when
  the cluster is created afresh. This change does not effect existing installs
  or upgrades.

  The value will be determined as follows
________________________________
  Zones | Default value | total nodes
______________________________
    1   |      3        |   3
    2   |      2        |   4
    3   |      1        |   3
    4   |      1        |   4

Closes #

Testing: Created a new cluster on vSphere, GCE and AWS and verified that it works for 1, 2, 3 & 4 zones

**GCE:
2 zones:
    "max_storage_nodes_per_zone": 2,**
root@gke-nrevanna-cluster-2-z-default-pool-0cf395ca-wp2l:/# pxctl status | grep gke-nrevanna| wc -l
4
root@gke-nrevanna-cluster-2-z-default-pool-0cf395ca-wp2l:/# pxctl status | grep gke-nrevanna| grep 'No Storage' |wc -l
0

**4 zones:
     "max_storage_nodes_per_zone": 1,**
root@gke-nrevanna-cluster-4-z-default-pool-65c80102-ntbg:/# pxctl status | grep gke-nrevanna| wc -l
8
root@gke-nrevanna-cluster-4-z-default-pool-65c80102-ntbg:/# pxctl status | grep gke-nrevanna| grep 'No Storage' |wc -l
4
root@gke-nrevanna-cluster-4-z-default-pool-65c80102-ntbg:/#

**EKS
3 zones:
    "max_storage_nodes_per_zone": 1**
[root@ip-192-168-47-179 /]# pxctl status | grep ip-192| wc -l
3

**2 zones
    "max_storage_nodes_per_zone": 2,**
[root@ip-192-168-6-129 /]# pxctl status| grep  ip-192-168 | wc -l
6
[root@ip-192-168-6-129 /]# pxctl status| grep  ip-192-168 | grep 'No Storage'| wc -l
2

**vSphere
1 zone
    "max_storage_nodes_per_zone": 3,**

### DIFF
--- a/pkg/cloudprovider/cloudprovider.go
+++ b/pkg/cloudprovider/cloudprovider.go
@@ -8,8 +8,8 @@ import (
 )
 
 const (
-	failureDomainZoneKey   = "failure-domain.beta.kubernetes.io/zone"
-	failureDomainRegionKey = "failure-domain.beta.kubernetes.io/region"
+	failureDomainZoneKey   = v1.LabelTopologyZone
+	failureDomainRegionKey = v1.LabelTopologyRegion
 )
 
 var (
--- a/pkg/controller/storagecluster/storagecluster.go
+++ b/pkg/controller/storagecluster/storagecluster.go
@@ -1123,9 +1123,9 @@ func (c *Controller) CreatePodTemplate(
 	return newTemplate, nil
 }
 
-// GetDefaultMaxStorageNodesPerZone aims to return a good value for MaxStorageNodesPerZone with the
+// getDefaultMaxStorageNodesPerZone aims to return a good value for MaxStorageNodesPerZone with the
 // intention of having at least 3 nodes in the cluster.
-func (c *Controller) GetDefaultMaxStorageNodesPerZone(zoneMap map[string]uint64) uint32 {
+func getDefaultMaxStorageNodesPerZone(zoneMap map[string]uint64) uint32 {
 	numZones := len(zoneMap)
 	switch numZones {
 	case 0, 1:
@@ -1235,7 +1235,7 @@ func (c *Controller) setStorageClusterDefaults(cluster *corev1.StorageCluster) e
 		toUpdate.Spec.CloudStorage.MaxStorageNodes == nil {
 		// Let's do this only when it's a fresh install of px
 		toUpdate.Spec.CloudStorage.MaxStorageNodesPerZone = new(uint32)
-		*toUpdate.Spec.CloudStorage.MaxStorageNodesPerZone = c.GetDefaultMaxStorageNodesPerZone(zoneMap)
+		*toUpdate.Spec.CloudStorage.MaxStorageNodesPerZone = getDefaultMaxStorageNodesPerZone(zoneMap)
 	}
 	c.Driver.SetDefaultsOnStorageCluster(toUpdate)
 
--- a/pkg/controller/storagecluster/storagecluster.go
+++ b/pkg/controller/storagecluster/storagecluster.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/libopenstorage/operator/pkg/cloudprovider"
 	"reflect"
 	"sort"
 	"strconv"
@@ -54,7 +55,6 @@ import (
 
 	"github.com/libopenstorage/operator/drivers/storage"
 	corev1 "github.com/libopenstorage/operator/pkg/apis/core/v1"
-	"github.com/libopenstorage/operator/pkg/cloudprovider"
 	"github.com/libopenstorage/operator/pkg/constants"
 	"github.com/libopenstorage/operator/pkg/util"
 	"github.com/libopenstorage/operator/pkg/util/k8s"
@@ -104,7 +104,8 @@ type Controller struct {
 	isStorkSchedDeploymentCreated bool
 	ctrl                          controller.Controller
 	// Node to NodeInfo map
-	nodeInfoMap map[string]*k8s.NodeInfo
+	nodeInfoMap       map[string]*k8s.NodeInfo
+	cloudProviderName string
 }
 
 // Init initialize the storage cluster controller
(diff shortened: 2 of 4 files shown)


pkg/controller/storagecluster/storagecluster.go
```diff
@@ -1155,6 +1161,27 @@ func (c *Controller) CreatePodTemplate(
 	return newTemplate, nil
 }
 
+// The aim of this function is to return a good value for MaStorageNodesPerZone with the
+// intention of having at least 3 nodes in the cluster.
+func getDefaultMaxStorageNodesPerZone(zoneMap map[string]uint64) uint32 {
+	numZones := len(zoneMap)
+	switch numZones {
+	case 1:
```
haibinxie, line 1169: help me understand the use case: if there is only one zone with 10 nodes, only 3 of them will be storage nodes? why not let all nodes be storage nodes

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -774,6 +774,12 @@ func (c *Controller) manage(
 		}
 	}
 
+	if cluster.Status.Phase == "" && cluster.Spec.CloudStorage.MaxStorageNodesPerZone == nil {
```
haibinxie, line 777: why not let it change? new zones may be added later

pkg/controller/storagecluster/storagecluster.go
nrevanna: Choosing the least number was the one with least negative impact. The consensus from the discussions was to start with 3 and then let the customer choose a different number if they choose to do so. The point of this enforcement is to introduce the customer to this option and avoid unnecessary additions of storage node during upgrades.

pkg/controller/storagecluster/storagecluster.go
nrevanna: Once the value is set, we don't have insight into if the customer modified it or it's the default value. May be the customer want's to continue to retain the value that exists. Since we are not in a position to determine that we will not let it change. 
If it's an existing cluster, we don't want to meddle with what the customer already has.
The only scenario we want to handle for an existing cluster is upgrade. I'll work on that and will send out a PR in a week.
The point of the default value is not to add too much intelligence instead just cover some obvious cases where we can start introducing the option with a non-zero value.

pkg/cloudprovider/cloudprovider.go
```diff
@@ -8,8 +8,8 @@ import (
 )
 
 const (
-	failureDomainZoneKey   = "failure-domain.beta.kubernetes.io/zone"
-	failureDomainRegionKey = "failure-domain.beta.kubernetes.io/region"
+	failureDomainZoneKey   = "topology.kubernetes.io/zone"
+	failureDomainRegionKey = "topology.kubernetes.io/region"
```
piyush-nimbalkar, line 12: Let's use constants from k8s - https://github.com/kubernetes/api/blob/f77fa25e8904fda257b3011a9fbbc07417d3d678/core/v1/well_known_labels.go#L22

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -1155,6 +1161,27 @@ func (c *Controller) CreatePodTemplate(
...
+	numZones := len(zoneMap)
+	switch numZones {
+	case 1:
+		// If there is a single Zone, have all 3 nodes in the same zone
+		return 3
+	case 2:
+		// If there are two zones, it'll be tricky since we'll always lose quorum when a zone
+		// goes down. Let's have 2 nodes in a zone so that we have a 4 node cluster.
+		return 2
+	case 3, 4, 5:
```
piyush-nimbalkar, line 1176: Can we use default case instead of listing 3, 4, 5?

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -774,6 +774,12 @@ func (c *Controller) manage(
 		}
 	}
 
+	if cluster.Status.Phase == "" && cluster.Spec.CloudStorage.MaxStorageNodesPerZone == nil {
+		// Let's do this only when it's a fresh install of px
+		cluster.Spec.CloudStorage.MaxStorageNodesPerZone = new(uint32)
+		*cluster.Spec.CloudStorage.MaxStorageNodesPerZone = getDefaultMaxStorageNodesPerZone(zoneMap)
```
piyush-nimbalkar, line 780: If the value is 0, then let's not set it and leave it as empty.

pkg/controller/storagecluster/storagecluster.go
piyush-nimbalkar, line 777: We should not set it if `cloudStorage.MaxStorageNodes` is set. We know that the user is restricting the cluster size to something, so that should do.

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -774,6 +774,12 @@ func (c *Controller) manage(
...
+	}
+
```
piyush-nimbalkar, line 782: This will pass the updated value to the portworx driver and create the correct px pod, but I don't think this will update the value in the storagecluster itself. 
We need to update the storagecluster object in k8s. I am thinking that may be we can move the entire zone calculation code to the SetDefaults.. method.

pkg/controller/storagecluster/storagecluster.go
haibinxie: Like Piyush said, this wont be persisted

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -773,10 +759,34 @@ func (c *Controller) manage(
...
+	if err != nil {
+		return fmt.Errorf("couldn't get node to storage cluster pods mapping for storage cluster %v: %v",
+			cluster.Name, err)
+	}
+	var nodesNeedingStoragePods, podsToDelete []string
+
+	nodeList, zoneMap, err := c.getZoneMap(cluster)
+	if err != nil {
+		return err
+	}
```
piyush-nimbalkar, line 785: nit: We can just call the UpdateDriver from the `SetDefaults` function. We don't need to do it here. 
Here, we just need the list of nodes. I would suggest just getting the node list and passing it to getZoneMap method. We don't even need to store the cloudProviderName that way.

piyush-nimbalkar (commented):
@nrevanna  can we add some UTs for the new changes? You can refer the SetDefaults* test here - https://github.com/libopenstorage/operator/blob/master/pkg/controller/storagecluster/controller_test.go#L766

Also, some of the UTs are failing in the travis build.

nrevanna:
> @nrevanna can we add some UTs for the new changes? You can refer the SetDefaults* test here - https://github.com/libopenstorage/operator/blob/master/pkg/controller/storagecluster/controller_test.go#L766
> 
> Also, some of the UTs are failing in the travis build.

Added a UT. Travis was failing for a code coverage issue. May be it got fixed with the UT.

haibinxie (commented):
LGTM

haibinxie (approved):

nrevanna:
Thanks @haibinxie and @piyush-nimbalkar  !

PR #714, opened 2022-08-10
[PWX-26180][PWX-26188] Use DaemonSet to deploy telemetry phonehome pods

pure-jliao:
Use DaemonSet instead of Deployment to create log uploader pods, stc node affinity applied.

Closes #

### DIFF
--- a/drivers/storage/portworx/component/telemetry.go
+++ b/drivers/storage/portworx/component/telemetry.go
@@ -10,6 +10,7 @@ import (
 	"github.com/hashicorp/go-version"
 	"github.com/sirupsen/logrus"
 	"google.golang.org/grpc"
+	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -26,6 +27,7 @@ import (
 	k8sutil "github.com/libopenstorage/operator/pkg/util/k8s"
 )
 
+// TODO: change all names to be px specific
 const (
 	// RoleNameSecretManager is name of role secret-manager
 	RoleNameSecretManager = "secret-manager"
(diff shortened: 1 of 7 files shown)


pkg/util/util.go
```diff
@@ -254,6 +254,42 @@ func DeepEqualDeployment(d1 *appsv1.Deployment, d2 *appsv1.Deployment) (bool, er
 	return true, nil
 }
 
+// DeepEqualDaemonSet compares if two daemonsets are same.
+func DeepEqualDaemonSet(d1, d2 *appsv1.DaemonSet) (bool, error) {
```
haibinxie, line 258: can we make a function DeepEqualPodTemplate and call it by DeepEqualDaemonset and DeepEqualDeployment to avoid duplicate code.

haibinxie (approved):
looks good if we can avoid duplicate code in DeepEqualDaemonSet

drivers/storage/portworx/component/telemetry.go
```diff
@@ -813,39 +814,26 @@ func (t *telemetry) createDeploymentPhonehomeCluster(
...
-	deployment.Spec.Replicas = &num
-	deployment.Spec.Template.Spec.ServiceAccountName = ServiceAccountNamePxTelemetry
-	pxutil.ApplyStorageClusterSettings(cluster, deployment)
+	daemonset.Spec.Template.Spec.ServiceAccountName = ServiceAccountNamePxTelemetry
+	pxutil.ApplyStorageClusterSettingsToPodSpec(cluster, &daemonset.Spec.Template.Spec)
 
-	existingDeployment, err := k8sutil.GetDeployment(t.k8sClient, deployment)
-	if err != nil {
+	existingDaemonSet := &appsv1.DaemonSet{}
+	if err := k8sutil.GetDaemonSet(t.k8sClient, daemonset.Name, daemonset.Namespace, existingDaemonSet); err != nil {
```
haibinxie, line 821: if the daemonset does not exist, we should not fail, you can refer to telemetry-java.go

and when doing deepEqual we shall handle the nil daemonset when it does not exist. so I was thinking about still have DeepEqualDaemonset and DeepEqualDeployment that checks nil then call DeepEqualPodTemplate.

pure-jliao: yes, I've verified that. when it doesn't exist, the result won't be nil but empty object, so deep equal will return false.

PR #715, opened 2022-08-10
PTX-12071 Implement method to determine Operator and PX versions

nikolaypopov:
Implement method to determine Operator and PX versions for future new CCM testing and upgrades

### DIFF
--- a/pkg/util/test/util.go
+++ b/pkg/util/test/util.go
@@ -90,6 +90,9 @@ const (
 	// PxMasterVersion is a tag for Portworx master version
 	PxMasterVersion = "3.0.0.0"
 
+	// PxOperatorMasterVersion is a tag for PX Operator master version
+	PxOperatorMasterVersion = "9.9.9.9"
+
 	// AksPVCControllerSecurePort is the PVC controller secure port.
 	AksPVCControllerSecurePort = "10261"
 )
@@ -97,6 +100,10 @@ const (
 // TestSpecPath is the path for all test specs. Due to currently functional test and
 // unit test use different path, this needs to be set accordingly.
 var TestSpecPath = "testspec"
+
+var (
+	opVer1_9_1, _ = version.NewVersion("1.9.1")
+)
 
 // MockDriver creates a mock storage driver
 func MockDriver(mockCtrl *gomock.Controller) *mock.MockDriver {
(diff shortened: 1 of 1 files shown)


pkg/util/test/util.go
```diff
@@ -2335,6 +2342,96 @@ func ValidateTelemetry(pxImageList map[string]string, cluster *corev1.StorageClu
...
+	var image string
+
+	NamespaceList, err := coreops.Instance().ListNamespaces(labelSelector)
+	if err != nil {
+		return image, err
+	}
+
+	for _, ns := range NamespaceList.Items {
+		operatorDeployment, err := appops.Instance().GetDeployment("portworx-operator", ns.Name)
+		if errors.IsNotFound(err) {
```
pure-jliao, line 2419: return if hit other errors

pkg/util/test/util.go
```diff
@@ -2335,6 +2342,96 @@ func ValidateTelemetry(pxImageList map[string]string, cluster *corev1.StorageClu
...
+	image, err := getPxOperatorImage()
+	if err != nil {
+		return nil, err
+	}
+
+	// We may run the automation on operator installed using private images,
+	// so assume we are testing the latest operator version if failed to parse the tag
+	var versionTag string
+	versionTag = strings.Split(image, ":")[1]
+	if strings.Contains(versionTag, "-dev") {
```
pure-jliao, line 2394: with `-dev` we could still parse the version, e.g. version 1.9.0-dev, and can be used to compare versions. so we could skip this step.

pkg/util/test/util.go
```diff
@@ -2335,6 +2342,96 @@ func ValidateTelemetry(pxImageList map[string]string, cluster *corev1.StorageClu
...
+func GetPxOperatorVersion() (*version.Version, error) {
+	image, err := getPxOperatorImage()
+	if err != nil {
+		return nil, err
+	}
+
+	// We may run the automation on operator installed using private images,
+	// so assume we are testing the latest operator version if failed to parse the tag
+	var versionTag string
+	versionTag = strings.Split(image, ":")[1]
```
pure-jliao, line 2393: nit: use last one instead. but should be fine without custom registry.

pkg/util/test/util.go
```diff
@@ -2335,6 +2342,96 @@ func ValidateTelemetry(pxImageList map[string]string, cluster *corev1.StorageClu
 	return ValidateTelemetryUninstalled(pxImageList, cluster, timeout, interval)
 }
 
+// GetPortworxVersion returns Portworx version from StorageCluster
```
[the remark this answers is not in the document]
nikolaypopov: Good point, should have looked and saved time and yeah there are import issues so I will copy this

pure-jliao (approved):

PR #719, opened 2022-08-15
PWX-26006 - fix pvccontroller crash on k8s 1.24+

haibinxie:

Closes #

### DIFF
--- a/drivers/storage/portworx/component/pvccontroller.go
+++ b/drivers/storage/portworx/component/pvccontroller.go
@@ -304,10 +304,13 @@ func (c *pvcController) createDeployment(
 		"--controllers=persistentvolume-binder,persistentvolume-expander",
 		"--use-service-account-credentials=true",
 	}
-	if pxutil.IsOpenshift(cluster) {
-		command = append(command, "--leader-elect-resource-lock=endpoints")
-	} else {
-		command = append(command, "--leader-elect-resource-lock=configmaps")
+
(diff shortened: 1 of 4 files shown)


drivers/storage/portworx/component/pvccontroller.go
```diff
@@ -304,10 +304,13 @@ func (c *pvcController) createDeployment(
 		"--controllers=persistentvolume-binder,persistentvolume-expander",
 		"--use-service-account-credentials=true",
 	}
-	if pxutil.IsOpenshift(cluster) {
-		command = append(command, "--leader-elect-resource-lock=endpoints")
-	} else {
-		command = append(command, "--leader-elect-resource-lock=configmaps")
+
+	if c.k8sVersion.LessThan(k8sutil.K8sVer1_24) {
```
pure-jliao, line 308: what about k8s 1.24? this arg is no longer needed?

piyush-nimbalkar: I think there is a default for the leader election lock type in k8s 1.24. It defaults to `leases`.

piyush-nimbalkar (approved):

pure-jliao (approved):

drivers/storage/portworx/component/pvccontroller.go
haibinxie: Yes, it has default value in 1.24. not sure if it has a default value before though as i could not find command help on specific version

PR #720, opened 2022-08-16
PTX-12097 Validate new Telemetry method and regression test changes

nikolaypopov:
Validate new Telemetry method and regression test changes

### DIFF
--- a/pkg/util/test/util.go
+++ b/pkg/util/test/util.go
@@ -109,6 +109,11 @@ const (
 // unit test use different path, this needs to be set accordingly.
 var TestSpecPath = "testspec"
 
+var (
+	pxVer2_12, _  = version.NewVersion("2.12.0")
+	opVer1_9_1, _ = version.NewVersion("1.9.1")
+)
+
 // MockDriver creates a mock storage driver
 func MockDriver(mockCtrl *gomock.Controller) *mock.MockDriver {
 	return mock.NewMockDriver(mockCtrl)
@@ -774,6 +779,12 @@ func ValidateUninstallStorageCluster(
 	// Validate deletion of Portworx ConfigMaps
 	if err := validatePortworxConfigMapsDeleted(cluster, timeout, interval); err != nil {
 		return err
+	}
+
+	// Verify telemetry secret is deleted on UninstallAndWipe
+	_, err := coreops.Instance().GetSecret("pure-telemetry-certs", cluster.Namespace)
+	if !errors.IsNotFound(err) {
+		return fmt.Errorf("telemetry secret pure-telemetry-certs was found when shouldn't have been")
 	}
 
 	return nil
@@ -2222,7 +2233,10 @@ func ValidateMonitoring(pxImageList map[string]string, cluster *corev1.StorageCl
 		return err
 	}
 
-	if err := ValidateTelemetry(pxImageList, cluster, timeout, interval); err != nil {
+	// Increasing timeout for Telemetry components as they take quite long time to initialize
+	defaultTelemetryRetryInterval := 30 * time.Second
+	defaultTelemetryTimeout := 10 * time.Minute
+	if err := ValidateTelemetry(pxImageList, cluster, defaultTelemetryTimeout, defaultTelemetryRetryInterval); err != nil {
 		return err
 	}
 
@@ -2286,8 +2300,8 @@ func ValidatePrometheus(pxImageList map[string]string, cluster *corev1.StorageCl
 	return nil
 }
 
-// ValidateTelemetryUninstalled validates telemetry component is uninstalled as expected
-func ValidateTelemetryUninstalled(pxImageList map[string]string, cluster *corev1.StorageCluster, timeout, interval time.Duration) error {
+// ValidateTelemetryUninstalled validates telemetry components are uninstalled as expected
+func ValidateTelemetryUninstalled(cluster *corev1.StorageCluster, timeout, interval time.Duration) error {
 	t := func() (interface{}, bool, error) {
 		_, err := appops.Instance().GetDeployment("px-metrics-collector", cluster.Namespace)
 		if !errors.IsNotFound(err) {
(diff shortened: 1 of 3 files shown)


test/integration_test/basic_test.go
```diff
@@ -439,67 +416,56 @@ func BasicUpgradeOperator(tc *types.TestCase) func(*testing.T) {
...
-}
+		// Create StorageCluster
+		cluster = ci_utils.DeployAndValidateStorageCluster(cluster, ci_utils.PxSpecImages, t)
+
+		// Validate Telemetry is not enabled by default
+		logrus.Info("Validate Telemetry is not enabled by default")
+		if cluster.Spec.Monitoring != nil {
+			if cluster.Spec.Monitoring.Telemetry != nil {
+				if cluster.Spec.Monitoring.Telemetry.Enabled {
+					require.False(t, cluster.Spec.Security.Enabled, "failed to validate default Telemetry status: expected: false, actual: %v", cluster.Spec.Monitoring.Telemetry.Enabled)
```
pure-jliao, line 439: nit: by default the field is nil, we expect it to be nil or false, so if it's enabled, we can call `require.FailNow()`. but this should work fine.

pkg/util/test/util.go
```diff
@@ -776,6 +781,12 @@ func ValidateUninstallStorageCluster(
 		return err
 	}
 
+	// Verify telemetry secret is deleted on UninstallAndWipe
```
pure-jliao, line 784: can we check the delete strategy to be UninstallAndWipe first then validate the secret deletion?

pkg/util/test/util.go
```diff
@@ -2337,13 +2351,31 @@ func ValidateTelemetryUninstalled(pxImageList map[string]string, cluster *corev1
...
+		return err
+	}
+	logrus.Infof("PX Operator version: %v", opVersion)
+
 	if cluster.Spec.Monitoring != nil &&
 		cluster.Spec.Monitoring.Telemetry != nil &&
 		cluster.Spec.Monitoring.Telemetry.Enabled {
+		logrus.Info("Telemetry is enabled in StorageCluster")
+		if pxVersion.GreaterThanOrEqual(pxVer2_12) && pxVersion.GreaterThanOrEqual(opVer1_9_1) {
+			return ValidateTelemetryEnabled(pxImageList, cluster, timeout, interval)
```
pure-jliao, line 2369: are you implementing a new validation function for ccm-go only? that case can we call it `ValidateTelemetryV2Installed` and change the old one to `ValidateTelemetryV1Installed` to be more clear?

pkg/util/test/util.go
```diff
@@ -2337,13 +2351,31 @@ func ValidateTelemetryUninstalled(pxImageList map[string]string, cluster *corev1
...
+		}
 		return ValidateTelemetryInstalled(pxImageList, cluster, timeout, interval)
 	}
 
-	return ValidateTelemetryUninstalled(pxImageList, cluster, timeout, interval)
+	logrus.Info("Telemetry is disabled in StorageCluster")
+	if pxVersion.GreaterThanOrEqual(pxVer2_12) && pxVersion.GreaterThanOrEqual(opVer1_9_1) {
+		return ValidateTelemetryDisabled(cluster, timeout, interval)
```
pure-jliao, line 2376: same here.

pkg/util/test/util.go
```diff
@@ -2578,6 +2610,255 @@ func ValidateAlertManagerDisabled(pxImageList map[string]string, cluster *corev1
...
+		return fmt.Errorf("failed to find image for envoy")
+	}
+	envoyImageName = util.GetImageURN(cluster, envoyImageName)
+	if registrationServiceDep.Spec.Template.Spec.Containers[1].Image != envoyImageName {
+		return fmt.Errorf("envoy image mismatch, image: %s, expected: %s",
+			registrationServiceDep.Spec.Template.Spec.Containers[1].Image,
+			envoyImageName)
+	}
+
+	// Verify px-telemetry-metrics-collector deployment and images
```
pure-jliao, line 2655: can we validate the deployment has a init container as well?

pkg/util/test/util.go
```diff
@@ -2578,6 +2610,255 @@ func ValidateAlertManagerDisabled(pxImageList map[string]string, cluster *corev1
...
+	logrus.Info("Validate px-telemetry-phonehome daemonset and images")
+	if err := appops.Instance().ValidateDaemonSet("px-telemetry-phonehome", cluster.Namespace, timeout); err != nil {
+		return err
+	}
+	telemetryPhonehomeDs, err := appops.Instance().GetDaemonSet("px-telemetry-phonehome", cluster.Namespace)
+	if err != nil {
+		return err
+	}
+
+	// Verify collector container image inside px-telemetry-phonehome
```
pure-jliao, line 2699: same here, the template should have an init container, we can just check the image.

pkg/util/test/util.go
```diff
@@ -2578,6 +2610,255 @@ func ValidateAlertManagerDisabled(pxImageList map[string]string, cluster *corev1
...
+	if err != nil {
+		return err
+	}
+
+	// Verify telemetry serviceaccounts
+	_, err = coreops.Instance().GetServiceAccount("px-telemetry", cluster.Namespace)
+	if err != nil {
+		return err
+	}
+
```
pure-jliao, line 2774: are we able to validate the clusterrole and clusterrolebinding as well?

pkg/util/test/util.go
```diff
@@ -2578,6 +2610,255 @@ func ValidateAlertManagerDisabled(pxImageList map[string]string, cluster *corev1
...
+		if !errors.IsNotFound(err) {
+			return "", true, fmt.Errorf("found px-telemetry-tls-certificate configmap, waiting for deletion")
+		}
+
+		// Verify telemetry serviceaccounts
+		_, err = coreops.Instance().GetServiceAccount("px-telemetry", cluster.Namespace)
+		if !errors.IsNotFound(err) {
+			return "", true, fmt.Errorf("found px-telemetry serviceaccount, waiting for deletion")
+		}
+
```
pure-jliao, line 2850: same here for clusterrole and clusterrolebinding

pure-jliao (approved):
LGTM after addressing comments.

test/integration_test/basic_test.go
nikolaypopov: If this is fine I will just leave it as is

pkg/util/test/util.go
nikolaypopov: In this func we only validate after the wipe, cluster and all components are gone at this point, the func that uninstalls cluster is setting the UninstallAndWipe strategy

pkg/util/test/util.go
nikolaypopov: done

pkg/util/test/util.go

pkg/util/test/util.go
nikolaypopov: Done

pkg/util/test/util.go
nikolaypopov: Done

pkg/util/test/util.go
nikolaypopov: Done

pkg/util/test/util.go
nikolaypopov: Done

PR #721, opened 2022-08-16
[PWX-26024] Telemetry ccm-go custom proxy support

pure-jliao:
Add custom proxy support. 
* All 3 telemetry components have its own envoy configmap, when using custom proxy, those configmaps should be updated accordingly
* the custom proxy should be in format of `host:port`, if it's not valid, envoy configmaps will be deleted, so that existing pod will continue to run, new pod will stuck at creation, until the proxy format got fixed.

Closes #

* log uploader using http proxy is tested
* working on register and collector support
* https validation in progress.


deploy/ccm/envoy-config-collector-custom-proxy.yaml
```diff
@@ -35,106 +35,73 @@ static_resources:
                   prefix: "/"
                 request_headers_to_add:
                 - header:
-                   key: "product-name"
-                   value: "portworx"
+                    key: "product-name"
```
haibinxie, line 38: nit: plz double check if extra space is needed

drivers/storage/portworx/component/telemetry.go
```diff
@@ -90,8 +96,11 @@ const (
 	stagingArcusRestProxyURL        = "rest.staging-cloud-support.purestorage.com"
 	stagingArcusRegisterProxyURL    = "register.staging-cloud-support.purestorage.com"
 
-	// port for log uploader, avoid 1970 as used by old ccm service, to be configurable in the future
-	logUploaderPort = 9090
+	// Port offset for telemetry components
+	ccmListeningPortOffset          = 89
+	collectorCloudSupportPortOffset = 999
```
haibinxie, line 101: can we just use port number, the offset is harder to track and avoid conflict.

drivers/storage/portworx/component/telemetry.go
```diff
@@ -554,23 +566,33 @@ func (t *telemetry) createCCMGoConfigMapRegister(
...
-	if proxy == "" {
-		replaceMap[configParameterRegisterProxyURL] = getArcusRegisterProxyURL(cluster)
+	if proxy != "" && t.usePxProxy {
+		address, port, err := pxutil.SplitPxProxyHostPort(proxy)
+		if err != nil {
+			logrus.Errorf("failed to get custom proxy address and port from proxy %s: %v", proxy, err)
+			return k8sutil.DeleteConfigMap(t.k8sClient, ConfigMapNameTelemetryRegisterProxy, cluster.Namespace, *ownerRef)
+		}
+		configFileName = configFileNameTelemetryRegisterCustomProxy
+		replaceMap[configParameterCloudSupportTCPProxyPort] = fmt.Sprint(registerCloudSupportPort + 2)
```
haibinxie, line 591: define a constant

drivers/storage/portworx/component/telemetry.go
```diff
@@ -554,23 +566,33 @@ func (t *telemetry) createCCMGoConfigMapRegister(
...
+		replaceMap[configParameterCloudSupportEnvoyInternalRedirectPort] = fmt.Sprint(registerCloudSupportPort + 4)
```
haibinxie, line 592: define a constant

haibinxie (approved):

drivers/storage/portworx/component/telemetry.go
pure-jliao: port number variables are defined locally currently, we cannot define const for those port values as we need stc to get the start port first, and it could be updated. what I could do is define some default ports, check whether the start port is the default one, if not we use another function to get the shifted ports. wdyt?

pure-jliao: > can we just use port number, the offset is harder to track and avoid conflict.

eventually I think we should have one place to track all port offsets to avoid conflict.

haibinxie: any reason to choose 89,999,3000, instead of 1,2,3,4,5?

pure-jliao: @haibinxie for ccm listening port we need to be consistent with portworx side, it used to be 9090 by default, and will shift with the start port 9001, which we pass with a `-r` arg. we need changes on both operator and px side.
collector we hard-coded 10000 as default previously, so still use 10000 by default, to avoid port conflict we need to shift it as well
for register and phonehome, we just chose a higher value as default in the template to avoid conflict, since when using custom proxy, by default all 12001~12006 will be used.

haibinxie (approved):

PR #722, opened 2022-08-17
PTX-11431 Fix master labels to support v1.24

nikolaypopov:
Copying function that determines master node from sched-ops to operator instead of updating sched-ops package due to huge vendor changes


pure-jliao (approved):
LGTM, this is just a workaround to fix the pipeline right? eventually we want to just update the vendor

nikolaypopov:
> LGTM, this is just a workaround to fix the pipeline right? eventually we want to just update the vendor

Yes, this is a workaround, until we can update sched-ops vendor to use the func from there, which also has the same fix

PR #726, opened 2022-08-18
PWX-25488: add option to disable some objects from controller list-wa…

haibinxie:
…tch cache to reduce memory usage

Closes #

### DIFF
--- a/cmd/operator/operator.go
+++ b/cmd/operator/operator.go
@@ -6,13 +6,17 @@ import (
 	"net/http"
 	_ "net/http/pprof"
 	"os"
+	"strings"
 	"time"
 
 	ocp_configv1 "github.com/openshift/api/config/v1"
 	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	log "github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
+	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	storagev1 "k8s.io/api/storage/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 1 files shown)


cmd/operator/operator.go
```diff
@@ -247,11 +257,57 @@ func run(c *cli.Context) {
...
+			objs = append(objs, &rbacv1.ClusterRole{})
+		case "ClusterRoleBinding":
+			objs = append(objs, &rbacv1.ClusterRoleBinding{})
+		case "Role":
+			objs = append(objs, &rbacv1.Role{})
+		case "RoleBinding":
+			objs = append(objs, &rbacv1.RoleBinding{})
+		case "StorageClass":
+			objs = append(objs, &storagev1.StorageClass{})
+		}
```
pure-jliao, line 298: nit: better to add else, if the type is not supported we should print a log

cmd/operator/operator.go
```diff
@@ -247,11 +257,57 @@ func run(c *cli.Context) {
...
+}
+
 func createManager(c *cli.Context, config *rest.Config) (manager.Manager, error) {
 	syncPeriod := defaultResyncPeriod
 	managerOpts := manager.Options{
-		SyncPeriod:         &syncPeriod,
-		MetricsBindAddress: fmt.Sprintf("0.0.0.0:%d", c.Int(flagMetricsPort)),
+		SyncPeriod:            &syncPeriod,
+		MetricsBindAddress:    fmt.Sprintf("0.0.0.0:%d", c.Int(flagMetricsPort)),
+		ClientDisableCacheFor: getObjects(c.String(flagDisableCacheFor)),
```
pure-jliao, line 310: if the input is empty, does it affect anything?

pure-jliao (commented):
LGTM, left some comments.

cmd/operator/operator.go
haibinxie: it does not, default is empty list.

cmd/operator/operator.go
haibinxie: sounds good

pure-jliao (approved):

PR #728, opened 2022-08-19
[PWX-25529] Adjust placement rule node labels for k8s 1.24+ support

pure-jliao:

Closes #

### DIFF
--- a/drivers/storage/portworx/portworx.go
+++ b/drivers/storage/portworx/portworx.go
@@ -631,6 +631,9 @@ func (p *portworx) storageNodeToCloudSpec(storageNodes []*corev1.StorageNode, cl
 
 // SetPortworxDefaults populates default storage cluster spec values
 func SetPortworxDefaults(toUpdate *corev1.StorageCluster, k8sVersion *version.Version) {
+	if k8sVersion == nil {
+		k8sVersion = pxutil.MinimumSupportedK8sVersion
+	}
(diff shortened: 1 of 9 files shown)


drivers/storage/portworx/portworx.go
```diff
@@ -631,6 +631,9 @@ func (p *portworx) storageNodeToCloudSpec(storageNodes []*corev1.StorageNode, cl
 
 // SetPortworxDefaults populates default storage cluster spec values
 func SetPortworxDefaults(toUpdate *corev1.StorageCluster, k8sVersion *version.Version) {
+	if k8sVersion == nil {
```
pure-jliao, line 634: without this it breaks too many unit test cases. in reality this should not be nil

haibinxie (approved):
Looks good, it would be nice to see an example of the new spec, like this csiDeployment_1.0_k8s124

https://github.com/libopenstorage/operator/blob/master/drivers/storage/portworx/testspec/csiDeployment_1.0_k8s120.yaml

pure-jliao:
> Looks good, it would be nice to see an example of the new spec in unit test, like this csiDeployment_1.0_k8s124
> 
> https://github.com/libopenstorage/operator/blob/master/drivers/storage/portworx/testspec/csiDeployment_1.0_k8s120.yaml

there are too many duplicated component specs to test in that case, currently px pod schedule and stc spec are tested so should be good.

PR #731, opened 2022-08-19
PTX-12493 Fix common registry issue for older operators

nikolaypopov:
Fix common registry issue for older then 1.9.1 operators

### DIFF
--- a/pkg/util/test/util.go
+++ b/pkg/util/test/util.go
@@ -114,8 +114,9 @@ const (
 var TestSpecPath = "testspec"
 
 var (
-	pxVer2_12, _ = version.NewVersion("2.12.0")
-	opVer1_10, _ = version.NewVersion("1.10.0")
+	pxVer2_12, _  = version.NewVersion("2.12.0")
+	opVer1_10, _  = version.NewVersion("1.10.0")
+	opVer1_9_1, _ = version.NewVersion("1.9.1")
 )
 
 // MockDriver creates a mock storage driver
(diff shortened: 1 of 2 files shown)


pkg/util/test/util.go
```diff
@@ -3232,6 +3216,51 @@ func validateAllStorageNodesInState(namespace string, status corev1.NodeConditio
 	}
 }
 
+// GetImageURN returns the complete image name based on the registry and repo
+func GetImageURN(cluster *corev1.StorageCluster, image string) string {
```
haibinxie, line 3220: Let's not copy this function, in the future if we change this function, test would not catch it.  you can write a function ValidateImage, for 1.9.1 call util.ImageURN, for lower version if image does not match, remove docker.io prefix and compare again.

nikolaypopov: I wanted to add the check for operator version inside `GetImageURN()` in `pkg/util/util.go` but couldn't due to import issue

pkg/util/util.go
```diff
@@ -83,7 +83,8 @@ func AddDefaultRegistryToImage(image string) string {
 	return DefaultImageRegistry + "/" + image
 }
 
-func getMergedCommonRegistries(cluster *corev1.StorageCluster) map[string]bool {
+// GetMergedCommonRegistries return merged common registries
```
haibinxie, line 86: change back to private function.

pkg/util/test/util.go
```diff
@@ -1958,78 +1928,93 @@ func validateCsiExtImages(cluster *corev1.StorageCluster, pxImageList map[string
...
+
 	for _, pod := range pods.Items {
-		foundImage := false
-		for _, container := range pod.Spec.Containers {
-			if container.Image == image {
-				foundImage = true
-				break
+		foundContainer := false
+		foundImage := ""
+		if containerName == "init-cont" {
```
haibinxie, line 1984: let's not check this hardcoded name here, it's already checked inside of the loop in line 1986. instead you can add all containers and init containers in an array, check the array, it will also reduce duplicate code in if and else

pkg/util/util.go
nikolaypopov: Done

haibinxie (approved):

PR #734, opened 2022-08-22
do not take docker.io as custom registry

haibinxie:

Closes #

### DIFF
--- a/pkg/migration/generate.go
+++ b/pkg/migration/generate.go
@@ -14,6 +14,8 @@ import (
 	pxutil "github.com/libopenstorage/operator/drivers/storage/portworx/util"
 	corev1 "github.com/libopenstorage/operator/pkg/apis/core/v1"
 	"github.com/libopenstorage/operator/pkg/constants"
+	"github.com/libopenstorage/operator/pkg/util"
+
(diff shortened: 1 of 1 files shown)


pkg/migration/generate.go
```diff
@@ -117,6 +119,11 @@ func (h *Handler) createStorageCluster(
 }
 
 func parseCustomImageRegistry(pxImage string, componentImages []string) (string, bool) {
+	// Let's remove docker.io prefix, so we don't take docker.io as custom image registry.
```
pure-jliao, line 122: add this case to case 1 at line 128
```
// 1. portworx/oci-monitor:image or docker.io/portworx/oci-monitor:image
```
and handle it in case 1 as well
```
if len(imageParts) == 2 && imageParts[0] == "portworx"  || imageParts[0] == util.DefaultImageRegistry {
```

pure-jliao (approved):
LGTM, have some minor suggestions.

pkg/migration/generate.go
haibinxie: there is no need to handle it, as the prefix is removed at this line already.

PR #740, opened 2022-08-25
[PWX-25380] Disable telemetry during DaemonSet migration if PX is 2.12+

pure-jliao:

Closes #

### DIFF
--- a/pkg/migration/generate.go
+++ b/pkg/migration/generate.go
@@ -16,6 +16,7 @@ import (
 	"github.com/libopenstorage/operator/pkg/constants"
 	"github.com/libopenstorage/operator/pkg/util"
 
+	k8sutil "github.com/libopenstorage/operator/pkg/util/k8s"
 	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	"github.com/sirupsen/logrus"
 	"gopkg.in/yaml.v2"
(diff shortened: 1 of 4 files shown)


pkg/migration/generate.go
```diff
@@ -936,6 +936,14 @@ func (h *Handler) addMonitoringSpec(cluster *corev1.StorageCluster, ds *appsv1.D
 	}
 	cluster.Status.DesiredImages.Telemetry = telemetryImage
 
+	// Telemetry components are removed from PX DaemonSet 2.12+ so it's not a valid cluster setup to migrate.
+	// CCM upgrade to a newer version during migration is not supported, so we disable telemetry if end up here.
+	if pxutil.GetPortworxVersion(cluster).GreaterThanOrEqual(pxutil.MinimumPxVersionCCMGO) {
+		cluster.Spec.Monitoring.Telemetry.Enabled = false
```
haibinxie, line 942: Let's still enable it if it was enabled.

haibinxie (approved):

pkg/migration/generate.go
haibinxie: if it's tested working, we should still enable it, otherwise it's ok to disable it.

test/integration_test/migration_test.go
```diff
@@ -366,9 +367,13 @@ func validateStorageClusterFromDaemonSet(
...
 		case "telemetry":
-			if cluster.Spec.Monitoring == nil ||
-				cluster.Spec.Monitoring.Telemetry == nil ||
-				!cluster.Spec.Monitoring.Telemetry.Enabled {
+			telemetryEnabled := cluster.Spec.Monitoring != nil &&
+				cluster.Spec.Monitoring.Telemetry != nil &&
+				cluster.Spec.Monitoring.Telemetry.Enabled
+			ccmGoSupported := pxutil.GetPortworxVersion(cluster).GreaterThanOrEqual(pxutil.MinimumPxVersionCCMGO)
+			if ccmGoSupported && telemetryEnabled {
+				cerr = fmt.Errorf("telemetry should been disabled")
```
nikolaypopov, line 375: nit: should have been*

nikolaypopov (approved):

PR #742, opened 2022-08-30
[PWX-26670] Filter out px-csi-ext pods when scanning pre-installed snapshot controllers

pure-jliao:
**What this PR does / why we need it**: If `installSnapshotController` is true in stc spec, during operator upgrade it will treat the existing operator deployed snapshot controller as external one, then remove it from px-csi-ext deployment.

Closes #


nikolaypopov (approved):

drivers/storage/portworx/component/csi.go
```diff
@@ -530,6 +530,11 @@ func (c *csi) findPreinstalledCSISnapshotController() error {
 		return err
 	}
 	for _, pod := range podList.Items {
+		// Ignore csi pods deployed by operator
```
haibinxie, line 533: can we change "app" into a constant and use it in csiDeploymentTemplateLabels initialization as well

haibinxie (approved):

PR #744, opened 2022-08-30
[PWX-26683] Fix bug during updating migration test spec images

pure-jliao:

Closes #

### DIFF
--- a/pkg/migration/cleanup.go
+++ b/pkg/migration/cleanup.go
@@ -3,7 +3,6 @@ package migration
 import (
 	"context"
 
-	corev1 "github.com/libopenstorage/operator/pkg/apis/core/v1"
 	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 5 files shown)


pkg/migration/cleanup.go
```diff
@@ -258,6 +260,9 @@ func (h *Handler) deleteMonitoringComponent(cluster *corev1.StorageCluster) erro
 	if err := h.deleteObject(prometheusOpAccountName, cluster.Namespace, &v1.ServiceAccount{}, cluster); err != nil {
 		return err
 	}
+	if err := h.deleteObject(component.TelemetryConfigMapName, cluster.Namespace, &v1.ConfigMap{}, cluster); err != nil {
```
haibinxie, line 263: there is no need to delete it, the configMap will be reused by the v1 telemetry. if it's v2 telemetry, the telemetry component will delete the configmap correct?

pure-jliao: if delete it, during reconcile it will be recreated. otherwise there is an issue when telemetry got disabled while approving the migration, reconcile never happened and the configmap doesn't have owner, so if stc got deleted the configmap won't be cleaned up

haibinxie (approved):

pkg/migration/cleanup.go
haibinxie: add the details in the comment when possible

PR #754, opened 2022-09-08
PWX-26230: fix portworx operator does not honor the PX version in the px-versions configmap

haibinxie:

Closes #

### DIFF
--- a/drivers/storage/portworx/portworx.go
+++ b/drivers/storage/portworx/portworx.go
@@ -189,114 +189,12 @@ func (p *portworx) SetDefaultsOnStorageCluster(toUpdate *corev1.StorageCluster) 
 			toUpdate.Spec.Version = release.PortworxVersion
 		}
 
-		toUpdate.Status.Version = toUpdate.Spec.Version
-
-		if autoUpdateStork(toUpdate) &&
-			(toUpdate.Status.DesiredImages.Stork == "" ||
-				pxVersionChanged ||
-				autoUpdateComponents(toUpdate)) {
-			toUpdate.Status.DesiredImages.Stork = release.Components.Stork
-		}
-
-		if autoUpdateAutopilot(toUpdate) &&
-			(toUpdate.Status.DesiredImages.Autopilot == "" ||
-				pxVersionChanged ||
-				autoUpdateComponents(toUpdate)) {
-			toUpdate.Status.DesiredImages.Autopilot = release.Components.Autopilot
-		}
-
-		if autoUpdateLighthouse(toUpdate) &&
-			(toUpdate.Status.DesiredImages.UserInterface == "" ||
-				pxVersionChanged ||
-				autoUpdateComponents(toUpdate)) {
-			toUpdate.Status.DesiredImages.UserInterface = release.Components.Lighthouse
-		}
-
-		if autoUpdateTelemetry(toUpdate) &&
-			(toUpdate.Status.DesiredImages.Telemetry == "" ||
-				pxVersionChanged ||
-				autoUpdateComponents(toUpdate)) {
-			toUpdate.Status.DesiredImages.Telemetry = release.Components.Telemetry
-		}
-
-		if autoUpdateTelemetryProxy(toUpdate) &&
-			(toUpdate.Status.DesiredImages.TelemetryProxy == "" ||
-				pxVersionChanged ||
-				autoUpdateComponents(toUpdate)) {
-			toUpdate.Status.DesiredImages.TelemetryProxy = release.Components.TelemetryProxy
-		}
-
-		if autoUpdateMetricsCollector(toUpdate) &&
-			(toUpdate.Status.DesiredImages.MetricsCollector == "" ||
-				(!pxutil.IsCCMGoSupported(pxutil.GetPortworxVersion(toUpdate)) &&
-					toUpdate.Status.DesiredImages.MetricsCollectorProxy == "") ||
-				(pxutil.IsCCMGoSupported(pxutil.GetPortworxVersion(toUpdate)) &&
-					toUpdate.Status.DesiredImages.TelemetryProxy == "") ||
-				pxVersionChanged ||
-				autoUpdateComponents(toUpdate)) {
-			toUpdate.Status.DesiredImages.MetricsCollector = release.Components.MetricsCollector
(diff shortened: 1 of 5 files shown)


drivers/storage/portworx/portworx.go
```diff
@@ -365,6 +263,118 @@ func (p *portworx) SetDefaultsOnStorageCluster(toUpdate *corev1.StorageCluster)
 	setDefaultAutopilotProviders(toUpdate)
 }
 
+func (p *portworx) updateDesiredImages(toUpdate *corev1.StorageCluster, pxVersionChanged bool, release *manifest.Version) {
+	toUpdate.Status.Version = toUpdate.Spec.Version
```
pure-jliao, line 267: can we move this out? not something related to desired image. also resetting the desired images to empty should also be part of this function.

drivers/storage/portworx/portworx.go
```diff
@@ -189,114 +189,12 @@ func (p *portworx) SetDefaultsOnStorageCluster(toUpdate *corev1.StorageCluster)
...
-
-		// Reset the component update strategy if it is 'Once', so that we don't
-		// upgrade components again during next reconcile loop
-		if toUpdate.Spec.AutoUpdateComponents != nil &&
-			*toUpdate.Spec.AutoUpdateComponents == corev1.OnceAutoUpdate {
-			toUpdate.Spec.AutoUpdateComponents = nil
+		// The version from manifest does not match desired version, this may be caused by wrong configMap version.
+		if release.PortworxVersion != "" && toUpdate.Spec.Version != "" && release.PortworxVersion != toUpdate.Spec.Version {
+			msg := fmt.Sprintf("Version in StorageCluster %s does not match version manifest %s, please edit StorageCluster or px-versions configMap.", toUpdate.Spec.Version, release.PortworxVersion)
+			p.warningEvent(toUpdate, util.FailedComponentReason, msg)
```
pure-jliao, line 195: if we hit this, do we still proceed? if we don't update desired images in this case, we should not reset them either I feel.

drivers/storage/portworx/portworx.go
```diff
@@ -189,114 +189,12 @@ func (p *portworx) SetDefaultsOnStorageCluster(toUpdate *corev1.StorageCluster)
...
-			}
-		}
-
-		// Reset the component update strategy if it is 'Once', so that we don't
-		// upgrade components again during next reconcile loop
-		if toUpdate.Spec.AutoUpdateComponents != nil &&
-			*toUpdate.Spec.AutoUpdateComponents == corev1.OnceAutoUpdate {
-			toUpdate.Spec.AutoUpdateComponents = nil
+		// The version from manifest does not match desired version, this may be caused by wrong configMap version.
+		if release.PortworxVersion != "" && toUpdate.Spec.Version != "" && release.PortworxVersion != toUpdate.Spec.Version {
```
pure-jliao, line 193: sorry but in what scenario will this happen? like the version is updated manually from stc spec? I see from line 189 if it's empty, it should be guaranteed to be the same.

drivers/storage/portworx/portworx_test.go
```diff
@@ -202,11 +203,56 @@ func TestGetStorkEnvMap(t *testing.T) {
...
+	recorder := record.NewFakeRecorder(100)
+	err := driver.Init(k8sClient, runtime.NewScheme(), recorder)
+	require.NoError(t, err)
+	cluster := &corev1.StorageCluster{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "px-cluster",
+			Namespace: namespace,
+		},
+		Spec: corev1.StorageClusterSpec{
+			Image: "test/image:2.10.0",
```
pure-jliao, line 234: hmm still confused, in this case shouldn't the version here overwrite the one in the configmap?

drivers/storage/portworx/portworx.go
haibinxie: I can probably rename the function, updateImageAndVersions. setting to empty is not here, because regardless what version, setting to empty should be called, hence it's outside of the function and it's always called.

drivers/storage/portworx/portworx.go
haibinxie: the scenario is only for px-versions configmap, user updates px-versions configmap, but does not update storageCluster, hence the version mismatches

drivers/storage/portworx/portworx_test.go
haibinxie: no it should not, we need to use same version in storagecluster and px-versions configmap, there is complaint that operator does not right version, and if the versions do not match there is no source of truth. in px-versions configmap scenario, to update px: 1. update px-versions configmap. 2. update storageCluster.spec.image.

drivers/storage/portworx/portworx.go
haibinxie: do you mean we should not reset to empty or something else?

pure-jliao: nvm so if the versions don't match, we only avoid updating component images, then everything else should still work fine with the old desired images.

drivers/storage/portworx/portworx_test.go
pure-jliao: ok so we should add one more validation here, like the desired stork image should not be updated in this case and should not be empty.

drivers/storage/portworx/portworx_test.go
```diff
@@ -202,11 +203,56 @@ func TestGetStorkEnvMap(t *testing.T) {
...
+	// Image version does not match configmap version
+	driver.SetDefaultsOnStorageCluster(cluster)
+	findEvent := strings.Contains(reflect.ValueOf(<-recorder.Events).String(), "does not match version manifest")
+	require.True(t, findEvent)
+	require.Equal(t, cluster.Status.Version, "")
+
+	// Image version matches configmap version
+	cluster.Spec.Image = "test/image:3.2.1"
+	driver.SetDefaultsOnStorageCluster(cluster)
+	require.Equal(t, cluster.Status.Version, "3.2.1")
```
pure-jliao, line 247: validate desired stork image here as well.

pure-jliao (approved):
LGTM after understanding the context. left one more comment about unit test validation.

drivers/storage/portworx/portworx_test.go
haibinxie: sounds good

PR #757, opened 2022-09-09
PTX-12273 Upgrade StorageCluster test with all components enabled

nikolaypopov:
Upgrade StorageCluster test with all components enabled

### DIFF
--- a/test/integration_test/basic_test.go
+++ b/test/integration_test/basic_test.go
@@ -116,12 +116,6 @@ var testStorageClusterBasicCases = []types.TestCase{
 			cluster, ok := objects[0].(*corev1.StorageCluster)
 			require.True(t, ok)
 			cluster.Name = "upgrade-stc-with-all-components-test"
-			alertManagerSecret, err := ci_utils.ParseSpecs("monitoring/alertmanager-secret.yaml")
-			require.NoError(t, err)
-
-			logrus.Infof("Creating alert manager secret")
-			err = ci_utils.CreateObjects(alertManagerSecret)
-			require.NoError(t, err)
 			return cluster
 		},
 		ShouldSkip: func(tc *types.TestCase) bool {
@@ -346,6 +340,22 @@ func BasicUpgradeStorageCluster(tc *types.TestCase) func(*testing.T) {
 		cluster, ok := testSpec.(*corev1.StorageCluster)
 		require.True(t, ok)
 
+		// Create AlertManager secret, if AlertManager is enabled
+		if cluster.Spec.Monitoring != nil {
+			if cluster.Spec.Monitoring.Prometheus != nil {
+				if cluster.Spec.Monitoring.Prometheus.AlertManager != nil {
+					if cluster.Spec.Monitoring.Prometheus.AlertManager.Enabled {
+						alertManagerSecret, err := ci_utils.ParseSpecs("monitoring/alertmanager-secret.yaml")
+						require.NoError(t, err)
+
+						logrus.Infof("Creating alert manager secret")
+						err = ci_utils.CreateObjects(alertManagerSecret)
+						require.NoError(t, err)
+					}
+				}
+			}
+		}
+
 		var lastHopURL string
 		for i, hopURL := range ci_utils.PxUpgradeHopsURLList {
 			// Get versions from URL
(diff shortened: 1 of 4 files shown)


test/integration_test/basic_test.go
```diff
@@ -107,6 +107,34 @@ var testStorageClusterBasicCases = []types.TestCase{
 		},
 		TestFunc: BasicUpgradeStorageCluster,
 	},
+	{
+		TestName:        "BasicUpgradeStorageClusterWithAllComponents",
```
pure-jliao, line 111: this should cover everything in the basic upgrade test case, if this test passes, we can run this test only in our pipeline, but let's keep the default upgrade case for now

test/integration_test/basic_test.go
```diff
@@ -107,6 +107,34 @@ var testStorageClusterBasicCases = []types.TestCase{
...
+		TestrailCaseIDs: []string{"C50241"},
+		TestSpec: func(t *testing.T) interface{} {
+			objects, err := ci_utils.ParseSpecs("storagecluster/storagecluster-with-all-components.yaml")
+			require.NoError(t, err)
+			cluster, ok := objects[0].(*corev1.StorageCluster)
+			require.True(t, ok)
+			cluster.Name = "upgrade-stc-with-all-components-test"
+			alertManagerSecret, err := ci_utils.ParseSpecs("monitoring/alertmanager-secret.yaml")
```
pure-jliao, line 119: can we parse and return the stc object only in this TestSpec function here without doing the alert manager secret creation inside? we can do this in the test function after getting the stc, check stc spec if alertmanager is enabled, then parse it then create.

test/integration_test/basic_test.go
```diff
@@ -107,6 +107,34 @@ var testStorageClusterBasicCases = []types.TestCase{
...
+			return cluster
+		},
+		ShouldSkip: func(tc *types.TestCase) bool {
+			if len(ci_utils.PxUpgradeHopsURLList) == 0 {
+				logrus.Info("--px-upgrade-hops-url-list is empty, cannot run BasicUpgradeStorageClusterWithAllComponents test")
+				return true
+			}
+			k8sVersion, _ := k8sutil.GetVersion()
+			pxVersion := ci_utils.GetPxVersionFromSpecGenURL(ci_utils.PxUpgradeHopsURLList[0])
+			return k8sVersion.GreaterThanOrEqual(k8sutil.K8sVer1_22) && pxVersion.LessThan(pxVer2_9)
```
pure-jliao, line 134: I don't remember exactly what's the case we want to skip the test. but is this still true?

test/integration_test/basic_test.go
```diff
@@ -352,6 +380,12 @@ func BasicUpgradeStorageCluster(tc *types.TestCase) func(*testing.T) {
 			lastHopURL = hopURL
 		}
 
+		// Delete AlertManager secret, if found
```
pure-jliao, line 383: can we do this inside of `UninstallAndValidateStorageCluster`? check if alert manager is enabled then delete the secret. as I noticed the migration test may leave stale alertmanager secret as well.

test/integration_test/testspec/storagecluster/storagecluster-with-all-components.yaml
```diff
@@ -0,0 +1,19 @@
+kind: StorageCluster
+apiVersion: core.libopenstorage.org/v1
```
pure-jliao, line 2: how about the pvc controller? and can we add some basic metadata here? so it's still a valid object that has all the fields required, test utils should be able to update the fields.

test/integration_test/basic_test.go
nikolaypopov: Moved

test/integration_test/basic_test.go
nikolaypopov: Done

test/integration_test/testspec/storagecluster/storagecluster-with-all-components.yaml
nikolaypopov: Added `pvc-controller` annotation to the spec

test/integration_test/basic_test.go
nikolaypopov: I think PX might have some sort of reqs, I honestly don't remember either, I left it as is for now

pure-jliao (approved):
LGTM! please test it first.

PR #758, opened 2022-09-09
[PWX-26110] Update vendors for sched-ops and controller-runtime to k8s 1.21.4 dependencies

pure-jliao:
* Update sched-ops vendor
* Update k8s dependency vendors to be consistent with sched-ops (1.21.4)
* Update controller-runtime to lowest k8s 1.21.4 supported version
* Replaced deprecated function
* Updated mock gen
* Fixed some unit test cases, seems with the new controller, objects get delete immediately before we could validate the fields.

Closes #


haibinxie (approved):
lgtm. one question, any specific reason to upgrade to 1.21 instead of newer like 1.24?

pure-jliao:
> 

originally we need to bump up sched-ops dependency, and it has k8s 1.21 vendors, to be consistent and avoid vendor conflicts.

PR #765, opened 2022-09-13
PTX-12868 Check pure-telemetry-certs for owner reference before deleting StorageCluster

nikolaypopov:
Check pure-telemetry-certs for owner reference before deleting StorageCluster

### DIFF
--- a/pkg/util/test/util.go
+++ b/pkg/util/test/util.go
@@ -413,9 +413,45 @@ func UninstallStorageCluster(cluster *corev1.StorageCluster, kubeconfig ...strin
 		if _, err = operatorops.Instance().UpdateStorageCluster(cluster); err != nil {
 			return err
 		}
+
+		defaultCheckReferenceTimeout := 30 * time.Second
+		defaultCheckReferenceInterval := 1 * time.Second
+		if err = checkTelemetrySecretGotOwnerReference(cluster, defaultCheckReferenceTimeout, defaultCheckReferenceInterval); err != nil {
+			return err
+		}
 	}
 
 	return operatorops.Instance().DeleteStorageCluster(cluster.Name, cluster.Namespace)
+}
+
+func checkTelemetrySecretGotOwnerReference(cluster *corev1.StorageCluster, timeout, interval time.Duration) error {
+	t := func() (interface{}, bool, error) {
+		secret, err := coreops.Instance().GetSecret("pure-telemetry-certs", cluster.Namespace)
+		if err != nil {
+			if errors.IsNotFound(err) {
+				return nil, false, nil
+			}
+			return nil, true, fmt.Errorf("failed to get secret pure-telemetry-certs")
+		}
+		logrus.Debugf("Found secret %s", secret.Name)
+
+		if len(secret.OwnerReferences) != 0 {
+			ownerRef := metav1.NewControllerRef(cluster, corev1.SchemeGroupVersion.WithKind("StorageCluster"))
+			for _, reference := range secret.OwnerReferences {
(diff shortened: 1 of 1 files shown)


pkg/util/test/util.go
```diff
@@ -413,11 +413,45 @@ func UninstallStorageCluster(cluster *corev1.StorageCluster, kubeconfig ...strin
...
+		}
 	}
 
 	return operatorops.Instance().DeleteStorageCluster(cluster.Name, cluster.Namespace)
 }
 
+func checkTelemetrySecretGotOwnerReference(cluster *corev1.StorageCluster, timeout, interval time.Duration) error {
+	t := func() (interface{}, bool, error) {
+		secret, err := coreops.Instance().GetSecret("pure-telemetry-certs", cluster.Namespace)
+		if err != nil {
```
pure-jliao, line 430: as we check when telemetry enabled so the secret must exist. so we retry for all errors. if not found then there's an issue.

pkg/util/test/util.go
```diff
@@ -413,11 +413,45 @@ func UninstallStorageCluster(cluster *corev1.StorageCluster, kubeconfig ...strin
...
+		defaultCheckReferenceInterval := 1 * time.Second
+		if err = checkTelemetrySecretGotOwnerReference(cluster, defaultCheckReferenceTimeout, defaultCheckReferenceInterval); err != nil {
+			return err
+		}
 	}
 
 	return operatorops.Instance().DeleteStorageCluster(cluster.Name, cluster.Namespace)
 }
 
+func checkTelemetrySecretGotOwnerReference(cluster *corev1.StorageCluster, timeout, interval time.Duration) error {
```
pure-jliao, line 427: if telemetry disabled, return nil directly, only check ref when telemetry enabled.

pkg/util/test/util.go
```diff
@@ -413,11 +413,45 @@ func UninstallStorageCluster(cluster *corev1.StorageCluster, kubeconfig ...strin
...
+	t := func() (interface{}, bool, error) {
+		secret, err := coreops.Instance().GetSecret("pure-telemetry-certs", cluster.Namespace)
+		if err != nil {
+			if errors.IsNotFound(err) {
+				return nil, false, nil
+			}
+		}
+		logrus.Debugf("Found secret %s", secret.Name)
+
+		if len(secret.OwnerReferences) != 0 {
```
pure-jliao, line 437:
```suggestion
		if len(secret.OwnerReferences) != 0 {
		ownerRef := metav1.NewControllerRef(cluster, pxutil.StorageClusterKind())
		for _, reference := range secret.OwnerReferences {
		    if reference.UID == ownerRef.UID {
		    }
		
```

pure-jliao:
btw can we fix this at the same PR? https://github.com/libopenstorage/operator/blob/master/pkg/util/test/util.go#L3241
see here for reference https://github.com/libopenstorage/operator/blob/px-rel-1.10.0/pkg/util/test/util.go#L1959
want to add more coverage.

pkg/util/test/util.go
nikolaypopov: Added a check for other errors

pkg/util/test/util.go
nikolaypopov: Won't do this as secret might exist from previously enabled telemetry

pkg/util/test/util.go
nikolaypopov: Done

pure-jliao (approved):
LGTM!

PR #771, opened 2022-09-15
PTX-12938 Modify custom-namespace test to run with all components enabled

nikolaypopov:
Modify custom-namespace test to run with all components enabled


test/integration_test/basic_test.go
```diff
@@ -295,6 +293,22 @@ func BasicInstallInCustomNamespace(tc *types.TestCase) func(*testing.T) {
 			}
 		}
 
+		// Create AlertManager secret, if AlertManager is enabled
+		if cluster.Spec.Monitoring != nil {
+			if cluster.Spec.Monitoring.Prometheus != nil {
+				if cluster.Spec.Monitoring.Prometheus.AlertManager != nil {
+					if cluster.Spec.Monitoring.Prometheus.AlertManager.Enabled {
+						alertManagerSecret, err := ci_utils.ParseSpecs("monitoring/alertmanager-secret-custom-namespace.yaml")
```
pure-jliao, line 301: are we able to change the namespace here after parsing instead of creating a new yaml spec?

test/integration_test/basic_test.go
```diff
@@ -45,18 +45,16 @@ var testStorageClusterBasicCases = []types.TestCase{
...
-					StartPort: func(val uint32) *uint32 { return &val }(17001),
-				},
-			}
+			objects, err := ci_utils.ParseSpecs("storagecluster/storagecluster-with-all-components.yaml")
+			require.NoError(t, err)
+			cluster, ok := objects[0].(*corev1.StorageCluster)
+			require.True(t, ok)
+			cluster.Name = "custom-ns-shifted-port-all-components"
+			cluster.Namespace = "custom-namespace"
+			cluster.Spec.StartPort = func(val uint32) *uint32 { return &val }(17001)
```
pure-jliao, line 57: nit: could simply create a bool pointer without function and looks more clear, but I'm ok with this

pure-jliao (approved):
LGTM, please verify the test could pass.

test/integration_test/basic_test.go
nikolaypopov: ParseSpecs unfortunately returns `func ParseSpecs(filename string) ([]runtime.Object, error)`, I tried and not sure how to change `namespace` in this `runtime.Object`

test/integration_test/basic_test.go
nikolaypopov: This is how it was previously done `StartPort: func(val uint32) *uint32 { return &val }` so I just left it

test/integration_test/basic_test.go
nikolaypopov: Done

PR #780, opened 2022-09-21
Add missing json labels for manifest generation

piyush-nimbalkar:
Planning to add automatic crd manifest generation so that we don't have to update both the golang object and crd schema. Going to use controller-tools to do that. Will add the generation in a later PR.

Closes #


haibinxie (approved):
This is a nice feature. once the other PR is finished, in future we can just regenerate the CRD in the release brach rather than manually merge the changes from master?

piyush-nimbalkar:
> This is a nice feature. once the other PR is finished, in future we can just regenerate the CRD in the release brach rather than manually merge the changes from master?

Yes, that's correct. We will have to define a process on how we cherrypick things. Generated files should be a separate commit and should not be cherrypicked but be rather generated again on respective branches.

PR #781, opened 2022-09-21
[PWX-26954] Enable telemetry by default

pure-jliao:
* Hold until we decide to enable it by default

Closes #

### DIFF
--- a/drivers/storage/portworx/portworx.go
+++ b/drivers/storage/portworx/portworx.go
@@ -696,10 +696,7 @@ func SetPortworxDefaults(toUpdate *corev1.StorageCluster, k8sVersion *version.Ve
 		}
 	}
 
-	if pxutil.IsTelemetryEnabled(toUpdate.Spec) && t.pxVersion.LessThan(pxutil.MinimumPxVersionCCM) {
-		toUpdate.Spec.Monitoring.Telemetry.Enabled = false // telemetry not supported for < 2.8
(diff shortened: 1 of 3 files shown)


drivers/storage/portworx/portworx.go
```diff
@@ -941,6 +938,33 @@ func setDefaultAutopilotProviders(
 	}
 }
 
+func setTelemetryDefaults(
+	toUpdate *corev1.StorageCluster,
+	pxVersion *version.Version,
+) {
+	if pxVersion.Equal(pxutil.MaximumPxVersion) {
+		return
+	}
```
piyush-nimbalkar, line 947: Let it be enabled if the version is not specified, right? It mostly means we are dealing with an unreleased version of PX.

pure-jliao: @piyush-nimbalkar  the issue is we call this  `SetPortworxDefault` first inside of `SetDefaultsOnStorageCluster`, where the image, versions got set later. so when no px image specified and before operator sets the default version, `GetPortworxVersion` will return this very large number. it's not the correct way imo and it breaks the unit test. in practice if no image specified, the reconcile loop should take care of this case.

pure-jliao: I feel `setTelemetryDefaults` along with some other default settings should be in set storage cluster default instead of portworx, and @haibinxie has plan to redesign the image and version related block, given this PR will lack full system testing if merged in 1.10, the change needs to be minimum and safe..

PR #783, opened 2022-09-22
[PWX-27142] Update ccm envoy config for cert validation

pure-jliao:

Closes #

### DIFF
--- a/deploy/ccm/envoy-config-collector-custom-proxy.yaml
+++ b/deploy/ccm/envoy-config-collector-custom-proxy.yaml
@@ -121,6 +121,13 @@ static_resources:
             name: tls_sds
             sds_config:
               path: /etc/envoy/tls_certificate_sds_secret.yaml
+          validation_context:
+            trusted_ca:
+              filename: /etc/ssl/certs/ca-certificates.crt
+            match_typed_subject_alt_names:
+            - san_type: DNS
+              matcher:
+                exact: REST_PROXY_URL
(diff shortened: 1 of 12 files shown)


deploy/ccm/envoy-config-collector-custom-proxy.yaml
```diff
@@ -121,6 +121,13 @@ static_resources:
             name: tls_sds
             sds_config:
               path: /etc/envoy/tls_certificate_sds_secret.yaml
+          validation_context:
+            trusted_ca:
+              filename: /etc/ssl/certs/ca-certificates.crt
+            match_typed_subject_alt_names:
+            - san_type: DNS
```
ezhang599, line 128: some indentation issue here. need two more spaces and so are two following lines. 
this seems to apply to all "match_typed_subject_alt_names" section, please help verify again.
correct format should be like (https://github.dev.purestorage.com/parts/ccm-go/blob/master/deploy/envoy-config-rest.yaml#L77)

pure-jliao: it's just different yaml conventions, should be the same. san_type is sub component of match_typed_subject_alt_names

ezhang599: cool, thanks for clarifying. I don't have any other issue now

ezhang599 (approved):

PR #785, opened 2022-09-23
PTX-13087 Run pxctl command with Security token if PX Security is enabled

nikolaypopov:
Run pxctl command with Security token if PX Security is enabled

### DIFF
--- a/pkg/util/test/util.go
+++ b/pkg/util/test/util.go
@@ -2684,7 +2684,7 @@ func ValidateTelemetryV1Disabled(cluster *corev1.StorageCluster, timeout, interv
 	}
 
 	// Validate Telemetry is Disabled in pxctl status
-	if err := validateTelemetryStatusInPxctl(false, cluster.Namespace); err != nil {
+	if err := validateTelemetryStatusInPxctl(false, cluster); err != nil {
 		return fmt.Errorf("failed to validate that Telemetry is Disabled in pxctl status, Err: %v", err)
 	}
 
@@ -3056,7 +3056,7 @@ func ValidateTelemetryV2Enabled(pxImageList map[string]string, cluster *corev1.S
 	}
 
 	// Validate Telemetry is Healthy in pxctl status
-	if err := validateTelemetryStatusInPxctl(true, cluster.Namespace); err != nil {
+	if err := validateTelemetryStatusInPxctl(true, cluster); err != nil {
 		return fmt.Errorf("failed to validate that Telemetry is Healthy in pxctl status, Err: %v", err)
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/util/test/util.go
```diff
@@ -3056,22 +3056,30 @@ func ValidateTelemetryV2Enabled(pxImageList map[string]string, cluster *corev1.S
...
 	logrus.Infof("All Telemetry components were successfully enabled/installed")
 	return nil
 }
 
-func validateTelemetryStatusInPxctl(telemetryShouldBeEnabled bool, namespace string) error {
+func validateTelemetryStatusInPxctl(telemetryShouldBeEnabled bool, cluster *corev1.StorageCluster) error {
 	listOptions := map[string]string{"name": "portworx"}
 	cmd := "pxctl status | grep Telemetry:"
 
+	if cluster.Spec.Security != nil {
```
pure-jliao, line 3071: && cluster.Spec.Security.Enabled
and move this inside of the retry block.

pkg/util/test/util.go
```diff
@@ -3107,6 +3115,18 @@ func validateTelemetryStatusInPxctl(telemetryShouldBeEnabled bool, namespace str
 	return nil
 }
 
+func getSecurityAdminToken(namespace string) (string, error) {
+	logrus.Debug("PX Security is enabled, getting token from px-admin-token secret")
+
+	secret, err := coreops.Instance().GetSecret("px-admin-token", namespace)
```
pure-jliao, line 3121: does this retry if network issue?

test/integration_test/basic_test.go
```diff
@@ -492,7 +492,7 @@ func BasicTelemetryRegression(tc *types.TestCase) func(*testing.T) {
 		if cluster.Spec.Monitoring != nil {
 			if cluster.Spec.Monitoring.Telemetry != nil {
 				if cluster.Spec.Monitoring.Telemetry.Enabled {
```
pure-jliao, line 494: nit: we can remove this condition check

pkg/util/test/util.go

pkg/util/test/util.go
nikolaypopov: I think its enough for it to be inside one retry loop

test/integration_test/basic_test.go
nikolaypopov: I will leave it for now as we are validating that its not enabled by default

pure-jliao: yes, but if it's enabled, it still return error. it's a duplicated condition check. but I'm ok with that.

pure-jliao (approved):

test/integration_test/basic_test.go
nikolaypopov: I misunderstood what you said, done ^^

PR #787, opened 2022-09-23
PTX-13089 Adding PX Auth validation on PX nodes

nikolaypopov:
Adding PX Auth validation on PX nodes

### DIFF
--- a/pkg/util/test/util.go
+++ b/pkg/util/test/util.go
@@ -110,8 +110,11 @@ const (
 	defaultTelemetryInPxctlValidationTimeout  = 5 * time.Minute
 	defaultTelemetryInPxctlValidationInterval = 30 * time.Second
 
-	defaultRunCmdInPxPodTimeout       = 25 * time.Second
-	defaultRunCmdInPxPodRetryInterval = 5 * time.Second
+	defaultPxAuthValidationTimeout  = 20 * time.Minute
+	defaultPxAuthValidationInterval = 30 * time.Second
+
+	defaultRunCmdInPxPodTimeout  = 25 * time.Second
+	defaultRunCmdInPxPodInterval = 5 * time.Second
 )
 
 // TestSpecPath is the path for all test specs. Due to currently functional test and
@@ -1186,7 +1189,7 @@ func validateComponents(pxImageList map[string]string, cluster *corev1.StorageCl
 	}
 
 	// Validate Security
-	previouslyEnabled := false
+	previouslyEnabled := false // NOTE: This is set to false by default as we are not expecting Security to be previously enabled here
 	if err = ValidateSecurity(cluster, previouslyEnabled, timeout, interval); err != nil {
 		return err
 	}
@@ -2425,6 +2428,10 @@ func ValidateSecurityEnabled(cluster *corev1.StorageCluster, storkDp *appsv1.Dep
 		return err
 	}
 
+	if err := validatePxAuthOnPxNodes(true, cluster); err != nil {
+		return fmt.Errorf("failed to validate PX Auth is enabled on PX nodes, Err: %v", err)
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


test/integration_test/basic_test.go
```diff
@@ -1011,10 +1011,8 @@ func BasicSecurityRegression(tc *types.TestCase) func(*testing.T) {
 
 		// Validate Security is not enabled by default
 		logrus.Info("Validate Security is not enabled by default")
-		if cluster.Spec.Security != nil {
-			if cluster.Spec.Security.Enabled {
-				require.False(t, cluster.Spec.Security.Enabled, "failed to validate Security is enabled: expected: false, actual: %v", cluster.Spec.Security.Enabled)
-			}
+		if cluster.Spec.Security != nil && cluster.Spec.Security.Enabled {
```
pure-jliao, line 1014: nit: the enabled check is not needed.
```
if cluster.Spec.Security != nil {
    require.False(t, cluster.Spec.Security.Enabled,
```

pkg/util/test/util.go
```diff
@@ -2735,12 +2746,15 @@ func runCmdInsidePxPod(pxPod *v1.Pod, cmd string, namespace string, ignoreErr bo
 		logrus.Debugf("[%s] Running command inside pod %s", pxPod.Name, cmds)
 		output, err := coreops.Instance().RunCommandInPod(cmds, pxPod.Name, "portworx", pxPod.Namespace)
 		if !ignoreErr && err != nil {
-			return nil, true, fmt.Errorf("[%s] failed to run command inside pod, command: %v, err: %v", pxPod.Name, cmds, err)
+			return "", true, fmt.Errorf("[%s] failed to run command inside pod, command: %v, err: %v", pxPod.Name, cmds, err)
+		}
+		if ignoreErr {
```
pure-jliao, line 2751: feel this block is not needed? we only retry when err != nil and not ignore, all the other cases it will not retry as long as it returns `output, false, nil`

pkg/util/test/util.go
```diff
@@ -3064,6 +3078,70 @@ func ValidateTelemetryV2Enabled(pxImageList map[string]string, cluster *corev1.S
...
 }
 
+func validatePxAuthOnPxNodes(pxAuthShouldBeEnabled bool, cluster *corev1.StorageCluster) error {
+	listOptions := map[string]string{"name": "portworx"}
+	cmd := "pxctl status"
+	cmdWithPxAuthToken := ""
+
+	logrus.Infof("Run pxctl status on all PX nodes to determine if PX Auth is enabled/disabled properly")
+	t := func() (interface{}, bool, error) {
+		if cluster.Spec.Security != nil && cluster.Spec.Security.Enabled {
```
pure-jliao, line 3088: we already have `pxAuthShouldBeEnabled` as an input, so just use `if pxAuthShouldBeEnabled`?

pure-jliao (approved):
LGTM overall after addressing comments!

pkg/util/test/util.go
nikolaypopov: I tried couple of combinations and it doesn't quite work as I needed it, sometimes I don't want to retry when there is an error, as I am expecting error.

pkg/util/test/util.go

test/integration_test/basic_test.go
nikolaypopov: Done

pkg/util/test/util.go
pure-jliao: sorry I mean return `output, false, err`, as long as we return false it doesn't retry I think. and we can keep output as well for debugging.

nikolaypopov: If there is an error, there is no output and the `DoRetryWithTimeout()` will panic, that was one of the issues I experienced so we would need to do another statement to check if output is empty I guess either way.

nikolaypopov: Nvm it works ^^

PR #789, opened 2022-10-04
[PWX-25353] Auto detect eks cloud environment

pure-jliao:

Closes #

still in progress, but ready for design wise review.


pkg/cloudprovider/cloudprovider.go
```diff
@@ -13,17 +16,70 @@ const (
...
+		if node.Spec.ProviderID != "" {
+			tokens := strings.Split(node.Spec.ProviderID, "://")
+			if len(tokens) == 2 {
+				cloudProviderName = tokens[0]
+				break
+			} // else provider id is invalid
+		}
+	}
+
+	instance = New(cloudProviderName)
```
Pure-AdamuKaapan, line 77: We should add a check if the cloud provider name never gets set? Or is this handled inside the `New()` method?

pkg/cloudprovider/cloudprovider.go
```diff
@@ -53,10 +118,15 @@ func (d *defaultProvider) GetZone(node *v1.Node) (string, error) {
...
+func (d *defaultProvider) IsK8sServiceProvider() bool {
+	return false
+}
+
+func initRegistry() {
 	providerRegistryLock.Lock()
 	defer providerRegistryLock.Unlock()
 
 	providerRegistry = make(map[string]Ops)
 	providerRegistry[azureName] = &azure{}
```
Pure-AdamuKaapan, line 130: [nit] it's strange having `AWSProviderName` be public but `azureName` be private.

pkg/cloudprovider/aws.go
```diff
@@ -0,0 +1,39 @@
...
+	// AWSProviderName AWS cloud provider name
+	AWSProviderName = "aws"
+)
+
+type aws struct {
+	isEks bool
+}
+
+func (a *aws) init(nodeList *v1.NodeList) error {
+	for _, node := range nodeList.Items {
```
Pure-AdamuKaapan, line 20: This seems very similar to the loop at line `https://github.com/libopenstorage/operator/pull/789/files#diff-a6af1de71edb43b701d83b9ee10e345cf3c6ef3fc738ee3132c3309b6958896dR65`... any reason we need it in both places and/or can we reduce code duplication somehow?

pkg/cloudprovider/azure.go
```diff
@@ -14,6 +14,10 @@ const (
 
 type azure struct{}
 
+func (a *azure) init(nodeList *v1.NodeList) error {
+	return nil
```
Pure-AdamuKaapan, line 18:
```suggestion
	return fmt.Errorf("azure cloud provider not yet implemented")
```

Pure-AdamuKaapan (commented):
Some feedback but generally makes sense

pkg/cloudprovider/aws.go
pure-jliao: at cloud provider init we try this to check the provider name. given different providers may use different way to init, here just aws provider we need to scan nodes again not necessary other providers. to make it clean and isolate the general init and specific provider init I put another loop here. don't have a better idea yet, but could be improved once we have all provider implementations in the future.

pkg/cloudprovider/azure.go
pure-jliao: we actually have azure cloudprovider needed somewhere else. the provider is needed, but just nothing required in this init so far.

pkg/cloudprovider/cloudprovider.go
pure-jliao: will import the names from cloudops vendor soon.

pkg/cloudprovider/cloudprovider.go
pure-jliao: if it's not a registered provider, the name can be registered as an empty provider. it's used somewhere else as well, don't want to break things.

pkg/cloudprovider/cloudprovider.go
```diff
@@ -53,10 +119,15 @@ func (d *defaultProvider) GetZone(node *v1.Node) (string, error) {
...
+}
+
+func initRegistry() {
 	providerRegistryLock.Lock()
 	defer providerRegistryLock.Unlock()
 
 	providerRegistry = make(map[string]Ops)
-	providerRegistry[azureName] = &azure{}
+	providerRegistry[cloudops.Azure] = &azure{}
+	providerRegistry[string(cloudops.AWS)] = &aws{}
```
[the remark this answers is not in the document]
pure-jliao: because we only have aws specified differently in cloudops vendor

pkg/cloudprovider/aws.go
```diff
@@ -0,0 +1,49 @@
...
+	return nil
+}
+
+func (a *aws) Name() string {
+	return string(cloudops.AWS)
+}
+
+func (a *aws) CheckCloudDrivePermission() error {
+	// TODO: check the permission here by doing dummy drive operations
+	logrus.Infof("JLIAO: doing aws permisison check")
```
Pure-AdamuKaapan, line 40: [nit] Remove before merge, maybe replace with `return fmt.Errorf("CheckCloudDrivePermission not implemented")` just for this PR

pkg/cloudprovider/cloudprovider.go
```diff
@@ -1,62 +1,140 @@
...
+func getK8sDistributionName() (string, error) {
+	k8sVersion, err := coreops.Instance().GetVersion()
+	if err != nil {
+		return "", err
 	}
-	return node.Labels[failureDomainZoneKey], nil
+
+	logrus.WithFields(logrus.Fields{
+		"info": k8sVersion,
+	}).Infof("JLIAO: get k8s version")
```
Pure-AdamuKaapan, line 114: Remove before merge or make a "formal" log line?

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -405,6 +413,27 @@ func (c *Controller) validateCloudStorageLabelKey(cluster *corev1.StorageCluster
...
+func (c *Controller) validatePreflightCheck(cluster *corev1.StorageCluster) error {
+	check, err := pxutil.DoPreflightCheck(cluster)
+	if !check || err != nil {
+		return err
+	}
+
+	// Do the preflight check for eks only for now to check the cloud drive permission
+	if cloudprovider.IsEKS() {
+		if err := cloudprovider.Instance().CheckCloudDrivePermission(); err != nil {
+			logrus.Errorf("permission check for eks cloud drive failed")
```
Pure-AdamuKaapan, line 425:
```suggestion
			logrus.WithError(err).Errorf("permission check for eks cloud drive failed")
```

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -405,6 +413,27 @@ func (c *Controller) validateCloudStorageLabelKey(cluster *corev1.StorageCluster
...
+			// TODO: set the failure status
+			return err
+		}
+		// TODO: set the success status
```
Pure-AdamuKaapan, line 429: Add a log line so we know the prereqs went well?

Pure-AdamuKaapan (approved):
Some minor nits (mainly changing logs) but overall LGTM!

pkg/cloudprovider/cloudprovider.go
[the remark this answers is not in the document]
piyush-nimbalkar: It's not actually. All of them are of type `ProviderType`.
https://github.com/libopenstorage/cloudops/blob/master/cloudops.go#L20

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -278,18 +277,27 @@ func (c *Controller) waitingForMigrationApproval(cluster *corev1.StorageCluster)
...
 	}
 	if err := c.validateCustomAnnotations(cluster); err != nil {
+		cluster.Status.Phase = string(corev1.ClusterOperationFailed)
 		return err
 	}
 	if err := c.validateCloudStorageLabelKey(cluster); err != nil {
+		cluster.Status.Phase = string(corev1.ClusterOperationFailed)
+		return err
+	}
+	if err := c.validatePreflightCheck(cluster); err != nil {
```
piyush-nimbalkar: Let's check this after `Driver.Validate` so we call the expensive cloud APIs at the very end.

pkg/cloudprovider/cloudprovider.go
pure-jliao: will address all log lines before merging, currently for easier debugging purpose.

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -405,6 +405,55 @@ func (c *Controller) validateCloudStorageLabelKey(cluster *corev1.StorageCluster
...
 }
 
+func (c *Controller) runPreflightCheck(cluster *corev1.StorageCluster) error {
+	check, ok := cluster.Annotations[pxutil.AnnotationPreflightCheck]
+	if !ok {
+		return nil
+	} else if check == "false" {
+		if cluster.Status.Phase == string(corev1.ClusterConditionTypePreflight)+string(corev1.ClusterOperationFailed) {
+			return fmt.Errorf("preflight check failed, please run it again")
+		}
```
piyush-nimbalkar, line 415: This should not result into returning an error, right? Shouldn't it just mean don't run the pre-flight check?

pkg/cloudprovider/aws.go
```diff
@@ -0,0 +1,107 @@
...
+
+func (a *aws) Name() string {
+	return string(cloudops.AWS)
+}
+
+func (a *aws) CheckCloudDrivePermission() error {
+	// check the permission here by doing dummy drive operations
+	logrus.Infof("JLIAO: doing aws permisison check")
+	volTypeGp2 := opsworks.VolumeTypeGp2
+	volID := "px-check-aws-cloud-permission"
```
piyush-nimbalkar, line 44: Let's add the cluster name or even better storagecluster uuid to the volume ID for uniqueness. Multiple px clusters coming up together could cause race conditions.

pkg/cloudprovider/aws.go
```diff
@@ -0,0 +1,107 @@
...
+	volSize := int64(1)
+
+	// Create
+	v, err := a.ops.Create(&ec2.Volume{
```
piyush-nimbalkar, line 48: We should handle already exists error gracefully and not fail. It could be a common issue during retries where the volume created succeeded but other operations failed.

pkg/cloudprovider/aws.go
```diff
@@ -0,0 +1,107 @@
...
+		return err
+	}
+
+	logrus.WithFields(logrus.Fields{
+		"volume": v,
+	}).Infof("JLIAO: volume created")
+
+	// Attach
+	vol := v.(*ec2.Volume)
+	path, err := a.ops.Attach(*vol.VolumeId, nil)
```
piyush-nimbalkar, line 64: Let's check if already attached before calling attach or handle the already attached error gracefully.

pkg/cloudprovider/aws.go
```diff
@@ -0,0 +1,107 @@
...
+	if err != nil {
+		logrus.Errorf("JLIAO: failed to attach aws volume: %v", err)
+		return err
+	}
+	logrus.WithFields(logrus.Fields{
+		"path": path,
+	}).Infof("JLIAO: volume created")
+
+	// Tag
+	if err := a.ops.ApplyTags(volID, map[string]string{"testkey": "testval"}); err != nil {
```
piyush-nimbalkar, line 74: Let's add slightly meaningful names to the tags/labels.

pkg/cloudprovider/aws.go
```diff
@@ -0,0 +1,107 @@
...
+		logrus.Errorf("JLIAO: failed to tag %v", err)
+		return err
+	}
+
+	// Expand
+	if _, err := a.ops.Expand(volID, 2); err != nil {
+		logrus.Errorf("JLIAO: failed to expand %v", err)
+		return err
+	}
```
piyush-nimbalkar, line 83: Let's skip this one for now. It will take a long time to finish the preflight and the expand won't block the initial installation either.

pkg/cloudprovider/cloudprovider.go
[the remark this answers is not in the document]
pure-jliao: see the code you sent, aws has `ProviderType`, the rest are string

pkg/controller/storagecluster/storagecluster.go
pure-jliao: idea is if it ran preflight and failed, in the next reconcile loop it should be still blocked. if no error returned here, it will proceed to the next step later. here user can fix their prerequisite, then trigger the preflight check again by changing the annotation, otherwise everything got blocked here.

drivers/storage/portworx/portworx.go
```diff
@@ -153,6 +155,15 @@ func (p *portworx) GetSelectorLabels() map[string]string {
 }
 
 func (p *portworx) SetDefaultsOnStorageCluster(toUpdate *corev1.StorageCluster) {
+	if cloudprovider.IsEKS() {
+		toUpdate.Annotations[pxutil.AnnotationIsEKS] = "true"
+		// Do preflight check for eks cloud permission once before px installation
+		// TODO: use stc status instead of annotation? but annotation can be set by UI in the future
```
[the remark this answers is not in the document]
pure-jliao: I feel we may still need the annotation, because we need a switch to control the preflight, to the user, to future UI. If not using annotation we may need another spec for the same purpose.

drivers/storage/portworx/portworx.go
```diff
@@ -153,6 +155,14 @@ func (p *portworx) GetSelectorLabels() map[string]string {
 }
 
 func (p *portworx) SetDefaultsOnStorageCluster(toUpdate *corev1.StorageCluster) {
+	if preflight.IsEKS() {
+		toUpdate.Annotations[pxutil.AnnotationIsEKS] = "true"
+		// Do preflight check for eks cloud permission once before px installation
+		if _, ok := toUpdate.Annotations[pxutil.AnnotationPreflightCheck]; !ok {
```
Pure-AdamuKaapan, line 161: [nit] maybe move this so there's a "requiresPreflight" check, set it to true here, and then in a separate if check you can apply the preflight check annotation. That way it scales to multiple different providers easier in the future.

pkg/cloudprovider/cloudprovider.go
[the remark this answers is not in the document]
Pure-AdamuKaapan: @piyush-nimbalkar actually, that was my first reaction too, but I just gave a quick try in the Go Playground and it actually is like @pure-jliao said?? Really throws me off

https://go.dev/play/p/UVTML4FeLEU

For example, this compiles fine, but obviously they're not both `int`s.
```
const (
	a int = 5
	b     = "asdf"
)
```

pkg/controller/storagecluster/storagecluster.go
Pure-AdamuKaapan: I think we'll need a doc for that, I could see myself getting tripped up by that: "I fixed what you said, why aren't you starting again??"

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -405,6 +406,56 @@ func (c *Controller) validateCloudStorageLabelKey(cluster *corev1.StorageCluster
...
+	toUpdate := cluster.DeepCopy()
+	toUpdate.Annotations[pxutil.AnnotationPreflightCheck] = "false"
+	var err error
+
+	// Do the preflight check for eks only for now to check the cloud drive permission
+	if preflight.IsEKS() {
+		if err = preflight.Instance().CheckCloudDrivePermission(); err != nil {
+			logrus.WithError(err).Errorf("permission check for eks cloud drive failed")
+		} else {
+			logrus.Infof("JLIAO: permission check for eks cloud drive passed")
```
Pure-AdamuKaapan, line 430: Remove or convert to a "proper" log line

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -405,6 +406,56 @@ func (c *Controller) validateCloudStorageLabelKey(cluster *corev1.StorageCluster
...
+		}
+	}
+
+	// TODO: validate cloud permission for other providers as well
+
+	if err != nil {
+		logrus.Infof("JLIAO: setting precheck status to failed")
```
Pure-AdamuKaapan, line 437: Same for these two, remove or convert to a "proper" log

pkg/preflight/aws.go
```diff
@@ -0,0 +1,133 @@
...
+func (a *aws) initCloudOps() error {
+	if a.ops != nil {
+		return nil
+	}
+
+	// Initialize aws client for cloud drive permission checks
+	// TODO: use provided credentials to create a client
+	client, err := awsops.NewClient()
+	if err != nil {
+		logrus.Infof("JLIAO: err: %v", err)
```
Pure-AdamuKaapan, line 34: Same for this file, remove all "JLIAO" lines or make them into more permanent ones

pkg/preflight/aws.go
```diff
@@ -0,0 +1,133 @@
...
+			}).Errorf("JLIAO: volume not attached")
+			return fmt.Errorf("JLIAO: volume not attached")
+		}
+		logrus.WithFields(logrus.Fields{
+			"volume": vol,
+		}).Infof("JLIAO: volume attached")
+	}
+
+	// Detach
+	if err := a.ops.Detach(*vol.VolumeId); err != nil {
```
Pure-AdamuKaapan, line 116: This flow makes me nervous: if we fail partway through, we leave an EBS volume still attached to the volume, or still created at the very least, and those costs can add up very quickly (I know from homelab experience 🙃 ).

Can we put the Detach and Delete inside a defer block? That way we'll always try to clean up after ourselves. It may be easier to split it up into multiple methods so you can have an individual defer on each one (have the creation process be one method, which will defer deletion, and then the attach method will be separate and defer detachment).

Pure-AdamuKaapan (changes requested):
I have a little nervous comment about the EBS preflight check part, but the rest looks overall good. A few other comments as well.

pkg/controller/storagecluster/storagecluster.go
pure-jliao: don't worry about all log lines, will do before merge.

drivers/storage/portworx/portworx_test.go
```diff
@@ -400,6 +402,55 @@ func TestSetDefaultsOnStorageCluster(t *testing.T) {
...
+		},
+	}
+
+	// TestCase: default cloud provider
+	err = preflight.InitPreflightChecker()
+	require.NoError(t, err)
+	driver.SetDefaultsOnStorageCluster(cluster)
+	_, ok := cluster.Annotations[pxutil.AnnotationIsEKS]
+	require.False(t, ok)
+	_, ok = cluster.Annotations[pxutil.AnnotationPreflightCheck]
```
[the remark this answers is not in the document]
pure-jliao: will use the annotation for now, until later we finalize all the workflow and found it's necessary to have a separate spec for that.

pkg/preflight/aws.go
```diff
@@ -0,0 +1,162 @@
...
+
+func (a *aws) cleanupVolumes() {
+	if len(a.volList) == 0 {
+		return
+	}
+
+	logrus.Infof("cleaning up volumes created in preflight check")
+	for vol := range a.volList {
+		err := a.ops.Detach(vol)
+		if err == nil {
```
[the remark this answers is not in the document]
pure-jliao: the preflight already passed at this step, so guaranteed to have permission to do those operatorion, we try our best to delete old volumes. for now if it fails we don't do anything.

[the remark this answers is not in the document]
Pure-AdamuKaapan: I agree, I'm also concerned about this if the Detach fails for some other reason like temporary AWS API errors, some other weird thing (I dunno if there's a "detach lock" feature), or who knows what, because then we could leave a disk just hanging forever.

Do we have a way to check for preflight disks after the fact to continually keep them cleaned up?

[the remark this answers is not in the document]
pure-jliao: idea solution should be when creating the volume, we apply a tag saying it's a preflight check volume, associated with the cluster uid or something. during the cleanup, operator should list volume based on the tag then delete all of them. don't know how complex it is, but definitely a more proper way. That's something we need to do before GA, but can be done in following PRs.

pure-jliao:
Update in PR: 
* fixed annotation nil map crash
* labels added when creating preflight check volumes: volume ID and cluster name
* do volume list based on volume labels (cluster name) then clean up after permission check

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -405,6 +406,54 @@ func (c *Controller) validateCloudStorageLabelKey(cluster *corev1.StorageCluster
 	return nil
 }
 
+func (c *Controller) runPreflightCheck(cluster *corev1.StorageCluster) error {
+	check, ok := cluster.Annotations[pxutil.AnnotationPreflightCheck]
+	if !ok || check == "skip" {
```
haibinxie, line 411: can we make skip, false etc case insensitive.

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -405,6 +406,54 @@ func (c *Controller) validateCloudStorageLabelKey(cluster *corev1.StorageCluster
...
+		return nil
+	} else if check == "false" {
+		if cluster.Status.Phase == string(corev1.ClusterConditionTypePreflight)+string(corev1.ClusterOperationFailed) {
```
haibinxie, line 414: define new phase for it, the concatenated string is not in good format, also it's not easy to refer in multiple places.

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -405,6 +406,54 @@ func (c *Controller) validateCloudStorageLabelKey(cluster *corev1.StorageCluster
...
+			return fmt.Errorf("please make sure your cluster meet all prerequisites and rerun preflight check")
+		}
+		return nil
```
haibinxie, line 417: shouldnt we return error here so it wont proceed with installation?

pure-jliao: the idea here is after each check, the annotation will be flipped to `false`, and we only return error when it says `PreflightFailed` in the status. otherwise we should proceed with the installation. Will have the preflight status in condition instead of Phase in future PRs.

pkg/controller/storagecluster/storagecluster.go
pure-jliao: will use condition instead of phase, will see how I can address this there.

haibinxie (approved):

pkg/controller/storagecluster/storagecluster.go
piyush-nimbalkar: Create a ticket for the conditions, if not already created.

pkg/preflight/preflight.go
```diff
@@ -0,0 +1,119 @@
...
+	ProviderName() string
+
+	// K8sDistributionName returns the k8s distribution providerName
+	K8sDistributionName() string
+
+	// CheckCloudDrivePermission checks if permissions for drive operation is granted
+	CheckCloudDrivePermission(cluster *corev1.StorageCluster) error
+}
+
+func (c checker) ProviderName() string {
```
piyush-nimbalkar, line 35: nit: Why not use a pointer receiver? 
`func (c *checker) ...`

pkg/preflight/preflight.go
```diff
@@ -0,0 +1,119 @@
...
+	if IsEKS() {
+		instance = &aws{
+			checker: *c,
+		}
+	}
+
+	return nil
+}
+
+func getCloudProviderName() (string, error) {
```
piyush-nimbalkar, line 87: Ideally pre-flight checker should check with cloudprovider package to find the cloud provider and distribution. It is not the responsibility of the checker to worry about finding that. But it's okay for now.

pkg/preflight/aws.go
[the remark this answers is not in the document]
piyush-nimbalkar: Create a ticket to track detach delete failures.

piyush-nimbalkar (approved):
This is good, @pure-jliao!
Minor comments and questions.

pkg/preflight/preflight.go
pure-jliao: Use to have all checks in cloudprovider until we decided to have a preflight interface. Now I only initial singleton for preflight, then cloudprovider can get everything from it. It's simpler with such initialization in one place. will think about if there's better way to do so.

PR #792, opened 2022-10-06
[PWX-27292] Change default phonehome listening port from 9090 to 9024

pure-jliao:

Closes #

### DIFF
--- a/drivers/storage/portworx/testspec/ccmGoPhonehomeConfigMap.yaml
+++ b/drivers/storage/portworx/testspec/ccmGoPhonehomeConfigMap.yaml
@@ -12,7 +12,7 @@ data:
   ccm.properties: |-
     {
       "product_name": "portworx",
-      "port": "10090",
+      "port": "10021",
       "node_name_key": "K8S_NODE_NAME",
       "envoy_port": "13002",
       "logupload": {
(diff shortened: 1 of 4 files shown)


jrivera-px (approved):
LGTM...

drivers/storage/portworx/testspec/ccmGoPhonehomeConfigMap.yaml
```diff
@@ -12,7 +12,7 @@ data:
   ccm.properties: |-
     {
       "product_name": "portworx",
-      "port": "10090",
+      "port": "10024",
```
piyush-nimbalkar, line 15: I believe this would be startPort + 20 = 10021, assuming the start port for this test is 10001.
@jrivera-px I am assuming you are doing the same in porx.

jrivera-px: Start port + 20 yeah...

pure-jliao: why +20? the 9021 is in use seems https://github.com/portworx/porx/blob/master/build/piglet-ubuntu/docker-entry-point.sh#L2575

piyush-nimbalkar: It will come out to 9021 only if the start port is 9001. We are talking of a different start port. If the start port is not 9001, then we just do start port + 20.

piyush-nimbalkar (approved):

PR #804, opened 2022-10-26
Create operator 1.10.1 release branch

haibinxie:


deploy/operator.yaml
```diff
@@ -20,7 +20,7 @@ spec:
     spec:
       containers:
       - name: portworx-operator
-        image: portworx/px-operator:1.10.0-dev
+        image: portworx/px-operator:1.10.1-dev
         imagePullPolicy: Always
         command:
         - /operator
```
pure-jliao, line 26: line 27, do we still need the `--verbose` here?

pure-jliao (approved):
one question regarding the verbose flag, otherwise LGTM

deploy/operator.yaml
haibinxie: any idea who uses this operator.yaml? the customer facing ones are spec gen and CSV which have been removed.

pure-jliao: I'm not sure but I think we can merge the change first.

piyush-nimbalkar: No one outside uses this. It is just for development.

PR #805, opened 2022-10-27
PWX-27538 add registry.k8s.io as common registry

haibinxie:
k8s will move from k8s.gcr.io to registry.k8s.io, add this new registry as common registry, so if an image has the prefix it's handled correctly when replacing with custom registry. this is just a proactive change. at this moment we may not have any image in the new registry yet.


pure-jliao (approved):
if we don't have images using this registry then should be good..

PR #807, opened 2022-10-31
PWX-27563: Setting stork as the default scheduler for px-csi-ext pods

pp511:

When CSI is enabled  PVC provisioning goes through px-csi-ext-* pods which is a deployment of 3 replicas. If these 3 PX nodes go offline none of the PVC creations would succeed even if the PX cluster is operational. In such cases we need to move the px-csi-ext pods to other nodes in the cluster.
This change will make Stork as the default scheduler for px-csi-ext-* pods. Change will be made in Stork to detect px-csi-ext pods are running on an offline PX node such that it can delete the csi-ext pod and schedule them on the online nodes.

Closes #https://portworx.atlassian.net/browse/PWX-27563

* Refer to https://portworx.atlassian.net/browse/PWX-27549 for details
* Tested  on dev setup that stork is the default scheduler for px-csi-ext-*  deployment.
* Ran the following job with these changes in this PR for regressions https://jenkins.pwx.dev.purestorage.com/job/test-op-csi-e2e/1/

### DIFF
--- a/drivers/storage/portworx/component/csi.go
+++ b/drivers/storage/portworx/component/csi.go
@@ -43,8 +43,6 @@ const (
 	CSIServiceName = "px-csi-service"
 	// CSIApplicationName name of the CSI application (deployment/statefulset)
 	CSIApplicationName = "px-csi-ext"
-	// CSIApplicationSchedulerName name of the CSI application pod scheduler
-	CSIApplicationSchedulerName = "stork"
 
 	csiProvisionerContainerName             = "csi-external-provisioner"
 	csiAttacherContainerName                = "csi-attacher"
@@ -511,11 +509,11 @@ func (c *csi) createDeployment(
 		healthMonitorControllerImage != existingHealthMonitorControllerContainerName ||
 		util.HasPullSecretChanged(cluster, existingDeployment.Spec.Template.Spec.ImagePullSecrets) ||
 		util.HasNodeAffinityChanged(cluster, existingDeployment.Spec.Template.Spec.Affinity) ||
+		util.HasSchedulerStateChanged(cluster, existingDeployment.Spec.Template.Spec.SchedulerName) ||
 		util.HaveTolerationsChanged(cluster, existingDeployment.Spec.Template.Spec.Tolerations) ||
 		util.HaveTopologySpreadConstraintsChanged(updatedTopologySpreadConstraints,
 			existingDeployment.Spec.Template.Spec.TopologySpreadConstraints) ||
 		hasCSITopologyChanged(cluster, existingDeployment)
-
 	if !c.isCreated || modified {
 		deployment := getCSIDeploymentSpec(cluster, csiConfig, ownerRef, provisionerImage, attacherImage,
 			snapshotterImage, resizerImage, snapshotControllerImage, healthMonitorControllerImage, updatedTopologySpreadConstraints)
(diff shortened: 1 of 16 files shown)


ggriffiths:
@pp511 do you know the behavior if stork is disabled in the storagecluster? Will k8s ignore the stork scheduler, or error out?

pp511:
@ggriffiths If stork is disabled then then csi pods will not come online. Default k8 scheduler will not be used. 
We should use the default scheduler when stork is disabled. Right? Please confirm and I will update the PR.

ggriffiths:
Yes, there might be cases where someone disables stork and enables CSI.

haibinxie (commented):
If node goes offline, k8s default scheduler will reschedule the pod, doen't that work for us?

pp511:
@ggriffiths Updated the PR to use Stork as the scheduler only when it is enabled in StorageCluster.
@haibinxie Currently Stork’s monitor controller monitors all the application pods that use Portworx PVCs and deletes such pods which are running on offline PX nodes. Something similar needs to be done for CSI pods where Stork needs to identify such px-csi-ext pods and use its scheduling logic even though there are no Portworx PVCs being used by them.

ggriffiths (approved):
LGTM

drivers/storage/portworx/component/csi.go
```diff
@@ -666,6 +668,10 @@ func getCSIDeploymentSpec(
 		},
 	}
 
+	if pxutil.IsStorkEnabled(cluster) {
```
piyush-nimbalkar, line 671: This won't handle if stork is disabled at runtime. This is not a common scenario, but we should handle it.
Currently the CSI deployment gets recreated only if certain things change in the deployment based on the storagecluster spec. Check line https://github.com/libopenstorage/operator/pull/807/files#diff-41fa95b9f2df1987f95180171e28b9d77de07ce97cf8d5c34dea99ca3c54e0c2R506

We will have to add some custom logic there for the scheduler field or check with @haibinxie to use the custom DeepCopy function he introduced.

piyush-nimbalkar: Can you verify what happens if the stork scheduler is removed at runtime? Try deleting one of the stork pods and see what happens.

ggriffiths: Ah yes, we need to make sure the px-csi-ext is bounced when stork is disabled or enabled.

pp511: @piyush-nimbalkar Added the changes to handle the scenario when stork is enabled or disabled at run time.

pkg/util/util.go
```diff
@@ -284,6 +287,22 @@ func HasNodeAffinityChanged(
 	return !reflect.DeepEqual(cluster.Spec.Placement.NodeAffinity, existingAffinity.NodeAffinity)
 }
 
+// HasSchedulerStateChanged checks if the stork has been enabled/disabled in the StorageCluster
+func HasSchedulerStateChanged(
+	cluster *corev1.StorageCluster,
+	previousSchedulerName string,
+) bool {
+	if cluster.Spec.Stork == nil {
+		return false
```
piyush-nimbalkar, line 296: If stork was enabled before and then the spec was set to empty, this should return true.

piyush-nimbalkar: Can you add UTs for this function?

pp511: Done

piyush-nimbalkar (approved):

piyush-nimbalkar:
> If node goes offline, k8s default scheduler will reschedule the pod, doen't that work for us?

@haibinxie here we are talking about portworx node going offline. Kubernetes won't reschedule the csi ext pod in that case.

PR #808, opened 2022-11-02
[PWX-25354] Set default cloud storage spec on EKS

pure-jliao:
* Set default eks cloud storage specs in `SetPortworxDefaults`
* As storage node per zone is needed, move max storage node per zone logic from controller API `setStorageClusterDefaults` (before set eks default) to Portworx API `SetDefaultsOnStorageCluster` (after set eks default)
* Fix unit test cases for changes above

Closes #

* Hold until max storage node got merged


drivers/storage/portworx/portworx.go
```diff
@@ -153,7 +155,176 @@ func (p *portworx) GetSelectorLabels() map[string]string {
...
+
+		nodeList := &v1.NodeList{}
+		err = p.k8sClient.List(context.TODO(), nodeList, &client.ListOptions{})
+		if err != nil {
+			return err
+		}
+
+		if pxutil.IsFreshInstall(toUpdate) {
+			// Let's do this only when it's a fresh install of px
+			// TODO:  get node list
```
Pure-AdamuKaapan, line 176: Is this a TODO for this PR, or later? If later, is there a PWX?

drivers/storage/portworx/util/util.go
```diff
@@ -1172,3 +1172,8 @@ func CountStorageNodes(
 
 	return storageNodesCount, nil
 }
+
+// IsFreshInstall checks whether it's a fresh Portworx install
+func IsFreshInstall(cluster *corev1.StorageCluster) bool {
+	return cluster.Status.Phase == ""
```
Pure-AdamuKaapan, line 1178: Would this cause any issues on retries? Is the phase updated if it fails the first time?

Pure-AdamuKaapan (approved):
Overall makes sense, left a couple questions but I think it's good

drivers/storage/portworx/portworx.go
pure-jliao: oh sorry, already resolved in this PR, will delete this.

drivers/storage/portworx/util/util.go
pure-jliao: good catch, after review I saw one corner case here, but not in scope of this ticket. I'm going to create a bug for that.
basically even if it's a fresh install, before stc sync during validation if it fails, phase will be set to Failed, then next retry it won't be able to set this default number of storage node. So far I don't have any easy solutions, but should be able to fix this once finish this https://portworx.atlassian.net/browse/PWX-27765. cc: @piyush-nimbalkar

PR #810, opened 2022-11-02
PWX-27591: Default value for maxStorageNodesPerZone

nrevanna:
The previous default of 3 storage nodes did not fit for all cluster
        sizes. This change will pick a default value during install based on
        1. If portworx.io/node-type=storage is specified, total number of
           storage nodes will be used to calculate maxStorageNodesPerZone
        2. If label is not specified, total nodes in the cluster will be the
           number of storage nodes
    
           maxStorageNodesPerZone = (total storage nodes)/ (number of zones)

Closes #

Testing: Tests added

### DIFF
--- a/pkg/controller/storagecluster/storagecluster.go
+++ b/pkg/controller/storagecluster/storagecluster.go
@@ -1234,7 +1234,7 @@ func getDefaultStorageNodesDisaggregatedMode(
 		}
 		return 0, false, err
 	}
-	return 0, true, nil
+	return minValue, true, nil
 }
 
 // getDefaultMaxStorageNodesPerZone aims to return a good value for MaxStorageNodesPerZone with the
--- a/pkg/controller/storagecluster/storagecluster.go
+++ b/pkg/controller/storagecluster/storagecluster.go
@@ -21,7 +21,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-
 	"math"
 	"reflect"
 	"sort"
@@ -82,14 +81,6 @@ const (
 	deprecatedCRDBasePath               = "/crds/deprecated"
 	storageClusterCRDFile               = "core_v1_storagecluster_crd.yaml"
 	minSupportedK8sVersion              = "1.12.0"
-	// NodeTypeKey is the key of the label used to set node as storage or storageless
-	NodeTypeKey = "portworx.io/node-type"
-	// StorageNodeValue is the value for storage node
-	StorageNodeValue = "storage"
-	// StoragelessNodeValue is the value for storage node
-	StoragelessNodeValue = "storageless"
-	// StoragePartitioningEnvKey is the storage spec environment variable used to set storage/storageless node type
-	StoragePartitioningEnvKey = "ENABLE_ASG_STORAGE_PARTITIONING"
 )
 
 var _ reconcile.Reconciler = &Controller{}
@@ -1193,7 +1184,7 @@ func getDefaultStorageNodesDisaggregatedMode(
 ) (uint64, bool, error) {
 	// Check if ENABLE_ASG_STORAGE_PARTITIONING is set to false. If not, we can look at the 'portworx.io/node-type' label
 	for _, envVar := range cluster.Spec.Env {
-		if envVar.Name == StoragePartitioningEnvKey {
+		if envVar.Name == util.StoragePartitioningEnvKey {
 			if envVar.Value == "false" {
 				return 0, false, nil
 			}
@@ -1202,21 +1193,21 @@ func getDefaultStorageNodesDisaggregatedMode(
 	}
 
 	// This will return a zone map of storage nodes in each zones
-	nodeTypeZoneMap := getZoneMap(nodeList, NodeTypeKey, StorageNodeValue)
-	if nodeTypeZoneMap == nil {
-		return 0, false, fmt.Errorf("count not find zone information")
+	nodeTypeZoneMap, err := getZoneMap(nodeList, util.NodeTypeKey, util.StorageNodeValue)
+	if err != nil {
+		return 0, false, err
 	}
 	totalNodes := uint64(0)
 	// We'll find the zone with least number of labels
 	minValue := uint64(math.MaxUint64)
 	prevKey := ""
-	prevValue := uint64(0)
+	prevValue := uint64(math.MaxUint64)
 	for key, value := range nodeTypeZoneMap {
 		totalNodes += value
 		if value < minValue {
 			minValue = value
 		}
-		if prevValue != uint64(0) && prevValue != value {
+		if prevValue != math.MaxUint64 && prevValue != value {
 			k8s.InfoEvent(
 				recorder, cluster, util.UnevenStorageNodesReason,
 				fmt.Sprintf("Uneven number of storage nodes labelled across zones."+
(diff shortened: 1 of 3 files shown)


pkg/controller/storagecluster/storagecluster.go
```diff
@@ -1195,12 +1234,24 @@ func (c *Controller) isPxImageBeingUpdated(toUpdate *corev1.StorageCluster) bool
...
 
-func getZoneMap(nodeList *v1.NodeList) map[string]uint64 {
+func getZoneMap(nodeList *v1.NodeList, filterLabelKey string, filterLabelValue string) map[string]uint64 {
 	cloudProviderName := getCloudProviderName(nodeList)
 	cloudProvider := cloudprovider.New(cloudProviderName)
 	zoneMap := map[string]uint64{}
 	for _, node := range nodeList.Items {
 		if zone, err := cloudProvider.GetZone(&node); err == nil {
+			if len(filterLabelKey) > 0 {
+				value, ok := node.Labels[filterLabelKey]
```
piyush-nimbalkar, line 1244: In px we look for `portworx.io/node-type` key suffix and not exact match. I don't remember the reason, but just to be consistent, we should do the same here.

piyush-nimbalkar (approved):
As discussed offline, let's log and raise an info event when there are unequal number of nodes with `node-type: storage` label in all zones.

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -1176,12 +1183,44 @@ func (c *Controller) getCurrentMaxStorageNodesPerZone(
...
 // intention of having at least 3 nodes in the cluster.
-func getDefaultMaxStorageNodesPerZone(nodeList *v1.NodeList) (uint32, error) {
-	zoneMap := getZoneMap(nodeList)
+func getDefaultMaxStorageNodesPerZone(nodeList *v1.NodeList, cluster *corev1.StorageCluster) (uint32, error) {
+	var storageNodes uint64
+	var err error
+
+	// Check if storage nodes are explicitly set using labels
+	storageNodes, err = getDefaultStorageNodesDisaggregatedMode(nodeList, cluster)
+	if err == nil {
```
haibinxie, line 1217: this does not handle unexpected error, we can use a bool to indicate whether it's disaggregated mode or not.

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -1195,12 +1234,24 @@ func (c *Controller) isPxImageBeingUpdated(toUpdate *corev1.StorageCluster) bool
...
 }
 
-func getZoneMap(nodeList *v1.NodeList) map[string]uint64 {
+func getZoneMap(nodeList *v1.NodeList, filterLabelKey string, filterLabelValue string) map[string]uint64 {
 	cloudProviderName := getCloudProviderName(nodeList)
 	cloudProvider := cloudprovider.New(cloudProviderName)
 	zoneMap := map[string]uint64{}
 	for _, node := range nodeList.Items {
 		if zone, err := cloudProvider.GetZone(&node); err == nil {
+			if len(filterLabelKey) > 0 {
```
haibinxie, line 1243: is the error safe to ignore?

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -1176,12 +1183,44 @@ func (c *Controller) getCurrentMaxStorageNodesPerZone(
...
-func getDefaultMaxStorageNodesPerZone(nodeList *v1.NodeList) (uint32, error) {
-	zoneMap := getZoneMap(nodeList)
+func getDefaultMaxStorageNodesPerZone(nodeList *v1.NodeList, cluster *corev1.StorageCluster) (uint32, error) {
+	var storageNodes uint64
+	var err error
+
+	// Check if storage nodes are explicitly set using labels
+	storageNodes, err = getDefaultStorageNodesDisaggregatedMode(nodeList, cluster)
+	if err == nil {
+		return uint32(storageNodes), nil
```
haibinxie, line 1218: if we set this as a fixed number, how can users scale storage nodes in the future?

pkg/controller/storagecluster/storagecluster.go
nrevanna: Thanks for pointing that. I hadn't noticed it.
@adityadani do you know why we have a suffix match and not an exact match?

pkg/controller/storagecluster/storagecluster.go
nrevanna: Currently it's only if the node is nil, which won't be the case. I'll change it to handle the error  to protect against future changes to GetZone.

pkg/controller/storagecluster/storagecluster.go
haibinxie: should we not set default maxStrorageNodesPerZone in disaggregated mode? as the user is managing it by the label anyway, feel like it's the same functionality but it adds more complexity to consider maxStrorageNodesPerZone while setting storage label.

nrevanna: If we don't set it, we'll be in the same situation as before. Like I mentioned before, if a node pool has the storage label and it's upgraded, the new nodes that come in as replacement could possible form a new storage node unintentionally. 
I agree with you that it's an extra knob which affect UX. But I'd argue that without adding this we'll be adding more confusion.
With this change we can make a universal claim that if you expect to increase the storage nodes in a cluster the one thing you have to change is maxStroageNodesPerZone. And then they'll continue with the previous expectation that labels should be correct.
Without this change we will be putting an if else. If you have label and if dont 'have. have labels we have different behaviour.

nrevanna: I'll include these scenarios in the documentation so that we avoid confusion and promote maxStorageNodesPerZone as the main knob.

haibinxie (approved):

pkg/controller/storagecluster/storagecluster.go
adityadani: I also don't remember the reason. In hindsight I think this should have been an exact match.

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -1176,22 +1186,91 @@ func (c *Controller) getCurrentMaxStorageNodesPerZone(
...
+	// We'll find the zone with least number of labels
+	minValue := uint64(math.MaxUint64)
+	prevKey := ""
+	prevValue := uint64(0)
+	for key, value := range nodeTypeZoneMap {
+		totalNodes += value
+		if value < minValue {
+			minValue = value
+		}
+		if prevValue != uint64(0) && prevValue != value {
```
pure-jliao, line 1219: if we have zone map with value e.g. 0,0,1, it will not raise the warning right? so probably init prevValue to maxuint64 as well.

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -1176,22 +1186,91 @@ func (c *Controller) getCurrentMaxStorageNodesPerZone(
...
+			totalNodes += value
+		}
+		if totalNodes > 0 {
+			k8s.InfoEvent(
+				recorder, cluster, util.AllStoragelessNodesReason,
+				fmt.Sprintf("%v nodes marked as storageless, none marked as storage nodes", totalNodes),
+			)
+			return 0, true, fmt.Errorf("storageless nodes found. None marked as storage node")
+		}
+		return minValue, false, nil
```
pure-jliao, line 1242: minValue can be used incorrectly by callers, why not just return 0? otherwise we can call `getZoneMap(nodeList, "", "")` again then calculate the number of storage nodes? that case we don't even need the bool return value in this function and still have a single function `getDefaultMaxStorageNodesPerZone`

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -1176,22 +1186,91 @@ func (c *Controller) getCurrentMaxStorageNodesPerZone(
 	return 0, fmt.Errorf("storage disabled")
 }
 
+func getDefaultStorageNodesDisaggregatedMode(
+	nodeList *v1.NodeList,
+	cluster *corev1.StorageCluster,
+	recorder record.EventRecorder,
+) (uint64, bool, error) {
+	// Check if ENABLE_ASG_STORAGE_PARTITIONING is set to false. If not, we can look at the 'portworx.io/node-type' label
+	for _, envVar := range cluster.Spec.Env {
```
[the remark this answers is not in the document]
nrevanna: Yes. That is like a feature flag that px uses for disaggregated mode. It predates this change.

pkg/controller/storagecluster/storagecluster.go
nrevanna: >> minValue can be used incorrectly by callers, why not just return 0?
I kept minValue since totalNodes is 0, minValue will be zero as well. I made it explicit to be clear. 

For you suggestion on merging with the caller: I feel we should keep the two decisions separate as modules. That will help us reuse this function in an integration test later.

pure-jliao (approved):

pkg/util/util.go
```diff
@@ -41,6 +41,10 @@ const (
 	MigrationCompletedReason = "MigrationCompleted"
 	// MigrationFailed is added to an event when the migration fails.
 	MigrationFailedReason = "MigrationFailed"
+	// UnevenStorageNodesReason is added to an event when there are uneven number of storage nodes are labelled across zones
+	UnevenStorageNodesReason = "UnevenStorageNodes"
+	// AllStoragelessNodesReason is added to an event when all the nodes in the cluster is  labelled as storageless
```
piyush-nimbalkar, line 46: * _are_ labeled

piyush-nimbalkar (approved):

PR #817, opened 2022-11-07
[PWX-27619] Use dry run for eks cloud permission check

pure-jliao:

Closes #

Will change vendor once cloudops PR merged.


pkg/preflight/aws.go
```diff
@@ -64,90 +71,90 @@ func (a *aws) CheckCloudDrivePermission(cluster *corev1.StorageCluster) error {
...
-			logrus.Infof("preflight volume %s deleted", vol)
+		err := a.ops.Delete(vol, nil)
+		if err != nil {
+			logrus.Warnf("preflight failed to delete volume %s: %v", vol, err)
 		}
 	}
 }
+
+func dryRunSucceeded(err error) bool {
+	return err != nil && strings.Contains(err.Error(), dryRunErrMsg)
```
piyush-nimbalkar, line 159: The AWS API returns an error even on success?

pkg/preflight/aws.go
```diff
@@ -64,90 +71,90 @@ func (a *aws) CheckCloudDrivePermission(cluster *corev1.StorageCluster) error {
...
 
-	// Create volume
-	volName := volumePrefix + cluster.Name + "-" + uuid.New()
-	labels := map[string]string{
-		labelVolumeName: volName,
-		labelClusterID:  cluster.Name,
-	}
-	v, err := a.ops.Create(a.getEC2VolumeTemplate(), labels)
+	// List volume on cluster ID first, Describe permission checked here first
+	result, err := a.ops.Enumerate(nil, map[string]string{labelClusterID: cluster.Name}, "")
```
piyush-nimbalkar, line 77: Can we use the storagecluster uuid as well for enumerate?
This is a corner case, but there could be two different px clusters (different k8s clusters) getting deployed at the same time with the same name.

piyush-nimbalkar (approved):
Looks good.

pkg/preflight/aws.go
pure-jliao: yes, the error type will be `DryRunOperation`

pkg/preflight/aws.go
pure-jliao: sure, good point, I'm adding another label, so it's easier to filter on name as well, which is more readable.

pure-jliao: sorry the cluster name is already in the volume name, I'll use cluster UID only.

PR #822, opened 2022-11-10
PWX-27783: Add more zone labels

nrevanna:
In vSphere, we allow customers to tag VMs with zone. Operator needs
   to recognize that and base it's decision of maxStorageNodesPerZone
   based on that. Currently, the change is in the default cloud
   provider. Clouds like GKE, EKS etc don't use that label. Hence we can
   have them also check and that will have no effect. This will solve
   the problem for vSphere. In a future release we can separate out the
   cloud providers.

Closes #

### DIFF
--- a/pkg/cloudprovider/cloudprovider.go
+++ b/pkg/cloudprovider/cloudprovider.go
@@ -10,7 +10,14 @@ import (
 const (
 	failureDomainZoneKey   = v1.LabelTopologyZone
 	failureDomainRegionKey = v1.LabelTopologyRegion
+	pxTopologyLabel        = "topology.portworx.io"
+	pxZoneLabel            = pxTopologyLabel + "/zone"
+	//Deprecated: having this only to support backward compatibility
+	pxZoneLabelDeprecated = "px/zone"
 )
+
+// ZoneLabelsPriority is the priority of zone labels that px uses to determine the zone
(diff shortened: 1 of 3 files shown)


pkg/cloudprovider/cloudprovider.go
```diff
@@ -10,8 +10,15 @@ import (
 const (
 	failureDomainZoneKey   = v1.LabelTopologyZone
 	failureDomainRegionKey = v1.LabelTopologyRegion
```
piyush-nimbalkar, line 12: Can we remove these constants now and just use `v1.LabelTopologyZone` directly in the tests?

pkg/cloudprovider/cloudprovider.go
```diff
@@ -10,8 +10,15 @@ import (
...
+	pxTopologyLabel        = "topology.portworx.io"
+	pxZoneLabel            = pxTopologyLabel + "/zone"
+	//Deprecated: having this only to support backward compatibility
+	pxZoneLabelDeprecated = "px/zone"
 )
 
+// ZoneLabelsPriority is the priority of zone labels that px uses to determine the zone
+var ZoneLabelsPriority = [4]string{pxZoneLabel, pxZoneLabelDeprecated, v1.LabelTopologyZone, v1.LabelZoneFailureDomain}
```
piyush-nimbalkar, line 20: Should we make it private if we are not exposing it outside this package?

piyush-nimbalkar (approved):

PR #827, opened 2022-11-17
PWX-27901: Count accurate number of total nodes

nrevanna:
Master nodes are counted towards total nodes while counting default
   value for maxStorageNodesPerZOne. This fixes that

Closes #

Testing:
Verified that maxStorageNodesPerZone is 5 for the below

> kn get nodes
NAME                          STATUS   ROLES                  AGE   VERSION
nrevanna-vms-luminous-foe-0   Ready    <none>                 33d   v1.21.1
nrevanna-vms-luminous-foe-1   Ready    <none>                 33d   v1.21.1
nrevanna-vms-luminous-foe-2   Ready    <none>                 33d   v1.21.1
nrevanna-vms-luminous-foe-3   Ready    control-plane,master   33d   v1.21.1
nrevanna-vms-luminous-foe-4   Ready    <none>                 33d   v1.21.1
nrevanna-vms-luminous-foe-5   Ready    <none>                 33d   v1.21.1

### DIFF
--- a/pkg/controller/storagecluster/storagecluster.go
+++ b/pkg/controller/storagecluster/storagecluster.go
@@ -1259,7 +1259,18 @@ func getDefaultMaxStorageNodesPerZone(
 		return 0, err
 	}
 	numZones := len(zoneMap)
-	storageNodes = uint64(len(nodeList.Items) / numZones)
+	totalPxNodes := 0
+	for _, node := range nodeList.Items {
+		shouldRun, _, err := k8s.CheckPredicatesForStoragePod(&node, cluster, nil)
+		if err != nil {
+			return 0, err
+		}
+		if shouldRun {
+			totalPxNodes++
+		}
+	}
+
+	storageNodes = uint64(totalPxNodes / numZones)
 	return uint32(storageNodes), nil
 }
 
--- a/pkg/controller/storagecluster/storagecluster.go
+++ b/pkg/controller/storagecluster/storagecluster.go
@@ -1236,7 +1236,7 @@ func getDefaultStorageNodesDisaggregatedMode(
 
 // getDefaultMaxStorageNodesPerZone aims to return a good value for MaxStorageNodesPerZone with the
 // intention of having at least 3 nodes in the cluster.
-func getDefaultMaxStorageNodesPerZone(
+func (c *Controller) getDefaultMaxStorageNodesPerZone(
 	nodeList *v1.NodeList,
 	cluster *corev1.StorageCluster,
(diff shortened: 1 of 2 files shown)


pkg/controller/storagecluster/storagecluster.go
```diff
@@ -1259,7 +1259,18 @@ func getDefaultMaxStorageNodesPerZone(
 		return 0, err
 	}
 	numZones := len(zoneMap)
-	storageNodes = uint64(len(nodeList.Items) / numZones)
+	totalPxNodes := 0
+	for _, node := range nodeList.Items {
+		shouldRun, _, err := k8s.CheckPredicatesForStoragePod(&node, cluster, nil)
```
piyush-nimbalkar, line 1264: Can we add a unit test for case where the node affinity does not match all the nodes and some nodes with taints for which storagecluster does not have tolerations?

piyush-nimbalkar (approved):
Can we create a PR for master as well? I know master has diverged from 1.10.1 a bit wrt this code, but sync up with @pure-jliao if you need context.

pkg/controller/storagecluster/storagecluster.go
haibinxie: shall we call nodeShouldRunStoragePod instead. in addition it looks at disable-storage annotation on storagecluster and if a node is being deleted or not.

haibinxie (approved):

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -1236,30 +1236,40 @@ func getDefaultStorageNodesDisaggregatedMode(
...
 	nodeList *v1.NodeList,
 	cluster *corev1.StorageCluster,
 	recorder record.EventRecorder,
 ) (uint32, error) {
 	if len(nodeList.Items) == 0 {
 		return 0, nil
 	}
+	filteredList := v1.NodeList{}
+	for _, node := range nodeList.Items {
+		shouldRun, _, err := c.nodeShouldRunStoragePod(&node, cluster)
```
pure-jliao, line 1249: are we not considering should continue running here? 
also @haibinxie I moved those functions from controller to portworx API, calling controller util function here might be hard to resolve.. do we have a better way for this? like adding additional conditions after `k8s.CheckPredicatesForStoragePod`? I feel this `nodeShouldRunStoragePod` is too complicated.

nrevanna: shouldContinue will return true for master nodes. So not using it.

PR #828, opened 2022-11-17
[PWX-27886][PWX-27891] DaemonSet migration feature and test improvement

pure-jliao:
* Check external snapshot controller during csi validation
* Add migration node affinity only once
* Add sleep before start migration because node initialization may be in progress 

Closes #

### DIFF
--- a/pkg/util/test/util.go
+++ b/pkg/util/test/util.go
@@ -1952,11 +1952,9 @@ func ValidateCsiEnabled(pxImageList map[string]string, cluster *corev1.StorageCl
 			return nil, true, err
 		}
 
-		// Validate CSI snapshot controller on non-ocp env, since ocp deploys its own snapshot controller
-		if !isOpenshift(cluster) {
-			if err := validateCSISnapshotController(cluster, pxImageList, timeout, interval); err != nil {
-				return nil, true, err
-			}
+		// Validate CSI snapshot controller
+		if err := validateCSISnapshotController(cluster, pxImageList, timeout, interval); err != nil {
+			return nil, true, err
 		}
 		return nil, false, nil
 	}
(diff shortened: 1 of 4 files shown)


Pure-AdamuKaapan (approved):
LGTM! All good fixes

pkg/util/test/util.go
```diff
@@ -2191,6 +2189,26 @@ func validateCSISnapshotController(cluster *corev1.StorageCluster, pxImageList m
...
+		}
+		for _, p := range podList.Items {
+			// ignore pods deployed by operator
+			if label, ok := p.Labels["app"]; ok && label == "px-csi-driver" {
+				continue
+			}
+			for _, c := range p.Spec.Containers {
+				if strings.Contains(c.Image, "/snapshot-controller:") {
+					logrus.Infof("found external snapshot controller in pod %s/%s", p.Namespace, p.Name)
+					installSnapshotController = false
```
haibinxie, line 2206: we should verify if external snapshot controller exists, operator installed snapshot controller should not exist.

test/integration_test/testspec/migration/simple-daemonset.yaml
```diff
@@ -124,13 +124,13 @@ spec:
             httpGet:
               host: 127.0.0.1
               path: /status
-              port: 9001
+              port: 9002
```
haibinxie, line 127: why do we need to change the port?

pkg/util/test/util.go
pure-jliao: yes, the validation is already underneath, line 2228, it would fail if found the snapshot controller deployed in csi deployment.

test/integration_test/testspec/migration/simple-daemonset.yaml
pure-jliao: seems we changed the start port, but forgot to shift the healthy check port, I found this when comparing the pod specs.

haibinxie (approved):

PR #829, opened 2022-11-18
linter-tools cleanup

zoxpx:
* fixing makefile rules for linter-tools  (golangci-lint, errcheck, staticcheck ...)
* removed unused gomock tool

Fixes / cleanup of make-rules for "linter tools":
* will download/install linter-tools only once, if they were missing from `$GOPATH/bin` directory
* switched to "go install" syntax to download/install the linter-tools
* removed unused `gomock` tool

Closes #  (none)

This should speed up the repetitive "make"-jobs, since we'll just use linter-tools from $GOPATH/bin directory instead of downloading them with every make-job.


zoxpx:
This is odd -- the travis-build keep failing on the following:

```
...
# golint check ...
level=error msg="Running error: context loading failed: failed to load packages: timed out to load packages: context deadline exceeded"
level=error msg="Timeout exceeded: try increasing it by passing --timeout option"
make: *** [Makefile:120: lint] Error 4
```

The corresponding (updated) Makefile is the following:
```
118: lint: $(GOPATH)/bin/golangci-lint
119:	# golint check ...
120:	@$(GOPATH)/bin/golangci-lint run ./...
```

So, it seems like `golangci-lint` utility is attempting to download something, and it's timing out?
Note also that the `golangci-lint` utility does not have a `--timeout` option.

I don't have a good explanation for this travis-build failures.

piyush-nimbalkar:
> This is odd -- the travis-build keep failing on the following:
> 
> ```
> ...
> # golint check ...
> level=error msg="Running error: context loading failed: failed to load packages: timed out to load packages: context deadline exceeded"
> level=error msg="Timeout exceeded: try increasing it by passing --timeout option"
> make: *** [Makefile:120: lint] Error 4
> ```
> 
> The corresponding (updated) Makefile is the following:
> 
> ```
> 118: lint: $(GOPATH)/bin/golangci-lint
> 119:	# golint check ...
> 120:	@$(GOPATH)/bin/golangci-lint run ./...
> ```
> 
> So, it seems like `golangci-lint` utility is attempting to download something, and it's timing out? Note also that the `golangci-lint` utility does not have a `--timeout` option.
> 
> I don't have a good explanation for this travis-build failures.

The golangci-lint does not have a timeout, but its `run` command has it.
```
golangci-lint run --timeout=5m
```

pure-jliao:
@zoxpx is there a reason we are updating the makefile here? can we try to run the linter in vendor instead `github.com/golangci/golangci-lint/cmd/golangci-lint`? we hit kind of issue before, a version of lint tool would run super slow and timeout. it's the binary issue. and also with go install, we may want to run go mod vendor to update the vendors as well?

zoxpx:
Thanks @pure-jliao -- I completely missed that `golangci-lint run --timeout=5m` syntax  (fixed in [36124b8](https://github.com/libopenstorage/operator/pull/829/commits/36124b82a865a8abc1a507b05f0fe673f5099d8b))

Regarding the `go install` cleanups -- the Makefile rules w/ this PR should now be uniform for all of them, and work as follows :
1. if linter-tool binary is missing, go install it into the `$GOPATH/bin/` directory
2. run the linter-tool on the sources

This should speed up our test-builds, since we won't compile the `golangci-lint` tool's code every time.

pure-jliao (approved):
LGTM, but @zoxpx could you help to run `go mod vendor` and check there's no update? and `make mockgen` to make sure it still works before merge?

zoxpx:
@pure-jliao -- thanks for the review, but I don't understand what do you want to do with `go mod vendor`, and how is it related to this PR?

As far as I understand, one would not do `go mod vendor` as a part of every build.  Rather, only when one imports or refreshes go-packages  (e.g. `go get -u XXX`), you'd want to follow up with the `go mod tidy; go mod vendor; git add -u ...` commands.
Did you want to run such vendor-update sequence with every build?

pure-jliao:
@zoxpx not with every build, but we saw sometimes changes like this would fail the travis CI built, especially I see go get with `GO111MODULE=off` is removed for receive, not super familiar with go install, so I'm not sure if there will be vendor updated on this. safer side personally would do a vendor update to make sure there's no vendor changed. this PR should be good.

zoxpx:
@pure-jliao -- when building linter-tools  (such as "staticcheck"), I don't think we should rely on the local `./vendor/...` directory.  It's best to use `env GO111MODULE=auto go install <github-ENDPOINT>@<VERSION>` to always get the exact version of the linter-tool (with the tool's dependencies), regardless of the sources in this project.
As far as I understand, this is pretty much what we got with this PR.

zoxpx:
Thanks for the reviews and sorry for the delays -- merging the PR

PR #830, opened 2022-11-19
PWX-26920: add autopilot gitops configuration support

trierra:
to add autopilot gitops configuration support

Closes #PWX-26920

### DIFF
--- a/drivers/storage/portworx/component/autopilot.go
+++ b/drivers/storage/portworx/component/autopilot.go
@@ -184,11 +184,16 @@ func (c *autopilot) createConfigMap(
 		for _, key := range keys {
 			val := cluster.Spec.Autopilot.GitOps.Params[key]
 			if key == "defaultReviewers" {
-				params += `
+				switch val.(type) {
+				default:
+					return fmt.Errorf("invalid spec type for defaultReviewers: %T. Expected array of strings", val)
+				case []string:
+					params += `
     defaultReviewers:`
-				for _, v := range val.([]string) {
-					params += fmt.Sprintf(`
+					for _, v := range val.([]string) {
+						params += fmt.Sprintf(`
       - "%s"`, v)
+					}
 				}
 				continue
 			}
--- a/drivers/storage/portworx/component/autopilot.go
+++ b/drivers/storage/portworx/component/autopilot.go
@@ -171,6 +171,37 @@ func (c *autopilot) createConfigMap(
 			provider.Name, provider.Type, params)
 	}
 
+	if cluster.Spec.Autopilot.GitOps != nil {
+		config += "\ngitops:"
+
+		keys := make([]string, 0, len(cluster.Spec.Autopilot.GitOps.Params))
+		for k := range cluster.Spec.Autopilot.GitOps.Params {
+			keys = append(keys, k)
+		}
+		sort.Strings(keys)
+
+		params := ""
+		for _, key := range keys {
+			val := cluster.Spec.Autopilot.GitOps.Params[key]
+			if key == "defaultReviewers" {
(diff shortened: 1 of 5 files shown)


pkg/apis/core/v1/storagecluster.go
```diff
@@ -527,6 +529,13 @@ type DataProviderSpec struct {
 	Params map[string]string `json:"params,omitempty"`
 }
 
+// GitOpsSpec contains the details for GitOps provider like github or bitbucket
+type GitOpsSpec struct {
+	Name   string                 `json:"name,omitempty"`
+	Type   string                 `json:"type"`
+	Params map[string]interface{} `json:"params"`
```
piyush-nimbalkar, line 536: nit: Let's add `omitempty`

drivers/storage/portworx/component/autopilot.go
```diff
@@ -171,6 +171,37 @@ func (c *autopilot) createConfigMap(
...
+		}
+		sort.Strings(keys)
+
+		params := ""
+		for _, key := range keys {
+			val := cluster.Spec.Autopilot.GitOps.Params[key]
+			if key == "defaultReviewers" {
+				params += `
+    defaultReviewers:`
+				for _, v := range val.([]string) {
```
piyush-nimbalkar, line 189: This will panic if the value is not an array of strings. We should check and fail properly with an error. Add a test for that.

pkg/apis/core/v1/storagecluster.go
trierra: I skipped that intentionally. Autopilot can't instantiate a gitops provider w/o params provided

drivers/storage/portworx/component/autopilot.go
haibinxie: as it's not strong typed, you may want to use comma separated string.

drivers/storage/portworx/component/autopilot.go
```diff
@@ -171,6 +171,37 @@ func (c *autopilot) createConfigMap(
...
+		keys := make([]string, 0, len(cluster.Spec.Autopilot.GitOps.Params))
+		for k := range cluster.Spec.Autopilot.GitOps.Params {
+			keys = append(keys, k)
+		}
+		sort.Strings(keys)
+
+		params := ""
+		for _, key := range keys {
+			val := cluster.Spec.Autopilot.GitOps.Params[key]
+			if key == "defaultReviewers" {
```
haibinxie, line 186: nit: define defaultReviewers as constant

haibinxie (approved):

pure-jliao:
I think there's one missing piece, run `make codegen` to generate the deep equal functions. @trierra

trierra:
thanks, will do @pure-jliao

pkg/apis/core/v1/storagecluster.go
piyush-nimbalkar: The omit empty will just not show the params field if not specified. Dropping it won't help autopilot anyway. It is not going to force the user to specify params here.

drivers/storage/portworx/component/autopilot.go
```diff
@@ -171,6 +172,42 @@ func (c *autopilot) createConfigMap(
...
+    %s:`, AutopilotDefaultReviewersKey)
+					for _, v := range val {
+						params += fmt.Sprintf(`
+      - "%s"`, v)
+					}
+				}
+				continue
+			}
+			params += fmt.Sprintf(`
+    %s: %s`, key, val.(string))
```
piyush-nimbalkar, line 202: Again, not a good idea to use `val.(string)`. This code will panic of the value is anything other than string.

PR #834, opened 2022-11-22
PWX-26656 add affinity to run pvc controller on storage nodes

haibinxie:
add pod affinity, so pvc controller pod will run on nodes with storage pod.

### DIFF
--- a/drivers/storage/portworx/component/pvccontroller.go
+++ b/drivers/storage/portworx/component/pvccontroller.go
@@ -366,6 +366,9 @@ func (c *pvcController) createDeployment(
 		return err
 	}
 
+	deployment := c.getPVCControllerDeploymentSpec(cluster, ownerRef, imageName, command, targetCPUQuantity,
+		updatedTopologySpreadConstraints)
+
 	modified := existingImage != imageName ||
 		!reflect.DeepEqual(existingCommand, command) ||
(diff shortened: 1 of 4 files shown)


drivers/storage/portworx/component/pvccontroller.go
```diff
@@ -476,6 +478,24 @@ func (c *pvcController) getPVCControllerDeploymentSpec(
...
 					},
 					Affinity: &v1.Affinity{
+						PodAffinity: &v1.PodAffinity{
+							RequiredDuringSchedulingIgnoredDuringExecution: []v1.PodAffinityTerm{
+								{
+									TopologyKey: "kubernetes.io/hostname",
+									LabelSelector: &metav1.LabelSelector{
+										MatchExpressions: []metav1.LabelSelectorRequirement{
+											{
+												Key:      "storage",
```
piyush-nimbalkar, line 488: Let's use the key and value constants from https://github.com/libopenstorage/operator/blob/master/pkg/constants/metadata.go#L11 to be consistent.

piyush-nimbalkar (approved):

pure-jliao (approved):
LGTM. Better to add the validation in automation as well.

piyush-nimbalkar:
@haibinxie let's wait before we merge this. The ideal fix should be in Portworx. If we do that, we don't need this fix.

PR #835, opened 2022-11-23
PWX-25288-pt2: Adding support for 'auto-tls'

zoxpx:
* adding new 'is-auto-tls' annotation
* this annotation initiates automatic SSL setup for porx
* gRPC-dialing now automatically adds K8s CA, since porx APIs will also switch to SSL-mode
* modified access to `portworx-service` to use K8s DNS, rather than ClusterIP

Additional:
* added support for `--trace` logging option

This is adding the support, and setup for "automatic SSL" for Portworx

Closes #  PWX-25288   (part 2)

This change depends on https://github.com/portworx/px-installer/pull/1497 for OCI-Monitor  (<< linked to ClusterIP change)
Unit-tests pending

### DIFF
--- a/cmd/operator/operator.go
+++ b/cmd/operator/operator.go
@@ -65,11 +65,6 @@ func main() {
 
 	app.Flags = []cli.Flag{
 		cli.BoolFlag{
-			Name:   "trace",
-			Usage:  "Enable trace logging",
-			Hidden: true,
-		},
-		cli.BoolFlag{
 			Name:  "verbose",
 			Usage: "Enable verbose logging",
 		},
@@ -131,12 +126,10 @@ func run(c *cli.Context) {
 		log.Fatalf("driver option is required")
 	}
 
-	if c.Bool("trace") {
-		log.SetLevel(log.TraceLevel)
-	} else if c.Bool("verbose") {
+	verbose := c.Bool("verbose")
+	if verbose {
 		log.SetLevel(log.DebugLevel)
 	} else {
-		// set INFO logs as default
 		log.SetLevel(log.InfoLevel)
 	}
 
--- a/cmd/operator/operator.go
+++ b/cmd/operator/operator.go
@@ -65,6 +65,11 @@ func main() {
 
 	app.Flags = []cli.Flag{
 		cli.BoolFlag{
+			Name:   "trace",
+			Usage:  "Enable trace logging",
+			Hidden: true,
+		},
+		cli.BoolFlag{
 			Name:  "verbose",
 			Usage: "Enable verbose logging",
 		},
@@ -126,10 +131,12 @@ func run(c *cli.Context) {
 		log.Fatalf("driver option is required")
 	}
 
-	verbose := c.Bool("verbose")
-	if verbose {
+	if c.Bool("trace") {
+		log.SetLevel(log.TraceLevel)
+	} else if c.Bool("verbose") {
 		log.SetLevel(log.DebugLevel)
 	} else {
+		// set INFO logs as default
 		log.SetLevel(log.InfoLevel)
 	}
 
--- a/drivers/storage/portworx/deployment.go
+++ b/drivers/storage/portworx/deployment.go
@@ -979,7 +979,7 @@ func (t *template) getArguments() []string {
 	}
 
 	if pxutil.IsTLSEnabledOnCluster(&t.cluster.Spec) {
-		logrus.Tracef("TLS is enabled! Getting oci-monitor arugments")
+		logrus.Tracef("TLS is enabled! Getting oci-monitor arguments")
 		if tlsArgs, err := pxutil.GetOciMonArgumentsForTLS(t.cluster); err == nil {
 			logrus.Tracef("oci-monitor arguments for TLS: %v\n", tlsArgs)
 			args = append(args, tlsArgs...)
@@ -989,7 +989,7 @@ func (t *template) getArguments() []string {
 	}
 
 	if t.autoTLSEnabled {
-		logrus.Tracef("AutoTLS is enabled! Setting oci-monitor arugments")
+		logrus.Tracef("AutoTLS is enabled! Setting oci-monitor arguments")
 		args = append(args, "--auto-tls")
 	}
 
--- a/drivers/storage/portworx/deployment.go
+++ b/drivers/storage/portworx/deployment.go
@@ -79,7 +79,6 @@ type template struct {
 	isIKS           bool
 	isOpenshift     bool
 	isK3s           bool
-	autoTLSEnabled  bool
 	runOnMaster     bool
 	imagePullPolicy v1.PullPolicy
 	serviceType     v1.ServiceType
@@ -131,7 +130,6 @@ func newTemplate(
 	t.serviceType = pxutil.ServiceType(cluster, "")
 	t.imagePullPolicy = pxutil.ImagePullPolicy(cluster)
 	t.startPort = pxutil.StartPort(cluster)
-	t.autoTLSEnabled = pxutil.IsAutoTLSEnabled(cluster)
 
 	return t, nil
 }
@@ -988,11 +986,6 @@ func (t *template) getArguments() []string {
 		}
 	}
 
-	if t.autoTLSEnabled {
-		logrus.Tracef("AutoTLS is enabled! Setting oci-monitor arguments")
-		args = append(args, "--auto-tls")
-	}
-
 	if pxutil.EssentialsEnabled() {
 		for args[len(args)-1] == "--oem" {
 			args = args[:len(args)-1]
@@ -1126,7 +1119,7 @@ func (t *template) getEnvList() []v1.EnvVar {
 	}
 
 	// We're setting this to signal to porx that tls is enabled
-	if pxutil.IsTLSEnabledOnCluster(&t.cluster.Spec) || t.autoTLSEnabled {
+	if pxutil.IsTLSEnabledOnCluster(&t.cluster.Spec) {
 		// Set:
 		//    env:
 		//    - name: PX_ENABLE_TLS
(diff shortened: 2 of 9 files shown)


zoxpx:
FYI -- the unit-test code is still pending!
I had issues running `make test` on my workstation (was hitting timeots) -- so please be aware that I might need to mis-use Travis to run UTs  (please expect some errors).

drivers/storage/portworx/util/util.go
```diff
@@ -253,6 +255,8 @@ var (
 	MinimumPxVersionCCMGO, _ = version.NewVersion("2.12")
 	// MinimumPxVersionMetricsCollector minimum PX version to install metrics collector
 	MinimumPxVersionMetricsCollector, _ = version.NewVersion("2.9.1")
+	// MinAutoTLSVersion is a minimal PX version that supports "auto-TLS" setup
+	MinAutoTLSVersion, _ = version.NewVersion("3.0.0")
```
pure-jliao, line 259: nit: use `MinimumPxVersionsAutoTLS` to be consistent with other const.

drivers/storage/portworx/util/util.go
```diff
@@ -132,6 +132,8 @@ const (
 	AnnotationClusterID = pxAnnotationPrefix + "/cluster-id"
 	// AnnotationPreflightCheck do preflight check before installing Portworx
 	AnnotationPreflightCheck = pxAnnotationPrefix + "/preflight-check"
+	// AnnotationIsAutoTLS annotation whether to set up cluster with automatic SSL/TLS credentials
+	AnnotationIsAutoTLS = pxAnnotationPrefix + "/is-auto-tls"
```
pure-jliao, line 136: nit: we have is-eks, is-iks flags, I feel probably it's better to simply use `auto-tls` instead of `is-auto-tls`? it's a feature to enable or disable.

drivers/storage/portworx/util/util.go
```diff
@@ -326,6 +330,22 @@ func IsHostPidEnabled(cluster *corev1.StorageCluster) bool {
 	return err == nil && enabled
 }
 
+// IsAutoTLS returns true if the AutoTLS-annotation is true value, and cluster version qualifies
+func IsAutoTLS(cluster *corev1.StorageCluster) bool {
```
pure-jliao, line 334: nit: `IsAutoTLSEnabled`

drivers/storage/portworx/util/util.go
```diff
@@ -326,6 +330,22 @@ func IsHostPidEnabled(cluster *corev1.StorageCluster) bool {
...
+	enabled, err := strconv.ParseBool(cluster.Annotations[AnnotationIsAutoTLS])
+	if err != nil || !enabled {
+		logrus.WithError(err).Tracef("> IsAutoTLS() early return FALSE")
+		return false
+	}
+
+	if cluster.Spec.Version == "latest" {
```
pure-jliao, line 341: I don't think we have a version set as latest. you can use `pxutil.GetPortworxVersion(cluster)`. if it's a staging tag, it should return a super large version

drivers/storage/portworx/util/util.go
```diff
@@ -701,7 +721,8 @@ func GetPortworxConn(sdkConn *grpc.ClientConn, k8sClient client.Client, namespac
 		return nil, fmt.Errorf("failed to get endpoint for portworx volume driver")
 	}
 
-	endpoint := pxService.Spec.ClusterIP
+	// note, using symbolic name for the `endpoint`, as SSL certificates won't have K8s service IP
+	endpoint := PortworxServiceName + "." + namespace + ".svc.cluster.local"
```
pure-jliao, line 725: I guess probably this is why the unit test is timing out, this messed up the mock calls there. I feel it's kind of risky to make this change here without test coverage, and if we can make it backward compatible.

drivers/storage/portworx/deployment.go
```diff
@@ -79,6 +79,7 @@ type template struct {
 	isIKS           bool
 	isOpenshift     bool
 	isK3s           bool
+	isAutoTLS       bool
```
pure-jliao, line 82: similar here, autoTLSEnabled might be better.

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -1261,6 +1262,15 @@ func (c *Controller) setStorageClusterDefaults(cluster *corev1.StorageCluster) e
 		return err
 	}
 
+	// set / reset TLS env-vars depending on the updated cluster settings
```
pure-jliao, line 1265: can we move this to `setSecuritySpecDefaults` in `SetPortworxDefaults`?

pure-jliao (commented):
another question is why do we add an annotation instead of adding a spec under the Security section? annotations are more like for internal usage and temporary workarounds.

zoxpx:
> another question is why do we add an annotation instead of adding a spec under the Security section? annotations are more like for internal usage and temporary workarounds.

@pure-jliao -- what did you have in mind?

AFAIK, we modeled the `securityContext:` YAML segment after Kubernetes, which is good because we don't have to teach our customers about it  (i.e. they'll already know _"what it means"_ and _"how to use it"_ based on K8s docs).
Did you want to add a new custom-setting there?  I would be concerned that this would be more confusing than useful.

Also note, I would expect that deploying with "automatic SSL certs" will become a new default once we stabilize and test the whole integration (e.g. 6 months or a year from now?) -- so eventually we'll need to add "the opposite switches", that'll turn off (or modify) the default behavior.

cmd/operator/operator.go
```diff
@@ -64,6 +64,11 @@ func main() {
 	app.Action = run
 
 	app.Flags = []cli.Flag{
+		cli.BoolFlag{
```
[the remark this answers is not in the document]
zoxpx: Note that we already use `--trace` (as well as `--debug`) on other products  (e.g. OCI-Monitor supports both flags).  Additionally, there were `logrus.Trace/Tracef` calls already in the code, so you won't have a way of getting these "lowest-level detailed" logs out.

If you want to "unify it", then you'd normally introduce something like `--logLevel <trace|debug|info|warn|error|fatal>` option, which we don't normally use anywhere else.

So, do you want me to remove this for the time being?

pure-jliao:
@zoxpx personally I feel we might want to have a storage cluster spec like this, we already have the tls spec section, why don't we expand that like
```
tls:
    enabled: true (or Auto)
    autoTLS: true
```
to make it as a new default, we can still enable it in the future, and specify the field explicitly for now to enable it.
But if we want to force enable it later and hide the switch not to confuse people, we can use annotation for now and delete it entirely later. My point is operator uses storage cluster spec to reconcile the object, it's better not putting all settings in the annotation, unless you feel it's not necessary to change the spec schema.

cmd/operator/operator.go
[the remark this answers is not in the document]
pure-jliao: operator too verbose is a known issue currently, and we have plan to update the log levels and offer one single parameter to control it. I see we do have 3 lines of trace level log in the code, but we're not printing those anyways. and as you suggested, I think `--logLevel <trace|debug|info|warn|error|fatal>` is the right approach, and needs to be done in a separate ticket. for now I believe using a proper debug level log should be good enough. wdyt @haibinxie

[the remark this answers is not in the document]
zoxpx: re _"we're not printing those anyways"_ -- you'd normally not display `--debug/--trace` log-messages, unless you are troubleshooting, or need more details in the logs (e.g. ask customer/QA to increase log-level, retest, and get you the logs).

For example, the `--trace` level on OCI-Monitor will start dumping all input/output data-structures it got from APIs.  This level is not suitable for "every day logs", but it is handy to troubleshoot at customer's sites without code-modifications  (e.g. when suspecting APIs/infrastructure is buggy/misconfigured).

drivers/storage/portworx/util/util.go
zoxpx: Thanks for the heads-up -- will take a look

cmd/operator/operator.go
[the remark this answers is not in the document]
piyush-nimbalkar: @zoxpx let us remove this option for now. We are going to add a `--log-level` option very soon where users should be able to set the value to different levels. We will deprecate the `--verbose` option then.

Unfortunately, you won't be able to set the value to `trace` during runtime until then.

drivers/storage/portworx/deployment.go
```diff
@@ -986,6 +988,11 @@ func (t *template) getArguments() []string {
 		}
 	}
 
+	if t.autoTLSEnabled {
+		logrus.Tracef("AutoTLS is enabled! Setting oci-monitor arugments")
```
piyush-nimbalkar, line 992: nit: arguments

drivers/storage/portworx/util/util.go
```diff
@@ -326,6 +330,17 @@ func IsHostPidEnabled(cluster *corev1.StorageCluster) bool {
 	return err == nil && enabled
 }
 
+// IsAutoTLSEnabled returns true if the AutoTLS-annotation is true value, and cluster version qualifies
+func IsAutoTLSEnabled(cluster *corev1.StorageCluster) bool {
+	enabled, err := strconv.ParseBool(cluster.Annotations[AnnotationAutoTLS])
```
piyush-nimbalkar, line 335: @zoxpx what do we want the default behavior to be if the annotation is not specified?
If this option is not specified, what would the non-auto way for the user to setup TLS?

zoxpx: I'll try to re-work the "trigger" -- I was thinking, if someone requests SSL/TLS, but they _do not_ supply their own certificates -- we'd set up with "automatic SSL certs".

I think the current `tls` -block in StorageCluster YAML will allow this.

zoxpx:
FYI -- with [93585f1](https://github.com/libopenstorage/operator/pull/835/commits/93585f1f8a7d35aab6e63865917c473d77d72889) we changed the trigger-point for the "auto-ssl":
* no longer using `portworx.io/is-auto-tls: "true"` annotation
* using this instead:
```
spec:
  security:
    auth:
      enabled: false
    enabled: true
    tls:
      enabled: true
```
   - note, this is "enabled security and TLS", but no SSL/CA certificates were provided -- thus, we'll set up certs automatically

piyush-nimbalkar (approved):

zoxpx:
Thanks for the review -- merging the PR

PR #838, opened 2022-11-28
PWX-27864: fix operator migration functional test on k8s 1.25

haibinxie:
Validated: https://jenkins.pwx.dev.purestorage.com/view/Newstack%20Dashboard/job/Operator/job/daemonset-migration-master/99/parameters/


pure-jliao:
deleted branch recovered, so not needed

haibinxie:
@pure-jliao this PR is to enable operator migration on k8s 1.25, i reopened it:)

nikolaypopov:
@haibinxie Are you sure that just ignoring k8s version will work? I am pretty sure we had that there for a reason, @pure-jliao can you confirm? We should build a private image from this PR and run the migration suite with it before merging, I am suspecting there would need to be some changes for the hardcoded daemonset specs to make it run on v1.25, as you can see this spec were generated for v1.21 https://github.com/libopenstorage/operator/blob/master/test/integration_test/testspec/migration/daemonset-with-all-components.yaml#L1 and this one doesn't even have k8s version https://github.com/libopenstorage/operator/blob/master/test/integration_test/testspec/migration/simple-daemonset.yaml#L1, this might cause some issue I believe

haibinxie:
> @haibinxie Are you sure that just ignoring k8s version will work? I am pretty sure we had that there for a reason, @pure-jliao can you confirm? We should build a private image from this PR and run the migration suite with it before merging, I am suspecting there would need to be some changes for the hardcoded daemonset specs to make it run on v1.25, as you can see this spec were generated for v1.21 https://github.com/libopenstorage/operator/blob/master/test/integration_test/testspec/migration/daemonset-with-all-components.yaml#L1 and this one doesn't even have k8s version https://github.com/libopenstorage/operator/blob/master/test/integration_test/testspec/migration/simple-daemonset.yaml#L1, this might cause some issue I believe

@nikolaypopov this fix was validated at https://jenkins.pwx.dev.purestorage.com/view/Newstack%20Dashboard/job/Operator/job/daemonset-migration-master/99/parameters/
this PR also changed prometheus spec, which was the main reason of the failure.

pure-jliao (approved):
If it's passing on master then it should be good.

PR #841, opened 2022-11-30
PWX-26920: changed parameters type & codegen fix

trierra:

Closes #


drivers/storage/portworx/component/autopilot.go
```diff
@@ -185,27 +185,25 @@ func (c *autopilot) createConfigMap(
...
 			params += fmt.Sprintf(`
-    %s: %s`, key, val.(string))
+    %s: %s`, key, val)
 		}
 
 		config += fmt.Sprintf(`
   name: %s
   type: %s
   params:%s`, cluster.Spec.Autopilot.GitOps.Name, cluster.Spec.Autopilot.GitOps.Type, params)
+		fmt.Println(config)
```
piyush-nimbalkar, line 206: nit: Remove print.

piyush-nimbalkar (approved):

pkg/apis/core/v1/storagecluster.go
```diff
@@ -531,9 +531,9 @@ type DataProviderSpec struct {
 
 // GitOpsSpec contains the details for GitOps provider like github or bitbucket
 type GitOpsSpec struct {
-	Name   string                 `json:"name,omitempty"`
-	Type   string                 `json:"type"`
-	Params map[string]interface{} `json:"params"`
+	Name   string            `json:"name,omitempty"`
+	Type   string            `json:"type"`
+	Params map[string]string `json:"params"`
```
pure-jliao, line 536: add omitempty as discussed? do we need omitempty for type as well? cc: @piyush-nimbalkar

drivers/storage/portworx/component/autopilot.go
trierra: oh, sorry

pkg/apis/core/v1/storagecluster.go
piyush-nimbalkar: Oh yes. Let's add omitempty for both `Type` and `Params`.
Without it the StorageCluster spec looks unnecessarily long.

trierra: we don't want the omitemty as this config is not usable w/o params. The whole point of the `omitempty` as I understand is allowing to proceed w/o this field. We can't let user create gitops type and omit params

pure-jliao (approved):

pkg/apis/core/v1/storagecluster.go
piyush-nimbalkar: That's not the purpose of `omitempty`. It only tells the yaml parser to omit the field if it is set to a default value based on the type of the field. For instance, it will not show the fields in the yaml for an empty map, empty string, false bool, 0 int, etc
It is not same as 'required', where it will stop the user from proceeding w/o this field.

drivers/storage/portworx/components_test.go
```diff
@@ -3189,6 +3189,81 @@ func TestAutopilotInstall(t *testing.T) {
 	require.Equal(t, expectedDeployment, autopilotDeployment)
 }
 
+func TestAutopilotInstallIncorrectSpec(t *testing.T) {
+	// testing the case, where the `defaultReviewers` has been set as []interface{} instead of []string
```
pure-jliao, line 3193: nit: update the comment.

pkg/apis/core/v1/storagecluster.go
```diff
@@ -533,7 +533,7 @@ type DataProviderSpec struct {
 type GitOpsSpec struct {
 	Name   string            `json:"name,omitempty"`
 	Type   string            `json:"type"`
```
pure-jliao, line 535: type in this case still can be empty string. do we add omitempty and validation as well?

pure-jliao (approved):

PR #843, opened 2022-12-01
PWX-28023 update prometheus CRD

haibinxie:
Prometheus operator image version is half year newer than prometheus CRD version. Update prometheus CRD version to the latest version, suppose prometheus CRD is backward compatible we shall always keep the CRD up to date.


pure-jliao (approved):
did we check if this fixes the automation?

haibinxie:
@pure-jliao no that's different issue, i will send another PR.

haibinxie:
@piyush-nimbalkar actually there is another bug, we would not update prometheus CRD once it's created, I am going to make a fix to that too. 

https://github.com/libopenstorage/operator/blob/master/drivers/storage/portworx/component/prometheus.go#L204

piyush-nimbalkar:
> @piyush-nimbalkar actually there is another bug, we would not update prometheus CRD once it's created, I am going to make a fix to that too.
> 
> https://github.com/libopenstorage/operator/blob/master/drivers/storage/portworx/component/prometheus.go#L204

We did not add update there because the prometheus could have been deployed by the customer. If we update their CRD to an incompatible version, it could cause issues in their setup.
We should update the CRD only if we have created it. May be add a label to the CRD when we create it? Next time when we add update logic, we do it only if we see the label. Wdyt?

piyush-nimbalkar (approved):

haibinxie:
@piyush-nimbalkar we want to update CRD on existing cluster deployed by operator which does not have the label hence we wont be able to distinguish who deployed it.

piyush-nimbalkar:
> there because the prometheus could have been deploy

I understand that, but not sure what we should be doing here. We risk affecting customer deployments especially on envs like rancher where prometheus is already installed.

haibinxie:
> > there because the prometheus could have been deploy
> 
> I understand that, but not sure what we should be doing here. We risk affecting customer deployments especially on envs like rancher where prometheus is already installed.

if we keep our CRD up to date it should be ok, as CRD should be backward compatible. old prometheus operator image may not pass vulnerability scan, we just need to keep CRD updated when updating prometheus operator image.

PR #844, opened 2022-12-02
PTX-14313 Pin AlertManager and Prometheus images

nikolaypopov:
Pin AlertManager and Prometheus images


pure-jliao (approved):
LGTM! thanks for the fix!

PR #846, opened 2022-12-02
PWX-26920 fix prometheus operator test automation failure

haibinxie:


deploy/crds/core_v1_storagecluster_crd.yaml
```diff
@@ -938,21 +938,131 @@ spec:
                               to an implementation-defined value. More info: https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/'
                             type: object
                       securityContext:
-                        type: object
```
pure-jliao, line 941: is the order formatted by IDE? it takes me a minute to find the type is moved to the bottom.

pure-jliao (approved):
LGTM. are those crd schema changes all necessary? if yes then should be good. also please verify the functional test can pass.

deploy/crds/core_v1_storagecluster_crd.yaml
haibinxie: I copied the entire securityContext section from prometheus CRD:
1. it's pass-through, easier for us to understand if it's entirely passthough
2. in case if there is new permission issue we can resolve on customer env

PR #850, opened 2022-12-05
PWX-25288-pt2: Unit-tests fix

zoxpx:
* will update /etc/hosts "in-place"

Original unit-tests were updating `/etc/hosts` file by using a different temporary file, and then renaming it as `/etc/hosts`

ISSUE:
* renaming `/etc/hosts` file does not work inside containers, since containers "lock" this file and will not allow renaming/moving it -- thus, the unit-tests would fail when ran in "containerized" environments
  - they do, however, allow appending/overwriting the file

As a fix, we're updating `/etc/hosts` file "in place" by overwriting it with the test-content

Closes # PWX-25288

### DIFF
--- a/drivers/storage/portworx/components_test.go
+++ b/drivers/storage/portworx/components_test.go
@@ -1,13 +1,13 @@
 package portworx
 
 import (
+	"bufio"
 	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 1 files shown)


mborodin-px (approved):
LGTM

zoxpx:
Thanks Max -- merging the PR

PR #858, opened 2022-12-08
Bump all CSI images to the latest and remove prov fork for PX 2.13+

ggriffiths:
Bumps all CSI images and removes prov fork for PX 2.13.0+

Closes #


test/integration_test/testspec/migration/daemonset-with-all-components.yaml
```diff
@@ -132,7 +132,7 @@ spec:
             - name: dbusmount
               mountPath: /var/run/dbus
         - name: csi-node-driver-registrar
-          image: k8s.gcr.io/sig-storage/csi-node-driver-registrar:v2.5.1
```
pure-jliao, line 135: fyi we don't need to update images here, they will be updated according to the release manifest during automation runs.

pure-jliao (approved):
LGTM, do we have the spec generator change as well?

haibinxie (approved):

test/integration_test/testspec/migration/daemonset-with-all-components.yaml
ggriffiths: Yes I'm aware, It's better to keep these up to date in case the release manifest cannot be reached or we switch away from that model.

PR is here-
https://github.com/portworx/px-installer/pull/1512

PR #865, opened 2022-12-16
[PWX-26914][PWX-27865][PWX-27923] add resources limit/requests to oci-mon and runc args

haibinxie:

Closes #

### DIFF
--- a/drivers/storage/portworx/deployment.go
+++ b/drivers/storage/portworx/deployment.go
@@ -802,6 +802,24 @@ func (t *template) getArguments() []string {
 		"-x", "kubernetes",
 	}
 
+	if t.cluster.Spec.Resources != nil && t.cluster.Spec.Resources.Limits != nil {
+		if memory, ok := t.cluster.Spec.Resources.Limits[v1.ResourceMemory]; ok {
+			bytes, b := memory.AsInt64()
+			if b {
+				args = append(args, "--memory")
+				args = append(args, fmt.Sprint(bytes))
+			} else {
+				logrus.Errorf("failed to parse memory %+v", memory)
+			}
+		}
+
+		if cpu, ok := t.cluster.Spec.Resources.Limits[v1.ResourceCPU]; ok {
+			f := cpu.AsApproximateFloat64()
+			args = append(args, "--cpus")
+			args = append(args, fmt.Sprintf("%f", f))
+		}
+	}
+
 	if t.cluster.Spec.Kvdb != nil {
 		if t.cluster.Spec.Kvdb.Internal {
 			args = append(args, "-b")
--- a/drivers/storage/portworx/deployment_test.go
+++ b/drivers/storage/portworx/deployment_test.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"os"
 	"path"
+	"strings"
 	"testing"
 
 	"github.com/golang/mock/gomock"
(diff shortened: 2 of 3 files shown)


drivers/storage/portworx/deployment.go
```diff
@@ -802,6 +802,24 @@ func (t *template) getArguments() []string {
...
+			if b {
+				args = append(args, "--memory")
+				args = append(args, fmt.Sprint(bytes))
+			} else {
+				logrus.Errorf("failed to parse memory %+v", memory)
+			}
+		}
+
+		if cpu, ok := t.cluster.Spec.Resources.Limits[v1.ResourceCPU]; ok {
+			f := cpu.AsApproximateFloat64()
```
Pure-AdamuKaapan, line 817: Does `AsApproximateFloat64` not return an "ok" value like the `AsInt64()` method?

drivers/storage/portworx/deployment_test.go
```diff
@@ -242,6 +243,43 @@ func TestPodSpecWithPortworxContainerResources(t *testing.T) {
...
+			v1.ResourceCPU:    resource.MustParse("8"),
+		},
+	}
+
+	podSpec, err = driver.GetStoragePodSpec(cluster, nodeName)
+	assert.NoError(t, err, "Unexpected error on GetStoragePodSpec")
+	assert.Len(t, podSpec.Containers, 1)
+	assert.Equal(t, *cluster.Spec.Resources, podSpec.Containers[0].Resources)
+	require.True(t, strings.Contains(strings.Join(podSpec.Containers[0].Args, " "), "--cpus 8"))
+	require.True(t, strings.Contains(strings.Join(podSpec.Containers[0].Args, " "), "--memory 8589934592"))
```
Pure-AdamuKaapan, line 282: What happens if we provide an invalid memory value (like some random text)? That's my concern with the parsing error earlier

Pure-AdamuKaapan (commented):
Overall LGTM, just have one error handling question

drivers/storage/portworx/deployment.go
haibinxie: Yes

drivers/storage/portworx/deployment_test.go
haibinxie: for memory it wont be set, for cpu actually I have not tried, i can add a unit test to it.

Pure-AdamuKaapan: Would be good to do yeah

haibinxie: actually in unit test you can never parse invalid value, as the quantity is an object. in real cluster, you could edit storageCluster to use invalid memory, however operator will never receive the updated stc due to it's invalid.

E1217 21:07:06.547436       1 reflector.go:140] sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:262: Failed to watch *v1.StorageCluster: failed to list *v1.StorageCluster: unable to parse quantity's suffix

Pure-AdamuKaapan: Ahhhh that makes sense: yeah, then I'm fine without testing that obviously.

Pure-AdamuKaapan (approved):
LGTM!

piyush-nimbalkar (approved):

PR #872, opened 2022-12-22
PWX-26156: enable csi by default on k8s 1.26

haibinxie:

Closes #

### DIFF
--- a/drivers/storage/portworx/portworx.go
+++ b/drivers/storage/portworx/portworx.go
@@ -349,7 +349,9 @@ func (p *portworx) SetDefaultsOnStorageCluster(toUpdate *corev1.StorageCluster) 
 			}
 		}
 
-		SetPortworxDefaults(toUpdate, p.k8sVersion)
+		if err := SetPortworxDefaults(toUpdate, p.k8sVersion); err != nil {
+			return err
+		}
 	}
 
 	removeDeprecatedFields(toUpdate)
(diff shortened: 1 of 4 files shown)


pkg/controller/storagecluster/storagecluster.go
```diff
@@ -1208,6 +1208,24 @@ func (c *Controller) CreatePodTemplate(
 func (c *Controller) setStorageClusterDefaults(cluster *corev1.StorageCluster) error {
 	toUpdate := cluster.DeepCopy()
 
+	if cluster.Spec.CSI == nil {
```
pure-jliao, line 1211: this should be changed in this function instead https://github.com/libopenstorage/operator/blob/312fa642d31e8c0fac62897d3849eef9fe0d89b6/drivers/storage/portworx/portworx.go#L1066
and unit test `SetDefaultsOnStorageCluster` in `component_test.go`

drivers/storage/portworx/portworx.go
```diff
@@ -847,13 +849,13 @@ func (p *portworx) storageNodeToCloudSpec(storageNodes []*corev1.StorageNode, cl
 }
 
 // SetPortworxDefaults populates default storage cluster spec values
-func SetPortworxDefaults(toUpdate *corev1.StorageCluster, k8sVersion *version.Version) {
+func SetPortworxDefaults(toUpdate *corev1.StorageCluster, k8sVersion *version.Version) error {
```
pure-jliao, line 852: nit: currently we only init p.k8sVersion once right? I don't know if we can get more benefit refreshing p.k8sVersion every reconcile loop in `SetDefaultsOnStorageCluster`, then we can pass the correct k8s version in and don't have to change this signature here. But this change looks good to me.

pure-jliao (approved):

drivers/storage/portworx/portworx.go
haibinxie: right now Init is only once, once csi is set to non-nil, there is no need to refresh k8s version anymore.

PR #873, opened 2022-12-22
[PWX-27765] Fix stc object and migration condition update issues during migration

pure-jliao:
* During migration always get live stc when updating
* Mark migration as completed in controller reconcile loop

Closes #

### DIFF
--- a/pkg/migration/migration.go
+++ b/pkg/migration/migration.go
@@ -174,18 +174,30 @@ func (h *Handler) processMigration(
 	// correctly as we go through different phases of migration we can use that instead of this
 	// internal annotation.
 	// Block the component migration until the portworx pod migration is finished.
-	cluster.Annotations[constants.AnnotationPauseComponentMigration] = "true"
-	if err := h.client.Update(context.TODO(), cluster, &client.UpdateOptions{}); err != nil {
+	liveCluster := &corev1.StorageCluster{}
+	if err := h.client.Get(context.TODO(), types.NamespacedName{Name: cluster.Name, Namespace: cluster.Namespace}, liveCluster); err != nil {
 		return err
 	}
-	// Unblock operator reconcile loop to start managing the storagecluster
-	util.UpdateStorageClusterCondition(cluster, &corev1.ClusterCondition{
-		Source: pxutil.PortworxComponentName,
-		Type:   corev1.ClusterConditionTypeMigration,
-		Status: corev1.ClusterConditionStatusInProgress,
-	})
-	if err := k8sutil.UpdateStorageClusterStatus(h.client, cluster); err != nil {
+	liveCluster.Annotations[constants.AnnotationPauseComponentMigration] = "true"
+	if err := h.client.Update(context.TODO(), liveCluster, &client.UpdateOptions{}); err != nil {
 		return err
+	}
+
+	// Unblock operator reconcile loop to start managing the storagecluster
+	liveCluster = &corev1.StorageCluster{}
+	if err := h.client.Get(context.TODO(), types.NamespacedName{Name: cluster.Name, Namespace: cluster.Namespace}, liveCluster); err != nil {
+		return err
+	}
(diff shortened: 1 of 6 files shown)


pkg/controller/storagecluster/storagecluster.go
```diff
@@ -778,6 +779,19 @@ func (c *Controller) updateStorageClusterStatus(
...
+	err := c.client.Get(context.TODO(),
+		types.NamespacedName{
+			Name:      cluster.Name,
+			Namespace: cluster.Namespace,
+		}, liveCluster)
+	if err != nil {
+		return err
+	}
+	toUpdate.Status.Phase = liveCluster.Status.Phase
+	toUpdate.Status.Conditions = liveCluster.Status.Conditions
```
piyush-nimbalkar, line 794: If we are getting overwriting the phase and conditions from what is in API server, then why are we calling `updateStorageClusterStatus`? Doesn't the caller want to do the same?

pkg/migration/migration.go
```diff
@@ -174,19 +174,31 @@ func (h *Handler) processMigration(
 	// correctly as we go through different phases of migration we can use that instead of this
 	// internal annotation.
 	// Block the component migration until the portworx pod migration is finished.
-	cluster.Annotations[constants.AnnotationPauseComponentMigration] = "true"
-	if err := h.client.Update(context.TODO(), cluster, &client.UpdateOptions{}); err != nil {
+	liveCluster := &corev1.StorageCluster{}
```
piyush-nimbalkar, line 177: Can we just use the `cluster` object here so that this latest cluster is passed along to the next steps in the migration? Otherwise the `cluster` object is going to be stale.

pkg/controller/storagecluster/storagecluster.go
pure-jliao: we call this controller `updateStorageClusterStatus` only when reconcile finished, while other conditions like migration or preflight are updated by the client directly and reflects in the stc object without calling this, otherwise when it hits error before reaching here the conditions will never be updated.
other components can update conditions and write to the object as well in component reconcile loops in the future, but here it should the only place where we iterate all conditions to conclude the phase, and supposed to be called when reconcile finished.

pkg/migration/migration.go
pure-jliao: after calling `processMigration` there's no other place we use the cluster object so I think it's ok. also since we have both migration handler and controller updating the storage cluster, most likely we always have to get the latest version, so it can be stale anytime.

pure-jliao: major issue to resolve here is conditions are updated by components, so we only overwrite the conditions here (phase should be optional). then underneath in `Driver.UpdateStorageClusterStatus`, px runtime condition will be updated first, then we iterate all conditions to conclude the phase.

pkg/controller/storagecluster/storagecluster.go
pure-jliao: major issue to resolve here is conditions are updated by components, so we only overwrite the conditions here (phase should be optional). then underneath in `Driver.UpdateStorageClusterStatus`, px runtime condition will be updated first, then we iterate all conditions to conclude the phase.

pure-jliao: found a better solution here, check and set conditions all in one place `Driver.UpdateStorageClusterStatus` if possible, including marking migration as completed. I think this makes more sense and easier to maintain. pr updated.

Pure-AdamuKaapan (commented):
LGTM, will let others approve as I'm still learning what's going on here

pkg/migration/migration.go
piyush-nimbalkar: Got it.
I saw that methods like `portworxNodesRollingUpdate` or `portworxNodesOnDeleteUpdate` were using it, but found out that they don't update the cluster. So it should be fine.

piyush-nimbalkar (approved):

PR #874, opened 2022-12-22
PWX-28162_pt3: Adding CSR auto-approval

zoxpx:
* adds new controller for CertificateSigningRequest (CSR) objects, which automatically approves TLS/SSL certificates requested by the OCI-Monitor
* functionality can be disabled by adding `portworx.io/disable-csr-approve: "true"` annotation, otherwise by default it will automatically approve PX-posted CSRs
* note that StorageCluster.Reconcile() needs to explicitly register the PX ClusterID with CSR-controller to handle only targetted (non-disabled) ClusterIDs

This is an improvement for the _"Automatic TLS setup"_ epic/work.

The TLS/SSL certificates currently need to be approved manually via `kubectl certificate approve px-node123`.
* as discussed w/ @piyush-nimbalkar, we are adding a dedicated controller that will monitor CertificateSigningRequest (CSR) objects, to automate this manual step, and automatically approve CSRs posted by the OCI-Monitor PODs
* note that after CSR is approved by PX-Operator, K8s will automatically attach the signed TLS/SSL cert to the CSR object
  - the PX installation watches CSRs, and will continue after TLS cert becomes available

Closes # PWX-28162  (part 3)

Additional dependent fixes:
* https://github.com/portworx/px-installer/pull/1525  -- independent fix for OCI-Monitor
* https://github.com/portworx/sched-ops/pull/294 -- `sched-ops` API update, **TODO** -- will need to vendor-in into the operator code


pkg/controller/csr/certificatesigningrequest.go
```diff
@@ -0,0 +1,225 @@
...
+		ret reconcile.Result
+		log = logrus.WithField("Request.Name", req.Name)
+	)
+
+	if !util.IsTLSEnabled() {
+		log.Debugf("CSR.Reconcile() bailing out  (TLS not enabled)")
+		return ret, nil
+	}
+
+	log.Infof("Reconciling CSR")
```
pure-jliao, line 108: maybe not good to stomp this

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -242,6 +243,8 @@ func (c *Controller) Reconcile(ctx context.Context, request reconcile.Request) (
 		return reconcile.Result{}, err
 	}
 
+	c.registerCSRAutoApproval(cluster)
```
pure-jliao, line 246: I'm not sure if this is the best place to do this? probably after waiting for migration approval? idk

pure-jliao (commented):
LGTM, will let others to take a look as well and approve

cmd/operator/operator.go
```diff
@@ -9,11 +9,23 @@ import (
 	"strings"
 	"time"
 
+	"github.com/libopenstorage/operator/drivers/storage"
```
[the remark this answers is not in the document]
zoxpx: Oh, I'm surprised, you were actually maintaining the import-lists (and gaps) manually?

Most of our golang-code elsewhere is formatted according to `goimports` command -- which separates "golang internal" vs "everything else" -imports (like I did here).

drivers/storage/portworx/util/util.go
```diff
@@ -136,6 +136,8 @@ const (
 	AnnotationClusterID = pxAnnotationPrefix + "/cluster-id"
 	// AnnotationPreflightCheck do preflight check before installing Portworx
 	AnnotationPreflightCheck = pxAnnotationPrefix + "/preflight-check"
+	// AnnotationDisableCSRAutoApprove annotation will disable CSR auto-approval
+	AnnotationDisableCSRAutoApprove = pxAnnotationPrefix + "/disable-csr-approve"
```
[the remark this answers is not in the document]
zoxpx: This is supposed to be rarely used configuration -- where customer _wants_ automatic TLS, but _does not want_ operator to automatically approve the certs.
* in this case, they will have to manually approve the certs (i.e. `kubectl certificate approve XXX`), and the blocked PX-installs will continue.

pkg/controller/storagecluster/storagecluster.go
zoxpx: I had to find a way how to bind the `StorageCluster`-controller updates, with CSR-controller
* note that StorageCluster YAML carries `disable-csr-approve` annotation that turns CSR-approval -controller on or off for that particular ClusterID
* so, we register the callback here, to configure the CSR-controller

pkg/controller/csr/certificatesigningrequest.go
zoxpx: Uh.. I'm not sure what you mean by this..   Could you please elaborate?

cmd/operator/operator.go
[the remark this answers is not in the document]
pure-jliao: IDE can sort the imports automatically, but here we want to separate the golang internal, portworx internal and everything else.

pkg/controller/csr/certificatesigningrequest.go
pure-jliao: do we want to print this in every reconcile loop every 20s? don't know if that's necessary.

cmd/operator/operator.go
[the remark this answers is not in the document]
zoxpx: Ok, if you guys keep re-sorting the lists manually, I'll revert my change.

pkg/controller/csr/certificatesigningrequest.go
zoxpx: Oh, I thought controller's `Reconcile()` triggers on K8s object's _changes_  (didn't realize it triggers on schedule, every 20 seconds).   If this is the case, what's the advantage of using K8s controllers in the fist place  (seems like cronjobs would do a similar job, and object watches would be more efficient?)

Other than this..
* the StorageCluster's `Reconcile()` will also drop an INFO-log every time it triggers
* also, since CSR-controller is a brand new functionality, I think it would be OK to have more verbose logging until we have better confidence the new code works w/o issues...

pkg/controller/csr/certificatesigningrequest.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+
+	if cnt := len(csr.Status.Certificate); cnt > 0 {
+		log.Debugf("CSR Certificate already signed  (%d certs attached)", cnt)
+		return ret, nil
+	}
+
+	// ignore no labels, or missing label name=portworx, or missing label cid
+	var clusterID string
+	if labs := csr.GetLabels(); labs == nil {
```
piyush-nimbalkar, line 150: nit: We can check `len(labs) == 0`. That way we check for both `nil` and empty map.

pkg/controller/csr/certificatesigningrequest.go
```diff
@@ -0,0 +1,225 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const (
+	ControllerName    = "px-csr-controller"
+	labelNameKey      = "name"
+	labelNameVal      = "portworx"
```
piyush-nimbalkar, line 33: To avoid defining constants in multiple places, can we make the existing constants public and use it in this controller?
Here is the reference to them: https://github.com/libopenstorage/operator/blob/master/drivers/storage/portworx/util/util.go#L604

pkg/controller/csr/certificatesigningrequest.go
```diff
@@ -0,0 +1,225 @@
...
+		Message:            "CSR failed px-operator validation process. Reason: " + reason,
+		LastUpdateTime:     metav1.Now(),
+		LastTransitionTime: metav1.Time{},
+	}
+	if approved {
+		cnd.Type = certv1.CertificateApproved
+		cnd.Reason = "cert validated by px-operator"
+		cnd.Message = "CSR approved by px-operator"
+	}
+	csr.Status.Conditions = append(csr.Status.Conditions, cnd)
```
piyush-nimbalkar, line 218: nit: Should we check if the condition of that type already exists and edit that instead of directly appending it?

pkg/controller/storagecluster/storagecluster.go
piyush-nimbalkar, line 246: Can we also unregister on StorageCluster delete so we no longer approve any further/pending CSRs with that ID? 
We can do that in the `deleteStorageCluster` function -  https://github.com/libopenstorage/operator/blob/ff3634adc3ffd93ad17dc7ba0e92d55682fc57a1/pkg/controller/storagecluster/storagecluster.go#L649

piyush-nimbalkar (approved):
Looks good, @zoxpx. Just a comment about unregister, rest are just nits.

pkg/controller/csr/certificatesigningrequest.go
```diff
@@ -0,0 +1,225 @@
...
+	scheme   *runtime.Scheme
+	recorder record.EventRecorder
+	ctrl     controller.Controller
+}
+
+// Init initializes CSR controller
+func (c *Controller) Init(mgr manager.Manager) error {
+	if ver, err := k8s.GetVersion(); err != nil {
+		return fmt.Errorf("could not determine K8s version: %s", err)
+	} else if ver.LessThanOrEqual(minSupportedK8sVersion) {
```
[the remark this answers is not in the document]
zoxpx: You are correct -- thanks for pointing it out
(although I don't think many people will run the patch-level 0 `k8s-1.19.0` version, but it's still a bug)

pkg/controller/csr/certificatesigningrequest.go
zoxpx: Note -- we don't actually call the `appendCondition()` "unconditionally" --
* if CSR has been approved/denied, we'll exit `Reconcile()`

ref from `Reconcile():137`
```go
	if approved, denied := c.getCertApprovalCondition(&csr.Status); approved || denied {
		log.WithField("approved", approved).WithField("denied", denied).
			Debugf("CSR already approved/denied")
		return ret, nil
	}
```

zoxpx:
**NOTE**: Piyush'es feedback addressed with [df4461c](https://github.com/libopenstorage/operator/pull/874/commits/df4461c0f76a463e891216d251b5e8093489219e) commit

zoxpx:
Thanks for the review guys -- merging the PR.

PR #875, opened 2022-12-22
[PWX-28349] Update STC status to Degraded when installation fails

pure-jliao:

Closes # 

### DIFF
--- a/pkg/controller/storagecluster/storagecluster.go
+++ b/pkg/controller/storagecluster/storagecluster.go
@@ -39,6 +39,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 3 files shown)


pkg/controller/storagecluster/storagecluster.go
```diff
@@ -234,7 +235,7 @@ func (c *Controller) Reconcile(ctx context.Context, request reconcile.Request) (
 
 	if err := c.validate(cluster); err != nil {
 		k8s.WarningEvent(c.recorder, cluster, util.FailedValidationReason, err.Error())
-		if updateErr := c.updateStorageClusterState(cluster, corev1.ClusterStateDegraded); updateErr != nil {
+		if updateErr := c.updateLiveStorageClusterState(cluster, corev1.ClusterStateDegraded); updateErr != nil {
```
haibinxie, line 238: nit: this is actually not needed, as the stc was just fetched a few lines above. https://github.com/libopenstorage/operator/blob/0ecb30b24fe96dcec287d507af4f127554e366a4/pkg/controller/storagecluster/storagecluster.go#L225

I think migration took longer and stc was probably updated somewhere.

haibinxie (approved):

pkg/controller/storagecluster/storagecluster.go
pure-jliao: yes probably this is not needed, most scenarios we don't need this but just incase the cluster is updated somewhere unknown, we won't getting the object update error saying object is updated. And in this function we only update the stc phase, so I think it's good to have.

piyush-nimbalkar (approved):

PR #879, opened 2023-01-11
PWX-27934: fix CSI socket path is not mounted in portworx container

Glitchfix:
add CSI mount path to portworx container

Signed-off-by: Shivanjan Chakravorty <schakravorty@purestorage.com>

With the new installation method with operator, CSI socket path is not mounted in portworx container.
The fix for this was to include the CSI volume mount path in while creating portworx container

### DIFF
--- a/drivers/storage/portworx/deployment.go
+++ b/drivers/storage/portworx/deployment.go
@@ -1233,6 +1233,7 @@ func (t *template) getEnvList() []v1.EnvVar {
 func (t *template) getVolumeMounts() []v1.VolumeMount {
 	volumeInfoList := getDefaultVolumeInfoList(t.pxVersion)
 	extensions := []func() []volumeInfo{
+		t.getCSIVolumeInfoList,
 		t.getK3sVolumeInfoList,
 		t.getPKSVolumeInfoList,
 		t.getBottleRocketVolumeInfoList,
@@ -1422,6 +1423,7 @@ func (t *template) getCSIVolumeInfoList() []volumeInfo {
 		volumeInfo{
 			name:         "csi-driver-path",
 			hostPath:     t.csiConfig.DriverBasePath(),
+			mountPath:    t.csiConfig.DriverBasePath(),
 			hostPathType: hostPathTypePtr(v1.HostPathDirectoryOrCreate),
 		},
 	)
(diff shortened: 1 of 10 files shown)


drivers/storage/portworx/deployment.go
```diff
@@ -1462,6 +1463,7 @@ func (t *template) getCSIVolumeInfoList() []volumeInfo {
 		volumeInfo{
 			name:         "csi-driver-path",
 			hostPath:     t.csiConfig.DriverBasePath(),
+			mountPath:    t.csiConfig.DriverBasePath(),
```
haibinxie, line 1466: as we discussed in the meeting, the kubeletPath should be retrieved from env var.

zoxpx:
**ATTN**: This commit breaks Portworx installs for all non-OpenShift Kubernetes deployments !!

The issue is that we are adding a new `/var/lib/kubelet/plugins/pxd.portworx.com` mount, which will get rejected since the parent kubelet-directory `/var/lib/kubelet` is already (automatically) mounted.
* we added code into `px-runc` binary to specifically check for such mount-overlaps, because they can lead into very bad consequences  (e.g. the host-system will hang, after `portworx.service` is restarted 5 to 10 times).

haibinxie:
> **ATTN**: This commit breaks Portworx installs for all non-OpenShift Kubernetes deployments !!
> 
> The issue is that we are adding a new `/var/lib/kubelet/plugins/pxd.portworx.com` mount, which will get rejected since the parent kubelet-directory `/var/lib/kubelet` is already (automatically) mounted.
> 
> * we added code into `px-runc` binary to specifically check for such mount-overlaps, because they can lead into very bad consequences  (e.g. the host-system will hang, after `portworx.service` is restarted 5 to 10 times).

what's your suggestion to fix, shall oci-mon take a change too?

zoxpx:
> what's your suggestion to fix, shall oci-mon take a change too?

The original issue was with a PX+CSI deployments on IBM's K8s clusters, which use a non-standard kubelet directory (~`/var/data/kubelet/`~ instead of `/var/lib/kubelet`).
As a better fix, I would suggest to detect the destination-cluster type as IBM-cluster, and change the CSI socket's location  (i.e. `env CSI_ENDPOINT=/var/data/kubelet/csi-plugins/pxd.portworx.com/csi.sock`)

vinayakshnd:
@Glitchfix pls close this PR in favor of correct fix done as part of https://github.com/libopenstorage/operator/pull/884

PR #880, opened 2023-01-11
[PWX-27920] Enable metrics collector batching to reduce memory usage

pure-jliao:
* revert previous changes which disables metrics collector by default
* update default metrics collector image and config

Closes #

### DIFF
--- a/drivers/storage/portworx/component/telemetry.go
+++ b/drivers/storage/portworx/component/telemetry.go
@@ -408,11 +408,15 @@ func (t *telemetry) deleteCCMGoPhonehomeCluster(
 	return nil
 }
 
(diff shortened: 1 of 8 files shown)


drivers/storage/portworx/component/telemetry.go
```diff
@@ -116,7 +116,6 @@ type telemetry struct {
 	isDeploymentRegistrationServiceCreated bool
 	isDaemonSetTelemetryPhonehonmeCreated  bool
 	usePxProxy                             bool
-	reconcileMetricsCollector              *bool
```
haibinxie, line 119: you may want to keep the code but set it to true just in case there is still issue in future

haibinxie (approved):

PR #884, opened 2023-01-13
PWX-28101 read custom kubelet path from env

Glitchfix:
Signed-off-by: Shivanjan Chakravorty <schakravorty@purestorage.com>

PX Operator allows to pass a non standard kubelet config path and setup the csi sidecars accordingly

PX Operator should allow to pass a non standard kubelet config path and setup the csi sidecars accordingly

Closes #879

### DIFF
--- a/drivers/storage/portworx/util/util.go
+++ b/drivers/storage/portworx/util/util.go
@@ -71,6 +71,8 @@ const (
 	EssentialsUserIDKey = "px-essen-user-id"
 	// EssentialsOSBEndpointKey is the secret key for Essentials OSB endpoint
 	EssentialsOSBEndpointKey = "px-osb-endpoint"
+	// EnvKeyKubeletDir env var to set custom kubelet directory
+	EnvKeyKubeletDir = "KUBELET_DIR"
 
 	// AnnotationIsPKS annotation indicating whether it is a PKS cluster
 	AnnotationIsPKS = pxAnnotationPrefix + "/is-pks"
@@ -438,6 +440,13 @@ func KubeletPath(cluster *corev1.StorageCluster) string {
 	if IsPKS(cluster) {
 		return "/var/vcap/data/kubelet"
 	}
+
+	for _, env := range cluster.Spec.Env {
+		if env.Name == EnvKeyKubeletDir && len(env.Value) > 0 {
+			return env.Value
+		}
+	}
+
 	return "/var/lib/kubelet"
 }
 


drivers/storage/portworx/util/util.go
```diff
@@ -438,6 +440,13 @@ func KubeletPath(cluster *corev1.StorageCluster) string {
 	if IsPKS(cluster) {
 		return "/var/vcap/data/kubelet"
 	}
+
+	for _, env := range cluster.Spec.Env {
```
haibinxie, line 444: can we move this before IsPKS check, it should be the same if uses choose to overwrite it in case of failures, I think the variations should also exist on PKS

Glitchfix: you are right, let me add that change

drivers/storage/portworx/util/util.go
```diff
@@ -435,9 +437,16 @@ func StartPort(cluster *corev1.StorageCluster) int {
 
 // KubeletPath returns the kubelet path
 func KubeletPath(cluster *corev1.StorageCluster) string {
+	for _, env := range cluster.Spec.Env {
```
haibinxie, line 440: Could you add some unit tests to validate CSI deployment/statefulset and PX pod take the env variable?
CSI test example: https://github.com/libopenstorage/operator/blob/master/drivers/storage/portworx/components_test.go#L5244

Portworx pod test example:
https://github.com/libopenstorage/operator/blob/master/drivers/storage/portworx/deployment_test.go#L99

Glitchfix: added unit test in both deployment and components

README.md
```diff
@@ -3,3 +3,45 @@
...
+
+If you get errors like 
+
+```sh
+/usr/local/go/src/math/erf.go:189:6: Erf defined in both Go and assembly
+/usr/local/go/src/math/erf.go:274:6: Erfc defined in both Go and assembly
+```
+
+
+try to switch to go 1.18.
```
haibinxie, line 31: we require go 1.19+

drivers/storage/portworx/deployment_test.go
```diff
@@ -96,6 +96,84 @@ func TestBasicRuncPodSpec(t *testing.T) {
...
+				RuntimeOpts: map[string]string{
+					"op1": "10",
+				},
+			},
+		},
+	}
+
+	driver := portworx{}
+
+	actual, err := driver.GetStoragePodSpec(cluster, nodeName)
```
haibinxie, line 163: In this function it should be validating CSI statefulset uses the path. 
https://github.com/libopenstorage/operator/blob/master/drivers/storage/portworx/components_test.go#L5431

haibinxie (approved):

PR #887, opened 2023-01-17
PWX-26914: Add resources config for stork, stork-scheduler and autopilot

haibinxie:

Closes #

### DIFF
--- a/deploy/crds/core_v1_storagecluster_crd.yaml
+++ b/deploy/crds/core_v1_storagecluster_crd.yaml
@@ -643,6 +643,30 @@ spec:
                                               type: string
                                             mode:
                                               type: integer
+                  resources:
+                    type: object
+                    description: Specifies the resource requirements for stork and stork scheduler.
+                    properties:
+                      requests:
+                        type: object
+                        description: Requested resources.
+                        properties:
+                          memory:
+                            type: string
+                            description: Requested memory.
+                          cpu:
+                            type: string
+                            description: Requested cpu.
+                      limits:
+                        type: object
+                        description: Resource limit.
+                        properties:
+                          memory:
+                            type: string
+                            description: Memory limit.
+                          cpu:
+                            type: string
+                            description: CPU limit.
               userInterface:
                 type: object
                 description: Contains spec of a user interface for the storage driver.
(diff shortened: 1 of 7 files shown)


Pure-AdamuKaapan (commented):
Makes sense to me, will let someone else review as well to sanity check

deploy/crds/core_v1_storagecluster_crd.yaml
```diff
@@ -643,6 +643,30 @@ spec:
                                               type: string
                                             mode:
                                               type: integer
+                  resources:
+                    type: object
+                    description: Specifies the compute resource requirements for the storage pod.
```
pure-jliao, line 648: guess you just copied it from storage pod resource setting? need to rephrase those descriptions.

pkg/controller/storagecluster/stork.go
```diff
@@ -1117,6 +1121,12 @@ func getStorkSchedDeploymentSpec(
 		deployment.Spec.Template.Spec.TopologySpreadConstraints = topologySpreadConstraints
 	}
 
+	// If resources is specified in the spec, the resources specified by annotation (such as portworx.io/stork-cpu)
```
pure-jliao, line 1124: there are 2 annotations stork-cpu and stork-scheduler-cpu, they might be different, after the spec change we unify them into one?

pure-jliao (approved):
LGTM. have one suggestion how we can handle the old annotations. I'm thinking we can remove the annotation check logic in autopilot and stork when setting the deployment specs, then check the old annotations in "set portworx default" function, if exists we set the new resource spec values, and remove the annotation from stc, so we can safely deprecate the annotations. only thing is stork has 2 separate annotations we may need to combine them into 1.

pkg/controller/storagecluster/stork.go
haibinxie: Yes, in storagecluster spec we can't configure them separately

deploy/crds/core_v1_storagecluster_crd.yaml
haibinxie: nice catch.

haibinxie:
> LGTM. have one suggestion how we can handle the old annotations. I'm thinking we can remove the annotation check logic in autopilot and stork when setting the deployment specs, then check the old annotations in "set portworx default" function, if exists we set the new resource spec values, and remove the annotation from stc, so we can safely deprecate the annotations. only thing is stork has 2 separate annotations we may need to combine them into 1.

Changing STC may cause confusion to the users.

PR #889, opened 2023-01-18
[PWX-27622] Use provided AWS credentials to run permission check on EKS

pure-jliao:
* When AWS credentials are provided in stc env, preflight check will use that to create a client to check cloud permission instead of using privileged instances.
* User can modify specified credentials and retrigger preflight check again.
* Credentials can be passed from string or k8s secret object

Closes #

Test manually:
1. without privileged instances, preflight respond 403 error in the condition list
2. with wrong credentials provided then rerun preflight, got a different 401 error in the condition list
3. update stc to use correct credentials then rerun preflight, px is deployed successfully
4. uninstall and add privileges to node group iam, remove aws credentials, redeploy portworx, preflight passed successfully

### DIFF
--- a/pkg/preflight/aws.go
+++ b/pkg/preflight/aws.go
@@ -1,16 +1,20 @@
 package preflight
 
 import (
+	"context"
 	"fmt"
+	"os"
 	"strings"
 
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/aws/aws-sdk-go/service/opsworks"
 	"github.com/pborman/uuid"
 	"github.com/sirupsen/logrus"
+	v1 "k8s.io/api/core/v1"
 
 	"github.com/libopenstorage/cloudops"
 	awsops "github.com/libopenstorage/cloudops/aws"
+	pxutil "github.com/libopenstorage/operator/drivers/storage/portworx/util"
 	corev1 "github.com/libopenstorage/operator/pkg/apis/core/v1"
 )
 
@@ -22,6 +26,9 @@ const (
 
 	// dryRunErrMsg dry run error response if succeeded, otherwise will be UnauthorizedOperation
 	dryRunErrMsg = "DryRunOperation"
+
+	awsAccessKeyEnvName = "AWS_ACCESS_KEY_ID"
+	awsSecretKeyEnvName = "AWS_SECRET_ACCESS_KEY"
 )
 
 var (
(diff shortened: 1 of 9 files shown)


pkg/preflight/aws.go
```diff
@@ -55,6 +73,45 @@ func (a *aws) initCloudOps() error {
...
+	secretKey, err := pxutil.GetValueFromEnvVar(context.TODO(), a.k8sClient, secretKeyEnv, cluster.Namespace)
+	if err != nil {
+		return err
+	}
+
+	// Set environment variables using credentials above
+	if err := os.Setenv(awsAccessKeyEnvName, accessKey); err != nil {
+		return err
+	}
+	if err := os.Setenv(awsSecretKeyEnvName, secretKey); err != nil {
```
haibinxie, line 107: is this a new or legacy behavior? is there any security concern about exposing it in env var?

haibinxie (approved):

pkg/preflight/aws.go
pure-jliao: @haibinxie based on doc https://docs.portworx.com/install-portworx/cloud/aws/aws-asg/
we can either grant the ec2 instance privileges or pass this user specified env vars to storage cluster spec, then the credentials will be passed to px pods.
the behavior has been there. If we are passing it to px pod, I don't see security concerns here.

piyush-nimbalkar: We should not be setting these values in the operator's memory. We do pass the values to the pods but not directly. We use the `env.valueFrom` field and don't directly add the value ourselves. Kubernetes takes the responsibility of fetching the secret and mounting it in the pod securely.

pkg/preflight/aws.go
```diff
@@ -34,13 +41,24 @@ type aws struct {
 	zone string
 }
 
-func (a *aws) initCloudOps() error {
-	if a.ops != nil {
+func (a *aws) initCloudOps(cluster *corev1.StorageCluster) error {
+	// Set env vars for user provided credentials: AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY
+	existingAccessKey := os.Getenv(awsAccessKeyEnvName)
+	existingSecretKey := os.Getenv(awsSecretKeyEnvName)
```
piyush-nimbalkar, line 47: As commented below, we should not set/get these in operator's env.

pkg/preflight/aws.go
pure-jliao: but this is operator pod, we cannot edit operator pod to add volumes and mounts. we can use stc.env.valueFrom to get the credential from storage cluster, then do we have other better ways to set this env in operator container? @piyush-nimbalkar

piyush-nimbalkar: My worry is that if someone gets access to the operator pod, they can exec into it and print env values.

I was hoping we will get the values from the secret and directly pass them to the AWS sdk client directly. But then I just saw that we are using the cloudops package which expects the creds in the env variables.
I don't see any other way apart from adding a new NewClient to cloudops to take the id and secret as parameters. Let's merge this change for now and explore adding a new method to cloudops.

piyush-nimbalkar (approved):

pkg/preflight/aws.go
pure-jliao: created https://portworx.atlassian.net/browse/PWX-28664 to track

pkg/preflight/aws.go
```diff
@@ -22,6 +26,9 @@ const (
 
 	// dryRunErrMsg dry run error response if succeeded, otherwise will be UnauthorizedOperation
 	dryRunErrMsg = "DryRunOperation"
+
+	awsAccessKeyEnvName = "AWS_ACCESS_KEY_ID"
+	awsSecretKeyEnvName = "AWS_SECRET_ACCESS_KEY"
```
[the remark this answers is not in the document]
pure-jliao: interesting thing is we don't have this constant anywhere..

PR #893, opened 2023-01-18
PWX28442: add log to debug PDB

haibinxie:


drivers/storage/portworx/util/util.go
```diff
@@ -1192,18 +1192,23 @@ func CountStorageNodes(
 				[]string{node.DataIp, node.MgmtIp, node.Hostname},
 			)
 			if err != nil {
-				logrus.Warnf("Unable to find kubernetes node name for nodeID %v: %v", node.Id, err)
+				logrus.Warnf("Unable to find kubernetes node name for nodeID %v: %v, node object: %+v", node.Id, err, node)
```
pure-jliao, line 1195: can we have another debug line for the whole object? this can dump a lot of logs on metro dr clusters.

pure-jliao (approved):

PR #894, opened 2023-01-19
PWX-28184: Add 'base-id' param to envoy startup to allow multiple envoys to run.

jrivera-px:
Run multiple envoys on the same system

Closes # PWX-28184

### DIFF
--- a/deploy/ccm/metrics-collector-deployment.yaml
+++ b/deploy/ccm/metrics-collector-deployment.yaml
@@ -34,6 +34,8 @@ spec:
           readOnly: true
       - args:
         - envoy
+        - "--base-id"
+        - "4"
         - --config-path
         - /config/envoy-config.yaml
         image: purestorage/telemetry-envoy:1.0.0
--- a/deploy/ccm/registration-service.yaml
+++ b/deploy/ccm/registration-service.yaml
@@ -34,6 +34,8 @@ spec:
             name: proxy-config
           args:
           - "envoy"
+          - "--base-id"
+          - "5"
           - "--config-path"
           - "/config/envoy-config-register.yaml"
       volumes:
--- a/drivers/storage/portworx/testspec/ccmGoCollectorDeployment.yaml
+++ b/drivers/storage/portworx/testspec/ccmGoCollectorDeployment.yaml
@@ -54,6 +54,8 @@ spec:
           readOnly: true
       - args:
         - envoy
+        - "--base-id"
+        - "4"
         - --config-path
         - /config/envoy-config.yaml
         image: docker.io/purestorage/envoy:1.2.3
--- a/drivers/storage/portworx/testspec/ccmGoRegisterDeployment.yaml
+++ b/drivers/storage/portworx/testspec/ccmGoRegisterDeployment.yaml
@@ -46,6 +46,8 @@ spec:
           readOnly: true
       - args:
         - envoy
+        - "--base-id"
+        - "5"
         - --config-path
         - /config/envoy-config-register.yaml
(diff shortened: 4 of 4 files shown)


pure-jliao (commented):
do we need this? I think we discussed before, if the registration pod is not using host network, this envoy instance won't conflict with other envoys on the host?
cc: @haibinxie @piyush-nimbalkar

haibinxie:
Yes, I remember this should not be an issue.

jrivera-px:
This appears to be a different issue.   The registration pods is failing to start failing connection to the envoy.   Envoy logs show "unable to bind domain socket with base_id=0, id=0, errno=98 (see --base-id option)".     I don't see a host port being used in this instance.

jrivera-px:
Actually are you suggesting that this should be removed : https://github.com/libopenstorage/operator/blob/15e4bc82eb77a52064cf619ddbd1bb169a9bb894/deploy/ccm/phonehome-cluster.yaml#L66?

pure-jliao:
I remember this phonehome envoy port is exposed because the registration service will use it for secret renewal purpose. as renewal will call the `rest.cloud-support.purestorage.com` instead of `register.cloud-support.purestorage.com`

jrivera-px:
So then I believe we will need the base-id change in this case.

pure-jliao:
my confusion is do we consistently hit this issue so that registration is not working? or is that a setup issue, probably due to there's some stale envoy setup or something on your cluster? do we always have this issue? it used to work fine right?

jrivera-px:
This is an Anthos setup we never tests on this on this deployment before.  On Anthos this issue appears to be consistent but we don't hit this on all deployments.  In fact only seen a similar issue on OCP     It does work fine on other deployments without this change.   This I believe is also related to localhost connection on the hostPost.    We had a similar issue on OCP (PWX-28184) which we caused when we added a hostNetwork and HostPort for for the uploader connection.  However when we did this it caused the issue base-id error with the envoy port 9901 and in the PWX notes I mentioned you changed the port to 9902.   But this change did not need to be permanent and we worked around it via the client connection.  So this issue appears to be different.

pure-jliao:
thanks @jrivera-px. IMO there's no harm to add base-id here for registration pod, you need to fix unit test specs as well https://github.com/libopenstorage/operator/blob/master/drivers/storage/portworx/testspec/ccmGoRegisterDeployment.yaml
my question is do we need to add a base id for metrics collector and phonehome pods as well, I feel we can do this on all envoy args to avoid potential issues. Also if we pick base-id=5, is it possible to conflict with let's say the customer has another envoy setup on the host out side of px with base-id=5? probably we can pick some large numbers to avoid this?

jrivera-px:
I think we only need to change the registry pod.     It has the potential conflict when it lands on the same node as a telemetry pod.  Since  the default base-id is 0 if we leave telemetry configs unchanged they will always start with the default and registry is always 5 we should never get a conflict.   Also  we only have one telemetry pod on a node so regardless of where the registration pod falls we should never see the conflict.  Is there a case where we would have two telemetry pods on the same node cause that would be hard to solve?

pure-jliao:
@jrivera-px we should have phonehome pods on all nodes, registration pod and metrics collector pod potentially can run on a same node as one of the phonehome, so it's good to make this change for metrics collector as well, and use a different id other than 5. potentially we can have all those 3 pods on a same node.

jrivera-px:
Oh ok.. I will need to find the config files which require the changes then and I will add them to this PR.

pure-jliao (approved):

PR #900, opened 2023-01-25
PTX-15076 Add retry for deleting StorageCluster

nikolaypopov:
Seen this timeout and fail job, adding retry with timeout for deleting StorageCluster

### DIFF
--- a/pkg/util/test/util.go
+++ b/pkg/util/test/util.go
@@ -112,6 +112,9 @@ const (
 	defaultPxAuthValidationTimeout  = 20 * time.Minute
 	defaultPxAuthValidationInterval = 30 * time.Second
 
+	defaultDeleteStorageClusterTimeout  = 3 * time.Minute
+	defaultDeleteStorageClusterInterval = 10 * time.Second
+
 	defaultRunCmdInPxPodTimeout  = 25 * time.Second
 	defaultRunCmdInPxPodInterval = 5 * time.Second
 )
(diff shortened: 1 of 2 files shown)


pkg/util/test/util.go
```diff
@@ -427,26 +430,41 @@ func UninstallStorageCluster(cluster *corev1.StorageCluster, kubeconfig ...strin
...
+				cluster.Spec.DeleteStrategy.Type != corev1.UninstallStorageClusterStrategyType) {
+			cluster.Spec.DeleteStrategy = &corev1.StorageClusterDeleteStrategy{
+				Type: corev1.UninstallAndWipeStorageClusterStrategyType,
+			}
+
+			if _, err := operatorops.Instance().UpdateStorageCluster(cluster); err != nil {
+				return nil, true, err
+			}
+
+			if err := validateTelemetrySecret(cluster, defaultTelemetrySecretValidationTimeout, defaultTelemetrySecretValidationInterval, false); err != nil {
```
pure-jliao, line 451: we already have retry inside with 30-min timeout, so if this fails, we should fail the test

pkg/util/test/util.go
```diff
@@ -427,26 +430,41 @@ func UninstallStorageCluster(cluster *corev1.StorageCluster, kubeconfig ...strin
...
-			return err
+
+		if cluster.Spec.DeleteStrategy == nil ||
+			(cluster.Spec.DeleteStrategy.Type != corev1.UninstallAndWipeStorageClusterStrategyType &&
+				cluster.Spec.DeleteStrategy.Type != corev1.UninstallStorageClusterStrategyType) {
+			cluster.Spec.DeleteStrategy = &corev1.StorageClusterDeleteStrategy{
+				Type: corev1.UninstallAndWipeStorageClusterStrategyType,
+			}
+
+			if _, err := operatorops.Instance().UpdateStorageCluster(cluster); err != nil {
```
pure-jliao, line 447: nit: it might be better to get the latest cluster here for the validation and deletion later? to avoid running into object version issues. I don't know if that would happen, so probably it's ok

pkg/util/test/util.go
```diff
@@ -112,6 +112,9 @@ const (
 	defaultPxAuthValidationTimeout  = 20 * time.Minute
 	defaultPxAuthValidationInterval = 30 * time.Second
 
+	defaultDeleteStorageClusterTimeout  = 30 * time.Minute
```
pure-jliao, line 115: do we really need 30 minutes to retry the stc deletion? it shouldn't take that long. is 10 minutes better?

pkg/util/test/util.go
```diff
@@ -427,26 +430,41 @@ func UninstallStorageCluster(cluster *corev1.StorageCluster, kubeconfig ...strin
...
+			}
+
+			if err := validateTelemetrySecret(cluster, defaultTelemetrySecretValidationTimeout, defaultTelemetrySecretValidationInterval, false); err != nil {
+				return nil, true, err
+			}
 		}
 
-		if err := validateTelemetrySecret(cluster, defaultTelemetrySecretValidationTimeout, defaultTelemetrySecretValidationInterval, false); err != nil {
-			return err
+		if err := operatorops.Instance().DeleteStorageCluster(cluster.Name, cluster.Namespace); err != nil {
```
haibinxie, line 456: the deletion actually returns immediately, the storageCluster is still there. we shouldn't need this fix. do you want to use UninstallAndValidateStorageCluster?

pkg/util/test/util.go
nikolaypopov: I don't know, network instabilities might cause `px-node-wiper` pods potentially can take longer to pull images, I thought its better to have a little bigger timeout than to fail due to these things. Our lab occasionally goes unstable in terms of resources ^^

pkg/util/test/util.go
nikolaypopov: The method we use for deleting and validating StorageCluster deletion is called `ci_utils.UninstallAndValidateStorageCluster(cluster, t)`, we are already using it and it has
```
	// Delete cluster
	logrus.Infof("Delete StorageCluster %s", cluster.Name)
	err := testutil.UninstallStorageCluster(cluster)
	require.NoError(t, err)

	// Validate cluster deletion
	logrus.Infof("Validate StorageCluster %s deletion", cluster.Name)
	err = testutil.ValidateUninstallStorageCluster(cluster, DefaultValidateUninstallTimeout, DefaultValidateUninstallRetryInterval)
	require.NoError(t, err)
```
The problem there is that some of these APIs are getting connection refused and are not going through, hence why I figured we should add more retries for such calls, I think its better retry some of these than to fail a long running job due to network issues. I am seeing more and more if these last couple days because something going on in the lab. Let me know if you and @pure-jliao want to sync up about this so we can discuss what is the best way to handle these and avoid test failures

pkg/util/test/util.go
nikolaypopov: The method we use for deleting and validating StorageCluster deletion is called `ci_utils.UninstallAndValidateStorageCluster(cluster, t)`, we are already using it and it has
```
	// Delete cluster
	logrus.Infof("Delete StorageCluster %s", cluster.Name)
	err := testutil.UninstallStorageCluster(cluster)
	require.NoError(t, err)

	// Validate cluster deletion
	logrus.Infof("Validate StorageCluster %s deletion", cluster.Name)
	err = testutil.ValidateUninstallStorageCluster(cluster, DefaultValidateUninstallTimeout, DefaultValidateUninstallRetryInterval)
	require.NoError(t, err)
```
The problem there is that some of these APIs are getting connection refused and are not going through, hence why I figured we should add more retries for such calls, I think its better retry some of these than to fail a long running job due to network issues. I am seeing more and more if these last couple days because something going on in the lab. Let me know if you and @pure-jliao want to sync up about this so we can discuss what is the best way to handle these and avoid test failures

pkg/util/test/util.go
nikolaypopov: Actually I just checked its 30 secs, so we are good
```
    defaultTelemetrySecretValidationTimeout  = 30 * time.Second
    defaultTelemetrySecretValidationInterval = time.Second
```

pkg/util/test/util.go
nikolaypopov: We do not need to get new `StorageCluster` object as `operatorops.Instance().DeleteStorageCluster(cluster.Name, cluster.Namespace)` only cares about name and namespace

pkg/util/test/util.go
haibinxie: in that case we can retry for a few times, but we dont have to wait for 30 mins, 3 mins should be good.

nikolaypopov: oh yeah you are right, I will change to 3 mins or less, I missed the point that this is just for deleting and not validating deletion

nikolaypopov: I have changed timeout to 3 mins with 10 second retry interval

haibinxie (approved):

pure-jliao (approved):
LGTM!

PR #904, opened 2023-01-27
Fix missing node name from the storage pod template

piyush-nimbalkar:

Closes #

### DIFF
--- a/pkg/controller/storagecluster/storagecluster.go
+++ b/pkg/controller/storagecluster/storagecluster.go
@@ -1189,6 +1189,7 @@ func (c *Controller) CreatePodTemplate(
 	}
 	k8s.AddOrUpdateStoragePodTolerations(&podSpec)
 
+	podSpec.NodeName = node.Name
 	newTemplate := v1.PodTemplateSpec{
 		ObjectMeta: metav1.ObjectMeta{
 			Namespace: cluster.Namespace,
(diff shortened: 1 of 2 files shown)


pp511 (approved):
lgtm.

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -1189,6 +1189,7 @@ func (c *Controller) CreatePodTemplate(
 	}
 	k8s.AddOrUpdateStoragePodTolerations(&podSpec)
 
+	podSpec.NodeName = node.Name
```
haibinxie, line 1192: can we move this to be inside of driver.GetStoragePodSpec?

haibinxie (approved):

pkg/controller/storagecluster/storagecluster.go
piyush-nimbalkar: I thought about that initially, but kept it out in the open to make it clear we have a node selector.
Also, it is more of the main controller's responsibility to ensure the pod runs on the node for which it is creating a pod template.
I can change if you still insist.

haibinxie: It's ok. there is another caller of driver.GetStoragePodSpec for migration, which does not use nodeName currently

haibinxie: Thought about if kvdb had the issue, actually nodeName is set in driver.GetKVDBPodSpec. it would be nice to put it there for consistency and in case it's missed in future code changes.

PR #916, opened 2023-02-02
[PWX-28664] Reset AWS credential env vars  after client creation

pure-jliao:
* aws credential env vars are set in operator container to init the preflight check client, which brings us security concerns
* with this change the env vars will be reset once the client is created

Closes #

verified manually on EKS


pkg/preflight/aws.go
```diff
@@ -99,19 +107,40 @@ func (a *aws) setAWSCredentialEnvVars(cluster *corev1.StorageCluster) error {
 	if err != nil {
 		return err
 	}
+	if accessKey == "" || secretKey == "" {
+		return fmt.Errorf("both AWS_ACCESS_KEY_ID and AWS_ACCESS_KEY_ID need to be provided")
```
piyush-nimbalkar, line 111: .. and *AWS_SECRET_ACCESS_KEY ..

Same above on line 98

pkg/preflight/aws.go
```diff
@@ -99,19 +107,40 @@ func (a *aws) setAWSCredentialEnvVars(cluster *corev1.StorageCluster) error {
...
 	}
+	a.credentialHash = credentialHash
+	a.ops = nil
 
 	logrus.Infof("environment variables set using AWS credentials provided")
 	return nil
 }
 
+func resetAWSCredentialEnvVars() {
+	if err := os.Setenv(awsAccessKeyEnvName, ""); err != nil {
```
piyush-nimbalkar, line 136: nit: We can call `os.Unsetenv`

piyush-nimbalkar (approved):

pkg/preflight/aws.go
```diff
@@ -37,29 +38,26 @@ var (
...
 		return err
 	}
 
-	// init aws client when
-	// 1. it's never created
-	// 2. user provided credentials got changed
-	if a.ops != nil &&
-		existingAccessKey == os.Getenv(awsAccessKeyEnvName) &&
-		existingSecretKey == os.Getenv(awsSecretKeyEnvName) {
+	if a.ops != nil {
```
haibinxie, line 54: why dont compare the value here? if the value changes the client wont be recreated?

pkg/preflight/aws.go
```diff
@@ -84,10 +86,16 @@ func (a *aws) setAWSCredentialEnvVars(cluster *corev1.StorageCluster) error {
 		}
 	}
 	if accessKeyEnv == nil && secretKeyEnv == nil {
+		// No credential provided or credentials are removed
 		logrus.Debugf("no aws credentials provided, will use instance privileges instead")
+		if a.credentialHash != "" {
+			a.credentialHash = ""
+			a.ops = nil
+			resetAWSCredentialEnvVars()
```
haibinxie, line 94: probably change the function name to remove or unset. The setAWSCredentialEnvVars function actually resets the value.

pkg/preflight/aws.go
pure-jliao: now all that is handled in `setAWSCredentialEnvVars`, the env vars are only set when it needs to create a new client, e.g. credentials removed, added or updated, otherwise we don't set the env.

PR #917, opened 2023-02-03
[PWX-28868] Restart telemetry phonehome DaemonSet when ccm ConfigMap gets updated

pure-jliao:
* when ccm config got updated during operator upgrade, phonehome doesn't reload the latest config, so restarting all phonehome pods as a workaround for now.

Closes #

### DIFF
--- a/drivers/storage/portworx/component/telemetry.go
+++ b/drivers/storage/portworx/component/telemetry.go
@@ -114,7 +114,7 @@ type telemetry struct {
 	isCCMGoSupported                       bool
 	isCollectorDeploymentCreated           bool
 	isDeploymentRegistrationServiceCreated bool
-	isDaemonSetTelemetryPhonehonmeCreated  bool
+	isDaemonSetTelemetryPhonehomeCreated   bool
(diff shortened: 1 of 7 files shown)


drivers/storage/portworx/component/telemetry.go
```diff
@@ -806,16 +808,35 @@ func (t *telemetry) createCCMGoConfigMapTLSCertificate(
...
+			Name:      ConfigMapNameTelemetryPhonehome,
+			Namespace: cluster.Namespace,
+		},
+		existingConfigMap,
+	)
+	if err == nil && reflect.DeepEqual(existingConfigMap.Data, data) {
+		dataChanged = false
+	}
+
+	return dataChanged, k8sutil.CreateOrUpdateConfigMap(
```
haibinxie, line 839: CreateOrUpdateConfigMap already calls get and has isModified flag, there is no need to make another get call in this function. you can return the flag from that function directly.

dsupure (approved):
LGTM!

PR #920, opened 2023-02-06
PWX-28608 add minReadySeconds for rolling updates

haibinxie:
During rolling update, wait for minReadySeconds after pod is ready before rolling update next pod

### DIFF
--- a/deploy/crds/core_v1_storagecluster_crd.yaml
+++ b/deploy/crds/core_v1_storagecluster_crd.yaml
@@ -140,6 +140,13 @@ spec:
                     type: object
                     description: Spec to control the desired behavior of storage cluster rolling update.
                     properties:
+                      minReadySeconds:
+                        description: Minimum number of seconds for which a newly created Portworx pod
+                          should be ready without any of its container crashing for it to
+                          be considered available. Defaults to 0 (pod will be considered available
+                          as soon as it is ready).
+                        format: int32
+                        type: integer
                       maxUnavailable:
                         x-kubernetes-int-or-string: true
                         description: >-
--- a/pkg/apis/core/v1/storagecluster.go
+++ b/pkg/apis/core/v1/storagecluster.go
@@ -300,6 +300,12 @@ type RollingUpdateStorageCluster struct {
 	// that at least 70% of original number of StorageCluster pods are available at
 	// all times during the update.
 	MaxUnavailable *intstr.IntOrString `json:"maxUnavailable,omitempty"`
+
+	// Minimum number of seconds for which a newly created Portworx pod should be ready
+	// without any of its container crashing, for it to be considered available.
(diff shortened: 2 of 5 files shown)


deploy/crds/core_v1_storagecluster_crd.yaml
```diff
@@ -140,6 +140,13 @@ spec:
                     type: object
                     description: Spec to control the desired behavior of storage cluster rolling update.
                     properties:
+                      minReadySeconds:
```
pure-jliao, line 143: do we make this term or there's a similar term on k8s side?

pkg/apis/core/v1/storagecluster.go
```diff
@@ -300,6 +300,12 @@ type RollingUpdateStorageCluster struct {
 	// that at least 70% of original number of StorageCluster pods are available at
 	// all times during the update.
 	MaxUnavailable *intstr.IntOrString `json:"maxUnavailable,omitempty"`
+
+	// Minimum number of seconds for which a newly created Portworx pod should be ready
+	// without any of its container crashing, for it to be considered available.
+	// Defaults to 0 (pod will be considered available as soon as it is ready)
+	// +optional
+	MinReadySeconds int32 `json:"minReadySeconds,omitempty"`
```
pure-jliao, line 308: could you run `make codegen` there should be deep copy function changes for this

pkg/controller/storagecluster/update.go
```diff
@@ -466,7 +466,12 @@ func (c *Controller) getUnavailableNumbers(
 			// for the purposes of update we ensure that the pod is
 			// both available and not terminating
 			if podutil.IsPodReady(pod) && pod.DeletionTimestamp == nil {
-				available = true
+				// Wait for MinReadySeconds after pod is ready.
+				if podutil.IsPodAvailable(pod, cluster.Spec.UpdateStrategy.RollingUpdate.MinReadySeconds, metav1.Now()) {
```
pure-jliao, line 470: we need to check the update strategy first, it can be `OnDelete` type and `RollingUpdate == nil`

deploy/crds/core_v1_storagecluster_crd.yaml
haibinxie: there is minReadySeconds in k8s deployment spec.

pkg/controller/storagecluster/update.go
haibinxie: this is already inside of rolling update. For other updates we don't check unavailable pods.

pkg/apis/core/v1/storagecluster.go
haibinxie: Good point, i have pushed a new commit.

pure-jliao (approved):
LGTM

PR #923, opened 2023-02-07
PWX-26956 handle telemetry cert owenrRef in storageCluster finalizer

haibinxie:
When storageCluster delete strategy is changed and then uninstalled, telemetry reconcile may not uninstall correctly as reconcile loop needs 30s to reconcile according to delete strategy. hence add the fix in finalizer of storagecluster.


pure-jliao (approved):
LGTM!

PR #925, opened 2023-02-08
PTX-15552 Fixing cluster creation if cluster already exists

nikolaypopov:


pure-jliao (approved):
LGTM

PR #934, opened 2023-02-16
PTX-15763 Fix MP operator upgrade logic

nikolaypopov:
Fix MP operator upgrade logic for unreleased dev images


test/integration_test/utils/px_operator.go
```diff
@@ -71,7 +71,18 @@ func getPXOperatorImageTag() (string, error) {
...
+					if env.Name == "OPERATOR_CONDITION_NAME" {
+						isMarketplaceDeployment = true
+					}
+				}
+				if isMarketplaceDeployment && strings.Contains(container.Image, "-dev") {
+					logrus.Infof("Looks like portworx-operator was installed via Openshift Marketplace, with unreleased image [%s]", container.Image)
+					tag = strings.Split(container.Image, ":")[1]
+					return tag, nil
+				}
+				logrus.Infof("Looks like portworx-operator was not installed via Openshift Marketplace, will use tag from the container image [%s]", container.Image)
```
pure-jliao, line 85: can we print the OPERATOR_CONDITION_NAME env value as well? will be easier to debug if the value is not expected.

test/integration_test/utils/marketplace.go
```diff
@@ -212,7 +210,8 @@ func DeploySubscription(name, namespace, opTag string, testCatalogSource *v1alph
 
 // ApproveInstallPlan finds correct install plan and approves it
 func ApproveInstallPlan(namespace, opTag string, subscription *v1alpha1.Subscription) error {
-	logrus.Infof("Looking for InstallPlan and setting Approved to true")
+	expectedClusterServiceVersion := fmt.Sprintf("portworx-operator.v%s", opTag)
```
pure-jliao, line 213: can we have `-dev` in `opTag` here?

pure-jliao (commented):
which part fixed the issue is not clear to me.. maybe have a quick sync tomorrow..

test/integration_test/utils/marketplace.go
nikolaypopov: No this won't have `-dev` postfix

test/integration_test/utils/px_operator.go
nikolaypopov: Done

pure-jliao (approved):
LGTM!

PR #935, opened 2023-02-16
[pwx-29718] Add leader elect resource args to pvc controller on all k8s versions

pure-jliao:
* Add pvc controller args to avoid conflict with other controller managers

Closes # pwx-29718


drivers/storage/portworx/component/pvccontroller.go
```diff
@@ -309,17 +309,14 @@ func (c *pvcController) createDeployment(
...
 	}
 
-	if c.k8sVersion.LessThan(k8sutil.K8sVer1_24) {
-		if pxutil.IsOpenshift(cluster) {
-			command = append(command, "--leader-elect-resource-lock=endpoints")
-		} else {
-			command = append(command, "--leader-elect-resource-lock=configmaps")
-		}
+	if pxutil.IsOpenshift(cluster) {
+		command = append(command, "--leader-elect-resource-lock=endpoints")
```
harsh-px, line 317: Is there benefit of using endpoints? Can't we always just use config maps now that we have the resource name fixed. I'm trying to keep the code simple and reduce branches.

drivers/storage/portworx/component/pvccontroller.go
```diff
@@ -309,17 +309,14 @@ func (c *pvcController) createDeployment(
...
+		fmt.Sprintf("--leader-elect-resource-namespace=%s", cluster.Namespace),
 	}
 
-	if c.k8sVersion.LessThan(k8sutil.K8sVer1_24) {
-		if pxutil.IsOpenshift(cluster) {
-			command = append(command, "--leader-elect-resource-lock=endpoints")
-		} else {
-			command = append(command, "--leader-elect-resource-lock=configmaps")
-		}
+	if pxutil.IsOpenshift(cluster) {
```
haibinxie, line 316: let's not use the check, just do this for all cluster.

		command = append(command, "--leader-elect-resource-name=portworx-pvc-controller")
		command = append(command, fmt.Sprintf("--leader-elect-resource-namespace=%s", cluster.Namespace))

drivers/storage/portworx/component/pvccontroller.go
```diff
@@ -309,17 +309,14 @@ func (c *pvcController) createDeployment(
 		"--leader-elect=true",
 		"--controllers=persistentvolume-binder,persistentvolume-expander",
 		"--use-service-account-credentials=true",
+		"--leader-elect-resource-name=portworx-pvc-controller",
+		fmt.Sprintf("--leader-elect-resource-namespace=%s", cluster.Namespace),
```
piyush-nimbalkar, line 313: I haven't followed all the threads on slack about this. But have we gone back and verified that older k8s versions also have these arguments? This would fail to come up in those versions if the args are absent.

drivers/storage/portworx/component/pvccontroller.go
piyush-nimbalkar: We should not use configmaps for new k8s versions. Those are deprecated in newer versions of k8s leader election. We should be using `leases`.
This is the reason we had the k8s version check because leases was added later and configmaps was the old way of doing leader election.

drivers/storage/portworx/component/pvccontroller.go
pure-jliao: we have bump the min supported k8s version to 1.22

haibinxie: i have checked all k8s versions 1.21+ have following two args. so let's just use the following args to use different resource name and not specify what type of leader election - leave to default. 

         command = append(command, "--leader-elect-resource-name=portworx-pvc-controller")
	command = append(command, fmt.Sprintf("--leader-elect-resource-namespace=%s", cluster.Namespace))

haibinxie (approved):

piyush-nimbalkar (approved):

PR #940, opened 2023-02-22
[PWX-26954] Enable telemetry by default on PX 2.12+

pure-jliao:
* Enable telemetry by default on PX 2.12+ (ccm-go) if custom proxy not specified or cluster is not air-gapped
* Update telemetry regression test

Closes #

telemetry regression test: telemetry is enabled by default https://jenkins.pwx.dev.purestorage.com/view/Newstack%20Dashboard/job/Operator/job/operator-integration-basic-master/207/console

### DIFF
--- a/drivers/storage/portworx/component/telemetry.go
+++ b/drivers/storage/portworx/component/telemetry.go
@@ -3,6 +3,9 @@ package component
 import (
 	"context"
 	"fmt"
+	"io"
+	"net/http"
+	"net/url"
 	"os"
 	"path"
 	"strings"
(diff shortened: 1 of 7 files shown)


drivers/storage/portworx/component/telemetry.go
```diff
@@ -1000,6 +1003,44 @@ func getArcusRegisterProxyURL(cluster *corev1.StorageCluster) string {
...
+			"appliance-id": {cluster.Status.ClusterUID},
+			"component-sn": {cluster.Name},
+		},
+	}
+	client := &http.Client{}
+	response, err := client.Do(request)
+	if err != nil {
+		logrus.WithError(err).Warnf("failed to access telemetry registration endpoint %s", endpoint)
+		return false, nil
+	} else if response.StatusCode != 200 {
```
haibinxie, line 1031: we should still enable telemetry as long as the network access passes. i am thinking about several scenarios this will not return success:
1. the version is changed from 1.0 to 2.0 in the url
2. internal server error
3. the format of the request header/body changes

haibinxie (approved):

drivers/storage/portworx/portworx.go
```diff
@@ -1287,6 +1286,46 @@ func setTLSDefaults(
...
+		return nil
+	}
+
+	// Only enable telemetry by default under conditions:
+	// 1. CCM Go is supported
+	// 2. Portworx initialized (cluster UUID ready)
+	// 3. telemetry is not specified explicitly
+	// 4. no air-gapped or custom proxy
+	if !pxutil.IsCCMGoSupported(pxVersion) ||
+		toUpdate.Status.ClusterUID == "" ||
```
Pure-AdamuKaapan, line 1307: Is this check run repeatedly/more than just at startup?

Pure-AdamuKaapan (commented):
I get the general idea, left some comments, will let someone more familiar approve though

drivers/storage/portworx/portworx.go
```diff
@@ -1287,6 +1286,46 @@ func setTLSDefaults(
 	}
 }
 
+func setTelemetryDefaults(
+	toUpdate *corev1.StorageCluster,
+	pxVersion *version.Version,
+) error {
+	// Disable telemetry if it's not supported but enabled explicitly
+	if pxutil.IsTelemetryEnabled(toUpdate.Spec) && pxVersion.LessThan(pxutil.MinimumPxVersionCCM) {
```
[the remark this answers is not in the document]
pure-jliao: yes, this block is checking if user is enabling telemetry but px version is too low. with the first part, we only disable it when user enables it, otherwise this telemetry section should be nil and won't show in the stc spec.

drivers/storage/portworx/portworx.go
pure-jliao: it is being checked repeatedly. operator only start telemetry reconcile once px is running, so something similar here.

nikolaypopov (approved):
LGTM

PR #943, opened 2023-02-23
Allow Prometheus customization of replicas,retention,PVC

siyingjin:
Allow Prometheus customization of 
* replicas
* retention, retentionSize
* PVC, volumeMounts, volumes

* [Ticket](https://portworx.atlassian.net/browse/PWX-24897)
* Previous [PR](https://github.com/libopenstorage/operator/pull/831) to customize Resources and SecurityContext

Test:

1. In StorageCluster spec, specify replicas, retention, retentionSize, storage, volumeMounts, volumes

Screen Shot 2023-02-22 at 7 16 49 PM

2. Wait for a few seconds, delete px-operator pod
3. Verify prometheus spec has been updated, and 2 prometheus pods running
Screen Shot 2023-02-22 at 7 21 50 PM

4. Describe prometheus pod
Screen Shot 2023-02-22 at 7 27 14 PM
Screen Shot 2023-02-22 at 7 32 53 PM
Screen Shot 2023-02-22 at 7 33 03 PM


deploy/crds/core_v1_storagecluster_crd.yaml
```diff
@@ -1146,6 +1146,2272 @@ spec:
                             type: boolean
                             description: Flag indicating whether AlertManager needs to be enabled and deployed
                               by the Storage operator.
+                      replicas:
```
pure-jliao, line 1149: is it necessary to have all of those specs under prometheus? did we confirm that?

drivers/storage/portworx/component/prometheus.go
```diff
@@ -656,9 +657,37 @@ func (c *prometheus) createPrometheusInstance(
...
+
+	if cluster.Spec.Monitoring.Prometheus.Retention != "" {
+		prometheusInst.Spec.Retention = cluster.Spec.Monitoring.Prometheus.Retention
+	}
+
+	if cluster.Spec.Monitoring.Prometheus.RetentionSize != "" {
+		prometheusInst.Spec.RetentionSize = cluster.Spec.Monitoring.Prometheus.RetentionSize
+	}
+
+	if cluster.Spec.Monitoring.Prometheus.Storage != nil {
```
pure-jliao, line 679: we should do some sort of validation here, e.g. if storage is specified but not volume/mount, should we raise an error? or let the prometheus operator to handle everything?

pkg/apis/core/v1/storagecluster.go
```diff
@@ -583,6 +584,23 @@ type PrometheusSpec struct {
...
+	// specified, then by default an EmptyDir will be used. If multiple storage options are specified,
+	// priority will be given as follows: EmptyDir, Ephemeral, and lastly VolumeClaimTemplate.
+	Storage *monitoringv1.StorageSpec `json:"storage,omitempty"`
+	// Volumes allows configuration of additional volumes on the output StatefulSet definition. Volumes specified will
+	// be appended to other volumes that are generated as a result of StorageSpec objects.
+	Volumes []v1.Volume `json:"volumes,omitempty"`
+	// VolumeMounts allows configuration of additional VolumeMounts on the output StatefulSet definition.
+	// VolumeMounts specified will be appended to other VolumeMounts in the prometheus container,
+	// that are generated as a result of StorageSpec objects.
+	VolumeMounts []v1.VolumeMount `json:"volumeMounts,omitempty"`
```
pure-jliao, line 603: do a `make codegen` locally to generate deep equal util functions for this crd.

pure-jliao (commented):
let's put the ticket number in commit message and PR title, so jira can pick those up.

drivers/storage/portworx/components_test.go
```diff
@@ -7408,6 +7408,54 @@ func TestPrometheusInstall(t *testing.T) {
...
+			Name: "testVlm",
+			VolumeSource: v1.VolumeSource{
+				PersistentVolumeClaim: &v1.PersistentVolumeClaimVolumeSource{
+					ClaimName: "testPxPvc",
+					ReadOnly:  false,
+				},
+			},
+		},
+	}
+	cluster.Spec.Monitoring.Prometheus.SecurityContext = nil
```
Pure-AdamuKaapan, line 7447: Should we validate that SecurityContext is nil in the created one as well? Or will it get filled in with defaults?

Pure-AdamuKaapan (commented):
Overall makes sense to me, left a couple minor comments. I'm still learning some of this code so I'll leave approving to Jiafeng or someone else.

deploy/crds/core_v1_storagecluster_crd.yaml
siyingjin: A lot of lines are from ```storage``` and ```volumes``` spec. I confirmed with @haibinxie we'll keep everything under ```storage``` to allow more customization options.

drivers/storage/portworx/component/prometheus.go
siyingjin: If the client doesn't specify ```storage```, prometheus is going to use default value (i.e. no storage/volumeMounts). Later we can add fallback storage if explicitly required.

drivers/storage/portworx/components_test.go
siyingjin: SecurityContext was set to some specific values in a previous unit test. So I set SecurityContext to nil so that the [fallback](https://github.com/libopenstorage/operator/blob/da2dfcbbd7ca1cf6a4e9d8a073ac113e614c7ce2/drivers/storage/portworx/component/prometheus.go#L657) logic will set ```runAsGroup``` and  ```fsGroup```, which is required for opening query file (I just added the error message in the PR description before we adding ```runAsGroup``` and ```fsGroup```)

pkg/apis/core/v1/storagecluster.go
siyingjin: Added codegen change.

drivers/storage/portworx/component/prometheus.go
pure-jliao: I mean when `storage` is specified, is volume and mount required as well? assume if user has storage and volume specified but not mount, what would happen in that case?

pkg/apis/core/v1/zz_generated.deepcopy.go
```diff
@@ -729,6 +730,30 @@ func (in *PrometheusSpec) DeepCopyInto(out *PrometheusSpec) {
 		*out = new(corev1.PodSecurityContext)
 		(*in).DeepCopyInto(*out)
 	}
+	if in.Replicas != nil {
```
haibinxie, line 733: Great this is handled

haibinxie (approved):
LGTM, please note to add functional tests prior to release. and after we have the release branch this should be cherry-picked to release branch.

drivers/storage/portworx/component/prometheus.go
siyingjin: Volume and Volumemount are not required. 
If we use automatic provisioning of storage by specifying the storageclass in ```storage```(StorageSpec) section. When creating the Prometheus object, a PersistentVolumeClaim is used for each Pod in the StatefulSet, and the storage should automatically be provisioned, mounted and used.
Volumes allows configuration of additional volumes on the output StatefulSet definition. Volumes specified will be appended to other volumes that are generated as a result of StorageSpec objects.
In my test, volume ```prometheus-px-prometheus-db``` is provisioned with storageclass ```px-repl3-sc```. Volume ```testvlm``` is additional volume configured through ```Volumes```.

pure-jliao (approved):
LGTM! great job!

PR #946, opened 2023-02-23
PWX-29351: migrate from k8s.gcr.io to registry.k8s.io

haibinxie:

Closes #


drivers/storage/portworx/util/csi_generator.go
```diff
@@ -260,9 +260,9 @@ func (c *CSIConfiguration) DriverBasePath() string {
 }
 
 func (g *CSIGenerator) getSidecarContainerVersionsV1_0() *CSIImages {
-	provisionerImage := "k8s.gcr.io/sig-storage/csi-provisioner:v3.3.0"
-	snapshotterImage := "k8s.gcr.io/sig-storage/csi-snapshotter:v6.1.0"
-	snapshotControllerImage := "k8s.gcr.io/sig-storage/snapshot-controller:v6.1.0"
+	provisionerImage := "registry.k8s.io/sig-storage/csi-provisioner:v3.3.0"
```
Pure-AdamuKaapan, line 263: Should this use the same constant as above? Replace with a `fmt.Sprintf(...`?

drivers/storage/portworx/util/csi_generator.go
```diff
@@ -272,8 +272,8 @@ func (g *CSIGenerator) getSidecarContainerVersionsV1_0() *CSIImages {
 	// For k8s 1.19 and earlier, use older versions
 	if g.kubeVersion.LessThan(k8sVer1_20) {
 		provisionerImage = "docker.io/openstorage/csi-provisioner:v2.2.2-1"
-		snapshotterImage = "k8s.gcr.io/sig-storage/csi-snapshotter:v3.0.3"
-		snapshotControllerImage = "k8s.gcr.io/sig-storage/snapshot-controller:v3.0.3"
+		snapshotterImage = "registry.k8s.io/sig-storage/csi-snapshotter:v3.0.3"
```
Pure-AdamuKaapan, line 275: Same here, use a constant?

drivers/storage/portworx/util/csi_generator.go
```diff
@@ -284,12 +284,12 @@ func (g *CSIGenerator) getSidecarContainerVersionsV1_0() *CSIImages {
 
 	return &CSIImages{
 		Attacher:                "docker.io/openstorage/csi-attacher:v1.2.1-1",
-		NodeRegistrar:           "k8s.gcr.io/sig-storage/csi-node-driver-registrar:v2.6.2",
+		NodeRegistrar:           "registry.k8s.io/sig-storage/csi-node-driver-registrar:v2.6.2",
```
Pure-AdamuKaapan, line 287: And here

drivers/storage/portworx/util/util.go
```diff
@@ -213,7 +213,7 @@ const (
 	// ErrMsgGrpcConnection error message if failed to connect to GRPC server
 	ErrMsgGrpcConnection = "error connecting to GRPC server"
 	// ImageNamePause is the container image to use for the pause container
-	ImageNamePause = "k8s.gcr.io/pause:3.1"
+	ImageNamePause = "registry.k8s.io/pause:3.1"
```
Pure-AdamuKaapan, line 216: Use constant?

Pure-AdamuKaapan (commented):
A couple places to change. Thanks much for doing this!

drivers/storage/portworx/util/csi_generator.go
haibinxie: sure, i can make the optimization, was just doing a string replacement in the code.

pure-jliao (approved):
LGTM, only question is for custom registry, why do we keep `k8s.gcr.io` in the path but not this new one.

haibinxie:
> LGTM, only question is for custom registry, why do we keep `k8s.gcr.io` in the path but not this new one.

k8s.gcr.io in the path was a bug, after operator was released we cannot change it anymore due to customer already using it on airgapped env. for the new path, we will change the airgapped script and operator to both not having it in path.

PR #954, opened 2023-03-01
PTX-16142 Change fmt.Printf() to logrus.Infof() in ValidateTelemetry() for PX version

nikolaypopov:
Change fmt.Printf() to logrus.Infof() in ValidateTelemetry() for PX version

### DIFF
--- a/pkg/util/test/util.go
+++ b/pkg/util/test/util.go
@@ -2862,12 +2862,12 @@ func ValidateTelemetry(pxImageList map[string]string, cluster *corev1.StorageClu
 	logrus.Info("Validate Telemetry components")
 	logrus.Info("Check PX and PX Operator versions to determine which telemetry to validate against..")
 	pxVersion := GetPortworxVersion(cluster)
-	fmt.Printf("PX Version: %v", pxVersion)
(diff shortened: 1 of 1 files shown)


pkg/util/test/util.go
```diff
@@ -2862,12 +2862,12 @@ func ValidateTelemetry(pxImageList map[string]string, cluster *corev1.StorageClu
 	logrus.Info("Validate Telemetry components")
 	logrus.Info("Check PX and PX Operator versions to determine which telemetry to validate against..")
 	pxVersion := GetPortworxVersion(cluster)
-	fmt.Printf("PX Version: %v", pxVersion)
+	logrus.Infof("PX Version: [%s]", pxVersion.String())
```
pure-jliao, line 2865: nit: pxVersion should be good without String()

pure-jliao (approved):

pkg/util/test/util.go
nikolaypopov: Both `GetPortworxVersion()` and `GetPxOperatorVersion()` return `*version.Version`, I figured I'd do same for both

PR #958, opened 2023-03-02
[pwx-26954] Disable telemetry if it's enabled with custom proxy

pure-jliao:
* Fresh install, telemetry will be enabled if
    * PX is 2.12+, so ccm-go is supported
    * telemetry is not disabled explicitly in the spec
    * no PX_HTTP_PROXY or PX_HTTPS_PROXY is specified in stc spec
    * no air-gapped, operator can reach to the registration endpoint
* Upgrade case:
    * if fresh install conditions above are met during upgrade, telemetry will be enabled by default
    * if already using telemetry enabled with proxy:
        * using PX_HTTP_PROXY in format of host:port or http://host:port, telemetry will be still enabled. this is the basic http proxy we have verified already, can be turn off at anytime
        * using http proxy but with auth, e.g. http://user:pass@host:port , telemetry will be disabled as this format is not supported yet
        * using PX_HTTPS_PROXY , telemetry will be disabled with any secure proxy. If http endpoint if provided, will be considered as PX_HTTP_PROXY

Closes # PWX-26954

### DIFF
--- a/drivers/storage/portworx/util/util.go
+++ b/drivers/storage/portworx/util/util.go
@@ -637,16 +637,26 @@ func GetClusterEnvVarValue(ctx context.Context, cluster *corev1.StorageCluster, 
 
 // GetPxProxyEnvVarValue returns the PX_HTTP(S)_PROXY environment variable value for a cluster.
 // Note: we only expect one proxy for the telemetry CCM container but we prefer https over http if both are specified
-func GetPxProxyEnvVarValue(cluster *corev1.StorageCluster) string {
+func GetPxProxyEnvVarValue(cluster *corev1.StorageCluster) (string, string) {
 	httpProxy := ""
 	for _, env := range cluster.Spec.Env {
-		if env.Name == EnvKeyPortworxHTTPSProxy {
-			return env.Value
-		} else if env.Name == EnvKeyPortworxHTTPProxy {
-			httpProxy = env.Value
+		key, val := env.Name, env.Value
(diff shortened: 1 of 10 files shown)


drivers/storage/portworx/util/util.go
```diff
@@ -1100,6 +1100,29 @@ func IsCCMGoSupported(pxVersion *version.Version) bool {
...
+	if !IsTelemetryEnabled(cluster.Spec) {
+		return nil
+	}
+
+	pxVersion := GetPortworxVersion(cluster)
+	if pxVersion.LessThan(MinimumPxVersionCCM) {
+		// PX version is lower than 2.8
+		return fmt.Errorf("telemetry is not supported on Portworx version: %s", pxVersion)
+	} else if IsCCMGoSupported(pxVersion) {
+		if proxy := GetPxProxyEnvVarValue(cluster); proxy != "" {
```
haibinxie, line 1114: I thought we will just disable it when https and/or username password are specified in the URL. we should not disable it if http is used.

pure-jliao: you're right, i missed the https part in the comment. for http we should not disable it.

haibinxie: actually I had a discussion with Bhavana and Prashant, they said we should also disable telemetry with http proxy too. could you confirm with Harsh?

haibinxie (approved):

PR #960, opened 2023-03-03
Catalog for 1.10.5 release

haibinxie:
Makefile is already 1.10.4 so no need to change

### DIFF
--- a/deploy/olm-catalog/portworx/1.10.5/portworx-certified.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/portworx/1.10.5/portworx-certified.clusterserviceversion.yaml
@@ -1,0 +1,540 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: ClusterServiceVersion
+metadata:
+  name: portworx-operator.v1.10.4
+  namespace: placeholder
+  annotations:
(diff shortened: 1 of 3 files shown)


deploy/olm-catalog/portworx/1.10.5/portworx-certified.clusterserviceversion.yaml
```diff
@@ -0,0 +1,540 @@
...
+          },
+          "spec": {}
+        }
+      ]
+spec:
+  displayName: Portworx Enterprise
+  version: 1.10.5
+  minKubeVersion: "1.12.0"
+  maturity: stable
+  replaces: portworx-operator.v1.10.5
```
pure-jliao, line 55: 1.10.4

pure-jliao (changes requested):

deploy/olm-catalog/portworx/1.10.5/portworx-certified.clusterserviceversion.yaml
haibinxie: fixed. thanks for catching this, i will create ticket to automate this

pure-jliao (approved):

PR #963, opened 2023-03-03
change previous version to 1.10.5 and copy 1.10.5 catalog

haibinxie:

Closes #


pure-jliao (commented):
we should merge this after 1.10.5 is released, otherwise don't know if build catalog would fail.

PR #967, opened 2023-03-08
[PWX-29531] Fix SecurityContextConstraints to be compatible with open…

siyingjin:
…shift

Add px-prometheus to portworx scc (which uses runAsAny) to bypass the MustRunAsRange restrictions of runAsUser and runAsGroup.

Ticket: https://portworx.atlassian.net/browse/PWX-29531

**Test**:
Tested with portworx fresh installation on openshift.
Tested with upgrading portworx operator with SCC change and verify openshift SCC is gone.
Tested tuning prometheus spec to add PVC.

### DIFF
--- a/drivers/storage/portworx/component/securitycontextconstraints.go
+++ b/drivers/storage/portworx/component/securitycontextconstraints.go
@@ -3,7 +3,6 @@ package component
 import (
 	"context"
 	"fmt"
-
 	"github.com/hashicorp/go-version"
 	ocp_secv1 "github.com/openshift/api/security/v1"
 	"github.com/sirupsen/logrus"
@@ -14,6 +13,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/utils/pointer"
+	"reflect"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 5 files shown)


drivers/storage/portworx/component/securitycontextconstraints.go
```diff
@@ -85,18 +85,28 @@ func (s *scc) Reconcile(cluster *opcorev1.StorageCluster) error {
...
 			err = s.k8sClient.Create(context.TODO(), &scc)
-		}
-
-		if err != nil {
-			return fmt.Errorf("failed to create %s security context constraints: %s", scc.Name, err)
+			if err != nil {
+				return fmt.Errorf("failed to create %s security context constraints: %s", scc.Name, err)
+			}
+		} else if err == nil {
+			if !reflect.DeepEqual(scc, existingScc) {
```
haibinxie, line 101: actually this will never be equal, i just checked on my cluster, the scc has these fields which are changed after it;s deployed:

metadata:
  creationTimestamp: "2023-03-07T01:30:11Z"
  generation: 1
  name: portworx
  resourceVersion: "6870447"
  uid: 4feeeb20-9d29-4c7e-b1ad-aac7af8ab883

Could you make this API as a public API in util, and then call this API to cleanup the object before comparing https://github.com/libopenstorage/operator/blob/master/pkg/dryrun/dryrun.go#L425

also write a log before updating object and verify that it's only updated once.

siyingjin: Resolved in rev-2

drivers/storage/portworx/component/securitycontextconstraints.go
```diff
@@ -85,18 +85,32 @@ func (s *scc) Reconcile(cluster *opcorev1.StorageCluster) error {
...
-		if err != nil {
-			return fmt.Errorf("failed to create %s security context constraints: %s", scc.Name, err)
+			if err != nil {
+				return fmt.Errorf("failed to create %s security context constraints: %s", scc.Name, err)
+			}
+		} else if err == nil {
+			resourceVersion := existingScc.ResourceVersion
+			pxutil.CleanupObject(&scc)
+			pxutil.CleanupObject(&existingScc)
+			logrus.Infof("Updating %s security context constraints", scc.Name)
```
haibinxie, line 104: looks good, you can merge the PR after verifying this only gets updated once.

haibinxie (approved):

PR #972, opened 2023-03-09
PWX-29609 PWX-26914 - do not pass resource limit to runc

haibinxie:

Closes #


pure-jliao (approved):
LGTM, we're still passing this to the container spec right?

haibinxie:
> LGTM, we're still passing this to the container spec right?

correct.

PR #974, opened 2023-03-10
[pwx-29114] Update default release manifest images

pure-jliao:

Closes #

### DIFF
--- a/drivers/storage/portworx/manifest/manifest.go
+++ b/drivers/storage/portworx/manifest/manifest.go
@@ -24,11 +24,11 @@ const (
 	envKeyReleaseManifestRefreshInterval = "PX_RELEASE_MANIFEST_REFRESH_INTERVAL_MINS"
 	// DefaultPortworxVersion is the default portworx version that will be used
 	// if none specified and if version manifest could not be fetched
-	DefaultPortworxVersion     = "2.9.1.4"
-	defaultStorkImage          = "openstorage/stork:2.11.0"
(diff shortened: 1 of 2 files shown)


drivers/storage/portworx/manifest/manifest.go
```diff
@@ -24,11 +24,11 @@ const (
 	envKeyReleaseManifestRefreshInterval = "PX_RELEASE_MANIFEST_REFRESH_INTERVAL_MINS"
 	// DefaultPortworxVersion is the default portworx version that will be used
 	// if none specified and if version manifest could not be fetched
-	DefaultPortworxVersion     = "2.9.1.4"
-	defaultStorkImage          = "openstorage/stork:2.11.0"
-	defaultAutopilotImage      = "portworx/autopilot:1.3.2.1"
+	DefaultPortworxVersion     = "2.13.0"
+	defaultStorkImage          = "openstorage/stork:2.12.4"
+	defaultAutopilotImage      = "portworx/autopilot:1.3.7"
 	defaultLighthouseImage     = "portworx/px-lighthouse:2.0.7"
```
nikolaypopov, line 30: Should we remove this, as we don't even deploy this using operator?

pure-jliao: let's not change it here. if it's enabled and there's no image at all then there's a problem.

haibinxie (approved):

PR #976, opened 2023-03-10
PWX-29565 set default scc priority to nil to unblock ocp 4.12 upgrade…

haibinxie:
…, add annotation to set priority

Closes #

### DIFF
--- a/drivers/storage/portworx/testspec/portworxSCC.yaml
+++ b/drivers/storage/portworx/testspec/portworxSCC.yaml
@@ -17,7 +17,6 @@ kind: SecurityContextConstraints
 metadata:
   name: portworx
   resourceVersion: "1"
-priority: 2
 readOnlyRootFilesystem: false
 requiredDropCapabilities: null
 runAsUser:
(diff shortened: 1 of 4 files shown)


drivers/storage/portworx/testspec/portworxSCC.yaml
```diff
@@ -17,7 +17,6 @@ kind: SecurityContextConstraints
 metadata:
   name: portworx
   resourceVersion: "1"
-priority: 2
```
pure-jliao, line 20: the `-` is missing

pure-jliao (approved):
LGTM after addressing the comments, and can pass CI

drivers/storage/portworx/testspec/portworxSCC.yaml
haibinxie: what do you mean? - means this line is removed

pure-jliao: Ohh nvm I thought it's a yaml list...

PR #977, opened 2023-03-13
Fix SCC update

siyingjin:
Before this change, SCC will be updated on every reconciliation due to deepEqual evaluation to false on empty field/pointers.
After the change, SCC will only be updated once on updating operator. 

Ticket: https://portworx.atlassian.net/browse/PWX-29531
Previous PR: https://github.com/libopenstorage/operator/pull/967

**Test**:
Previous portworx scc: users field doesn't contain px-prometheus service account
After pulling the operator change, 
Screen Shot 2023-03-13 at 12 46 24 PM
Confirmed 1. scc is updated only 1 time. 2.portworx scc contains px-prometheus service account. 3. px-prometheus pod up and running

### DIFF
--- a/drivers/storage/portworx/component/securitycontextconstraints.go
+++ b/drivers/storage/portworx/component/securitycontextconstraints.go
@@ -81,6 +81,35 @@ func (s *scc) crdExists(cluster *opcorev1.StorageCluster) bool {
 	}
 }
 
+func deepEqualScc(scc, existingScc *ocp_secv1.SecurityContextConstraints) bool {
+	if !(scc.Priority == nil && existingScc.Priority == nil ||
+		scc.Priority != nil && existingScc.Priority != nil && *scc.Priority == *existingScc.Priority) ||
+		!(scc.AllowPrivilegeEscalation == nil && existingScc.AllowPrivilegeEscalation == nil ||
+			scc.AllowPrivilegeEscalation != nil && existingScc.AllowPrivilegeEscalation != nil &&
+				*scc.AllowPrivilegeEscalation == *existingScc.AllowPrivilegeEscalation) {
+		return false
+	}
+
+	// Skip comparing auto-generated fields
+	sccCopy := scc.DeepCopy()
+	sccCopy.ResourceVersion = existingScc.ResourceVersion
+	sccCopy.ObjectMeta.UID = existingScc.ObjectMeta.UID
+	sccCopy.ObjectMeta.Generation = existingScc.ObjectMeta.Generation
+	sccCopy.ObjectMeta.CreationTimestamp = existingScc.ObjectMeta.CreationTimestamp
+	sccCopy.ObjectMeta.ManagedFields = existingScc.ObjectMeta.ManagedFields
+	sccCopy.ObjectMeta.GenerateName = existingScc.ObjectMeta.GenerateName
+	sccCopy.ObjectMeta.Finalizers = existingScc.ObjectMeta.Finalizers
+	sccCopy.ObjectMeta.OwnerReferences = existingScc.ObjectMeta.OwnerReferences
+	sccCopy.TypeMeta.Kind = existingScc.TypeMeta.Kind
+	sccCopy.TypeMeta.APIVersion = existingScc.TypeMeta.APIVersion
+	sccCopy.Priority = existingScc.Priority
+	sccCopy.AllowPrivilegeEscalation = existingScc.AllowPrivilegeEscalation
+	if scc.Groups == nil {
+		sccCopy.Groups = make([]string, 0)
+	}
+	return reflect.DeepEqual(sccCopy, existingScc)
+}
+
 func (s *scc) Reconcile(cluster *opcorev1.StorageCluster) error {
 	// Note SCC does not belong to namespace hence we don't set owner reference to StorageCluster.
 	// By design cross-namespace owner reference is invalid.
@@ -98,12 +127,9 @@ func (s *scc) Reconcile(cluster *opcorev1.StorageCluster) error {
 				return fmt.Errorf("failed to create %s security context constraints: %s", scc.Name, err)
 			}
 		} else if err == nil {
-			resourceVersion := existingScc.ResourceVersion
-			pxutil.CleanupObject(&scc)
-			pxutil.CleanupObject(&existingScc)
-			logrus.Infof("Updating %s security context constraints", scc.Name)
-			if !reflect.DeepEqual(scc, existingScc) {
-				scc.ResourceVersion = resourceVersion
+			if !deepEqualScc(&scc, &existingScc) {
+				scc.ResourceVersion = existingScc.ResourceVersion
+				logrus.Infof("Updating %s security context constraints", scc.Name)
 				err = s.k8sClient.Update(context.TODO(), &scc)
 				if err != nil {
 					return fmt.Errorf("failed to update %s security context constraints: %s", scc.Name, err)


drivers/storage/portworx/component/securitycontextconstraints.go
```diff
@@ -99,12 +105,20 @@ func (s *scc) Reconcile(cluster *opcorev1.StorageCluster) error {
...
-			if !reflect.DeepEqual(scc, existingScc) {
+			sccPriority := *scc.Priority
+			existingSccPriority := *existingScc.Priority
+			sccAllowPrivilegeEscalation := *scc.AllowPrivilegeEscalation
+			existingSccAllowPrivilegeEscalation := *existingScc.AllowPrivilegeEscalation
+			cleanupSccObject(&scc)
+			cleanupSccObject(&existingScc)
+			if sccPriority != existingSccPriority ||
+				sccAllowPrivilegeEscalation != existingSccAllowPrivilegeEscalation ||
+				!equality.Semantic.DeepDerivative(scc, existingScc) {
```
haibinxie, line 116: could you change this to deepEqual. deepDerivative will not compare values that are nil in first parameter, it is not a good fit for this case as we care comparing the object entirely.

drivers/storage/portworx/component/securitycontextconstraints.go
```diff
@@ -99,12 +105,20 @@ func (s *scc) Reconcile(cluster *opcorev1.StorageCluster) error {
...
 				scc.ResourceVersion = resourceVersion
+				scc.Priority = &sccPriority
 				err = s.k8sClient.Update(context.TODO(), &scc)
+				scc.AllowPrivilegeEscalation = &sccAllowPrivilegeEscalation
+				logrus.Infof("Updating %s security context constraints", scc.Name)
```
haibinxie, line 121: move this to one line before the k8sClient.Update API call

drivers/storage/portworx/component/securitycontextconstraints.go
```diff
@@ -99,11 +113,16 @@ func (s *scc) Reconcile(cluster *opcorev1.StorageCluster) error {
 			}
 		} else if err == nil {
 			resourceVersion := existingScc.ResourceVersion
-			pxutil.CleanupObject(&scc)
```
pure-jliao, line 102: I have a feeling we don't need to do a obj cleanup, do we? If we have a proper customized deep equal function, and setting the latest resource version, can't we update the priority and AllowPrivilegeEscalation directly?

siyingjin: existingScc has some system generated artifacts like timestamps, which are not in scc, but for comparison, we want to ignore these fields. Are you suggesting to skip comparing those fields like timestamps

drivers/storage/portworx/component/securitycontextconstraints.go
```diff
@@ -81,6 +81,20 @@ func (s *scc) crdExists(cluster *opcorev1.StorageCluster) bool {
 	}
 }
 
+func cleanupSccObject(constraints *ocp_secv1.SecurityContextConstraints) {
+	pxutil.CleanupObject(constraints)
+	constraints.Priority = nil
+	constraints.AllowPrivilegeEscalation = nil
+}
+
+func sccValuesEqual(scc *ocp_secv1.SecurityContextConstraints, existingScc *ocp_secv1.SecurityContextConstraints) bool {
```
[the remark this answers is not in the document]
siyingjin: Are there any places we want to use this function outside of securitycontextconstraints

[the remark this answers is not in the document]
pure-jliao: not really, we can keep it in this file but make the function similar, so in the future we can move it to utils when needed.

drivers/storage/portworx/component/securitycontextconstraints.go
pure-jliao: yes, we are using our customized deep equal function anyway, we can do the comparison without those fields, so we don't need to reset anything.

drivers/storage/portworx/component/securitycontextconstraints.go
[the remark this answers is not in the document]
siyingjin: makes sense to me, will update that

drivers/storage/portworx/component/securitycontextconstraints.go
siyingjin: Make sense to me.

drivers/storage/portworx/component/securitycontextconstraints.go
siyingjin: Yes, I'll change that

drivers/storage/portworx/component/securitycontextconstraints.go
siyingjin: Will update

drivers/storage/portworx/component/securitycontextconstraints.go
```diff
@@ -81,6 +81,20 @@ func (s *scc) crdExists(cluster *opcorev1.StorageCluster) bool {
 	}
 }
 
+func cleanupSccObject(constraints *ocp_secv1.SecurityContextConstraints) {
+	pxutil.CleanupObject(constraints)
+	constraints.Priority = nil
+	constraints.AllowPrivilegeEscalation = nil
```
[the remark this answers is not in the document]
siyingjin: As discussed in the other conversation, we will skip comparing without cleanup

pure-jliao (approved):
LGTM!

drivers/storage/portworx/component/securitycontextconstraints.go
```diff
@@ -81,6 +81,35 @@ func (s *scc) crdExists(cluster *opcorev1.StorageCluster) bool {
...
+	scc.ObjectMeta.ManagedFields = existingScc.ObjectMeta.ManagedFields
+	scc.ObjectMeta.GenerateName = existingScc.ObjectMeta.GenerateName
+	scc.ObjectMeta.Finalizers = existingScc.ObjectMeta.Finalizers
+	scc.ObjectMeta.OwnerReferences = existingScc.ObjectMeta.OwnerReferences
+	scc.TypeMeta.Kind = existingScc.TypeMeta.Kind
+	scc.TypeMeta.APIVersion = existingScc.TypeMeta.APIVersion
+	scc.Priority = existingScc.Priority
+	scc.AllowPrivilegeEscalation = existingScc.AllowPrivilegeEscalation
+	if scc.Groups == nil {
+		// Update will override nil Groups
```
haibinxie, line 107: this may cause issue if we update groups from non-empty (and non-default) to nil, what's the value that nil is overwritten to, we may need special handling for default value

drivers/storage/portworx/component/securitycontextconstraints.go
```diff
@@ -81,6 +81,35 @@ func (s *scc) crdExists(cluster *opcorev1.StorageCluster) bool {
...
+	if !(scc.Priority == nil && existingScc.Priority == nil ||
+		scc.Priority != nil && existingScc.Priority != nil && *scc.Priority == *existingScc.Priority) ||
+		!(scc.AllowPrivilegeEscalation == nil && existingScc.AllowPrivilegeEscalation == nil ||
+			scc.AllowPrivilegeEscalation != nil && existingScc.AllowPrivilegeEscalation != nil &&
+				*scc.AllowPrivilegeEscalation == *existingScc.AllowPrivilegeEscalation) {
+		return false
+	}
+
+	// Overwrite auto-generated fields to existing scc value to bypass DeepEqual check
+	scc.ObjectMeta.UID = existingScc.ObjectMeta.UID
```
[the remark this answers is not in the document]
siyingjin: Will make a deep copy.

drivers/storage/portworx/component/securitycontextconstraints.go
```diff
@@ -81,6 +81,35 @@ func (s *scc) crdExists(cluster *opcorev1.StorageCluster) bool {
 	}
 }
 
+func deepEqualSccTemplate(scc, existingScc *ocp_secv1.SecurityContextConstraints) bool {
+	scc.ResourceVersion = existingScc.ResourceVersion
```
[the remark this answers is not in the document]
siyingjin: Fixed

drivers/storage/portworx/component/securitycontextconstraints.go
siyingjin: Will use empty slice

drivers/storage/portworx/component/securitycontextconstraints.go
```diff
@@ -81,6 +81,35 @@ func (s *scc) crdExists(cluster *opcorev1.StorageCluster) bool {
...
+	sccCopy.ObjectMeta.ManagedFields = existingScc.ObjectMeta.ManagedFields
+	sccCopy.ObjectMeta.GenerateName = existingScc.ObjectMeta.GenerateName
+	sccCopy.ObjectMeta.Finalizers = existingScc.ObjectMeta.Finalizers
+	sccCopy.ObjectMeta.OwnerReferences = existingScc.ObjectMeta.OwnerReferences
+	sccCopy.TypeMeta.Kind = existingScc.TypeMeta.Kind
+	sccCopy.TypeMeta.APIVersion = existingScc.TypeMeta.APIVersion
+	sccCopy.Priority = existingScc.Priority
+	sccCopy.AllowPrivilegeEscalation = existingScc.AllowPrivilegeEscalation
+	if scc.Groups == nil {
+		sccCopy.Groups = make([]string, 0)
```
pure-jliao, line 108: a little bit confused, what was the change here before? the `Groups` field can be nil from existing scc, and deep equal comparing empty slice and nil will be false.

siyingjin: scc can be nil, but existing scc is auto converted to empty slice, so I set sccCopy's value to empty slice.

siyingjin:
Tested on ocp, update will only be executed once.

haibinxie (approved):
LGTM

PR #981, opened 2023-03-14
PWX-26175-r2: Fixing license extension tests

zoxpx:
Adding cleanup of test volume/snapshot

The tests include decommission of the px-nodes
* however, the test will sporadically fail, if the volume/snapshot has been created on the node we're attempting to decommission
* as a fix, we'll clean up the test volume/snapshot, so decommission should not complain about "destroying unique volumes"

Closes #  PWX-26175  (part 2)


zoxpx:
Looks like something went wrong with this PR  (it is showing too many files as modified) -- will close this PR, and reopen a new one

PR #983, opened 2023-03-14
PWX-28335: define securityContext in container spec since podsecurity…

ezhang-px:
…policy was not supported

PodSecurityPolicy (PSP) is deprecated from K8s 1.21 and starting from 1.25, it's even removed from k8s. And hence we need to clean up the spec generator for autopilot inside operator, like removing the non-existing PSP definition and invalid attachment to autopilot's clusterrole spec, and also replacing the previous policies with securityContext items to the container level. 
Previously we attach policy like _allowPrivilegeEscalation: false_ to autopilot, therefore this is maintained in current securityContext definition. Also in [autopilot deployment manifest](https://github.com/portworx/autopilot/blob/master/manifests/autopilot-deployment.yaml#L83), we have _privileged: false_ defined for securityContext, introduce from operator here as well.

P.S.: an example of the PSP attached on autopilot's clusterrole when running PX 2.12 in older version of k8s (1.20)

`apiVersion: policy/v1beta1
kind: PodSecurityPolicy
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"policy/v1beta1","kind":"PodSecurityPolicy","metadata":{"annotations":{},"name":"px-operator"},"spec":{"allowPrivilegeEscalation":false,"fsGroup":{"rule":"RunAsAny"},"readOnlyRootFilesystem":false,"runAsUser":{"rule":"RunAsAny"},"seLinux":{"rule":"RunAsAny"},"supplementalGroups":{"rule":"RunAsAny"},"volumes":["secret"]}}
  creationTimestamp: "2023-03-10T19:51:14Z"
  managedFields:
  - apiVersion: policy/v1beta1
    fieldsType: FieldsV1
    fieldsV1:
      f:metadata:
        f:annotations:
          .: {}
          f:kubectl.kubernetes.io/last-applied-configuration: {}
      f:spec:
        f:allowPrivilegeEscalation: {}
        f:fsGroup:
          f:rule: {}
        f:runAsUser:
          f:rule: {}
        f:seLinux:
          f:rule: {}
        f:supplementalGroups:
          f:rule: {}
        f:volumes: {}
    manager: kubectl-client-side-apply
    operation: Update
    time: "2023-03-10T19:51:14Z"
  name: px-operator
  resourceVersion: "1708"
  uid: 8f9f22a2-c830-41f4-81f6-17d3c691617a
spec:
  allowPrivilegeEscalation: false
  fsGroup:
    rule: RunAsAny
  runAsUser:
    rule: RunAsAny
  seLinux:
    rule: RunAsAny
  supplementalGroups:
    rule: RunAsAny
  volumes:
  - secret`

[PWX-28335](https://portworx.atlassian.net/browse/PWX-28335)

[PWX-28335]: https://portworx.atlassian.net/browse/PWX-28335?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


drivers/storage/portworx/component/autopilot.go
```diff
@@ -265,12 +264,6 @@ func (c *autopilot) createClusterRole() error {
 					Resources: []string{"*"},
 					Verbs:     []string{"*"},
 				},
-				{
```
piyush-nimbalkar, line 268: If customers are running in a pre-1.26 k8s and have PSPs enabled, we will still need this. 
To keep the support for those, we should remove this only for 1.26 and above.

ezhang-px: Good point. Originally I thought by removing PSP entirely and specifying the rules in securitycontext at container level should be also compatible for customer running pre-1.2.6.

Then in order to only remove for 1.26 and above, can you suggest if there is any control in operator currently like checking what version of k8s cluster that customer is running on? @pure-jliao @haibinxie

piyush-nimbalkar: The `Initialize()` function in this file take the current k8s version as an input. You can use that and check the k8s version against the required k8s version to remove or keep the PSPs.

Eg -https://github.com/libopenstorage/operator/blob/master/drivers/storage/portworx/component/pvccontroller.go#L394

ezhang-px: Huh, nice pointer! Thanks a lot @piyush-nimbalkar. Change is on the way

drivers/storage/portworx/component/autopilot.go
```diff
@@ -253,25 +255,37 @@ func (c *autopilot) createServiceAccount(
...
+				Verbs:     []string{"*"},
+			},
+			{
+				APIGroups:     []string{"policy"},
+				Resources:     []string{"podsecuritypolicies"},
+				ResourceNames: []string{constants.RestrictedPSPName},
+				Verbs:         []string{"use"},
+			},
+		}
+	}
```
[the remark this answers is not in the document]
ezhang-px: Yep, this is more efficient. Will follow, thanks

drivers/storage/portworx/components_test.go
```diff
@@ -4154,6 +4154,38 @@ func TestAutopilotInvalidCPU(t *testing.T) {
 		fmt.Sprintf("%v %v Failed to setup Autopilot.", v1.EventTypeWarning, util.FailedComponentReason))
 }
 
+func TestAutopilotSecurityContext(t *testing.T) {
```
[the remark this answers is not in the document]
ezhang-px: Ok, I'll add this coverage as well. Good point!

drivers/storage/portworx/component/autopilot.go
```diff
@@ -253,25 +255,37 @@ func (c *autopilot) createServiceAccount(
 }
 
 func (c *autopilot) createClusterRole() error {
+	var rules []rbacv1.PolicyRule
+	if c.k8sVersion.GreaterThanOrEqual(k8sutil.K8sVer1_26) {
```
haibinxie, line 259: PSP was deprecated on 1.25 not 1.26

ezhang-px: Thanks for catching it. Sorry for the typo. Fixed with latest commit

drivers/storage/portworx/components_test.go
```diff
@@ -3156,6 +3163,8 @@ func TestAutopilotInstall(t *testing.T) {
 		},
 	}
 
+	cluster := stcSpec
```
piyush-nimbalkar, line 3166: nit: This is not copying the object but just assigning the pointer to another variable. It will result in modification of both objects. If you really want to copy the object use `DeepCopy` here and on line 3238.
```
cluster := stcSpec.DeepCopy()
```

piyush-nimbalkar (approved):

drivers/storage/portworx/components_test.go
ezhang-px: You're right. We don't want the original object messed up by copying it. DeepCopy is required here. Fixed it. Thanks

haibinxie (approved):

PR #993, opened 2023-03-23
Add CRD for portworx diags collection

piyush-nimbalkar:

Closes # https://portworx.atlassian.net/browse/PWX-29389


pkg/apis/portworx/v1/portworxdiag.go
```diff
@@ -0,0 +1,101 @@
...
+	// Labels of the volumes to be selected.
+	Labels map[string]string `json:"labels,omitempty"`
+}
+
+// Allows selecting one or more volumes for which diags need to be  collected.
+// It will take a union of volume IDs and volume labels and collect diags for
+// all that match. Currently it will find the Portworx nodes that contain the
+// replicas of the selected volumes and wherever the volume is attached or
+// being consumed from.
+type VolumeSelector struct {
```
haibinxie, line 50: does this collect node diags of the nodes that have the replicas? not just logs related to the volume activity?

pkg/apis/portworx/v1/portworxdiag.go
```diff
@@ -0,0 +1,101 @@
...
+	// PortworxDiagResourceName is name for "portworxdiag" resource.
+	PortworxDiagResourceName = "portworxdiag"
+	// PortworxDiagResourcePlural is plural for "portworxdiag" resource.
+	PortworxDiagResourcePlural = "portworxdiags"
+)
+
+// PortworxDiagSpec is the spec used to define a portworx diag.
+type PortworxDiagSpec struct {
+	// Configuration for diags collection of the main Portworx component.
+	Portworx *PortworxComponent `json:"portworx,omitempty"`
```
haibinxie, line 17: in the design spec, we also discussed about these flags:
  collectK8SObjects: true | false
  collectPodLogs: true | false
  cloudUpload: true | false

pkg/apis/portworx/v1/portworxdiag.go
```diff
@@ -0,0 +1,101 @@
...
+// being consumed from.
+type VolumeSelector struct {
+	// Ids of the volumes to be selected.
+	IDs []string `json:"ids,omitempty"`
+	// Labels of the volumes to be selected.
+	Labels map[string]string `json:"labels,omitempty"`
+}
+
+// PortworxDiagStatus is the status of a portworx diag.
+type PortworxDiagStatus struct {
```
haibinxie, line 58: we may need path to the collected files and cloud upload status.

pkg/apis/portworx/v1/portworxdiag.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+// Status of the diags collection from a single node.
+type NodeStatus struct {
+	// ID of the node for which the diag status is reported.
+	NodeID string `json:"nodeId,omitempty"`
+	// One word status of the diags collection on the node.
+	Status string `json:"status,omitempty"`
+	// Optional message used to give the reason for any failure.
+	Message string `json:"message,omitempty"`
```
haibinxie, line 74: path to the collected files, that resides on the daemon running on the node.

pkg/apis/portworx/v1/portworxdiag.go
piyush-nimbalkar: I checked with Jose and he told me that we don't get a path for the diags uploaded to Pure1 fuse. Until we find a better way to work with the Pure1 team to organize the diags differently, there is no point in giving a path.

The current diags are directly uploaded just under a cluster_uuid folder in fuse.

pkg/apis/portworx/v1/portworxdiag.go
piyush-nimbalkar: In this version, as the interim solution we are not going to collect the diags in the diag pod, but just run `pxctl sv diags` and let the pure1 log uploaded upload the diags to s3->fuse.

pkg/apis/portworx/v1/portworxdiag.go
piyush-nimbalkar: Yes all nodes that have replicas of the volume and where the volumes is attached. In case of shared volumes even the nodes where the application is running (clients of shared volumes).

pkg/apis/portworx/v1/portworxdiag.go
piyush-nimbalkar: That's correct and we will add those flag when we implement those parts.
As described in the [design doc for pxone](https://docs.google.com/document/d/1vWrujUfycoOJ2BAeeDoHiSrhVDRneHRCTBpgNOzYji8/edit#), we are only going to collect/upload what we have in porx and not adding anything new due to time constraints.

As described in your broader cluster wide diags doc, we will start adding support for the other options later.

pkg/apis/portworx/v1/portworxdiag.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:name="Status",type="string",JSONPath=".status.phase",description="Status of the Portworx diag collection."
+// +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp",description="Age of the diag resource."
```
ggriffiths, line 80: nit: Maybe add a short name for the CRD:
```
// +kubebuilder:resource:scope=Namespaced,shortName=pxdiag
```
or pxd

ggriffiths (approved):
Looks good to me. Just one suggestion..

Makefile
```diff
@@ -170,6 +171,16 @@ codegen:
 	@echo "Generating CRD"
 	(GOFLAGS="" hack/update-codegen.sh)
 
+# This generates the deepcopy functions only for the portworx.io APIs. We should change
+# the path to "./..." to start generating the deepcopy functions for the core APIs too.
+generate:
```
pure-jliao, line 176: make this part of `codegen` as well?

deploy/crds/portworx.io_portworxdiags.yaml
pure-jliao, line 1: nit: file name not consistent though, not sure if we have a naming convention here. maybe add `_crd` suffix to be consistent. is this file auto generated?

pkg/apis/portworx/portworx.go
```diff
@@ -0,0 +1,9 @@
+// Package v1 contains API Schema definitions for the portworx API group
+// +k8s:deepcopy-gen=package,register
+// +groupName=portworx.io
+package portworx
```
pure-jliao, line 4: operator was designed for any storage drivers before right? then we're adding `pkg/apis/portworx`, moving forward are we still considering that when adding new components? because this path doesn't seem to be generic.

pkg/apis/portworx/v1/portworxdiag.go
pure-jliao: do we have some sort of identifier to match the uploaded diags with this diag object? e.g. put diag ID somewhere in the file path.

pkg/apis/portworx/portworx.go
piyush-nimbalkar: Yes, it was meant to be generic. That is why we created a new API group for the portworx APIs like the Diag so if needed we can move them to a different git repo/container and does not affect any of the 'operator' APIs. To avoid creating multiple containers we are putting the diags controller in the portworx operator. But again, both the controller and API can be moved out easily if needed.

Moving forward we should follow the same guideline. If it is very generic and belongs to the `core` operator functionality then it can go under core. For portworx specific CRs we should use the `portworx.io` group.

deploy/crds/portworx.io_portworxdiags.yaml
piyush-nimbalkar: It is auto-generated. Let me check if I can specify a name to the file.

pkg/apis/portworx/v1/portworxdiag.go
piyush-nimbalkar: It creates individual ZIPs per node, so it won't be easy to unzip all the available diags in the folder to find out which one it belongs to. I had initially proposed adding a unique ID to the name of the zip, but it was agreed upon to just use the timestamp instead of an identifier. So, I modified the design doc accordingly.

deploy/crds/portworx.io_portworxdiags.yaml
piyush-nimbalkar: Could not find a way to name the file. I guess we will have to live with it for now. 
When we move to use controller-gen for the existing operator specs, the filenames would be consistent.

pure-jliao (approved):
LGTM!

PR #995, opened 2023-03-24
[pwx-30005] Do Pure1 connectivity check when telemetry secret is not created

pure-jliao:
* behavior doc: https://docs.google.com/document/d/1d55gu_2Jdpm21zvjbKN1VIg3pxtjR99rGqDYjuZwIVw/edit?usp=sharing

Closes # PWX-30005


drivers/storage/portworx/component/telemetry.go
```diff
@@ -1004,41 +1008,59 @@ func getArcusRegisterProxyURL(cluster *corev1.StorageCluster) string {
...
 	endpoint := getArcusRegisterProxyURL(cluster)
 	logrus.Debugf("checking whether telemetry registration endpoint %s is accessible on cluster %s",
 		endpoint, cluster.Name)
 
-	url, err := url.Parse(fmt.Sprintf("https://%s:443/auth/1.0/ping", endpoint))
-	if err != nil {
-		return false, err
-	}
-
+	url, _ := url.Parse(fmt.Sprintf("https://%s:443/auth/1.0/ping", endpoint))
```
haibinxie, line 1021: although this will not return error for now, from coding convention perspective, let's still handle the error, if you don't want to return it, just log if error is not nil in case it happens

drivers/storage/portworx/component/telemetry.go
```diff
@@ -1004,41 +1008,59 @@ func getArcusRegisterProxyURL(cluster *corev1.StorageCluster) string {
...
-		logrus.WithFields(logrus.Fields{
-			"code": response.StatusCode,
-			"body": string(body),
-		}).Warnf("failed to access telemetry registration endpoint %s", endpoint)
-		return false, nil
+	if httpProxy != "" {
+		if !strings.HasPrefix(strings.ToLower(httpProxy), "http://") {
+			httpProxy = "http://" + httpProxy
+		}
+		proxyURL, _ := url.Parse(httpProxy)
```
haibinxie, line 1036: same the error should be handled

drivers/storage/portworx/component/telemetry.go
```diff
@@ -1004,41 +1008,59 @@ func getArcusRegisterProxyURL(cluster *corev1.StorageCluster) string {
...
 	}
-
-	return true, nil
+	for i := 1; i <= arcusPingRetry; i++ {
+		response, err := client.Do(request)
+		warnMsg := fmt.Sprintf("failed to access telemetry registration endpoint %s", endpoint)
+		if err != nil {
+			logrus.WithError(err).Warnf(warnMsg)
+		} else if response.StatusCode != 200 {
+			body, _ := io.ReadAll(response.Body)
```
haibinxie, line 1047: the returned error should be logged.

drivers/storage/portworx/component/telemetry.go
```diff
@@ -1004,41 +1008,59 @@ func getArcusRegisterProxyURL(cluster *corev1.StorageCluster) string {
...
+			response.Body.Close()
+			logrus.WithFields(logrus.Fields{
+				"code": response.StatusCode,
+				"body": string(body),
```
haibinxie, line 1051: this means it's accessible, we don't need to retry on this. I think it should return true immediately.

drivers/storage/portworx/component/telemetry.go
```diff
@@ -1004,41 +1008,59 @@ func getArcusRegisterProxyURL(cluster *corev1.StorageCluster) string {
 }
 
 // CanAccessArcusRegisterEndpoint checks if telemetry registration endpoint is reachable
-func CanAccessArcusRegisterEndpoint(cluster *corev1.StorageCluster) (bool, error) {
+// return true immediately if it can reach to Arcus
+// return false after failing 5 times in a row
+func CanAccessArcusRegisterEndpoint(
```
haibinxie, line 1013: Let's do some functional tests for this API. you may refer to this function for tests

func TestRemoteManifestWithProxy(t *testing.T) {

haibinxie (commented):
Looks good, could you write a unit test function so we can test the expected behaviors according to the table in the design doc. for example, a test function that takes all those columns as parameter and a parent function that can call the test function with different values in each row.

drivers/storage/portworx/component/telemetry.go
pure-jliao: the error can be traced when sending the request, if we hit error here the url will be empty, thus when sending the request, it can be logged properly. same for parsing http proxy.

drivers/storage/portworx/component/telemetry.go
pure-jliao: here the error is nil, error messages are in the response body, and logged below already.

drivers/storage/portworx/component/telemetry.go
pure-jliao: we discussed before to send the request properly, and only consider it's a successful ping when got 200 OK response.

haibinxie: Right, i remember that, do you know when it will not return 200? technically we don't care if it's 200, as long as the network is accessible

drivers/storage/portworx/component/telemetry.go
haibinxie: for parsing proxy url, just looked the code of http.ProxyURL, if proxyUrl is empty it will be treated as no proxy, from coding convention perspective it's easier to track when an error is handled in place.

	// If Proxy is nil or returns a nil *URL, no proxy is used.
	Proxy func(*Request) (*url.URL, error)

haibinxie (approved):

drivers/storage/portworx/component/telemetry.go
pure-jliao: handled parsing proxy url

PR #998, opened 2023-03-27
[pwx-30005] Do Pure1 connectivity check when telemetry secret is not created

pure-jliao:

Closes #


drivers/storage/portworx/component/telemetry.go
```diff
@@ -109,6 +110,9 @@ const (
 	defaultCollectorPort    = 10000
 	defaultRegisterPort     = 12001
 	defaultPhonehomePort    = 12002
+
+	arcusPingInterval = 6 * time.Second
```
dsupure, line 114: Do we need 30 secs in total?

dsupure (approved):
LGTM!

haibinxie (approved):

drivers/storage/portworx/component/telemetry.go
pure-jliao: yes that's the decision after discussion.

PR #1001, opened 2023-03-29
PWX-29398 : Report an appropriate error instead of panicking when there are 0 nodes available for PX install

rge-pure:
GetZoneMap() would return zero size if nodeList.Items would be zero. Then number of zones equals to 0, which trigger panic (divided by zero). Should change the panic to a better message showing that zero nodes are available to install.

[PWX-29398](https://portworx.atlassian.net/browse/PWX-29398)

Check node list length and add corresponding error instead of nil panic.
Add related unit test.


pkg/controller/storagecluster/storagecluster.go
```diff
@@ -1321,6 +1321,9 @@ func getZoneMap(nodeList *v1.NodeList, filterLabelKey string, filterLabelValue s
 			return nil, err
 		}
 	}
+	if len(nodeList.Items) == 0 {
+		return nil, fmt.Errorf("node list is empty")
```
pure-jliao, line 1325: can we do this validation at where we're dividing 0? I can see based on some filter key or values, it might be valid to have an empty map returned here.

pure-jliao (commented):
also could you open a ticket to track cherry-picking this change to master? incase it get lost. https://portworx.atlassian.net/browse/PWX-27944 this ticket will be a dependency.

pkg/controller/storagecluster/storagecluster.go
rge-pure: Thanks Jiafeng, I will do that!

rge-pure:
> also could you open a ticket to track cherry-picking this change to master? incase it get lost. https://portworx.atlassian.net/browse/PWX-27944 this ticket will be a dependency.

That's a good idea. I am going to do that!

pure-jliao (approved):
code change looks good to me, have @nrevanna to take a look as well. btw the ticket created should be for master branch not release branch. usually we should do master first then cherry pick to release, here because they diverged already, so we need to track the effort for cherry picking to master.

rge-pure:
Thanks Jiafeng! Appreciate your advice!

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -1285,6 +1285,9 @@ func (c *Controller) getDefaultMaxStorageNodesPerZone(
 		return 0, err
 	}
 	numZones := len(zoneMap)
+	if numZones == 0 {
```
zoxpx, line 1288: nit: small remark -- could we change this info `if numZones =< 0 {` ?

When I see `if ==` checks like this, I always try to generalize and see if _"we can cover more ground, without paying more"_ (so, won't cost us anything to check `if =<` instead -- just in very unlikely (impossible?) case that `len()` function returns a negative value).

zoxpx (approved):
Looks OK -- one minor remark listed below

pkg/controller/storagecluster/storagecluster.go
rge-pure: Thanks Zoran for helping me out. I can always learn from your comments!

pkg/controller/storagecluster/controller_test.go
```diff
@@ -9606,3 +9606,95 @@ func latestRevision(revs *appsv1.ControllerRevisionList) *appsv1.ControllerRevis
...
+					v1.LabelTopologyZone:   "foo-pxzone2",
+					v1.LabelTopologyRegion: "foo"},
+			},
+			Spec: v1.NodeSpec{
+				ProviderID: "azure://",
+			}},
+	}
+	storageNodes, err = controller.getDefaultMaxStorageNodesPerZone(&nodeList, cluster, recorder)
+	require.Equal(t, uint32(0), storageNodes)
+	require.Equal(t, fmt.Errorf("node list is empty and no MaxStorageNodesPerZone can be obtained"), err)
```
nrevanna, line 9699: nits:
- Create an error for this using errors.New(). e.g ErrNodeListEmpty
- The message can simply say "no nodes were available for px installation" and not talk about maxSNPZ

nrevanna (approved):

pkg/controller/storagecluster/controller_test.go
rge-pure: Thanks Naveen, I will do that

PR #1010, opened 2023-04-04
PWX-28826: Handle pre-flight check for DMthin 

jrivera-px:
Modify operator to execute a pre-flight on each node in the cluster.  This pre-flight will actually be executed by oci-monitor and px-runc installed on the host.   The pre-flight will determine if DMthin can be enabled on the cluster.   We do this by having operator launch a daemonset using the storage config with an extra pre-flight option.  This will cause oci-monitor to run in pre-flight mode which executes checks via px-runc on each node.   The results of the check are passed to oci-monitor via a json file. They are read  in and added to the StorageNode obj and passed to operator.  Once the checks are done this pre-flight daemonset is deleted by the operator.   The operator processes the checks returned in the StorageNode obj to determine if the existing storage config can be modified to enable DMthin (add: -T dmthin param).   If the checks fail then the storage config is not updated and the cluster starts with the original stc.

Closes #  
PWX-28826

px-runc side is in this PR: https://github.com/portworx/porx/pull/11018
px-installer side is in this PR: https://github.com/portworx/px-installer/pull/1679

### DIFF
--- a/drivers/storage/portworx/portworx.go
+++ b/drivers/storage/portworx/portworx.go
@@ -174,11 +174,9 @@ func (p *portworx) Validate(cluster *corev1.StorageCluster) error {
 
 	var storageNodes []*corev1.StorageNode
 
-	/*
-		Wait for pre-flight checks to complete.  All the pre-flight pods are running.
-		However the checks being performed may not have completed.  Keep checking the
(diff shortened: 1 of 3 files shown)


jrivera-px:
Does not contain all the changes...

PR #1014, opened 2023-04-05
PWX-28826: Handle pre-flight check for DMthin

jrivera-px:
Modify operator to execute a pre-flight on each node in the cluster. This pre-flight will actually be executed by oci-monitor and px-runc installed on the host. The pre-flight will determine if DMthin can be enabled on the cluster. We do this by having operator launch a daemonset using the storage config with an extra pre-flight option. This will cause oci-monitor to run in pre-flight mode which executes checks via px-runc on each node. The results of the check are passed to oci-monitor via a json file. They are read in and added to the StorageNode obj and passed to operator. Once the checks are done this pre-flight daemonset is deleted by the operator. The operator processes the checks returned in the StorageNode obj to determine if the existing storage config can be modified to enable DMthin (add: -T dmthin param). If the checks fail then the storage config is not updated and the cluster starts with the original stc.

Closes #
PWX-28826

px-runc side is in this PR: https://github.com/portworx/porx/pull/11018
px-installer side is in this PR: https://github.com/portworx/px-installer/pull/1679


pkg/controller/storagecluster/storagecluster.go
```diff
@@ -441,13 +437,49 @@ func (c *Controller) runPreflightCheck(cluster *corev1.StorageCluster) error {
...
+
 	condition := &corev1.ClusterCondition{
 		Source: pxutil.PortworxComponentName,
 		Type:   corev1.ClusterConditionTypePreflight,
 	}
+
+	// XXX - test
+	//err = fmt.Errorf("pre-flighr check force failure")
+
 	if err != nil {
```
piyush-nimbalkar, line 483: We would be capturing the wrong error here in case deleting storagenodes fail here. Also, if the cloud drive permission pre-check fails on line 436, we will overwrite that error.

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -465,6 +497,9 @@ func (c *Controller) runPreflightCheck(cluster *corev1.StorageCluster) error {
 			return err
 		}
 
+		// XXX - Sleep to avoid race cond check NOTE: in setStorageClusterDefaults()
+		time.Sleep(30 * time.Second)
+
```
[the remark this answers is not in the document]
jrivera-px: This is actually something that Jiafeng had me add for my testing to avoid the race condition.  At the time I was having issues getting the daemonset to start.    We can remove that for sure.   He mentioned that the same issue can happen here that is happening in setStorageClusterDefaults().

pkg/preflight/utils.go
```diff
@@ -11,8 +11,9 @@ func IsEKS() bool {
 
 // RequiresCheck returns whether a preflight check is needed based on the platform
 func RequiresCheck() bool {
+	return true
```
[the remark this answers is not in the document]
jrivera-px: No the way Harsh and Prabir thought this should go is that we would always run the pre-flight and we have a check for EKS stuff in IsEKS() condition blocks.

pkg/controller/storagecluster/storagecluster.go
jrivera-px: yeah thx I noticed that while I was debugging UTs issues.    I have that change coming in PRs which fixes the driver UT.   Basically when I cleanup the storageNodes used for pre-flight I am using a different err.

drivers/storage/portworx/portworx.go
```diff
@@ -88,7 +89,67 @@ func (p *portworx) Init(
...
+		logrus.Infof("pre-flight: Completed [%v] In Progress [%v] Total [%v]", completed, inProgress, total)
+
+		if total != 0 && completed == total {
+			logrus.Infof("pre-flight: completed...")
+			break
+		}
+
+		// Add five minute timeout.  If we do reconcile loop check we will need a different way.
+		cnt++
+		if cnt == 100 { // 3s * 100 = 300s (5 mins)
```
piyush-nimbalkar, line 127: This may not be enough because the first time we would be pulling the px-enterprise image as well. Should we set it to 10mins?

pkg/preflight/utils.go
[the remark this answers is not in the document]
piyush-nimbalkar: I am assuming we will fix this later and run it only on cloud or if the metadata device is already provided. Correct?

piyush-nimbalkar (approved):

drivers/storage/portworx/portworx.go
jrivera-px: incremented the time for now and logged https://portworx.atlassian.net/browse/PWX-30419 to fix this.

PR #1017, opened 2023-04-16
Use reconcile loop to check pre-flight status.  Timeout after 15m.

jrivera-px:
When executing pre-flight check we would wait for the checks to complete blocking the reconcile loop.   This prevented any other state change from being processed.  So this change corrected that.  Use cluster conditions to determine if we are still executing a pre-flight.  If so check the status each time though the loop.   Add a 15 min timeout based on the time the DS has been running.    

Closes #

Adding this as a separate PR to PWX-28826 branch so the changes can be clearly viewed.      This PR is separate https://github.com/libopenstorage/operator/pull/1014 which has been approved however it has the blocking code.  This merge will correct that.

### DIFF
--- a/drivers/storage/portworx/portworx.go
+++ b/drivers/storage/portworx/portworx.go
@@ -106,29 +106,38 @@ func (p *portworx) Validate(cluster *corev1.StorageCluster) error {
 		logrus.Debugf("pre-flight: container already running...")
 	}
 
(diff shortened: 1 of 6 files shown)


jrivera-px:
Closing this PR for now.   Update https://portworx.atlassian.net/browse/PWX-28826 with comment.

PR #1019, opened 2023-04-18
PWX-30496:  if '-T dmthin' exists in stc before preflight is ran and preflight fails don't start.   

jrivera-px:
If the user intended to use dmthin.    The '-T dmthin' will exist in the stc before preflight is ran.  If preflight fails in this case don't start.
Closes #
PWX-30496


drivers/storage/portworx/preflight.go
```diff
@@ -180,8 +185,36 @@ func (u *preFlightPortworx) RunPreFlight() error {
...
+			}
+		}
+	}
+
+	// Check for pre-existing DMthin in misc-args
+	if mArgs, err := pxutil.MiscArgs(u.cluster); err == nil {
+		checkArgs(mArgs)
+	} else {
+		logrus.Warnf("runPreFlight: error parsing misc args, skipping existing PX-StoreV2 param check: %v", err)
+	}
```
piyush-nimbalkar, line 203: We are already checking the pod args below. Why do we need to check the cluster's misc args again?

drivers/storage/portworx/preflight.go
```diff
@@ -245,9 +278,31 @@ func (u *preFlightPortworx) ProcessPreFlightResults(recorder record.EventRecorde
...
 	if passed {
-		// Enable DMthin via misc args
-		u.cluster.Annotations[pxutil.AnnotationMiscArgs] = strings.TrimSpace(u.cluster.Annotations[pxutil.AnnotationMiscArgs] + " -T dmthin")
-		k8sutil.InfoEvent(recorder, u.cluster, util.PassPreFlight, "Enabling DMthin")
+		if !u.hardFail { // Enable DMthin via misc args if not enabled already
+			u.cluster.Annotations[pxutil.AnnotationMiscArgs] = strings.TrimSpace(u.cluster.Annotations[pxutil.AnnotationMiscArgs] + " -T dmthin")
+			// Only set if CloudStorage is nil
+			if u.cluster.Spec.CloudStorage == nil {
+				// Add 64G metadata drive.
+				cmetadata := "type=gp2,size=64"
```
piyush-nimbalkar, line 286: We should use gp3 here.

drivers/storage/portworx/preflight.go
```diff
@@ -245,9 +278,31 @@ func (u *preFlightPortworx) ProcessPreFlightResults(recorder record.EventRecorde
 	}
 
 	if passed {
-		// Enable DMthin via misc args
-		u.cluster.Annotations[pxutil.AnnotationMiscArgs] = strings.TrimSpace(u.cluster.Annotations[pxutil.AnnotationMiscArgs] + " -T dmthin")
-		k8sutil.InfoEvent(recorder, u.cluster, util.PassPreFlight, "Enabling DMthin")
+		if !u.hardFail { // Enable DMthin via misc args if not enabled already
+			u.cluster.Annotations[pxutil.AnnotationMiscArgs] = strings.TrimSpace(u.cluster.Annotations[pxutil.AnnotationMiscArgs] + " -T dmthin")
+			// Only set if CloudStorage is nil
+			if u.cluster.Spec.CloudStorage == nil {
```
piyush-nimbalkar, line 284: - Let's add the metadata device in both cases where the user has passed `-T dmthin` or we added it for them.
- We need to handle the case where cloud storage is not nil. 
- We should not overwrite the metadata device if already provided by the user.
```
if u.cluster.Spec.CloudStorage == nil {
  u.cluster.Spec.CloudStorage = &corev1.CloudStorageSpec{}
}

if u.cluster.Spec.CloudStorage.SystemMdDeviceSpec != nil {
  u.cluster.Spec.CloudStorage.SystemMdDeviceSpec = &cmetadata
}
```

drivers/storage/portworx/preflight.go
jrivera-px: Args can be added in the spec correct.  So we should check them both..

piyush-nimbalkar (approved):
Let's add the metadata device if the check passes irrespective if the user added the dmthin arg or not.

PR #1030, opened 2023-05-05
PWX-29131: container image -url parsing fix

zoxpx:

The `AddDefaultRegistryToImage()` is used to parse all image-URL's specified in `px-versions` CM, but it had issues:
* most URLs will be prefixed w/ `docker.io`  (unless form well-known repos) -- so private registry-servers cannot be used
* no support for IPs (IPv4, IPv6)

As a fix, we'll...
* prefix the URL w/ `docker.io` only if URL does not start w/ hostname or IP
* additionally, append `:latest` if no image-tag was provided

Closes # PWX-29131

See JIRA-ticket for PX-specific test info

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -3,8 +3,10 @@ package util
 import (
 	"context"
 	"fmt"
+	"net"
 	"path"
 	"reflect"
+	"regexp"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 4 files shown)


pkg/util/util.go
```diff
@@ -105,12 +110,43 @@ func AddDefaultRegistryToImage(image string) string {
...
-		if strings.HasPrefix(image, k) || strings.HasPrefix(image, "gcr.io") || strings.HasPrefix(image, "k8s.gcr.io") {
+	// check if image has version-tag included, add ":latest" if missing
+	if strings.Count(image, ":") < 1 {
+		image = image + ":latest"
+	}
+
+	parts := strings.Split(image, "/")
+
+	// no separate parts, always add registry-server
+	if len(parts) <= 0 {
```
piyush-nimbalkar, line 121: nit: We don't need this check as we are already checking if the image is empty above. This will be true only if the image is empty.

piyush-nimbalkar (approved):

zoxpx:
Thanks for the review Piyush -- merging the PR, cherry-picking for `px-rel-23.5.0`

PR #1034, opened 2023-05-09
Use 23.4.0 as the previous version

piyush-nimbalkar:

- We need to change the base registry to a released version and create the new test catalog on top of that released catalog.
- Update the `replaces` to use a released version.

Closes #


zoxpx (approved):
Looks good -- thanks!

PR #1037, opened 2023-05-10
PWX-25280: OCP dynamic plugin support in operator

nikita-bhatia:
https://portworx.atlassian.net/browse/PWX-25280

This PR supports the feature to add dynamic plugin to OCP.
> During installation of portworx, configs of plugin are created in operator's namespace
> During uninstall of portworx, configs are deleted
> Console Plugin enable/disable option support from operator overview page

Tested :
1. Installation ofportworx deploys all configs for plugin
2. Enabling the plugin from installed PX operator page, adds the 'Portworx' tab
3. Disabling the plugin from installed PX operator page, removes the 'Portworx' tab
4. Uninstalling Portworx from OCP deletes all configs of plugin.


drivers/storage/portworx/component/plugin.go
```diff
@@ -0,0 +1,359 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/record"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"strings"
+)
+
+const (
+	PluginComponentName       = "Portworx Plugin"
+	PluginName                = "portworx"
+	PluginDeloymentName       = "portworx-plugin"
```
piyush-nimbalkar, line 28: nit: * Plugin**Deployment**Name

drivers/storage/portworx/component/plugin.go
```diff
@@ -0,0 +1,359 @@
...
+	}
+
+	//delete nginx configs
+	if err := k8s.DeleteDeployment(p.client, NginxDeploymentName, cluster.Namespace, *ownerRef); err != nil {
+		errList = append(errList, err.Error())
+	}
+	if err := k8s.DeleteConfigMap(p.client, NginxConfigMapName, cluster.Namespace, *ownerRef); err != nil {
+		errList = append(errList, err.Error())
+	}
+	if err := k8s.DeleteService(p.client, NginxServiceName, cluster.Namespace, *ownerRef); err != nil {
```
piyush-nimbalkar, line 163: Can you verify if the certificate secret that gets created with both the services also gets deleted when the service is deleted?

drivers/storage/portworx/component/plugin.go
```diff
@@ -0,0 +1,359 @@
...
+
+	_, err = k8s.CreateOrUpdateConfigMap(p.client, cm, ownerRef)
+	return err
+}
+
+func (p *plugin) createConsolePlugin(filename, storageNs string, ownerRef *metav1.OwnerReference) error {
+
+	if err := console.AddToScheme(p.scheme); err != nil {
+		return err
+	}
```
piyush-nimbalkar, line 227: Let's add this in `cmd/operator.go` so we don't have to add it here every time.

drivers/storage/portworx/component/plugin.go
```diff
@@ -0,0 +1,359 @@
...
+		return nil
+	}
+
+	if len(newOwners) == 0 {
+		logrus.Infof("Deleting %s/%s consolePlugin", storageNs, name)
+		return p.client.Delete(context.TODO(), consolePlugin)
+	}
+	consolePlugin.OwnerReferences = newOwners
+	logrus.Infof("Disowning %s/%s consolePlugin", storageNs, name)
+	return p.client.Update(context.TODO(), consolePlugin)
```
piyush-nimbalkar, line 298: Just like other k8s objects, can you move this to the k8s util instead of keeping it here?

drivers/storage/portworx/component/plugin.go
```diff
@@ -0,0 +1,359 @@
...
+        location / {
+          proxy_pass http://portworx-api.` + storageNs + `.svc.cluster.local:9021;
+        }
+      }
+    }`,
+		}
+	}
+}
+
+func isVersionGreaterOrEqual(version string, targetVersion string) bool {
```
piyush-nimbalkar, line 330: Use `version "github.com/hashicorp/go-version"` package to parse and compare versions. Search for the usage in the operator code and you will find examples.

drivers/storage/portworx/component/plugin.go
```diff
@@ -0,0 +1,359 @@
...
+}
+
+func (p *plugin) createDeployment(filename, storageNs string, ownerRef *metav1.OwnerReference) error {
+	deployment, err := k8s.GetDeploymentFromFile(filename, pxutil.SpecsBaseDir())
+	if err != nil {
+		return err
+	}
+	deployment.Namespace = storageNs
+	deployment.OwnerReferences = []metav1.OwnerReference{*ownerRef}
+	return k8s.CreateOrUpdateDeployment(p.client, deployment, ownerRef)
```
piyush-nimbalkar, line 195: Unfortunately the CreateOrUpdateDeployment does not work like the other funcs for other objects. As k8s adds some default values in the spec, the diff between what is in k8s api server and what we read from file is always non-nil. So we end up calling api server to update the deployment every time.

To avoid this we introduced a very crude way of keeping track of the deployment using flags. Check other components on how we deal with deployments.


Also, we should add extra things to the deployment like the image registry, image pull secret, toleration, etc. Again, check other components.

piyush-nimbalkar (commented):
Send the PR against master branch first and then we can cherrypick it to 23.5 branch.

drivers/storage/portworx/component/plugin.go
```diff
@@ -0,0 +1,359 @@
...
+	return DefaultComponentPriority
+}
+
+func (c *plugin) IsPausedForMigration(cluster *corev1.StorageCluster) bool {
+	return util.ComponentsPausedForMigration(cluster)
+}
+
+func (p *plugin) IsEnabled(cluster *corev1.StorageCluster) bool {
+	operator := &ocpconfig.ClusterOperator{}
+	err := p.client.Get(
```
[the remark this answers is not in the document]
nikita-bhatia: If this fails because type ClusterOperator object is not present, will that not be a good check for existence of OCP cluster?

drivers/storage/portworx/component/plugin.go
```diff
@@ -0,0 +1,338 @@
...
+		if len(cluster.Spec.Placement.Tolerations) > 0 {
+			deployment.Spec.Template.Spec.Tolerations = make([]v1.Toleration, 0)
+			for _, toleration := range cluster.Spec.Placement.Tolerations {
+				deployment.Spec.Template.Spec.Tolerations = append(
+					deployment.Spec.Template.Spec.Tolerations,
+					*(toleration.DeepCopy()),
+				)
+			}
+		}
+	}
```
piyush-nimbalkar, line 240: We can use `util.ApplyStorageClusterSettingsToPodSpec` to apply these settings to the deployment.

drivers/storage/portworx/component/plugin.go
```diff
@@ -0,0 +1,338 @@
...
+			Name:      deploymentName,
+			Namespace: cluster.Namespace,
+		},
+		existingDeployment,
+	)
+	if getErr != nil && !errors.IsNotFound(getErr) {
+		return getErr
+	}
+
+	equal, _ := util.DeepEqualPodTemplate(&deployment.Spec.Template, &existingDeployment.Spec.Template)
```
piyush-nimbalkar, line 213: We should be doing a comparison after we apply the below settings to the deployment. Otherwise it will always show up as different.

drivers/storage/portworx/component/plugin.go
```diff
@@ -0,0 +1,338 @@
...
+			}
+		}
+	}
+
+	if !equal {
+		if err := k8s.CreateOrUpdateDeployment(p.client, deployment, ownerRef); err != nil {
+			return err
+		}
+	}
+	p.isOperatorCreated = true
```
piyush-nimbalkar, line 247: This flag is not even used anywhere. We should use to avoid creating the Deployment above if already created.

Also, we are using these method for two deployments and using a common flag. We should have a separate flag for each.

drivers/storage/portworx/component/plugin.go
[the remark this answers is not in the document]
nikita-bhatia: Updated code as suggested

drivers/storage/portworx/component/plugin.go
nikita-bhatia: Update as suggested

PR #1043, opened 2023-05-15
Remove resources limits in order to avoid OOM issues with this collector pod con metrics-collector.

camartinez04:
Remove resources limits in order to avoid OOM issues with this collector pod con metrics-collector.

Closes # PWX-31146


harsh-px:
I don't think we want to do this. The OOM was happening due to a bug which is fixed now.

PR #1045, opened 2023-05-17
PWX-25280: OCP dynamic plugin support in operator

nikita-bhatia:
https://portworx.atlassian.net/browse/PWX-25280

This PR supports the feature to add dynamic plugin to OCP.
> During installation of portworx, configs of plugin are created in operator's namespace
> During uninstall of portworx, configs are deleted
> Console Plugin enable/disable option support from operator overview page

Tested :
1. Installation ofportworx deploys all configs for plugin
2. Enabling the plugin from installed PX operator page, adds the 'Portworx' tab
3. Disabling the plugin from installed PX operator page, removes the 'Portworx' tab
4. Uninstalling Portworx from OCP deletes all configs of plugin.


drivers/storage/portworx/component/plugin.go
```diff
@@ -0,0 +1,351 @@
...
+		return true
+	} else if p.isPluginSupported == "" {
+		gvk := schema.GroupVersionKind{
+			Kind:    ClusterOperatorKind,
+			Version: ClusterOperatorVersion,
+		}
+		exists, err := coreops.Instance().ResourceExists(gvk)
+		if err != nil {
+			logrus.Error(err)
+			p.isPluginSupported = FALSE
```
piyush-nimbalkar, line 94: If there was an error we should not set this flag and try again, right? Something could be temporarily wrong with the API server and we would want to get the proper confirmation from the API server whether the resource exists or not.

drivers/storage/portworx/component/plugin.go
```diff
@@ -0,0 +1,351 @@
...
+			operator := &ocpconfig.ClusterOperator{}
+			err := p.client.Get(
+				context.TODO(),
+				types.NamespacedName{
+					Name: OpenshiftAPIServer,
+				},
+				operator,
+			)
+
+			if errors.IsNotFound(err) {
```
piyush-nimbalkar, line 108: We are not handling all the errors here. This could lead to a nil pointer error below as the operator object could be nil on error.

Similar to above comment, let's only set the flag if we are able to successfully call the API.
```
if err != nil {
  logrus.Error(err)
  if errors.IsNotFound() {
    p.isPluginSupported = FALSE
  }
  return false
}
```

drivers/storage/portworx/component/plugin.go
```diff
@@ -0,0 +1,351 @@
...
+
+			for _, v := range operator.Status.Versions {
+				if v.Name == OpenshiftAPIServer && isVersionSupported(v.Version) {
+					p.isPluginSupported = TRUE
+					return true
+				}
+			}
+		}
+	}
+
```
piyush-nimbalkar, line 121: We should be setting the flag to `false` here, right?

drivers/storage/portworx/component/plugin.go
```diff
@@ -0,0 +1,351 @@
...
+func (c *plugin) Priority() int32 {
+	return DefaultComponentPriority
+}
+
+func (c *plugin) IsPausedForMigration(cluster *corev1.StorageCluster) bool {
+	return util.ComponentsPausedForMigration(cluster)
+}
+
+func (p *plugin) IsEnabled(cluster *corev1.StorageCluster) bool {
+	if p.isPluginSupported == TRUE {
```
piyush-nimbalkar, line 84: nit: Can we use a pointer instead of using such a custom enum?

```
type plugin struct {
        ...
	isPluginSupported         *bool
}

func (p *plugin) IsEnabled(cluster *corev1.StorageCluster) bool {
        if p.isPluginSupported != nil {
                return p.isPluginSupported
        }
        ...
}

```

drivers/storage/portworx/component/plugin.go
nikita-bhatia: agreed, corrected

drivers/storage/portworx/component/plugin.go
nikita-bhatia: we are setting desired value for each case inside the else block, this is default return value which might never be called

drivers/storage/portworx/component/plugin.go
```diff
@@ -0,0 +1,356 @@
...
+					Name: OpenshiftAPIServer,
+				},
+				operator,
+			)
+
+			if err != nil {
+				logrus.Error(err)
+				if errors.IsNotFound(err) {
+					p.isPluginSupported = boolPtr(false)
+					return false
```
piyush-nimbalkar, line 108: The `return false` should be outside the IsNotFound block. Otherwise when there is an error which is !NotFound, it would go ahead to the for loop and result into a nil pointer error.

piyush-nimbalkar (approved):
LGTM

drivers/storage/portworx/component/plugin.go
nikita-bhatia: agreed, moved it out of block.

PR #1044, opened 2023-05-17
PWX-31199: 23.5.0 cherry-pick for autopilot RBAC/security fixes

ezhang-px:
**What this PR does / why we need it**:
Cherry-picked two previous commits in master about deprecating the use of podsecuritypolicies and limiting number of RBAC permissions for 23.5.0 release

**Which issue(s) this PR fixes** (optional)
Closes #

**Special notes for your reviewer**:

PR #1046, opened 2023-05-17
PWX-31199: 23.5.0 cherry-pick for autopilot RBAC/security fixes

ezhang-px:
Cherry-picked two previous commits in master about deprecating the use of podsecuritypolicies and limiting number of RBAC permissions for 23.5.0 release. 
Also set a cpu limit of 250m for autopilot besides the requests 

Closes #

This PR was mostly reviewed and approved in another closed one (https://github.com/libopenstorage/operator/pull/1044) and added two more commits to resolve the cpu limits and nonroot user issue
https://github.com/libopenstorage/operator/commit/d5ac06d6bc58e466de48e669e0b1712fbdb3314d
https://github.com/libopenstorage/operator/commit/2a2c6f30d5e77f656d006156fd537c694c845428


ezhang-px:
We need to hold on the run as non root change. Found issue when setting the securitycontext to a non-root user, autopilot complained about not able to access /var/cores (700 by root only) for generating support bundle

drivers/storage/portworx/testspec/autopilotDeployment.yaml
```diff
@@ -75,6 +82,15 @@ spec:
                     operator: Exists
                   - key: node-role.kubernetes.io/worker
                     operator: Exists
```
zoxpx, line 84: Can a node be both `master` and a `worker`?
* is this intended for a single-node systems?

drivers/storage/portworx/testspec/autopilotDeployment.yaml
```diff
@@ -75,6 +82,15 @@ spec:
...
+              - matchExpressions:
+                  - key: px/enabled
+                    operator: NotIn
+                    values:
+                      - "false"
+                  - key: node-role.kubernetes.io/control-plane
+                    operator: Exists
+                  - key: node-role.kubernetes.io/worker
+                    operator: Exists
```
zoxpx, line 93: Are these rules correct?

It seems like we are looking for..
* non-master nodes (workers?) that are NOT tagged as `control-plane`, or
* worker nodes that ARE tagged as `control-plane`

zoxpx (approved):
I think this change looks OK (it has previously been reviewed as a `master`-branch commit)

Please double-check the comment below, before the commit.

drivers/storage/portworx/testspec/autopilotDeployment.yaml
piyush-nimbalkar: Yes, it can be. IBM cloud deploys such a cluster with both master and worker labels and that is why we had to add this.

drivers/storage/portworx/testspec/autopilotDeployment.yaml
piyush-nimbalkar: The k8s master node labels was changed from `../master` to `../control-plane`. This is basically covering the case where a master/control-plane node is also a worker node and we need to support it and treat it like a worker node.

zoxpx: Would it make sense to simplify the rules, e.g.
1. if `worker`  (that's essentially combined rule#2 and rule#3), or
2. if `NOT master and NOT control-plane` (that's rule#1)

PR #1047, opened 2023-05-18
PTX-17751 Fix when Telemetry is enabled by default by operator 23.3+ race condition

nikolaypopov:
Fix when Telemetry is enabled by default by operator 23.3+ during fresh installs and upgrades race condition:
* Added timeout of 2 minutes to wait after the cluster is ready, this how much it takes for PX Operator to perform checks and decide if to enable/disable Telemetry
* Updating Telemetry state to new state for PX Operator 23.3+ due to new default Telemetry behavior

In addition I fixed:
* Marketplace `BasicInstallViaOcpMarketplace` for when `-dev` Operator image is used, it was validating against wrong version
* Refactored some Operator functions to take in params instead of taking defaults, so we can use those in other tests to validate Operator objects
* Fix some issues with doing unnecessary checks and validations of Operator in `BasicUpgradeStorageCluster` between hops in special cases
* Added missing error return in `UpdateAndValidatePxOperatorViaMarketplace()` when used in `BasicUpgradeOperatorViaOcpMarketplace` test, it was failing silently and giving wrong update status


pkg/util/test/util.go
```diff
@@ -2963,6 +2980,69 @@ func ValidateTelemetry(pxImageList map[string]string, cluster *corev1.StorageClu
 	return ValidateTelemetryV1Disabled(cluster, timeout, interval)
 }
 
+// updateTelemetryStatus NOTE: is a workaround to get Telemetry again
+// from StorageCluster after validating that PX is online and upgraded
+// because there are multiple condition for when it can be disabled/enabled after sometime
```
piyush-nimbalkar, line 2985: Can we add more details on why exactly we are doing this and what are those conditions when telemetry can be enabled/disabled? It would make it easier for the next person to understand.

pkg/util/test/util.go
```diff
@@ -2963,6 +2980,69 @@ func ValidateTelemetry(pxImageList map[string]string, cluster *corev1.StorageClu
...
+func updateTelemetryStatus(pxImageList map[string]string, cluster *corev1.StorageCluster) (*corev1.StorageCluster, error) {
+	// Makin sure cluster is online
```
piyush-nimbalkar, line 2987: nit: *Making

pkg/util/test/util.go
```diff
@@ -2963,6 +2980,69 @@ func ValidateTelemetry(pxImageList map[string]string, cluster *corev1.StorageClu
...
+	// Validate storagenodes are upgraded
+	if err := ValidateAllStorageNodesAreUpgraded(pxImageList, liveCluster); err != nil {
+		return nil, err
+	}
+
+	// Check Telemetry state before timeout
+	if liveCluster.Spec.Monitoring != nil && liveCluster.Spec.Monitoring.Telemetry != nil {
+		logrus.Debugf("Telemetry state in the StorageCluster [%s] before sleep is [%v]", liveCluster.Name, liveCluster.Spec.Monitoring.Telemetry.Enabled)
+	} else {
+		logrus.Debugf("Telemetry state in the StorageCluster [%s] sleep is [nil]", liveCluster.Name)
```
piyush-nimbalkar, line 3002: nit: *before sleep

test/integration_test/basic_test.go
```diff
@@ -415,19 +415,29 @@ func BasicUpgradeOperator(tc *types.TestCase) func(*testing.T) {
...
+			pxOperatorImage, err := ci_utils.GetPxOperatorImage(pxOperatorDeployment)
 			require.NoError(t, err)
+			if len(lastHopImage) == 0 {
+				// This is initially deployed PX Operator version
+				lastHopImage = pxOperatorImage
+			}
 
 			if pxOperatorImage == hopImage {
-				logrus.Infof("Skipping upgrade of PX Operator from %s to %s", pxOperatorImage, hopImage)
+				logrus.Infof("Skipping upgrade of PX Operator from [%s] to [%s]", pxOperatorImage, hopImage)
```
piyush-nimbalkar, line 434: Just a suggestion - You can use `%q` to get a quoted string in the output if you want.
For example,
```
logrus.Infof("Skipping upgrade of PX Operator from %q to %q", pxOperatorImage, hopImage)
```
Output will be:
```
INFO Skipping upgrade of PX Operator from '23.4.0' to '23.5.0'
```

piyush-nimbalkar (approved):
Looks good.

pkg/util/test/util.go

pkg/util/test/util.go

PR #1058, opened 2023-06-01
PWX-31488 & PWX-31501: Crashloop backoff pods remained after, specify…

jrivera-px:
…ing pre-flight skip.  Defer cleanup declaration needs to be done before wait loop.

So the current implementation of pre-flight actually blocks and waits for pre-flight pods to finish/exit.  It has a timeout of 5 mins.   This is known issue and we look to fix this in the next release.  so if the pods get into a crash loop backoff and don't exit it will be 5 mins before operator will continue.  

Setting the "skip" annotation to bypass the pre-flight will not take affect until after the pre-flight block exits.     The problem was even after the blocking loop timeout and the "skip" was processed.   The pre-flight pods remained in crashloop back and the only way to remove them was to kill the daemonset.    The reason for this was because the blocking loop was being executed before the defer which was doing the cleanup.   So when the blocking loop timed out and returned an error there was no cleanup.   Moving the defer to before the blocking loop timeout fixes this issue.   

Closes #
PWX-31488 & PWX-31501
At this time once pre-flight is started and gets into crashloop backoff,  we cannot fix the code to handle the pre-flight "skip" until the 5 min timeout.

### DIFF
--- a/drivers/storage/portworx/portworx.go
+++ b/drivers/storage/portworx/portworx.go
@@ -108,6 +108,14 @@ func (p *portworx) Validate(cluster *corev1.StorageCluster) error {
 		logrus.Debugf("pre-flight: container already running...")
 	}
 
+	defer func() {
+		// Clean up the pre-flight pods
+		logrus.Infof("pre-flight: cleaning pre-flight ds...")
+		if derr := preFlighter.DeletePreFlight(); derr != nil {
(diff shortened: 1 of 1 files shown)


zoxpx (approved):
Looks harmless enough -- thanks Jose

PR #1061, opened 2023-06-06
PWX-30943: removed the memory limit for prometheus

ezhang-px:
Currently we observed failure of Prometheus deployment on testing cluster with memory limit of 800Mi as defined previously. And hence after some investigation:
 1. current testing cluster needs at least 1.5Gi to stabilize
 2. a previous test with cluster of 800-nodes used up to 3.7Gi
 3. customers are expected to deploy clusters at least with double the size of the testing cluster in 1

We decided to increase this limit to 4Gi so to cover most of the use case so far. More extensive investigation on whether a better (lower) limit could be specified will be carried out for next release. 

More details described in the ticket itself.

Edited Jun 6, 17:57 : ignore the statement in the description as we made decision to go with unbounded
https://github.com/libopenstorage/operator/pull/1061#issuecomment-1579214180

Closes #

### DIFF
--- a/drivers/storage/portworx/component/prometheus.go
+++ b/drivers/storage/portworx/component/prometheus.go
@@ -635,7 +635,7 @@ func (c *prometheus) createPrometheusInstance(
 		prometheusInst.Spec.Resources.Limits = cluster.Spec.Monitoring.Prometheus.Resources.Limits
 	} else {
 		prometheusInst.Spec.Resources.Limits = map[v1.ResourceName]resource.Quantity{
-			v1.ResourceMemory:           resource.MustParse("800Mi"),
+			v1.ResourceMemory:           resource.MustParse("4Gi"),
 			v1.ResourceCPU:              resource.MustParse("1"),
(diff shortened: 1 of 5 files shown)


zoxpx:
Hey @piyush-nimbalkar -- what do you think if we _removed_ the memory limit for Prometheus altogether?

ezhang-px:
> Hey @piyush-nimbalkar -- what do you think if we _removed_ the memory limit for Prometheus altogether?

Had offline discussion with @piyush-nimbalkar and we agreed to set no limit for Prometheus memory for now. Due to the reason that even this current 4Gi is partly experiential partly hypothesized, so customers might still run into the memory issue and the pod crash when used on even larger unforeseen cluster.

The ideal solution will be that we have a way to monitor the CPU/memory for components whose resource usage grows with cluster size and dynamically adjust it only when required, but not set to a static value like we are doing right now.

So removing this limit and let customer to decide how much limit they want to put by editing the stc spec when Prometheus is running on their clusters.

piyush-nimbalkar (approved):

PR #1064, opened 2023-06-07
PWX-31585: added new certStoreType kv pair to accommodate changes in latest ccm-go

ezhang-px:
Starting from ccm-go 1.0.9, it introduced an extra config field "certStoreType" to differentiate different approach for storing appliance certificate when cloud CA returns it to ccm-go. And in order to accommodate their side of change, we also need to introduce this field into ccm-go's configmap to use the latest image with vulnerabilities fixed. 

More context at this commend (https://portworx.atlassian.net/browse/PWX-31585?focusedCommentId=248697)

Closes #

### DIFF
--- a/deploy/ccm/config_properties_px.yaml
+++ b/deploy/ccm/config_properties_px.yaml
@@ -27,6 +27,7 @@ cloud:
   auth:
     version: "1.0"
 productName: "portworx"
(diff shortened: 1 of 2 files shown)


zoxpx (approved):
Looks harmless enough -- thanks Eugene

PR #1079, opened 2023-06-19
Operator's Windows container support 

nikita-bhatia:
Changes done : 
1. All px deployments will be appended with 
            “nodeSelector:
                 kubernetes.io/os: linux”
2. csidriver pxd.portworx.com  modified to have ‘attachedrequired = true
3. DaemonSet added for Windows
4.  Windows Storage Class added 
5. Cluster role the px.csi.driver has additional permission on volumeattachments/status 

Testing :
Tested on OCP cluster for with mix of linux and windows node

Few changes will be taken in next phase , this is initial PR to get Portworx support on windows node.


drivers/storage/portworx/component/windows.go
```diff
@@ -0,0 +1,163 @@
...
+func (w *windows) Priority() int32 {
+	return DefaultComponentPriority
+}
+
+func (w *windows) IsPausedForMigration(cluster *corev1.StorageCluster) bool {
+	return util.ComponentsPausedForMigration(cluster)
+}
+
+func (w *windows) IsEnabled(cluster *corev1.StorageCluster) bool {
+	w.isWindowsNode = true
```
piyush-nimbalkar, line 59: Isn't there going to be a way to automatically check if this is windows or not? We shouldn't be running this DS on non-windows envs, right?

drivers/storage/portworx/component/windows.go
```diff
@@ -0,0 +1,163 @@
...
+	if !equal && w.isWindowsNode {
+		if err := k8s.CreateOrUpdateDaemonSet(w.client, daemonSet, nil); err != nil {
+			return err
+		}
+	}
+	w.isCreated = true
+	return nil
+}
+
+func (w *windows) createStorageClass() error {
```
piyush-nimbalkar, line 141: Can we move this to the storage_class component instead of adding it to this windows component?

drivers/storage/portworx/component/windows.go
```diff
@@ -0,0 +1,163 @@
...
+			Name:      WindowsDaemonSetName,
+			Namespace: cluster.Namespace,
+		},
+		existingDaemonSet,
+	)
+	if getErr != nil && !errors.IsNotFound(getErr) {
+		return getErr
+	}
+
+	// pxutil.ApplyStorageClusterSettingsToPodSpec(cluster, &daemonSet.Spec.Template.Spec)
```
piyush-nimbalkar, line 129: We need to add this, otherwise the DS will get created on nodes where pod is not supposed to be deployed or it won't honor the image registry, pull policy, etc from the stroageclass.

drivers/storage/portworx/component/windows.go
```diff
@@ -0,0 +1,163 @@
...
+
+	equal, _ := util.DeepEqualPodTemplate(&daemonSet.Spec.Template, &existingDaemonSet.Spec.Template)
+	if !equal && w.isWindowsNode {
```
piyush-nimbalkar, line 132: We should use the `isCreated` flag here.

drivers/storage/portworx/component/windows.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+func (w *windows) createDaemonSet(filename string, ownerRef *metav1.OwnerReference, cluster *corev1.StorageCluster) error {
+	daemonSet, err := k8s.GetDaemonSetFromFile(filename, pxutil.SpecsBaseDir())
+	if err != nil {
+		return err
+	}
+
+	daemonSet.Namespace = cluster.Namespace
+	// daemonSet.OwnerReferences = []metav1.OwnerReference{*ownerRef}
```
piyush-nimbalkar, line 114: Why not add this back?

drivers/storage/portworx/component/prometheus.go
```diff
@@ -511,6 +511,9 @@ func getPrometheusOperatorDeploymentSpec(
 				},
 				Spec: v1.PodSpec{
 					ServiceAccountName: PrometheusOperatorServiceAccountName,
+					NodeSelector: map[string]string{
```
piyush-nimbalkar, line 514: Can this be part of the NodeAffinity for all components, including the windows daemonset we are creating?

drivers/storage/portworx/component/csi.go
```diff
@@ -1035,7 +1038,7 @@ func (c *csi) createCSIDriver(
 				Name: csiConfig.DriverName,
 			},
 			Spec: storagev1.CSIDriverSpec{
-				AttachRequired:       boolPtr(false),
+				AttachRequired:       boolPtr(true),
```
piyush-nimbalkar, line 1041: Can this always be `true` or is only needed for windows driver?
cc: @ggriffiths

drivers/storage/portworx/component/csi.go
```diff
@@ -675,13 +678,13 @@ func getCSIDeploymentSpec(
...
-				Image:           attacherImage,
+				Image:           "registry.k8s.io/sig-storage/csi-attacher:v4.3.0",
 				ImagePullPolicy: imagePullPolicy,
 				Args: []string{
-					"--v=3",
+					"--v=5",
 					"--csi-address=$(ADDRESS)",
 					"--leader-election=true",
-					"--leader-election-type=" + leaderElectionType,
+					"--http-endpoint=:8080",
```
piyush-nimbalkar, line 687: Is this arg needed only for windows?

drivers/storage/portworx/component/csi.go
```diff
@@ -675,13 +678,13 @@ func getCSIDeploymentSpec(
 			deployment.Spec.Template.Spec.Containers,
 			v1.Container{
 				Name:            csiAttacherContainerName,
-				Image:           attacherImage,
+				Image:           "registry.k8s.io/sig-storage/csi-attacher:v4.3.0",
```
piyush-nimbalkar, line 681: This image should come from the csi_generator like the other csi images. Let's not hard code it here.

drivers/storage/portworx/component/csi.go
```diff
@@ -582,10 +582,10 @@ func getCSIDeploymentSpec(
 	topologySpreadConstraints []v1.TopologySpreadConstraint,
 ) *appsv1.Deployment {
 	replicas := int32(3)
-	leaderElectionType := "leases"
+	// leaderElectionType := "leases"
 	provisionerLeaderElectionType := "leases"
 	if csiConfig.IncludeEndpointsAndConfigMapsForLeases {
-		leaderElectionType = "configmaps"
+		// leaderElectionType = "configmaps"
```
piyush-nimbalkar, line 588: Let's remove the commented code if we don't need it.

drivers/storage/portworx/util/csi_generator.go
```diff
@@ -190,6 +190,9 @@ func (g *CSIGenerator) GetCSIConfiguration() *CSIConfiguration {
 		cv.IncludeCsiDriverInfo = true
 	}
 
+	cv.IncludeAttacher = true
```
piyush-nimbalkar, line 193: Let's make this conditional and not set to true all the time.

drivers/storage/portworx/util/csi_generator.go
```diff
@@ -190,6 +190,9 @@ func (g *CSIGenerator) GetCSIConfiguration() *CSIConfiguration {
...
+	cv.IncludeCsiDriverInfo = true
```
piyush-nimbalkar, line 194: This is anyway true for k8s >= 1.14. I don't think you need this line.

drivers/storage/portworx/component/windows.go
nikita-bhatia: The only reason this is commented right now is because this is overwrting nodeselector from storagecluster, for windows support we have nodeaffinity as os=linux and for this daemonset, we need nodeaffinity as os=windows.

This is planned to be taken in next PR.

drivers/storage/portworx/component/windows.go
nikita-bhatia: The reason why this was kept here is we can have control over creation of this storage class on detection of windows only, do we still want to move it?

drivers/storage/portworx/component/windows.go
nikita-bhatia: yes added the check

drivers/storage/portworx/component/csi.go
nikita-bhatia: yes

drivers/storage/portworx/component/csi.go
ggriffiths: This will result in the CSI driver requiring attachment objects for both windows and linux, which is not needed.

There's no way to add this for windows only, as there's one CSIDriver object in a cluster for Portworx.

drivers/storage/portworx/component/csi.go
ggriffiths: Hold off on adding the attacher sidecar until I see why we need it. I am not convinced we need volume attachment objects and the volume attacher for windows support.

deploy/windows/win.yaml
```diff
@@ -0,0 +1,130 @@
...
+      securityContext: {}
+      terminationGracePeriodSeconds: 30
+      tolerations:
+      - key: os
+        value: Windows
+      volumes:
+        - hostPath:
+            path: \\.\pipe\csi-proxy-filesystem-v1
+            type: ''
+          name: csi-proxy-fs-pipe-v1
```
ggriffiths, line 101: Looks like you copied this code from the SMB CSI Driver:
https://github.com/kubernetes-csi/csi-driver-smb/blob/d18eaf64b2ebd2ec299d1090c4262d02cd0ac161/deploy/csi-smb-node-windows.yaml#L120

Let's remove deadcode we're not using when copying external code

ggriffiths: Same for other volumes below... remove volumes that are not mounted anywhere

drivers/storage/portworx/component/csi.go
pnookala-px: We don't need the above sidecar for now.

drivers/storage/portworx/component/csi.go
pnookala-px: This might not be required.

drivers/storage/portworx/util/csi_generator.go
pnookala-px: Not required for now.

PR #1081, opened 2023-06-21
PWX-31842: PKS mounts-leak fix

zoxpx:
* mounting changing OSD-mount to `-v /var/lib/osd:/var/vcap/store/lib/osd:shared`
* removing offending "pxlogs" (/var/vcap/store/lib/osd/log) mount
* changing `PRE-EXEC` to remove the /var/lib/osd/driver directory
* fixing UTs

ISSUE: the `/var/lib/osd` + `/var/lib/osd/log` mounts were causing the "mounts leak" issue on PKS platform

As a fix, we are:
* we are changing the OSD-mount to `/var/vcap/store/lib/osd`
* removing the `/var/lib/osd/log` mount
* we had to use `PRE-EXEC` to remove /var/lib/osd/driver directory, as the `pwx.sock` file inside causes 2-minute delays for every `pxctl` command

Closes # PWX-31842

### DIFF
--- a/drivers/storage/portworx/testspec/pks.yaml
+++ b/drivers/storage/portworx/testspec/pks.yaml
@@ -51,7 +51,7 @@ spec:
             - name: "PX_TEMPLATE_VERSION"
               value: "v4"
             - name: "PRE-EXEC"
-              value: "if [ ! -x /bin/systemctl ]; then apt-get update; apt-get install -y systemd; fi"
+              value: "rm -fr /var/lib/osd/driver"
           livenessProbe:
             periodSeconds: 30
             initialDelaySeconds: 840 # allow image pull in slow networks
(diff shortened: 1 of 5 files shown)


drivers/storage/portworx/testspec/pks.yaml
```diff
@@ -51,7 +51,7 @@ spec:
             - name: "PX_TEMPLATE_VERSION"
               value: "v4"
             - name: "PRE-EXEC"
-              value: "if [ ! -x /bin/systemctl ]; then apt-get update; apt-get install -y systemd; fi"
+              value: "rm -fr /var/lib/osd/driver"
```
piyush-nimbalkar, line 54: So, we don't need the old PRE-EXEC command?

piyush-nimbalkar (approved):

drivers/storage/portworx/testspec/pks.yaml
zoxpx: That's correct -- it was added initially to support Ubuntu 14 PKS stemcells, which did not have the `systemd` package installed.  These are "long gone" now -- standard support for Ubuntu 14 LTS version ended beginning of 2019.

zoxpx:
Thanks Piyush -- merging this PR, cherry-picking for `px-rel-23.6.0`

PR #1082, opened 2023-06-21
PWX-31931: Only execute pre-flight on fresh PX install.

jrivera-px:
Preflight was being executed on upgrade and was changing the configuration from BTRSF to DMthin when it was not supposed to.    Only execute pre-flight on fresh PX install.  
Closes #
PWX-31931

### DIFF
--- a/pkg/controller/storagecluster/storagecluster.go
+++ b/pkg/controller/storagecluster/storagecluster.go
@@ -640,13 +640,10 @@ func (c *Controller) syncStorageCluster(
 			return fmt.Errorf("preflight check failed for StorageCluster %v/%v: %v", cluster.Namespace, cluster.Name, err)
 		}
 	} else {
-		if cluster.Annotations == nil {
-			cluster.Annotations = make(map[string]string)
-		}
-		// Only set to false if its not set or empty
-		if _, ok := cluster.Annotations[pxutil.AnnotationPreflightCheck]; !ok {
-			cluster.Annotations[pxutil.AnnotationPreflightCheck] = "false"
-		} else if cluster.Annotations[pxutil.AnnotationPreflightCheck] == "" {
+		if _, exists := cluster.Annotations[pxutil.AnnotationPreflightCheck]; !exists {
+			if cluster.Annotations == nil {
+				cluster.Annotations = make(map[string]string)
+			}
 			cluster.Annotations[pxutil.AnnotationPreflightCheck] = "false"
 		}
 	}
--- a/pkg/controller/storagecluster/storagecluster.go
+++ b/pkg/controller/storagecluster/storagecluster.go
@@ -640,6 +640,9 @@ func (c *Controller) syncStorageCluster(
 			return fmt.Errorf("preflight check failed for StorageCluster %v/%v: %v", cluster.Namespace, cluster.Name, err)
 		}
 	} else {
+		if cluster.Annotations == nil {
+			cluster.Annotations = make(map[string]string)
+		}
 		cluster.Annotations[pxutil.AnnotationPreflightCheck] = "false"
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/controller/storagecluster/storagecluster.go
```diff
@@ -631,14 +631,16 @@ func (c *Controller) syncStorageCluster(
...
+	if pxutil.IsFreshInstall(cluster) && pxutil.GetPortworxVersion(cluster).GreaterThanOrEqual(pxVer30) { // Preflight should only run on PX version 3.0.0 and above
 		// If preflight failed, or previous check failed, reconcile would stop here until issues got resolved
 		if err := c.runPreflightCheck(cluster); err != nil {
 			if updateErr := c.updateStorageClusterState(cluster, corev1.ClusterStateDegraded); updateErr != nil {
 				logrus.Errorf("failed to update StorageCluster status. %v", updateErr)
 			}
 			return fmt.Errorf("preflight check failed for StorageCluster %v/%v: %v", cluster.Namespace, cluster.Name, err)
 		}
+	} else {
+		cluster.Annotations[pxutil.AnnotationPreflightCheck] = "false"
```
harsh-px, line 643: why is the explicit setting to false needed now and not before?

jrivera-px: Didn't have it before ... With the new operator we set this to true cause it does not exist.  If we skip pre-flight it never gets set to false and will stay true.   I am explicitly setting it for consistency.  We will skip pre-flight anyways but wanted to make sure the annotation was set.

piyush-nimbalkar (approved):

pkg/controller/storagecluster/storagecluster.go
piyush-nimbalkar: @jrivera-px let's set it to false only if it is not set or value is empty.

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -631,14 +631,19 @@ func (c *Controller) syncStorageCluster(
...
 			if updateErr := c.updateStorageClusterState(cluster, corev1.ClusterStateDegraded); updateErr != nil {
 				logrus.Errorf("failed to update StorageCluster status. %v", updateErr)
 			}
 			return fmt.Errorf("preflight check failed for StorageCluster %v/%v: %v", cluster.Namespace, cluster.Name, err)
 		}
+	} else {
+		if cluster.Annotations == nil {
+			cluster.Annotations = make(map[string]string)
+		}
+		cluster.Annotations[pxutil.AnnotationPreflightCheck] = "false"
```
piyush-nimbalkar: This should change to - 
```
		if _, exists := cluster.Annotations[pxutil.AnnotationPreflightCheck]; !exists {
                         if cluster.Annotations == nil {
              			cluster.Annotations = make(map[string]string)
                         }
                         cluster.Annotations[pxutil.AnnotationPreflightCheck] = "false"
		}
		

```

piyush-nimbalkar: We don't want to set the value if one already exists. Basically, if the user or the operator has already changed the value, we don't want to change it again.

jrivera-px: Actually yeah I did another commit after that .. which changed it like this `if cluster.Annotations == nil {
			cluster.Annotations = make(map[string]string)
		}
		// Only set to false if its not set or empty
		if _, ok := cluster.Annotations[pxutil.AnnotationPreflightCheck]; !ok {
			cluster.Annotations[pxutil.AnnotationPreflightCheck] = "false"
		} else if cluster.Annotations[pxutil.AnnotationPreflightCheck] == "" {
			cluster.Annotations[pxutil.AnnotationPreflightCheck] = "false"
`

PR #1087, opened 2023-06-23
PWX-31603: Initial operator code for PortworxDiag objects

Pure-AdamuKaapan:
Fills in the core logic of the PortworxDiag controller, including updating the status and cluster UUID of the PXDiag, creating the necessary pods, and watching for updates.

Heavily related to https://github.com/portworx/px-installer/pull/1776

Feel free to message with questions or advice, I haven't worked on reconciler-based or k8s-client-based code in a little while so I'm sure it's messy.


pkg/controller/portworxdiag/rbac.go
```diff
@@ -0,0 +1,121 @@
...
+	if err := createRoleIfNotExist(client, ns); err != nil {
+		return err
+	}
+	if err := createRoleBindingIfNotExist(client, ns); err != nil {
+		return err
+	}
+	return nil
+}
+
+func createServiceAccountIfNotExist(c client.Client, ns string) error {
```
piyush-nimbalkar, line 45: We already have utility methods to CreateOrUpdateXXX for different objects 
 in utils. - https://github.com/libopenstorage/operator/blob/master/pkg/util/k8s/k8s.go#L197

Eg usage - https://github.com/libopenstorage/operator/blob/master/drivers/storage/portworx/component/portworx_basic.go#L136
All components use these method to create objects.

pkg/controller/portworxdiag/rbac.go
```diff
@@ -0,0 +1,121 @@
...
+		ObjectMeta: metav1.ObjectMeta{Name: roleName, Namespace: ns},
+		Rules: []rbacv1.PolicyRule{
+			{
+				APIGroups: []string{"portworx.io"},
+				Resources: []string{"portworxdiags", "portworxdiags/status"},
+				Verbs:     []string{"*"},
+			},
+			{
+				APIGroups: []string{"core.libopenstorage.org"},
+				Resources: []string{"storageclusters"},
```
piyush-nimbalkar, line 82: Don't you need to storagenode objects as well? We are listing storagenods if the rpc is not available.
Let's just put * here instead of listing all the objects under `core.libopenstorage.org`.

pkg/controller/portworxdiag/pod.go
```diff
@@ -0,0 +1,135 @@
...
+
+func makeDiagPodTemplate(diag *portworxv1.PortworxDiag, ns string, nodeName string, nodeID string) (*v1.PodTemplateSpec, error) {
+	svcLinks := true
+	terminationGP := int64(10)
+	privileged := true
+
+	diagImage := "docker.io/portworx/oci-monitor"
+	diagTag := "akrpantest" // TODO: change once px-installer PR is merged
+
+	if img := os.Getenv("DIAG_IMAGE"); len(img) > 0 {
```
piyush-nimbalkar, line 87: Let's just have only one env variable called `DIAG_IMAGE` which includes the image tag.

pkg/controller/portworxdiag/pod.go
```diff
@@ -0,0 +1,135 @@
...
+		diagImage = img
+	}
+	if tag := os.Getenv("DIAG_IMAGE_TAG"); len(tag) > 0 {
+		diagTag = tag
+	}
+
+	return &v1.PodTemplateSpec{
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: "portworx-diag",
```
piyush-nimbalkar, line 96: nit: *px-diag for consistency

pkg/controller/portworxdiag/pod.go
```diff
@@ -0,0 +1,135 @@
...
+			Namespace:    ns,
+			Labels: map[string]string{
+				"name": PortworxDiagLabel,
+			},
+		},
+		Spec: v1.PodSpec{
+			NodeName:                      nodeName,
+			HostPID:                       true, // We do need this
+			HostNetwork:                   true, // Do we need this?
```
piyush-nimbalkar, line 105: Can you create a separate ticket to figure out if we need some of these and then remove them as part of that ticket?

pkg/controller/portworxdiag/pod.go
```diff
@@ -0,0 +1,135 @@
...
+			DNSPolicy:                     v1.DNSClusterFirst,
+			EnableServiceLinks:            &svcLinks, // Do we need this?
+			ServiceAccountName:            serviceAccountName,
+			TerminationGracePeriodSeconds: &terminationGP,
+			Tolerations:                   tolerations(),
+			Volumes:                       volumes(),
+			Containers: []v1.Container{
+				{
+					Name:            "diag-collector",
+					Image:           fmt.Sprintf("%s:%s", diagImage, diagTag),
```
piyush-nimbalkar, line 116: Let's use the custom registry specified in the storagecluster. All the other components use this method to add the image registry to the image - https://github.com/libopenstorage/operator/blob/master/pkg/util/util.go#L169


Similarly we need to add the image pull secret if it is specified in the storagecluster spec.

pkg/controller/portworxdiag/pod.go
```diff
@@ -0,0 +1,135 @@
...
+		{Name: "dbusmount", MountPath: "/var/run/dbus"},
+		{Name: "varlibosd", MountPath: "/var/lib/osd", MountPropagation: &mp},
+		{Name: "diagsdump", MountPath: "/var/cores"},
+		{Name: "etcpwx", MountPath: "/etc/pwx"},
+		{Name: "journalmount1", MountPath: "/var/run/log", ReadOnly: true},
+		{Name: "journalmount2", MountPath: "/var/log", ReadOnly: true},
+	}
+}
+
+func tolerations() []v1.Toleration {
```
piyush-nimbalkar, line 67: Can you check if this method gives you all the required tolerations - https://github.com/libopenstorage/operator/blob/master/pkg/util/k8s/pod.go#L18?

pkg/controller/portworxdiag/pod.go
```diff
@@ -0,0 +1,135 @@
...
+		{Key: "node.kubernetes.io/network-unavailable", Operator: v1.TolerationOpExists, Effect: v1.TaintEffectNoSchedule},
+	}
+}
+
+func makeDiagPodTemplate(diag *portworxv1.PortworxDiag, ns string, nodeName string, nodeID string) (*v1.PodTemplateSpec, error) {
+	svcLinks := true
+	terminationGP := int64(10)
+	privileged := true
+
+	diagImage := "docker.io/portworx/oci-monitor"
```
piyush-nimbalkar, line 84: Can we make this constant public and use it - https://github.com/libopenstorage/operator/blob/master/drivers/storage/portworx/portworx.go#L39?

pkg/controller/portworxdiag/pod.go
```diff
@@ -0,0 +1,135 @@
...
+				Path: path,
+				Type: &hpType,
+			},
+		},
+	}
+}
+
+func volumes() []v1.Volume {
+	return []v1.Volume{
+		v1Volume("dockersock", "/var/run/docker.sock", v1.HostPathUnset),
```
piyush-nimbalkar, line 30: Let's check if we need all these volumes and mounts in the diags pod. You can create a separate ticket and then do it later.

piyush-nimbalkar:
@Pure-AdamuKaapan I know this has a lot of comments. Mostly minor and pointing to re-use of existing code. But the overall direction seems correct.

pkg/controller/portworxdiag/portworxdiag.go
```diff
@@ -107,8 +153,351 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+	}
+	err = c.client.Status().Patch(context.TODO(), diag, client.RawPatch(types.JSONPatchType, body))
+	if err != nil {
+		return fmt.Errorf("failed to add empty status object for PortworxDiag CR: %v", err)
+	}
+	return nil
+}
+
+func (c *Controller) patchClusterUUIDIfChanged(diag *diagv1.PortworxDiag, stc *corev1.StorageCluster) error {
+	if diag.Status.ClusterUUID == stc.Status.ClusterUID {
```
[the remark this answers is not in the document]
Pure-AdamuKaapan: Surprisingly, in my testing `add` in the patch worked for both adding a new value and updating the existing one... it doesn't seem right but it was what worked, whereas `replace` only works if it already exists.

pkg/controller/portworxdiag/portworxdiag.go
```diff
@@ -107,8 +153,351 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+}
+
+func (c *Controller) syncPortworxDiag(diag *diagv1.PortworxDiag) error {
+	logrus.Info("Enter syncPortworxDiag")
+	stc, err := c.fetchSTC()
+	if err != nil {
+		logrus.WithError(err).Error("Failed to find StorageCluster object")
+		return fmt.Errorf("failed to find StorageCluster object: %v", err)
+	}
+
```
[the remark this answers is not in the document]
Pure-AdamuKaapan: The Portworx Diag object doesn't have an overall status field (though it may be good to add one). Should we add it, or just set it to failed and maybe add an event instead?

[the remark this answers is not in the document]
piyush-nimbalkar: It would be good to have a message field in `PortworxDiagStatus` and populate that. Let's add that field, mark the phase as Failed and also raise the event.

pkg/controller/portworxdiag/rbac.go
Pure-AdamuKaapan: Discussed offline, we will instead just use the Portworx SA and make sure it has permission to read/write all the portworx v1 objects we need

pkg/controller/portworxdiag/pod.go
Pure-AdamuKaapan: Yes it does! :grin:

pkg/controller/portworxdiag/portworxdiag.go
```diff
@@ -107,8 +153,351 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+			if err != nil {
+				logrus.WithError(err).Warnf("Failed to delete completed diags collection pod, it may still hang around")
+				// Don't exit, keep trying to clean up the rest
+				// TODO: raise an event/warning
+			}
+		}
+	}
+
+	// Finally, update the overall phase with all the latest info
+	err = c.patchPhaseIfChanged(diag)
```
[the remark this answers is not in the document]
Pure-AdamuKaapan: Good point, and pods aren't created instantaneously so it's not like a node's phase will have changed between the two. I'll try to combine them.

pkg/controller/portworxdiag/pod.go
Pure-AdamuKaapan: https://portworx.atlassian.net/browse/PWX-32177

pkg/controller/portworxdiag/pod.go
Pure-AdamuKaapan: https://portworx.atlassian.net/browse/PWX-32179

pkg/controller/portworxdiag/portworxdiag.go
```diff
@@ -68,7 +114,7 @@ func (c *Controller) StartWatch() error {
 	}
 
 	return c.ctrl.Watch(
-		&source.Kind{Type: &portworxv1.PortworxDiag{}},
+		&source.Kind{Type: &diagv1.PortworxDiag{}},
```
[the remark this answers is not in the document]
Pure-AdamuKaapan: I'm having some difficulty with this one and am hoping to sync up with you at some point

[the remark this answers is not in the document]
Pure-AdamuKaapan: Filed https://portworx.atlassian.net/browse/PWX-32193 to tackle this later

drivers/storage/portworx/util/util.go
```diff
@@ -1249,6 +1249,90 @@ func CountStorageNodes(
...
+		}
+
+		nodeNameToNodeID[n.Name] = n.Status.NodeUID
+		nodeIDToNodeName[n.Status.NodeUID] = n.Name
+	}
+
+	return nodeNameToNodeID, nodeIDToNodeName, nil
+}
+
+func GetStorageNodeMapping(
```
ggriffiths, line 1313: nit: add comment for exported method

pkg/controller/portworxdiag/pod.go
```diff
@@ -0,0 +1,144 @@
...
+		v1Volume("procmount", "/proc", v1.HostPathUnset),
+		v1Volume("sysdmount", "/etc/systemd/system", v1.HostPathUnset),
+		v1Volume("dbusmount", "/var/run/dbus", v1.HostPathUnset),
+		v1Volume("varlibosd", "/var/lib/osd", v1.HostPathUnset),
+		v1Volume("diagsdump", "/var/cores", v1.HostPathUnset),
+		v1Volume("etcpwx", "/etc/pwx", v1.HostPathUnset),
+		v1Volume("journalmount1", "/var/run/log", v1.HostPathUnset),
+		v1Volume("journalmount2", "/var/log", v1.HostPathUnset),
+		v1Volume("registration-dir", "/var/lib/kubelet/plugins_registry", v1.HostPathDirectoryOrCreate),
+		v1Volume("csi-driver-path", "/var/lib/kubelet/plugins/pxd.portworx.com", v1.HostPathDirectoryOrCreate),
```
ggriffiths, line 50: this kubelet path here can change based on where we're running - 
https://github.com/libopenstorage/operator/blob/4002f44a87d6ee54765b1a883eae7d5e1e08e175/drivers/storage/portworx/deployment.go#L1475C30-L1475C30

I'm not sure if that matters here, but just FYI

pkg/controller/portworxdiag/portworxdiag.go
```diff
@@ -107,8 +138,394 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+	case diagv1.DiagStatusCompleted:
+		fallthrough
+	case diagv1.DiagStatusFailed:
+		logrus.Infof("PortworxDiag %s in namespace %s is already in status %s, no work to do", diag.Name, diag.Namespace, diag.Status.Phase)
+		return nil
+	default:
+		break
+	}
+
+	logrus.Info("Enter syncPortworxDiag")
```
ggriffiths, line 438: nit: do we still want this as Info? Wondering if this could be spammy. However, if `syncPortworxDiag` is only called on a diag CR change/etc, this should be fine.

ggriffiths (approved):
Looks good, just minor nits and a question

pkg/controller/portworxdiag/portworxdiag.go
Pure-AdamuKaapan: It's called only on changes, or on a 30 second interval I think. However, we have a fast exit for already completed/failed ones, so it's only spammy while the Diag object is still running :wink:

I'll leave it as is while we're still early in development but if it ends up being spammy we can change it later (I have lots more work to do in these files)

pkg/controller/portworxdiag/pod.go
Pure-AdamuKaapan: I think we're hoping to remove this volume/mount entirely so I'll leave it for now but good reminder, thank you!

drivers/storage/portworx/component/portworx_basic.go
```diff
@@ -379,6 +379,11 @@ func (c *portworxBasic) createClusterRole() error {
 					Resources: []string{"*"},
 					Verbs:     []string{"*"},
 				},
+				{
+					APIGroups: []string{"portworx.io"},
+					Resources: []string{"portworxdiags", "portworxdiags/status"},
```
pp511, line 384: Do we need to mention portworxdiags/status explicitly here if we have portworxdiags in the resource list?

drivers/storage/portworx/component/portworx_basic.go
```diff
@@ -379,6 +379,11 @@ func (c *portworxBasic) createClusterRole() error {
...
+					Verbs:     []string{"*"},
```
pp511, line 385: I think we should explicitly mention the list of verbs instead of using wildcard here

drivers/storage/portworx/component/portworx_basic.go
Pure-AdamuKaapan: Yes, I believe that was a change in how Kubernetes handles object RBAC, it's to prevent non-controllers from accidentally updating the status of controlled objects

drivers/storage/portworx/component/portworx_basic.go
Pure-AdamuKaapan: Can do

PR #1089, opened 2023-06-24
PWX-31971 verify pre-flight results

jrivera-px:
The issue we had here was that pre-flight was not running to completion and resulted in an empty pre-flight check results list.  The results list contains any failed checks from pre-flight.    Since the list was empty it was being interpreted as no failed checks and this incorrectly enables DMthin.   Added code to handle cases when the pre-flight pod would return invalid results due to a failed or incomplete run by checking the list and its content more thoroughly. 
Closes #


drivers/storage/portworx/preflight.go
```diff
@@ -295,24 +295,43 @@ func (u *preFlightPortworx) RunPreFlight() error {
...
-		logrus.Infof("storageNode[%s]: %#v ", node.Name, node.Status.Checks)
-		if len(node.Status.Checks) > 1 {
-			for _, check := range node.Status.Checks {
-				if check.Type != "status" {
-					msg := fmt.Sprintf("%s pre-flight check ", check.Type)
-					if check.Success {
-						msg = msg + "passed: " + check.Reason
-						k8sutil.InfoEvent(recorder, u.cluster, util.PassPreFlight, msg)
+	passed := false
+	if len(storageNodes) > 0 {
```
alicelyu-px, line 299: Shall we handle `len(storageNodes) = 0` case as an error? 
The current code would return nil with no action taken, wonder if this is the desired behavior.

alicelyu-px, line 299: Nit: we can use "early returns" or "early continue" to eliminate some of the nesting. 
may also consider factoring out single node status check to shorten this function and improve maintainability and readability.

For example: 
```
func (u *preFlightPortworx) ProcessPreFlightResults(recorder record.EventRecorder, storageNodes []*corev1.StorageNode) error {
	logrus.Infof("pre-flight: process pre-flight results...")

	if len(storageNodes) == 0 {
		logrus.Errorf("pre-flight: storageNodes list empty, unable to process pre-flight results")
		return nil
	}

	passed := u.processNodesChecks(recorder, storageNodes)

	if passed {
		u.processPassedChecks(recorder)
	} else {
		u.processFailedChecks(recorder)
	}

	return nil
}
```

alicelyu-px (approved):
lgtm. Thanks for making the refactor!

PR #1090, opened 2023-06-25
PWX-31987: Add operator version to oci-mon/px environment.

jrivera-px:
Add the operator version as an environment variable to OCI-MON/PX pods.   This will allow use take action based on the Operator version if need be.  
Closes #
PWX-31987

Note: the addition of this will cause a PX restart.   @zoxpx is there a way to prevent the PX restart when someone updates the operator?

### DIFF
--- a/drivers/storage/portworx/deployment.go
+++ b/drivers/storage/portworx/deployment.go
@@ -27,6 +27,7 @@ import (
 	"github.com/libopenstorage/operator/pkg/preflight"
 	"github.com/libopenstorage/operator/pkg/util"
 	k8sutil "github.com/libopenstorage/operator/pkg/util/k8s"
+	opVersion "github.com/libopenstorage/operator/pkg/version"
 	coreops "github.com/portworx/sched-ops/k8s/core"
 )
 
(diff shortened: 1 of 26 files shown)


zoxpx (commented):
I have some reservations regarding this PR.

I think this PR will cause adding `OPERATOR_VERSION` env-variable to ALL OCI-Mon's in parallel:
* this will probably lead to the restarts of the all OCI-Monitor PODs (in parallel) and also
* the restart of all `portworx.service` services on all the nodes (in parallel)

AFAIK, we have a number of customers that do not anticipate that the `portworx.service`s will get restarted when they upgrade the `px-operator`.
* in fact, they need forward warnings, and scheduled maintenance windows to restart their `portworx.service` storage backends

piyush-nimbalkar:
> I have some reservations regarding this PR.
> 
> I think this PR will cause adding `OPERATOR_VERSION` env-variable to ALL OCI-Mon's in parallel:
> 
> * this will probably lead to the restarts of the all OCI-Monitor PODs (in parallel) and also
> * the restart of all `portworx.service` services on all the nodes (in parallel)
> 
> AFAIK, we have a number of customers that do not anticipate that the `portworx.service`s will get restarted when they upgrade the `px-operator`.
> 
> * in fact, they need forward warnings, and scheduled maintenance windows to restart their `portworx.service` storage backends

I agree with @zoxpx here. Let's not trigger a portworx pod restart on operator upgrades. 
If oci-mon needs to know about the operator version, it can directly get it from the operator deployment by query k8s. 

Also, let's understand the use cases that need oci-mon to know about operator version. This creates a two way version dependency where operator already depends on oci-mon version to take certain decisions and now we are opening the conditional logic from oci-mon -> operator.

harsh-px:
Good points above. Jose, let's close this for now and revisit if it becomes a pressing requirement. After recent PXStoreV2 conversations, looks like we are no longer blocked on this.

PR #1096, opened 2023-06-27
Operator default images should not include csi-health-monitor-controller

nikita-bhatia:
https://portworx.atlassian.net/browse/PWX-31944
In an airgapped environment, we were including the csi-health-monitor-controller.
We should never start this container by default. 

Disabling the csi-health-monitor-controller until further notice.


drivers/storage/portworx/util/csi_generator.go
```diff
@@ -203,9 +203,12 @@ func (g *CSIGenerator) GetCSIConfiguration() *CSIConfiguration {
 	}
 
 	// IncludeExternalHealthMonitor only with PX 2.10.0+ and k8s 1.21+
-	if g.kubeVersion.GreaterThanOrEqual(k8sVer1_21) && g.pxVersion.GreaterThanOrEqual(pxVer2_10) {
```
ggriffiths, line 206: Can we instead keep this block, but set to false with a comment above?

ggriffiths (changes requested):
Just a few minor changes

drivers/storage/portworx/util/csi_generator.go
nikita-bhatia: done

ggriffiths (approved):

PR #1101, opened 2023-06-28
PWX-32048: Fixes annotation not being properly added for FACD topology installs

Pure-AdamuKaapan:
What this PR does / why we need it:
Previously the check would initially pass but the annotation wouldn't be added, resulting in failures down the line. This small restructure fixes it all up and adds the annotation before the pre-flight check, which is also more consistent with the entire code flow. Many thanks to @piyush-nimbalkar  !

PWX-32048

### DIFF
--- a/drivers/storage/portworx/portworx.go
+++ b/drivers/storage/portworx/portworx.go
@@ -647,6 +647,22 @@ func (p *portworx) SetDefaultsOnStorageCluster(toUpdate *corev1.StorageCluster) 
 
 	setAutopilotDefaults(toUpdate)
 	setTLSDefaults(toUpdate)
+
+	if pxutil.IsFreshInstall(toUpdate) {
+		// Check for FACD variable on a new install: if so, add an annotation saying
+		// it's okay to continue.
+		facdTopologyEnv := ""
+		for _, env := range toUpdate.Spec.Env {
+			if env.Name == "FACD_TOPOLOGY_ENABLED" {
+				facdTopologyEnv = strings.ToLower(env.Value)
+				break
+			}
+		}
+
+		if facdTopologyEnv == "true" {
+			toUpdate.Annotations[pxutil.AnnotationFACDTopology] = "true"
+		}
+	}
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


drivers/storage/portworx/portworx.go
```diff
@@ -647,6 +647,22 @@ func (p *portworx) SetDefaultsOnStorageCluster(toUpdate *corev1.StorageCluster)
...
+		facdTopologyEnv := ""
+		for _, env := range toUpdate.Spec.Env {
+			if env.Name == "FACD_TOPOLOGY_ENABLED" {
+				facdTopologyEnv = strings.ToLower(env.Value)
+				break
+			}
+		}
+
+		if facdTopologyEnv == "true" {
+			toUpdate.Annotations[pxutil.AnnotationFACDTopology] = "true"
```
piyush-nimbalkar, line 663: Should we make sure that annotations is not nil?
```
if toUpdate.Annotations == nil {
  toUpdate.Annotations = make(map[string]string)
}
toUpdate.Annotations[pxutil.AnnotationsFACDTopology] = "true"
```

drivers/storage/portworx/portworx.go
```diff
@@ -745,31 +761,20 @@ func (p *portworx) validateFACDTopology(cluster *corev1.StorageCluster) error {
...
+	if facdTopologyEnv != "true" {
+		return nil
+	}
 
-		if !isValid {
-			// Fail the request!
-			return fmt.Errorf("FACD topology cannot be enabled on existing clusters")
+	// FACD topology is enabled. See if we already have the annotation marking this as valid.
+	// TODO: once we support upgrading to FACD topology, it will also be valid after whatever version supports that
+	if cluster.Annotations != nil {
```
piyush-nimbalkar, line 770: nit: We don't need the nil check. Golang allows you to query a nil map.
The following line won't fail even if `cluster.Annotations` are nil and would return the correct output.
```
if v, ok := cluster.Annotations[pxutil.AnnotationFACDTopology]; ok && v == "true" { ... }
```

piyush-nimbalkar (approved):
Any UTs that needed updating?

drivers/storage/portworx/portworx.go
Pure-AdamuKaapan: That gets guaranteed at the top of the function luckily, so we're good: https://github.com/libopenstorage/operator/pull/1101/files#diff-cc28f01bf9bf855945fe56dc2e65f61cf8e914889d3f93bad610daf12dff8f4eR399-R401

drivers/storage/portworx/portworx.go
Pure-AdamuKaapan: Oh good point, I was nervous cuz I'd hit an "assignment to entry in nil map" panic before, but that's assignment, not reading.

Pure-AdamuKaapan:
Lol, yes, unit tests need updating apparently. Will fix that nit and the tests.

piyush-nimbalkar (approved):

PR #1102, opened 2023-06-28
PWX-32073: Update csi-provisioner operation timeout from 10s to 5m

ggriffiths:
cherry-pick of https://github.com/libopenstorage/operator/pull/1099

### DIFF
--- a/drivers/storage/portworx/component/csi.go
+++ b/drivers/storage/portworx/component/csi.go
@@ -598,6 +598,7 @@ func getCSIDeploymentSpec(
 			"--leader-election=true",
 			"--default-fstype=ext4",
 			"--extra-create-metadata=true",
+			"--timeout=5m",
 		}
 	} else {
 		args = []string{


drivers/storage/portworx/component/csi.go
```diff
@@ -598,6 +598,7 @@ func getCSIDeploymentSpec(
 			"--leader-election=true",
 			"--default-fstype=ext4",
 			"--extra-create-metadata=true",
+			"--timeout=5m",
```
adityadani, line 601: Since we maintain two different CSI provisioner sidecars, can you double check if this option will work for the older csi-provisioner image? This is to ensure, if a customer on PX 2.12 upgrades their operator to 23.5.1

adityadani (approved):

drivers/storage/portworx/component/csi.go
ggriffiths: Yeah, this timeout flag has been there since provisioner 1.1.0:
https://github.com/kubernetes-csi/external-provisioner/blame/v1.1.0/cmd/csi-provisioner/csi-provisioner.go#L60

PR #1111, opened 2023-06-30
PWX-32127: Only add telemetry phonehome volume if px is at least 3.0

ezhang-px:
Currently change to introduce telemetry phonehome volume should only be applied to cluster greater than or equal to 3.0, otherwise we'll have regression failure for cluster running with Java CCM when px-telemetry-phonehome configmap is not part of the cluster

Closes #

### DIFF
--- a/drivers/storage/portworx/deployment.go
+++ b/drivers/storage/portworx/deployment.go
@@ -1282,7 +1282,11 @@ func (t *template) getVolumeMounts() []v1.VolumeMount {
 		t.getPKSVolumeInfoList,
 		t.getBottleRocketVolumeInfoList,
 		t.GetVolumeInfoForTLSCerts,
-		t.getTelemetryPhoneHomeVolumeInfoList,
+	}
+	// Only add telemetry phonehome volume mount if PX is at least 3.0
+	pxVer30, _ := version.NewVersion("3.0")
+	if t.pxVersion.GreaterThanOrEqual(pxVer30) {
+		extensions = append(extensions, t.getTelemetryPhoneHomeVolumeInfoList)
 	}
 	for _, fn := range extensions {
 		volumeInfoList = append(volumeInfoList, fn()...)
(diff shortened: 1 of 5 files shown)


drivers/storage/portworx/deployment.go
```diff
@@ -1282,7 +1282,11 @@ func (t *template) getVolumeMounts() []v1.VolumeMount {
 		t.getPKSVolumeInfoList,
 		t.getBottleRocketVolumeInfoList,
 		t.GetVolumeInfoForTLSCerts,
-		t.getTelemetryPhoneHomeVolumeInfoList,
+	}
+	// Only add telemetry phonehome volume mount if PX is at least 3.0
+	pxVer30, _ := version.NewVersion("3.0")
+	if t.pxVersion.GreaterThanOrEqual(pxVer30) {
+		extensions = append(extensions, t.getTelemetryPhoneHomeVolumeInfoList)
```
harsh-px, line 1289: Can we add UT coverage for this? This function is getting complex

harsh-px (changes requested):

drivers/storage/portworx/deployment.go
```diff
@@ -1344,7 +1348,11 @@ func (t *template) getVolumes() []v1.Volume {
 		t.getPKSVolumeInfoList,
 		t.getBottleRocketVolumeInfoList,
 		t.GetVolumeInfoForTLSCerts,
-		t.getTelemetryPhoneHomeVolumeInfoList,
+	}
+	// Only add telemetry phonehome volume if PX is at least 3.0
+	pxVer30, _ := version.NewVersion("3.0")
+	if t.pxVersion.GreaterThanOrEqual(pxVer30) {
+		extensions = append(extensions, t.getTelemetryPhoneHomeVolumeInfoList)
```
harsh-px, line 1355: Please add UT coverage.

harsh-px (approved):

PR #1112, opened 2023-06-30
PWX-32131: Remove ccm-phonehome-config mount.  This is a mount which …

jrivera-px:
…does not exists yet.

The addition of this mount to the spec for fixing PWX-31626.  Caused pre-flight to fail cause the mount location does not exist when pre-flight is run.   Remove it.  
Closes #
PWX-32131
This looks like more of a change than it is.    I changed the "RunPreFlight" function by pulled out the code which creates the preflight daemonset and moved into a separate function called "CreatePreFlightDaemonsetSpec" and just call that from "RunPreFlight".   This allowed me to use the "CreatePreFlightDaemonsetSpec" in a UT.

### DIFF
--- a/drivers/storage/portworx/preflight.go
+++ b/drivers/storage/portworx/preflight.go
@@ -208,6 +208,14 @@ func (u *preFlightPortworx) RunPreFlight() error {
 		logrus.Infof("runPreflight: running pre-flight with existing PX-StoreV2 param, hard fail check enabled")
 	}
 
+	// Remove phone-home cm volume mount, does not exist yet
+	for i, volumeMount := range preflightDS.Spec.Template.Spec.Containers[0].VolumeMounts {
+		if volumeMount.Name == "ccm-phonehome-config" {
+			preflightDS.Spec.Template.Spec.Containers[0].VolumeMounts = append(preflightDS.Spec.Template.Spec.Containers[0].VolumeMounts[:i],
+				preflightDS.Spec.Template.Spec.Containers[0].VolumeMounts[i+1:]...)
+			break
+		}
+	}
 	// Add pre-flight param
 	preflightDS.Spec.Template.Spec.Containers[0].Args = append([]string{"--pre-flight"},
 		preflightDS.Spec.Template.Spec.Containers[0].Args...)
--- a/drivers/storage/portworx/preflight.go
+++ b/drivers/storage/portworx/preflight.go
@@ -47,6 +47,8 @@ type PreFlightPortworx interface {
 	ProcessPreFlightResults(recorder record.EventRecorder, storageNodes []*corev1.StorageNode) error
 	// DeletePreFlight deletes the pre-flight daemonset
 	DeletePreFlight() error
+	// CreatePreFlightDaemonsetSpec is used to create the pre-fligh daemonset pod spec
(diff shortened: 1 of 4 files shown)


ezhang-px (approved):

drivers/storage/portworx/preflight.go
```diff
@@ -208,6 +208,14 @@ func (u *preFlightPortworx) RunPreFlight() error {
...
 	}
 
+	// Remove phone-home cm volume mount, does not exist yet
+	for i, volumeMount := range preflightDS.Spec.Template.Spec.Containers[0].VolumeMounts {
+		if volumeMount.Name == "ccm-phonehome-config" {
+			preflightDS.Spec.Template.Spec.Containers[0].VolumeMounts = append(preflightDS.Spec.Template.Spec.Containers[0].VolumeMounts[:i],
+				preflightDS.Spec.Template.Spec.Containers[0].VolumeMounts[i+1:]...)
+			break
+		}
+	}
```
harsh-px, line 218: is this covered by existing UTs?

jrivera-px: No not yet... Adding one now.

harsh-px: Instead of this volume mount getting added and then removed, can you explore not adding it in the first place itself? This will improve debugability if someone  is trying to figure out the spec that is used for pre-flight where they just need to look at one function (the getter for the pod spec).

jrivera-px: Not now.. I just spent time.. .adding a unit test.

ezhang-px (approved):

drivers/storage/portworx/preflight.go
```diff
@@ -208,6 +155,14 @@ func (u *preFlightPortworx) RunPreFlight() error {
 		logrus.Infof("runPreflight: running pre-flight with existing PX-StoreV2 param, hard fail check enabled")
 	}
 
+	// Remove phone-home cm volume mount, does not exist yet
+	for i, volumeMount := range preflightDS.Spec.Template.Spec.Containers[0].VolumeMounts {
```
piyush-nimbalkar, line 159: Don't we need to remove the volume from the pod spec?

piyush-nimbalkar (commented):
I probably missed this earlier, but we are using GetStoragePodSpec to get the pre-check pod spec. This will include the csi driver registrar container too, right? I don't think we need to run that.

Should we just have separate method to create the pre-install pod spec which reuses most of the code already written to generate the storage pod spec?

drivers/storage/portworx/preflight.go
jrivera-px: Yes, we are running the CSI pod also and did plan on removing that later but don't want to that in this release.  So yeah when I look into that I can look into just using functions in the deployment code to just get what we need.   

The volume is there yeah but we don't have to remove the volume so long as its not mounted its ok.    But looking at the code now .. @harsh-px suggested trying "explore not adding the mounts in the first place".  So seeing if that can be done using the pre-flight annotation.

jrivera-px: @piyush-nimbalkar @harsh-px changed the code to use the pre-flight annotation to not add the phonehome configmap.     Let me know if you see any issues.  thx

piyush-nimbalkar (approved):

zoxpx:
Thanks Jose

PR #1114, opened 2023-07-06
PWX-31968: Pass in port to be used for readiness endpoint when runnin…

jrivera-px:
…g the pre-flight pod.  The port will be used to determine when pre-flight pod is up and running.  Will not conflict with standard oci endpoint ports

The pre-flight pod is the oci-mon pod running in a different "pre-flight" enabled by a startup parameter. The issue we have is the pre-flight pod inherited all the runtime oci-mon ports which meant that if oci-mon pod was running when pre-flightpod was running the the readiness connection for pre-flight was connecting to the oci-mon pod and causing inaccurate pre-flight readiness values. The fix is to use another port and when we create the pre-flight pod spec in the operator we are passing that port in now as a parameter to the pre-flight pod startup.  
 
Closes #
PWX-31968
This is dependant on an installer  change also for the same PWX:  https://github.com/portworx/px-installer/pull/1816


drivers/storage/portworx/preflight.go
```diff
@@ -208,8 +209,12 @@ func (u *preFlightPortworx) RunPreFlight() error {
 		logrus.Infof("runPreflight: running pre-flight with existing PX-StoreV2 param, hard fail check enabled")
 	}
 
+	preFltEndPtPort := component.GetCCMListeningPort(u.cluster) + 1 // preflight endpoint port  +1 CCM uploader port
+	logrus.Infof("runPreflight: Setting port: %d", preFltEndPtPort)
+	preflightDS.Spec.Template.Spec.Containers[0].ReadinessProbe.ProbeHandler.HTTPGet.Port = intstr.FromInt(preFltEndPtPort)
```
zoxpx, line 214: wow.. you're putting a lot of faith in this code -- if any of the 8 `.<object>` happens to be `nil` -- this will cause a panic.

Are we guaranteed that all sub-objects can't be `nill` (and also `len(.Containers) > 1`)?

jrivera-px: This is the OCI-Mon container spec created by the deployment.go code ensures those exists.   I can add a routine with checks but the main thing I want to do is change the port.  In fact there will be a change coming that just generate pre-flight spec using the same routines that generate the OCI-Mon spec.   So we will be will basically only starting one container.. cause right now we also start the csi container along with this.   I will add some checks in there to be safe 👍.

jrivera-px: Oh and the initial objects "preflightDS.Spec.Template.Spec" are actually created above.   So I will check the passed in podSpec.. Thanks Z...

zoxpx (approved):
Thanks Jose

jrivera-px:
Thanks dude...

PR #1115, opened 2023-07-06
PWX-32131: Remove ccm-phonehome-config mount.  This is a mount which …

jrivera-px:
…… (#1112)

* PWX-32131: Remove ccm-phonehome-config mount.  This is a mount which does not exists yet.

* WX-32131: Add to TestValidate routine to make sure that the phonehome cm is removed from pre-flight pod spec.  Expose routine to get the pre-flight podSpec and inspect it in the UT.

* WX-32131: missing file from previous commit.

* PWX-32131: Use pre-flight annotation setting to skip adding phonehome volume.

---------

Closes #

This is cherry-pick from the release branch

### DIFF
--- a/drivers/storage/portworx/portworx_test.go
+++ b/drivers/storage/portworx/portworx_test.go
@@ -163,6 +163,45 @@ func TestValidate(t *testing.T) {
 	require.Contains(t, <-recorder.Events,
 		fmt.Sprintf("%v %v %s", v1.EventTypeNormal, util.PassPreFlight, "Enabling PX-StoreV2"))
 	require.Contains(t, *cluster.Spec.CloudStorage.SystemMdDeviceSpec, DefCmetaData)
+
+	//
+	// Validate Pre-flight Daemonset Pod Spec
+	//
+	cluster.Spec.Image = "portworx/oci-image:3.0.0"
+	cluster.Annotations = map[string]string{
+		pxutil.AnnotationPreflightCheck: "true",
+	}
+	err = driver.SetDefaultsOnStorageCluster(cluster)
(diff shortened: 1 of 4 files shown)


drivers/storage/portworx/portworx_test.go
```diff
@@ -163,6 +163,45 @@ func TestValidate(t *testing.T) {
...
+	err = driver.SetDefaultsOnStorageCluster(cluster)
+	require.NoError(t, err)
+
+	podSpec, err := driver.GetStoragePodSpec(cluster, "")
+	require.NoError(t, err)
+
+	// phone-home cm volume mount, should not exists since pre-flight is enabled
+	var errChk error
+	for _, volumeMount := range podSpec.Containers[0].VolumeMounts {
+		if volumeMount.Name == "ccm-phonehome-config" {
```
zoxpx, line 183: nit: you could replace the block with the assert-check (also, check all the containers) -- e.g. :

```go
	// phone-home cm volume mount, should not exists since pre-flight is enabled
	for _, cnt := range podSpec.Containers {
	    for _, volumeMount := range cnt.VolumeMounts {
		require.True(t, volumeMount.Name != "ccm-phonehome-config")
            }
        }
        // also potentially
        require.Equals(t, 1, len(podSpec.Containers))
```

zoxpx (approved):
I think this looks OK -- thanks Jose.

Some minor remarks (optimization suggestions) listed below.

drivers/storage/portworx/portworx_test.go
jrivera-px: I can make that change .. thanks @zoxpx

PR #1117, opened 2023-07-07
Missing commits 23.7.0

jrivera-px:

Missing commits from px-rel-23.5.1
PWX-32096: Only execute pre-flight if running on EKS. (#1107)
Only add telemetry phonehome volume if px is at least 3.0 (#1113)
 PWX-32131: Remove ccm-phonehome-config mount.  This is a mount which … (#1112)

Closes #


ezhang-px (approved):
LGTM

PR #1122, opened 2023-07-11
Update  "replaces" version to 23.5.0 for 23.5.1 release

nikita-bhatia:

Closes #

### DIFF
--- a/deploy/olm-catalog/portworx/23.5.1/portworx-certified.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/portworx/23.5.1/portworx-certified.clusterserviceversion.yaml
@@ -54,7 +54,7 @@ spec:
   version: 23.5.1
   minKubeVersion: "1.21.0"
   maturity: stable
-  replaces: portworx-operator.v23.4.0
+  replaces: portworx-operator.v23.5.0
   maintainers:
   - name: Portworx
     email: support@portworx.com


sharma-tapas (approved):

zoxpx:
Thanks for fixing this Nikita

nikita-bhatia:
> 23.5.0 had issues with preflight, so we had decided to purposely skip it. That's we put replaces as 23.4.0 so that users on 23.4.0 will directly get upgraded to 23.5.1.
> 
> Can you confirm with the QA team on how the upgrade hops were tested?


@piyush-nimbalkar we were hitting following issue when trying to update with replaces as 23.4.0, 

ERROR: /root/workspace/Release/px-operator/operator/deploy/olm-catalog/portworx/23.5.1/portworx-certified.clusterserviceversion.yaml file has wrong previous version in 'replaces' parameter!
Expected version: 23.5.0
Actual version:   replaces: portworx-operator.v23.4.0

piyush-nimbalkar:
> > 23.5.0 had issues with preflight, so we had decided to purposely skip it. That's we put replaces as 23.4.0 so that users on 23.4.0 will directly get upgraded to 23.5.1.
> > Can you confirm with the QA team on how the upgrade hops were tested?
> 
> @piyush-nimbalkar we were hitting following issue when trying to update with replaces as 23.4.0,
> 
> ERROR: /root/workspace/Release/px-operator/operator/deploy/olm-catalog/portworx/23.5.1/portworx-certified.clusterserviceversion.yaml file has wrong previous version in 'replaces' parameter! Expected version: 23.5.0 Actual version: replaces: portworx-operator.v23.4.0

I discussed this with Maksym. We should have passed 23.4.0 as the previous version when starting the release pipeline to avoid this error. The replaces in the CSV and the above assertion would have then matched. 
We should do this only when we know that we have to skip a particular release due to some bug.

PR #1131, opened 2023-07-12
PWX-29718: Prevents multiple diags from running at once

Pure-AdamuKaapan:
If multiple PortworxDiag objects are created, we only want to run one at once as concurrent pxctl diags requests aren't yet supported.

Eventually we may make this more detailed to handle this on a node-level basis, but for now this is enough as multiple diag objects at once is probably pretty uncommon.

Also does a minor fix for me calling the k8s event function wrong.

PWX-29718


ggriffiths (approved):
lgtm

PR #1133, opened 2023-07-12
PWX-32297: Use version extracted from manifest URL.

jrivera-px:
Use the version extracted from manifest URL as the PX version.
Closes #
PWX-32297
Need to add Unit test for the "GetPortworxVersion".


nikolaypopov (approved):
LGTM

drivers/storage/portworx/util/util.go
```diff
@@ -554,22 +556,35 @@ func GetPortworxVersion(cluster *corev1.StorageCluster) *version.Version {
...
-					pxVersionStr = getPortworxVersionFromManifestURL(manifestURL)
-					pxVersion, err = version.NewSemver(pxVersionStr)
-					if err != nil {
-						logrus.Warnf("Invalid PX version %s extracted from release manifest url: %v", pxVersionStr, err)
-					}
-				}
-			}
+			pxVersion = nil
+		} else {
+			logrus.Infof("Using version extracted from image name: %s", pxVersionStr)
```
nikolaypopov, line 561: Let return here if we got the version, no need to check annotation or PX release manifest

jrivera-px: 👍

nikolaypopov (approved):

PR #1137, opened 2023-07-13
PWX-31968: Pass in port to be used for readiness endpoint when runnin…

jrivera-px:
…… (#1114)

* PWX-31968: Pass in port to be used for readiness endpoint when running the pre-flight pod.  The port will be used to determine when pre-flight pod is up and running.  Will not conflict with standard oci endpoint ports

* PWX-31968: Add validity checks to the code before accessing objects.  Also '--pre-flight-port' will only work with upgrade Oci/PX so add version check.

* PWX-32131: Fix compile issues after merging in latest from master

---------

Cherry-pick from master.
Closes #
PWX-31968


zoxpx (approved):
Thanks Jose

PR #1141, opened 2023-07-16
PWX-27672: Reducing Privileged-requirements for Portworx PODs

zoxpx:
* Portworx PODs (oci-monitors) no longer using `privileged:true` - but using 5 capabilities instead
* adding new `STC.security.privileged = true` switch, to roll back to the original `privileged:true` security-setting
* we're also disabling all Bidirectional-mounts, and converting them to regular-mounts  (warning will indicate to change STC...privileged=true to enable back)

Reducing the privileges required to run the Portworx PODs
* no longer require `privileged:true`, but using 5 targeted capabilities instead

CAVEATS:
* note that this changes the DEFAULT how the PX is being installed
  - to mitigate the risks, we have added a simple `STC.security.privileged = true` switch that can revert to the old default
* also using `container.apparmor.security.beta.kubernetes.io/portworx = unconfirmed` annotation, to turn off the AppArmor profiles (i.e. Ubuntu/SuSe.. platforms), as they would interfere with Px install
* finally, automatically converting all `Bidirectional` (shared) mounts into the regular mounts
  - the warning message will hint at the `STC...privileged = true` switch, to restore the use of the Bidirectional mounts if required

Closes # PWX-27672

Note, the Portworx POD-spec now looks like this:

```yaml
apiVersion: v1
kind: Pod
metadata:
  annotations:
    container.apparmor.security.beta.kubernetes.io/portworx: unconfined
spec:
  containers:
  - name: portworx
    image: docker.io/portworx/oci-monitor:2.13.7
    securityContext:
      privileged: false
      capabilities:
        add:
        - SYS_ADMIN
        - SYS_PTRACE
        - SYS_RAWIO
        - SYS_MODULE
        - LINUX_IMMUTABLE
...
```
* note, `privileged: false` and using 5 distinct capabilities
* new `AppArmor` annotation to turn off AppArmor profiles (AppArmor was normally enforced for all non-privileged containers, but it blocks the PX installs)

---

UPDATE:

> To switch to the old behavior/defaults, use the new `privileged:true` switch:
> 
> ```yaml
> apiVersion: core.libopenstorage.org/v1
> kind: StorageCluster
> spec:
>   security:
>     privileged: false
> ...
> ```
> * using this switch will restore our old default w/ using `privileged: true` for our containers
> * it also restores the use of the Bidirectional (shared) mounts

This has been **modified** w/ https://github.com/libopenstorage/operator/pull/1141/commits/6bad619e9b476d77ff87c2a4dfdca533e40eaaf3. as requested by the product management !!!
* we will keep the old/original security setup (privileged oci-mon) as a default
* the NEW security-setup can be activated via `portworx.io/privileged: "false"` annotation

### DIFF
--- a/pkg/apis/core/v1/storagecluster.go
+++ b/pkg/apis/core/v1/storagecluster.go
@@ -199,10 +199,9 @@ type CSITopologySpec struct {
 
 // SecuritySpec is used to define the security configuration for a cluster.
 type SecuritySpec struct {
-	Enabled    bool      `json:"enabled,omitempty"` // CHECKME: remove this flag?
-	Privileged bool      `json:"privileged,omitempty"`
-	Auth       *AuthSpec `json:"auth,omitempty"`
-	TLS        *TLSSpec  `json:"tls,omitempty"`
+	Enabled bool      `json:"enabled,omitempty"`
+	Auth    *AuthSpec `json:"auth,omitempty"`
+	TLS     *TLSSpec  `json:"tls,omitempty"`
 }
 
 // AuthSpec lets the user define authorization (RBAC) configurations
--- a/pkg/apis/core/v1/storagecluster.go
+++ b/pkg/apis/core/v1/storagecluster.go
@@ -199,9 +199,10 @@ type CSITopologySpec struct {
 
 // SecuritySpec is used to define the security configuration for a cluster.
 type SecuritySpec struct {
-	Enabled bool      `json:"enabled,omitempty"`
-	Auth    *AuthSpec `json:"auth,omitempty"`
-	TLS     *TLSSpec  `json:"tls,omitempty"`
+	Enabled    bool      `json:"enabled,omitempty"` // CHECKME: remove this flag?
+	Privileged bool      `json:"privileged,omitempty"`
+	Auth       *AuthSpec `json:"auth,omitempty"`
+	TLS        *TLSSpec  `json:"tls,omitempty"`
 }
 
 // AuthSpec lets the user define authorization (RBAC) configurations
--- a/pkg/controller/storagecluster/storagecluster.go
+++ b/pkg/controller/storagecluster/storagecluster.go
@@ -88,10 +88,13 @@ const (
 var _ reconcile.Reconciler = &Controller{}
 
 var (
-	controllerKind       = corev1.SchemeGroupVersion.WithKind("StorageCluster")
-	crdBaseDir           = getCRDBasePath
-	deprecatedCRDBaseDir = getDeprecatedCRDBasePath
-	dmThinRexp           = regexp.MustCompile(`-T *dmthin`)
+	controllerKind        = corev1.SchemeGroupVersion.WithKind("StorageCluster")
+	crdBaseDir            = getCRDBasePath
+	deprecatedCRDBaseDir  = getDeprecatedCRDBasePath
+	dmThinRexp            = regexp.MustCompile(`-T *dmthin`)
+	containersAppArmorOff = map[string]bool{
+		"portworx": true,
+	}
 )
 
 // Controller reconciles a StorageCluster object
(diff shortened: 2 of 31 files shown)


pkg/apis/core/v1/storagecluster.go
```diff
@@ -199,9 +199,10 @@ type CSITopologySpec struct {
 
 // SecuritySpec is used to define the security configuration for a cluster.
 type SecuritySpec struct {
-	Enabled bool      `json:"enabled,omitempty"`
-	Auth    *AuthSpec `json:"auth,omitempty"`
-	TLS     *TLSSpec  `json:"tls,omitempty"`
+	Enabled    bool      `json:"enabled,omitempty"` // CHECKME: remove this flag?
+	Privileged bool      `json:"privileged,omitempty"`
```
harsh-px, line 203: Let's not add a first class spec parameter to enable a back door to go back to the previous method. 

Okay to keep a new annotation so achieve the same effect.

harsh-px:
What is the testing required for the unidirectional mounts?

pkg/apis/core/v1/storagecluster.go
zoxpx: Just to check if I understood correctly -- you'd want to move the `Security: Privileged` tag into the annotation?

Note that this is not a backdoor -- I picked this method since it is the "closest" to how people normally enable the privileged PODs, which is...
```
spec:
   containers:
      securityContext:
         privileged: true
```

zoxpx:
> What is the testing required for the unidirectional mounts?

I'm hoping the PR will be pretty easy to test -- since PX will either fail to install/work, or it'll work "as usual" (i.e. expecting "catastrophic errors" only - which will be easy to spot)

Regarding the ~Bidirectional~ -> Unidirectional mounts change -
* this means that people won't be able to _"explicitly mount pxd-volumes from the inside of the `portworx.service` container to /var/lib/osd/..."_ --
   - so `pxctl mount testVol --path /var/liib/osd/mounts/testVol` won't actually work any more (mount-action will throw _"not a shared mount"_ error)
   - sharing socket-files, dumping logs, etc.. this should all work as usual
* the Kubernetes-mounted volumes should still work, as these are being set up on the `portworx.service` level

Sadly-- there is no workaround for this.  Kubernetes lets only `privileged` PODs use Bidirectional mounts -- so if one needs to set them up, we'll need that `privileged:true` configuration option (or, annotation?)

zoxpx:
**UPDATE**:
As of commit [6bad619](https://github.com/libopenstorage/operator/pull/1141/commits/6bad619e9b476d77ff87c2a4dfdca533e40eaaf3) -- setting up the `privileged:false` is NO LONGER DEFAULT:

To activate the new behavior w/ `privileged:false`, use the following annotation:
```yaml
kind: StorageCluster
metadata:
  annotations:
    portworx.io/privileged: "false"
```
* by default, we will use normal security defaults, as we normally do (meaning, `privileged` portworx POD)

zoxpx:
Ping?

nikita-bhatia (approved):
Changes looks good to me

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -1237,27 +1240,37 @@ func (c *Controller) CreatePodTemplate(
...
-			Namespace: cluster.Namespace,
-			Labels:    c.StorageClusterSelectorLabels(cluster),
+			Namespace:   cluster.Namespace,
+			Labels:      c.StorageClusterSelectorLabels(cluster),
+			Annotations: make(map[string]string),
 		},
 		Spec: podSpec,
 	}
 
+	if !pxutil.IsPrivileged(cluster) {
```
piyush-nimbalkar, line 1250: Can we add a UT for this too?

piyush-nimbalkar (approved):

zoxpx:
note, rebased against latest `master`
* since we got follow-up PR https://github.com/libopenstorage/operator/pull/1181 lined up, we should be ready to commit the changes

zoxpx:
Woohoo.. Travis passed OK -- merging the PR, will cherry-pick into `px-rel-23.7.0` mixed with [#1181](https://github.com/libopenstorage/operator/pull/1181)

PR #1145, opened 2023-07-17
PTX-18776 Improving validating for stork-scheduler image

nikolaypopov:
Improving validating for stork-scheduler image


jrivera-px (approved):
LGTM...

PR #1146, opened 2023-07-18
PWX-31068: Add GCE/Azure cloud provider to storage spec so it can be picked…

jrivera-px:
… up by PX and sent via callhome.

Send the GCE/Azure cloud provider tag to PX so it can be included in callhome data.

Closes #
PWX-31068
This dependant on PX PR:  https://github.com/portworx/porx/pull/11536.    AWS and VSHPHERE code already exists and unit test indicate that it will work also.


pkg/preflight/utils.go
```diff
@@ -9,6 +9,16 @@ func IsEKS() bool {
 	return Instance().ProviderName() == string(cloudops.AWS) && Instance().K8sDistributionName() == eksDistribution
 }
 
+// IsGKE returns whether the cloud environment is running GKE
+func IsGKE() bool {
+	return Instance().ProviderName() == string(cloudops.GCE) && Instance().K8sDistributionName() == gkeDistribution
+}
+
+// IsAzure() returns whether the cloud environment is running Azure
+func IsAzure() bool {
```
piyush-nimbalkar, line 18: I believe we don't need this after the above comment. The ProviderName already returns the azure provider. In future we can have a method to check if AKS or not.

drivers/storage/portworx/portworx.go
```diff
@@ -430,6 +430,10 @@ func (p *portworx) SetDefaultsOnStorageCluster(toUpdate *corev1.StorageCluster)
 
 	if preflight.IsEKS() {
 		toUpdate.Annotations[pxutil.AnnotationIsEKS] = "true"
+	} else if preflight.IsGKE() {
+		toUpdate.Annotations[pxutil.AnnotationIsGKE] = "true"
+	} else if preflight.IsAzure() {
+		toUpdate.Annotations[pxutil.AnnotationIsAKS] = "true"
```
piyush-nimbalkar, line 436: Azure does not mean AKS. It could be openshift on Azure or just vanilla k8s deployment.

drivers/storage/portworx/deployment_test.go
```diff
@@ -1757,6 +1757,117 @@ func TestPodSpecWithCloudStorageSpecOnEKS(t *testing.T) {
...
+
+	// Reset preflight for other tests
+	coreops.SetInstance(coreops.New(fakek8sclient.NewSimpleClientset()))
+	err = preflight.InitPreflightChecker(k8sClient)
+	require.NoError(t, err)
+}
+
+func TestPodSpecWithCloudStorageSpecOnAzure(t *testing.T) {
+	fakeK8sNodes := &v1.NodeList{Items: []v1.Node{
+		{ObjectMeta: metav1.ObjectMeta{Name: "node1"}, Spec: v1.NodeSpec{ProviderID: "azure://node-id-1"}},
```
piyush-nimbalkar, line 1826: Have we verified this in an azure cluster that it sets the providerID to `azure://...`?

piyush-nimbalkar (changes requested):

drivers/storage/portworx/portworx.go
jrivera-px: Ok then will add a way to just add the the provider if it exists.

drivers/storage/portworx/deployment_test.go
jrivera-px: I didn't have an azure setup but this this test code is based off existing azure test code for "pkg/preflight/preflight_test.go".  Which I am assuming works.    Based on that the provider should get set.

pkg/preflight/utils.go
jrivera-px: Yes, that makes sense.

drivers/storage/portworx/portworx.go
```diff
@@ -432,6 +432,19 @@ func (p *portworx) SetDefaultsOnStorageCluster(toUpdate *corev1.StorageCluster)
...
+
+	if len(preflight.Instance().ProviderName()) > 0 {
+		if toUpdate.Spec.CloudStorage == nil {
+			toUpdate.Spec.CloudStorage = &corev1.CloudStorageSpec{}
+		}
+		// Cloud provider exists, if not set then set it
+		if toUpdate.Spec.CloudStorage.Provider == nil || len(*toUpdate.Spec.CloudStorage.Provider) == 0 {
+			cloudProvider := preflight.Instance().ProviderName()
+			toUpdate.Spec.CloudStorage.Provider = &cloudProvider
+		}
```
piyush-nimbalkar, line 447: Do we need this to be set in the storage cluster spec? We already send the cloud provider when we create the portworx pod spec. We use the same `isXXX` annotations to do that - https://github.com/libopenstorage/operator/blob/master/drivers/storage/portworx/deployment.go#L948

jrivera-px: Yeah we need to set this here.   The  function which ends up getting called https://github.com/libopenstorage/operator/blob/dd9ed40f2b159016161d1247b435260edb88f4f8/drivers/storage/portworx/deployment.go#L825 "getCloudProvider()" only uses annotation to determine the provider.     Also Harsh told me to not use Annotation and decouple that cause it will be difficult to keep track of the combinations.

piyush-nimbalkar: Got it. Let's move this code to https://github.com/libopenstorage/operator/blob/master/drivers/storage/portworx/portworx.go#L1171.

The above code already handles a similar case for vsphere. Also, we don't want to set the cloud provider if the user is not even using cloud drives.

jrivera-px: Ok done...

piyush-nimbalkar (approved):
Thanks @jrivera-px for accommodating all the changes.

PR #1149, opened 2023-07-19
PWX-32170 Add more attributes to the storagenode status

piyush-nimbalkar:
- Added OS and Kernel Version to the StorageNode status
- Added Storage and KVDB as explicit attributes of the StorageNode instead of deriving it from the conditions or node capacity. 
- If STATUS_STORAGE_DOWN is returned by PX node, then now we return Degraded status instead of Online.

Closes # https://portworx.atlassian.net/browse/PWX-32170


pp511 (approved):
Looks good.
Nit..Probably we could add UTs to account for missing code coverage in status.go

piyush-nimbalkar:
> Looks good. Nit..Probably we could add UTs to account for missing code coverage in status.go

Discussed this with @pp511 offline. The status.go changes are already tested but codecov is not recognizing that is being tested.

PR #1155, opened 2023-07-20
PWX-31981: Remove depricated '-T dmthin' misc argument

jrivera-px:
When upgrading PX to 3.0 with op 23.5.0, storage spec was updated with -T dmthin.   This was bug and was corrected with other fixes however the spec still contained the -T dmthin.   Even though its ignored we should remove it to avoid confusion.
Closes #
PWX-31981

### DIFF
--- a/pkg/controller/storagecluster/storagecluster.go
+++ b/pkg/controller/storagecluster/storagecluster.go
@@ -90,6 +90,7 @@ var (
 	controllerKind       = corev1.SchemeGroupVersion.WithKind("StorageCluster")
 	crdBaseDir           = getCRDBasePath
 	deprecatedCRDBaseDir = getDeprecatedCRDBasePath
+	dmThinRexp           = regexp.MustCompile(`-T *dmthin`)
 )
 
 // Controller reconciles a StorageCluster object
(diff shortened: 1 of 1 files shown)


zoxpx:
FYI: Looks like Travis crashed again (logs just end abruptly) -- you might want to restart the Travis-build

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -622,6 +623,25 @@ func (c *Controller) createOrUpdateDeprecatedCRD() error {
...
 }
 
+func (c *Controller) miscCleanUp(cluster *corev1.StorageCluster) error {
+	// cleanup  depricated -T dmthin
+	if _, miscExists := cluster.Annotations[pxutil.AnnotationMiscArgs]; miscExists {
+		// remove '-T dmthin' from misc args if it exists.  It would have been added due to a bug
+		miscAnnotations := cluster.Annotations[pxutil.AnnotationMiscArgs]
+		dmThinStr := `-T *dmthin`
+		if dmExists, err := regexp.MatchString(dmThinStr, miscAnnotations); err == nil {
+			rexp := regexp.MustCompile(dmThinStr)
```
zoxpx, line 633: Could you please move this local variable into global's?
* note, regexp-objects are immutable, and traditionally "expensive to create" -- so always a good idea to keep them defined as "global variables"

zoxpx: uh.. also, you're actually creating the same "heavy" regex-object _twice_  (first implicitly on line 632, then explicitly on line 633)

Recommend rewriting as:
```go
// globals...
var myMagicalRexp = regexp.MustCompile(`-T *dmthin`)

func (c *Controller) miscCleanUp(cluster *corev1.StorageCluster) error {
...
   	m := myMagicalRexp.FindStringSubmatch(miscAnnotations)
        if len(m) > 0 {
           // regex matched...

```

zoxpx (approved):
Looks OK -- some optimizations recommended below

PR #1159, opened 2023-07-21
PWX-31625 : Disable pvc-controller for openshift cluster

nikita-bhatia:
Disabling PVC controller for openshift since Openshift already has kube-controller maanger.
Having additional kube-controller ends up spamming vcenter with storage tasks.

Tests done : 

- Tested on Rh Openshift cluster 4.12.22
- pvc-controller is not created and portworx is deployed fine


piyush-nimbalkar (changes requested):
We would want to deploy the pvc controller if portworx is not deployed in the kube-system namespace. Without that portworx volume will not get deployed.

@nikita-bhatia to verify this ticket, you will have to create a non CSI px volume and check that it gets provisioned. Just deploying px won't tell us anything.

ggriffiths (changes requested):
Needs additional testing as per earlier comment

camartinez04:
I can create new PVCs without PX PVC Controller Manager being present on OpenShift (note that we need access to the master nodes) with Portworx deployed on a non kube-system namespace such as portworx. It uses the master node controller. @ggriffiths

camartinez04:
This issue is still happening with more people getting affected

zoxpx:
Woops -- looks like I missed this PR, and just re-fixed it @ https://github.com/libopenstorage/operator/pull/1297

I'll just close this one.

PR #1161, opened 2023-07-24
PWX-31962: Fixing up test name and provider logic

nrevanna:
TestStorageClusterBasicDMthin has the same prefix with another test
    TestStorageClusterBasic. This is causing the dmthin test to run in
    unintended jenkins jobs.

    There is an additional way of knowing the provider which is to use
    the value from --cloud-provider

Closes #


nikolaypopov (approved):
LGTM

PR #1162, opened 2023-07-24
PTX-18214 Add validation for kvdb image

nikolaypopov:
Add validation for kvdb image


zoxpx (approved):
I think it looks OK

Note, I would have recommended a more "brutal" (and simpler?) test:
1. deploy to `portworx` namespace, and using `STC.spec.customImageRegistry`
2. check everything got deployed
3. check all POD's image-URLs from `portworx` namespace are coming from registry specified via `customImageRegistry`

nikolaypopov:
> I think it looks OK
> 
> Note, I would have recommended a more "brutal" (and simpler?) test:
> 
> 1. deploy to `portworx` namespace, and using `STC.spec.customImageRegistry`
> 2. check everything got deployed
> 3. check all POD's image-URLs from `portworx` namespace are coming from registry specified via `customImageRegistry`

Those test already exists, this is the logic to validate this kvdb image wether its air-gapped or not, custom registry, multi registry or not. This is more of a default logic how we validate any scenarios for this particular component (KVDB)

PR #1163, opened 2023-07-25
PWX-29715: Implemented volume selector for PortworxDiags

Pure-AdamuKaapan:
Implements the ability to select which nodes to collect diags on by providing volume IDs or volume label selectors.

As this is a somewhat intensive operation, I also reorganized things slightly so that we will only run the calculation of which nodes to run on once: after that has been calculated, we just use the list of nodes in the diag object status.

PWX-29715

### DIFF
--- a/pkg/controller/portworxdiag/portworxdiag.go
+++ b/pkg/controller/portworxdiag/portworxdiag.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/libopenstorage/openstorage/api"
 	pxutil "github.com/libopenstorage/operator/drivers/storage/portworx/util"
 	apiextensionsops "github.com/portworx/sched-ops/k8s/apiextensions"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


ggriffiths (commented):
Are UTs going to be covered in a separate task? Looks like there's quite a big test coverage diff with this PR.

Pure-AdamuKaapan:
Yes, I've started working on unit tests in a separate branch. Thanks for checking!

pkg/controller/portworxdiag/portworxdiag.go
```diff
@@ -231,13 +232,84 @@ func getNodeIDToStatusMap(nodeStatuses []diagv1.NodeStatus) map[string]string {
...
+	}
+
+	// Inspect all volumes by labels
+	if len(diag.Spec.Portworx.VolumeSelector.Labels) > 0 {
+		volumeInspectResponse, err := volumeClient.InspectWithFilters(
+			ctx,
+			&api.SdkVolumeInspectWithFiltersRequest{
+				Labels: diag.Spec.Portworx.VolumeSelector.Labels,
+				Options: &api.VolumeInspectOptions{
+					Deep: true,
```
ggriffiths, line 278: Did we verify a deep inspect is needed in order to see ReplicaSets?

Deep inspecting all volumes is a very expensive operation and has caused system slowness

pkg/controller/portworxdiag/portworxdiag.go
```diff
@@ -292,9 +371,49 @@ func shouldPodBeOnNode(nodeID string, nodeName string, nodes []v1.Node, diag *di
 	return true // All labels in selector passed
 }
 
+func (c *Controller) getNodesToHavePods(nodes *v1.NodeList, diag *diagv1.PortworxDiag, stc *corev1.StorageCluster, nodeIDToNodeName map[string]string, nodeIDToStatus map[string]string) (map[string]bool, error) {
```
ggriffiths, line 374: nit: Can we add a comment above this line? Or maybe rename to something more descriptive? Perhaps `nodesSelectedForPods` or get `getSelectedNodesForPods`

I was confused initially what this function was doing without reading the function.

ggriffiths (commented):
Mostly looks good, just one question and one nit.

pkg/controller/portworxdiag/portworxdiag.go
Pure-AdamuKaapan: My initial pass at the code it seemed like it, but I'm not 100% sure. I can try without it and see.

Pure-AdamuKaapan: .... given I accidentally didn't include it for single volumes and it still worked, apparently not! I'll remove that

ggriffiths (approved):
lgtm

PR #1165, opened 2023-07-25
PTX-18659 Fix Telemetry status regex

nikolaypopov:
Fix Telemetry status regex for when we can have other debug lines in pxctl status output

### DIFF
--- a/pkg/util/test/util.go
+++ b/pkg/util/test/util.go
@@ -3555,15 +3555,21 @@ func validateTelemetryStatusInPxctl(telemetryShouldBeEnabled bool, cluster *core
 				return nil, true, fmt.Errorf("got error while trying to get Telemetry status from pxctl, Err: %v", err)
 			}
 
-			if telemetryShouldBeEnabled && !strings.Contains(output, "Healthy") {
-				return nil, true, fmt.Errorf("[%s (%s)] Telemetry is enabled and should be Healthy in pxctl status on PX node, but got [%s]", pxPod.Spec.NodeName, pxPod.Name, strings.TrimSpace(output))
-			} else if !telemetryShouldBeEnabled && !strings.Contains(output, "Disabled") {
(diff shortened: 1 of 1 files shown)


ggriffiths (approved):

pkg/util/test/util.go
```diff
@@ -3555,15 +3555,21 @@ func validateTelemetryStatusInPxctl(telemetryShouldBeEnabled bool, cluster *core
...
 			}
 
-			if telemetryShouldBeEnabled && !strings.Contains(output, "Healthy") {
-				return nil, true, fmt.Errorf("[%s (%s)] Telemetry is enabled and should be Healthy in pxctl status on PX node, but got [%s]", pxPod.Spec.NodeName, pxPod.Name, strings.TrimSpace(output))
-			} else if !telemetryShouldBeEnabled && !strings.Contains(output, "Disabled") {
-				return nil, true, fmt.Errorf("[%s (%s)] Telemetry is not enabled and should be Disabled in pxctl status on PX node, but got [%s]", pxPod.Spec.NodeName, pxPod.Name, strings.TrimSpace(output))
-			} else if !strings.Contains(output, "Disabled") && !strings.Contains(output, "Healthy") {
-				return nil, true, fmt.Errorf("[%s (%s)] Telemetry is Enabled=%v, but pxctl on PX node returned unexpected status [%s]", pxPod.Spec.NodeName, pxPod.Name, telemetryShouldBeEnabled, output)
+			// Find Telemetry status line
+			telemetryStatusString := regexp.MustCompile("Telemetry:.*").FindString(output)
```
jrivera-px, line 3559: Might want to change the regex to pickup line "beginning with"...  So maybe use "^Telemetry:.*" just in case we add warning or something later on...

jrivera-px (approved):
Looks ok besides the one comment on the regex

pkg/util/test/util.go

PR #1166, opened 2023-07-25
PTX-18831 Improving logic for when Telemetry should be enabled/disabled

nikolaypopov:
Improving logic for when Telemetry should be enabled/disabled

### DIFF
--- a/pkg/util/test/util.go
+++ b/pkg/util/test/util.go
@@ -3,9 +3,11 @@ package test
 import (
 	"context"
 	"crypto/rand"
+	cryptoTls "crypto/tls"
 	"encoding/base64"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"net/url"
 	"os"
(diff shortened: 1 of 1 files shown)


pkg/util/test/util.go
```diff
@@ -3017,88 +3044,231 @@ func ValidateTelemetry(pxImageList map[string]string, cluster *corev1.StorageClu
...
 	}
+	return false
+}
 
-	// Update Telemetry status
-	if cluster.Spec.Monitoring != nil && cluster.Spec.Monitoring.Telemetry != nil {
-		cluster.Spec.Monitoring.Telemetry.Enabled = newCluster.Spec.Monitoring.Telemetry.Enabled
-	} else {
-		cluster.Spec.Monitoring = newCluster.Spec.Monitoring
+func getArcusTelemetryLocation(cluster *corev1.StorageCluster) string {
```
ezhang-px, line 3233: This function will make our default location to be external(prod) unless specified in the annotation. Hope that's the expectation here

ezhang-px (commented):
some of the APIs (GetPxProxyEnvVarValue, CanAccessArcusRegisterEndpoint, ParsePxProxyURL, IsCCMGoSupported) are redefined in other packages: drivers/storage/portworx/component, drivers/storage/portworx/util with even different implementation, can we simply call them for the test purpose instead of rewriting?

nikolaypopov:
> some of the APIs (GetPxProxyEnvVarValue, CanAccessArcusRegisterEndpoint, ParsePxProxyURL, IsCCMGoSupported) are redefined in other packages: drivers/storage/portworx/component, drivers/storage/portworx/util with even different implementation, can we simply call them for the test purpose instead of rewriting?

I tried that and it is not possible due to `import cycle not allowed`, also as per @piyush-nimbalkar we do not want to re-use much of operator functions, just in case a bug is introduced there.

pkg/util/test/util.go
nikolaypopov: It is expected, this is a general validation for both production and staging telemetry

ezhang-px:
> > some of the APIs (GetPxProxyEnvVarValue, CanAccessArcusRegisterEndpoint, ParsePxProxyURL, IsCCMGoSupported) are redefined in other packages: drivers/storage/portworx/component, drivers/storage/portworx/util with even different implementation, can we simply call them for the test purpose instead of rewriting?
> 
> I tried that and it is not possible due to `import cycle not allowed`, also as per @piyush-nimbalkar we do not want to re-use much of operator functions, just in case a bug is introduced there.

Got you, thanks for the explanation. @jrivera-px also described more context to me offline about what's going yesterday. Just one catch here about the CanAccessArcusRegisterEndpoint, I think in the current release, it's already updated, so is it on purpose that we have to use the code before the https changes?
https://github.com/libopenstorage/operator/blob/px-rel-23.7.0/drivers/storage/portworx/component/telemetry.go#L1036

ezhang-px (approved):
LGTM

PR #1167, opened 2023-07-26
PWX-32539: When preflight enables PX-storeV2 for vsphere.  It should …

jrivera-px:
…add a default 'type=eagerzeroedthick,size=64' SystemMdDeviceSpec if none is provided.

Preflight should add a default 'type=eagerzeroedthick,size=64' SystemMdDeviceSpec if none is provided.  When running on vsphere.
Closes #
PWX-32539

### DIFF
--- a/drivers/storage/portworx/preflight.go
+++ b/drivers/storage/portworx/preflight.go
@@ -33,7 +33,7 @@ const (
 	// PxPreFlightServiceAccountName name of portworx pre flight service account
 	PxPreFlightServiceAccountName = "px-pre-flight"
 	// DefCmetaData default metadata cloud device for DMthin AWS
-	DefCmetaData = "type=gp3,size=64"
+	DefCmetaAWS = "type=gp3,size=64"
 	// DefCmetaVsphere default metadata cloud device for DMthin Vsphere
 	DefCmetaVsphere = "type=eagerzeroedthick,size=64"
 )
@@ -355,7 +355,7 @@ func (u *preFlightPortworx) processPassedChecks(recorder record.EventRecorder) {
 	}
 
 	if u.cluster.Spec.CloudStorage.SystemMdDeviceSpec == nil {
-		cmetaData := DefCmetaData
+		cmetaData := DefCmetaAWS
 		if pxutil.IsVsphere(u.cluster) {
 			cmetaData = DefCmetaVsphere
 		}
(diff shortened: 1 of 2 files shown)


drivers/storage/portworx/preflight.go
```diff
@@ -32,8 +32,10 @@ const (
 	pxPreFlightDaemonSetName          = "px-pre-flight"
 	// PxPreFlightServiceAccountName name of portworx pre flight service account
 	PxPreFlightServiceAccountName = "px-pre-flight"
-	// DefCmetaData default metadata cloud device for DMthin
+	// DefCmetaData default metadata cloud device for DMthin AWS
 	DefCmetaData = "type=gp3,size=64"
+	// DefCmetaVsphere default metadata cloud device for DMthin Vsphere
+	DefCmetaVsphere = "type=eagerzeroedthick,size=64"
```
prabirpaul, line 38: I think we should lazyzeroedthick. Dmthin nodes typically take longer to init. And i suspect eagerzeroedthick might add to that time.

prabirpaul (approved):

drivers/storage/portworx/preflight.go
nrevanna: lazy zeroed thick apparently has a little performance hit overhead during it's lifetime since there will be zeroes written during IO writes. Do we prefer a little perf overhead later or during initialization?

drivers/storage/portworx/preflight.go
```diff
@@ -32,8 +32,10 @@ const (
 	pxPreFlightDaemonSetName          = "px-pre-flight"
 	// PxPreFlightServiceAccountName name of portworx pre flight service account
 	PxPreFlightServiceAccountName = "px-pre-flight"
-	// DefCmetaData default metadata cloud device for DMthin
+	// DefCmetaData default metadata cloud device for DMthin AWS
 	DefCmetaData = "type=gp3,size=64"
```
nrevanna, line 36: nit: rename this to DefCmetaAWS

nrevanna (approved):

prabirpaul (approved):

drivers/storage/portworx/preflight.go
nrevanna: Had an offline discussion and we'll keep it EZT for now until we do some perf testing.

PR #1168, opened 2023-07-26
[PWX-30455] feat: volume-mount override via custom mount

Glitchfix:
override volume mounts in stc to mount it different host path

Closes #[PWX-30455](https://portworx.atlassian.net/browse/PWX-30455)

[PWX-30455]: https://portworx.atlassian.net/browse/PWX-30455?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/drivers/storage/portworx/deployment.go
+++ b/drivers/storage/portworx/deployment.go
@@ -1358,8 +1358,7 @@ func (t *template) mountsFromVolInfo(vols []volumeInfo) []v1.VolumeMount {
 
 	for _, v := range t.cluster.Spec.Volumes {
 		if _, ok := mountPathSet[v.MountPath]; ok {
-			logrus.Warnf("Found mountPath conflict for volume %s at %s, volume will be ignored", v.Name, v.MountPath)
-			continue
+			logrus.Warnf("Found mountPath conflict for volume %s at %s", v.Name, v.MountPath)
 		}
 		volumeMounts = append(volumeMounts, v1.VolumeMount{
 			Name:             pxutil.UserVolumeName(v.Name),
--- a/drivers/storage/portworx/deployment.go
+++ b/drivers/storage/portworx/deployment.go
@@ -1295,7 +1295,37 @@ func (t *template) getVolumeMounts() []v1.VolumeMount {
 	for _, fn := range extensions {
 		volumeInfoList = append(volumeInfoList, fn()...)
 	}
+	volumeInfoList = t.overrideVolumeMount(volumeInfoList)
 	return t.mountsFromVolInfo(volumeInfoList)
+}
+
+func (t *template) overrideVolumeMount(volInfos []volumeInfo) []volumeInfo {
+	volPaths := map[string]int{}
+
+	for i := range volInfos {
+		volPaths[volInfos[i].mountPath] = i
+	}
+
+	for i := range t.cluster.Spec.Volumes {
+		v := t.cluster.Spec.Volumes[i]
+		existingVolIdx, ok := volPaths[v.MountPath]
+		if !ok {
+			volInfos = append(volInfos, volumeInfo{
+				name:             v.Name,
+				readOnly:         v.ReadOnly,
+				hostPath:         v.HostPath.String(),
+				mountPath:        v.MountPath,
+				mountPropagation: v.MountPropagation,
+				hostPathType:     v.HostPath.Type,
+				configMapType:    v.ConfigMap,
+			})
+		} else {
+			volInfos[existingVolIdx].hostPath = v.HostPath.String()
+			volInfos[existingVolIdx].mountPropagation = v.MountPropagation
+		}
+	}
+
+	return volInfos
 }
 
 func (t *template) mountsFromVolInfo(vols []volumeInfo) []v1.VolumeMount {
--- a/drivers/storage/portworx/deployment_test.go
+++ b/drivers/storage/portworx/deployment_test.go
@@ -4455,7 +4455,7 @@ func Test_template_overrideVolumeMount(t *testing.T) {
 		Path: "/test1",
 	}
 	hp2 := &v1.HostPathVolumeSource{
-		Path: "/test1",
+		Path: "/test2",
 	}
 	additionalMounts := append(testVols, volumeInfo{
 		name:             "test",
--- a/drivers/storage/portworx/deployment_test.go
+++ b/drivers/storage/portworx/deployment_test.go
@@ -4436,3 +4436,105 @@ func assertContainerEqual(t *testing.T, expected, actual v1.Container) {
 	assert.ElementsMatch(t, expected.Env, actual.Env)
 	assert.ElementsMatch(t, expected.VolumeMounts, actual.VolumeMounts)
 }
+
+func Test_template_overrideVolumeMount(t *testing.T) {
+	testVols := getCommonVolumeList(pxutil.MinimumPxVersionAutoTLS)
+	hp0 := &v1.HostPathVolumeSource{
+		Path: "/test0",
+	}
+	overRideVols := append(testVols, []volumeInfo{}...)
+	for i := range overRideVols {
+		if overRideVols[i].name == "diagsdump" {
+			overRideVols[i].hostPath = hp0.String()
+			overRideVols[i].mountPropagation = mountPropagationModePtr(v1.MountPropagationBidirectional)
+			break
+		}
+	}
+
+	hp1 := &v1.HostPathVolumeSource{
+		Path: "/test1",
+	}
+	hp2 := &v1.HostPathVolumeSource{
+		Path: "/test1",
+	}
+	additionalMounts := append(testVols, volumeInfo{
+		name:             "test",
+		hostPath:         hp2.String(),
+		mountPath:        "/test2",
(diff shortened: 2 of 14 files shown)


drivers/storage/portworx/deployment_test.go
```diff
@@ -4436,3 +4436,105 @@ func assertContainerEqual(t *testing.T, expected, actual v1.Container) {
...
+			overRideVols[i].mountPropagation = mountPropagationModePtr(v1.MountPropagationBidirectional)
+			break
+		}
+	}
+
+	hp1 := &v1.HostPathVolumeSource{
+		Path: "/test1",
+	}
+	hp2 := &v1.HostPathVolumeSource{
+		Path: "/test1",
```
nikita-bhatia, line 4458: shoul this be "/test2" ?

Glitchfix: wouldn't matter actually that's just a dummy name being used

Glitchfix: changed it thought

nikita-bhatia (approved):
LGTM

drivers/storage/portworx/deployment.go
```diff
@@ -1295,9 +1295,39 @@ func (t *template) getVolumeMounts() []v1.VolumeMount {
 	for _, fn := range extensions {
 		volumeInfoList = append(volumeInfoList, fn()...)
 	}
+	volumeInfoList = t.overrideVolumeMount(volumeInfoList)
 	return t.mountsFromVolInfo(volumeInfoList)
 }
 
+func (t *template) overrideVolumeMount(volInfos []volumeInfo) []volumeInfo {
```
zoxpx, line 1302: Why is this called `overrideVolumeMount`?
* we are not _overriding_ the internal mounts-list
* instead, we are _filtering_ (and extending) the list that was provided as input

drivers/storage/portworx/deployment.go
```diff
@@ -1295,9 +1295,39 @@ func (t *template) getVolumeMounts() []v1.VolumeMount {
...
+	volPaths := map[string]int{}
+
+	for i := range volInfos {
+		volPaths[volInfos[i].mountPath] = i
+	}
+
+	for i := range t.cluster.Spec.Volumes {
```
zoxpx, line 1309: Looks like we'll take the `volInfos` first, then we'll _append_ all the previous `cluster.Spec.Volumes`

This seems to be clashing with how the px-runc is working on the "lower level":
* "default" mounts "go in" first
* "custom mounts" will go after the "default mounts" -- and have the opportunity to override the defaults

Is this code behaving the same?

drivers/storage/portworx/deployment.go
```diff
@@ -1295,9 +1295,39 @@ func (t *template) getVolumeMounts() []v1.VolumeMount {
...
+				readOnly:         v.ReadOnly,
+				hostPath:         v.HostPath.String(),
+				mountPath:        v.MountPath,
+				mountPropagation: v.MountPropagation,
+				hostPathType:     v.HostPath.Type,
+				configMapType:    v.ConfigMap,
+			})
+		} else {
+			volInfos[existingVolIdx].hostPath = v.HostPath.String()
+			volInfos[existingVolIdx].mountPropagation = v.MountPropagation
```
zoxpx, line 1324: Why are we overriding _only_ the host-path and the mount-propagation?

Shouldn't we at least copy the mount-options (i.e. read-only)?

zoxpx (commented):
Could we please add a "manual test" to prove that the code-change is working?

For example:
1. add the following into the STC:

```yaml
kind: StorageCluster
spec:
  volumes:
  - hostPath:
      path: /apps/px/cores
    mountPath: /var/cores
    name: px-cores
```

2. get the POD-yaml from the resulting deployment, and
3. show that the POD is using the overridden mount (either as the only mounting to `/var/cores` destination, or as the _last one_ mounting to this destination).

drivers/storage/portworx/deployment.go
Glitchfix: for existing mounts, if a stc yaml overrides the mount options then it could affect the functionality of porx
imagine someone makes the oci log file to be read only then we do not get to write the logs

drivers/storage/portworx/deployment.go
Glitchfix: the ticket requires allowing to existing mount paths with a custom hostpath
this function is triggered at last as this will allow us to add more mount paths as well as modify any existing mount paths that is utilized by our application

drivers/storage/portworx/deployment.go
Glitchfix: we are not just filtering but also modifying the host path property inside the function

zoxpx:
Ping?

Glitchfix:
Hi @zoxpx @ggriffiths 

I have added the output of the requested changes as well 

image: `cnbuautomation800/px-operator:pwx30455`

stc
```
# SOURCE: https://install.portworx.com/?operator=true&mc=false&kbver=1.25.0&ns=portworx&b=true&c=px-cluster-4d617504-5340-4a93-81d6-9255d71136c8&stork=true&csi=true&mon=true&tel=true&st=k8s&promop=true
kind: StorageCluster
metadata:
  name: px-30455
  namespace: px
  annotations:
    portworx.io/install-source: "https://install.portworx.com/?operator=true&mc=false&kbver=1.25.0&ns=portworx&b=true&c=px-cluster-4d617504-5340-4a93-81d6-9255d71136c8&stork=true&csi=true&mon=true&tel=true&st=k8s&promop=true"
spec:
  deleteStrategy:
    type: UninstallAndWipe
  volumes:
  - hostPath:
      path: /tmp/extra-mount
    mountPath: /tmp/extra-mount
    name: extra-mount
  - hostPath:
      path: /tmp/px-cores
    mountPath: /var/cores
    name: px-cores
  image: portworx/oci-monitor:2.13.7
  imagePullPolicy: Always
  kvdb:
    internal: true
  storage:
    useAll: true
  secretsProvider: k8s
  stork:
    enabled: true
    args:
      webhook-controller: "true"
  autopilot:
    enabled: true
  csi:
    enabled: true
  monitoring:
    telemetry:
      enabled: true
    prometheus:
      enabled: true
      exportMetrics: true
```

ls inside one of the node
```
 ~]# ls -la /tmp/extra-mount/
total 8
drwxr-xr-x   2 root root 4096 Aug 23 00:49 .
drwxrwxrwt. 15 root root 4096 Aug 23 00:59 ..
~]# ls -la /tmp/px-cores/
total 12
drwxr-xr-x   3 root root 4096 Aug 23 00:54 .
drwxrwxrwt. 15 root root 4096 Aug 23 01:00 ..
drw-r--r--   2 root root 4096 Aug 23 00:49 .alerts
```

zoxpx:
Uh.. that doesn't seem right  (I'm glad we tested this)
* the `/tmp/px-cores` should have a lot more files -- (same as our original `/var/cores` has)
* so, looks like the "override" did not work correctly

Could you run `jq .mounts /opt/pwx/oci/config.json` on your node, and search for `/cores`?
* for the override to work correctly, there should be only 1 mount present (`source:/tmp/px-cores, destination:/var/cores`)

zoxpx:
Ok.. did some extra testing, using px-operator image `cnbuautomation800/px-operator:pwx30455`


TEST's input modification -- using 2 modifications with the same destination (`/var/cores`)

```yaml
kind: StorageCluster
spec:
  volumes:
  - hostPath:
      path: /mnt/px-cores
    mountPath: /var/cores
    name: px-cores
  - hostPath:
      path: /mnt/px-cores22
    mountPath: /var/cores
    name: px-cores22
```


**RESULT= FAIL**: Portworx POX has 

* [minor issue] some excess host-paths (which is OK)
* [CRITICAL] it is MISSING container-mount for `user-px-cores22`

```yaml
kind: Pod
spec:
  containers:
  - name: portworx
    volumeMounts:
    - mountPath: /var/cores
      name: diagsdump
...
  volumes:
  - hostPath:
      path: /var/cores
      type: ""
    name: diagsdump
  - hostPath:
      path: /mnt/px-cores
      type: ""
    name: user-px-cores
  - hostPath:
      path: /mnt/px-cores22
      type: ""
    name: user-px-cores22
```


**RESULT= FAIL**: Actual `portworx.service` did NOT override the `/var/cores` mount
* see `jq .mounts /opt/pwx/oci/config.json` on the node -- the mount is still the original `/var/cores:/var/cores`

```json
[
  {
    "destination": "/var/cores",
    "type": "bind",
    "source": "/var/cores",
    "options": [
      "rbind",
      "rprivate"
    ]
  },
  ...
]
```

Please DO NOT merge this PR!

Glitchfix:
Hi, @zoxpx appoligies for the confusion

the later results I sent you uses this image

`cnbuautomation800/px-operator:pwx30455-2`

the previous image was missing this commit https://github.com/libopenstorage/operator/pull/1168/commits/e1053612765f04114b6edd9e82590a823f1ee3e1

it uses the same storage cluster file and doesn't crash

harsh-px:
Closing this as we will use https://github.com/libopenstorage/operator/pull/1226 for this fix.

PR #1170, opened 2023-07-27
PWX-32567: pre-flight pods do not run on master so don't create pre-f…

jrivera-px:
…light storage node entry for master.

Preflight validation was failing because all storageNode entries are expected to have results from the preflight pod.  Before preflight is run storageNode entries are created for all the nodes in the cluster.   However when pre-flight is run it is not run on the master node.  So the storageNode results were empty from that node and this caused validation to fail.    So before preflight is run don't create a storageNode entry for master node since pre-flight will not run on that node anyways.    
Closes #
PWX-32567

### DIFF
--- a/pkg/controller/storagecluster/storagecluster.go
+++ b/pkg/controller/storagecluster/storagecluster.go
@@ -63,6 +63,7 @@ import (
 	"github.com/libopenstorage/operator/pkg/preflight"
 	"github.com/libopenstorage/operator/pkg/util"
 	"github.com/libopenstorage/operator/pkg/util/k8s"
+	coreops "github.com/portworx/sched-ops/k8s/core"
 )
 
 const (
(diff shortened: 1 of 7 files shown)


nrevanna (commented):
This looks good to me.
- Can we have a ticket to add a unit test for this?
- Added Harsh for second pair of eyes on this.

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -447,8 +448,10 @@ func (c *Controller) runPreflightCheck(cluster *corev1.StorageCluster) error {
 		err = c.client.List(context.TODO(), k8sNodeList)
 		if err == nil {
 			for _, node := range k8sNodeList.Items {
-				logrus.Infof("Create pre-flight storage node entry for node: %s", node.Name)
-				c.createStorageNode(cluster, node.Name)
+				if !coreops.Instance().IsNodeMaster(node) {
```
nrevanna, line 451: @jrivera-px what's still not clear to me is - How is it that the list contains the master node sometimes and not always?

nrevanna (approved):

pkg/controller/storagecluster/storagecluster.go
jrivera-px: The list always contained the master node.   In the past the checks we made did not consider the it a failed validation if the master node checks was nil.

PR #1171, opened 2023-07-27
PWX-32103: Adds end-to-end unit tests for the PortworxDiag operator code

Pure-AdamuKaapan:
The previous unit tests were white-box tests on specific internal methods.
These are black-box unit tests testing the end-to-end functionality as pods/CRDs change.

PWX-32103


pkg/controller/portworxdiag/controller_test.go
```diff
@@ -25,13 +38,199 @@ import (
...
+const (
+	etcHostsFile       = "/etc/hosts"
+	tempEtcHostsMarker = "### px-operator unit-test"
+)
+
+var (
+	podCreateCount = 0
+)
+
+// setupEtcHosts sets up given ip/hosts in `/etc/hosts` file for "local" DNS resolution  (i.e. emulate K8s DNS)
```
ggriffiths, line 58: Is there any way to avoid this in the tests? Are we trying to emulate k8s DNS because the production code references the PX Service? 

If so, maybe we can make the connection information configurable instead and point to localhost:9021 for testing?

Pure-AdamuKaapan: There may be a way to avoid it, but this is also what the StorageCluster controller tests do, so it's already an evil we have :P

ggriffiths: Ah ok, can we externalize this here then? 
https://github.com/libopenstorage/operator/blob/master/pkg/util/test/util.go

At least in this case, fixes to any issues here would go in one spot.

ggriffiths (commented):
PR looks good to me once we get travis and everything passing. Might take a few re-runs.

ggriffiths:
Btw I'm not sure if you're looking into the check failures, but there's some go errors in `Build Operator` -
https://jenkins.pwx.dev.purestorage.com/job/CBT/job/cbt-operator-build/1157/console

ggriffiths (approved):
lgtm once checks are passing.

PR #1172, opened 2023-07-27
PTX-19249 Add delete StorageCluster after each test

nikolaypopov:
Add delete StorageCluster after each test

### DIFF
--- a/test/integration_test/basic_test.go
+++ b/test/integration_test/basic_test.go
@@ -109,7 +109,7 @@ var testStorageClusterBasicCases = []types.TestCase{
 		TestFunc: BasicUpgradeOperator,
 	},
 	{
-		TestName:        "BasicInstall",
+		TestName:        "BasicInstallWithNodeAffinity",
 		TestrailCaseIDs: []string{"C50962", "C51022", "C50236"},
 		TestSpec: ci_utils.CreateStorageClusterTestSpecFunc(&corev1.StorageCluster{
 			ObjectMeta: metav1.ObjectMeta{Name: "test-stc"},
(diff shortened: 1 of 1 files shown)


test/integration_test/basic_test.go
```diff
@@ -325,16 +325,26 @@ func BasicInstallInCustomNamespace(tc *types.TestCase) func(*testing.T) {
 			err = ci_utils.ValidateObjectsAreTerminated([]runtime.Object{ns}, false)
 			require.NoError(t, err)
 		}
+
+		// Delete and validate StorageCluster deletion
+		ci_utils.UninstallAndValidateStorageCluster(cluster, t)
```
ggriffiths, line 330: Maybe do a BeforeEach/AfterEach here to ensure this happens?

ggriffiths: nevermind, discussed offline

nikolaypopov: I think this won't work with the methods we have, let's test `px-wipe` after each test for all the different combinations we have

test/integration_test/basic_test.go
```diff
@@ -109,7 +109,7 @@ var testStorageClusterBasicCases = []types.TestCase{
 		TestFunc: BasicUpgradeOperator,
 	},
 	{
-		TestName:        "BasicInstall",
+		TestName:        "BasicInstallWithNodeAffinity",
```
jrivera-px, line 112: Why are we using node affinity now?

jrivera-px (approved):

test/integration_test/basic_test.go
nikolaypopov: To make it more clear ^^

ggriffiths (approved):

PR #1174, opened 2023-07-28
PWX-32193 Watch for diag pod changes and reconcile PortworxDiag obj

piyush-nimbalkar:

- The diag pods are created with controller reference to the corresponding PortworxDiag object.
- The added controller watch will ensure that if there is a change in the diag pods, the diag reconciler is called with the parent PortworxDiag object.

Closes # https://portworx.atlassian.net/browse/PWX-32193


Pure-AdamuKaapan (approved):
LGTM! Thanks so much for helping squeeze this in.

PR #1175, opened 2023-07-28
PTX-18228 Adding validation for OCP Dynamic Plugin

nikolaypopov:
Adding validation for OCP Dynamic Plugin


nikita-bhatia (approved):
LGTM

PR #1179, opened 2023-08-03
PTX-19346 Adding workaround for the known Telemetry port issue for specific Operator upgrades

nikolaypopov:
Adding workaround for the known Telemetry port issue for specific Operator upgrades

### DIFF
--- a/pkg/util/test/util.go
+++ b/pkg/util/test/util.go
@@ -3308,16 +3308,41 @@ func ValidateTelemetry(pxImageList map[string]string, cluster *corev1.StorageClu
 	}
 
 	if pxVersion.GreaterThanOrEqual(minimumPxVersionCCMGO) && opVersion.GreaterThanOrEqual(opVer1_10) {
-		if shouldTelemetryBeEnabled(cluster) {
-			return ValidateTelemetryV2Enabled(pxImageList, cluster, timeout, interval)
-		}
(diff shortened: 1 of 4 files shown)


pkg/util/test/util.go
```diff
@@ -3309,15 +3309,27 @@ func ValidateTelemetry(pxImageList map[string]string, cluster *corev1.StorageClu
 
 	if pxVersion.GreaterThanOrEqual(minimumPxVersionCCMGO) && opVersion.GreaterThanOrEqual(opVer1_10) {
 		if shouldTelemetryBeEnabled(cluster) {
-			return ValidateTelemetryV2Enabled(pxImageList, cluster, timeout, interval)
```
Rohit-PX, line 3312: Just create an uber method called `ValidateTelemetryV2` with a parameter for enabled or disabled and call the corresponding functions inside it.

It would like this:
```
ValidateTelemetryV1(pxImageList, cluster, timeout, interval, shouldTelemetryBeEnabled(cluster))
and 
ValidateTelemetryV2(pxImageList, cluster, timeout, interval, shouldTelemetryBeEnabled(cluster))
```

nikolaypopov: Done

Rohit-PX (approved):

PR #1182, opened 2023-08-03
Fix slack message action for PR Forks

ggriffiths:
Uses the `pull_request_target` event, which uses secrets from the base repository instead of the forked repository.


zoxpx (approved):
Seems harmless enough -- thanks Grant

PR #1189, opened 2023-08-08
PWX-32800: ignore proxy's root CA cert for now

ezhang-px:

Currently operator fails to refresh all components' image tags due to not able to access manifest host and has to fall back to default tags. This will introduce two issues:
1. our torpedo jobs will perform the version check for all images and this mismatch (default vs actual from manifest) will break those jobs
2. when customers use this operator to deploy px, they'll get the default (usually obsolete) images instead of the ones corresponding to the px version that they want

Operator fails due to it's using HTTPS proxy to access the manifest url without the proxy's root CA cert. Since current release is not intended to use cert but instead use basic authentication (user, password) of the proxy, ignore the requirement for now and will introduce proxy's CA cert as part of operator's trusted CA store along with corresponding changes inside envoy config in the next release

Closes #

### DIFF
--- a/drivers/storage/portworx/manifest/common.go
+++ b/drivers/storage/portworx/manifest/common.go
@@ -1,6 +1,7 @@
 package manifest
 
 import (
+	"crypto/tls"
 	"io"
 	"net/http"
 	"net/url"
@@ -44,6 +45,9 @@ func getManifestFromURL(manifestURL string, proxy string) ([]byte, error) {
 
 		client := &http.Client{
 			Transport: &http.Transport{
+				TLSClientConfig: &tls.Config{
+					InsecureSkipVerify: true,
+				},
 				Proxy: http.ProxyURL(proxyURL),
 			},
 		}


zoxpx:
Let's please _NOT_ check this in:
* the actual root-cause for PXW-32800 was _"wrong test-setup"_ (see https://portworx.atlassian.net/browse/PWX-32800?focusedCommentId=257573)
* this change would have been OK -- IF it was triggered by an annotation (e.g. introduce `portowrx.io/tls-skip-verify: true` annotation, and use it here)

drivers/storage/portworx/manifest/common.go
```diff
@@ -44,6 +45,9 @@ func getManifestFromURL(manifestURL string, proxy string) ([]byte, error) {
 
 		client := &http.Client{
 			Transport: &http.Transport{
+				TLSClientConfig: &tls.Config{
+					InsecureSkipVerify: true,
```
nikolaypopov, line 49: Should we take this true/false value from StorageCluster env var `DO_NOT_VERIFY_CA` instead of hardcoding? This way when we will allow user to have CA certs, we don't have to change this

zoxpx: Uh.. it's not that we used `DO_NOT_VERIFY_CA` env-var that widely (only limited support in `porx` and no support in operator).  We typically use annotations for such configuration changes -- which is why I suggested the annotation.

This particular annotation could work initially for Telemetry-proxy, but eventually across the whole PX cluster.

PR #1191, opened 2023-08-08
PWX-32732 : use portworx-restricted based on IsPrivileged flag

nikita-bhatia:
Do not use portworx-restricted SCC by default, use the old way and allow to enable/disable it via annotation

Closes # https://portworx.atlassian.net/browse/PWX-32732


drivers/storage/portworx/component/csi.go
```diff
@@ -332,6 +326,18 @@ func (c *csi) createClusterRole(
...
+		clusterRole.Rules = append(
+			clusterRole.Rules,
+			rbacv1.PolicyRule{
+				APIGroups:     []string{"security.openshift.io"},
+				Resources:     []string{"securitycontextconstraints"},
+				ResourceNames: []string{PxRestrictedSCCName},
+				Verbs:         []string{"use"},
+			},
+		)
+	}
```
zoxpx, line 339: This does not look right -- shouldn't it be something like:

```go
        sccName := PxSCCName
	if !pxutil.IsPrivileged(cluster) {
            sccName := PxRestrictedSCCName
        }
	clusterRole.Rules = append(
		clusterRole.Rules,
		rbacv1.PolicyRule{
			APIGroups:     []string{"security.openshift.io"},
			Resources:     []string{"securitycontextconstraints"},
			ResourceNames: []string{sccName},
			Verbs:         []string{"use"},
		},
	)
```

drivers/storage/portworx/testspec/csiClusterRole_k8s_1.14.yaml
```diff
@@ -54,10 +54,6 @@ rules:
 - apiGroups: ["coordination.k8s.io"]
   resources: ["leases"]
   verbs: ["*"]
-- apiGroups: ["security.openshift.io"]
-  resources: ["securitycontextconstraints"]
-  resourceNames: ["portworx-restricted"]
-  verbs: ["use"]
```
zoxpx, line 60: same here...

zoxpx (changes requested):
I think this PR will require some more work (see below)

drivers/storage/portworx/component/csi.go
nikita-bhatia: done

drivers/storage/portworx/testspec/csiClusterRole_k8s_1.11.yaml
```diff
@@ -51,10 +51,6 @@ rules:
 - apiGroups: ["coordination.k8s.io"]
   resources: ["leases"]
   verbs: ["*"]
-- apiGroups: ["security.openshift.io"]
-  resources: ["securitycontextconstraints"]
-  resourceNames: ["portworx-restricted"]
-  verbs: ["use"]
 - apiGroups: ["policy"]
```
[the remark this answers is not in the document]
nikita-bhatia: changed back to default SCCname, also added UTs with isPrivileged anotation "false"

drivers/storage/portworx/component/lighthouse.go
```diff
@@ -147,75 +147,79 @@ func (c *lighthouse) createServiceAccount(
...
+				APIGroups:     []string{"policy"},
+				Resources:     []string{"podsecuritypolicies"},
+				ResourceNames: []string{constants.PrivilegedPSPName},
+				Verbs:         []string{"use"},
 			},
 		},
-	)
+	}
+
+	return k8sutil.CreateOrUpdateClusterRole(c.k8sClient, clusterRole)
```
zoxpx, line 222: This seems to be a new addition... was this a mistake, or was this line required ?

zoxpx (approved):
This looks good -- thanks Nikita  (one question logged below)

drivers/storage/portworx/component/lighthouse.go
nikita-bhatia: This method was returning **k8sutil.CreateOrUpdateClusterRole** , changed it to last line to add conditional statement of SCC name before creating the cluster role.

PR #1193, opened 2023-08-09
PWX-32122: PortworxDiag integration tests

Pure-AdamuKaapan:
Adds integration tests for the Portworx Diag CR collection feature.

Also some related minor fixes:
* Fixes some issues when updating go modules regarding a recent reorganization in prometheus-operator structs
* Fixes a minor (non-feature-impacting) bug that results in a PortworxDiag not being marked "Pending" if multiple exist at once

PWX-32122

The main tests are written out already, I'm just filling out a few of the edge cases like nodes being disabled, etc.


test/integration_test/portworx_diag_test.go
```diff
@@ -0,0 +1,910 @@
...
+
+func DiagCollection_StopNodes(tc *types.TestCase) func(*testing.T) {
+	return func(t *testing.T) {
+		testSpec := tc.TestSpec(t)
+		dtc, ok := testSpec.(*diagTestSpec)
+		require.True(t, ok)
+
+		// TODO: any other conditions to try?
+		stoppedNode := ci_utils.AddLabelToRandomNode(t, "px/service", "stop")
+		// disabledNode := ci_utils.AddLabelToRandomNode(t, "px/service", "disable")
```
piyush-nimbalkar, line 563: Let's remove the commented lines. We don't need to test disable as it would give the same behavior.

piyush-nimbalkar (approved):
Very nicely written tests!

PR #1195, opened 2023-08-09
PWX-32850: When preflight is unsupported we need to make sure we are …

jrivera-px:
…setting the pre-flight annotation to false.

When preflight was not supposed to be run.   We were only setting the annotation to false when the annotation did not exist.  Problem here was it existed and was set to "true".    This prevented the cluster configuration changes from being applied since it continually though preflight should be run.      The fix is to make sure preflight annotation is set to false when pre-flight is not supported.  
Closes #
PWX-32850


nrevanna (approved):
LGTM.
I tested this image and it worked for my usecase too.

PR #1196, opened 2023-08-10
PWX-32800: proxy with http protocal also supports user/password

ezhang-px:
proxy with http protocol should also support using user/password basic authentication

Closes #


nikolaypopov:
Do you think we can also make similar change for our test logic as well as these 2 go together https://github.com/libopenstorage/operator/blob/7e5c4ffe16f31c3080c60aed7f4c11d6b0ba1e51/pkg/util/test/util.go#L3429

nikolaypopov (approved):

PR #1202, opened 2023-08-11
fixup for https proxy with basic authn

ezhang-px:
The previous implementation of using basic auth for https proxy was based on a wrong assumption that this https proxy also needs to be SSL-enabled (with https:// prefix/protocol). 

Removing them because https proxy can also be non-SSL enabled (http://) and also no TLS context is required in envoy config for this case.

Closes #


zoxpx (approved):
I think this looks OK -- thanks Eugene

nikolaypopov (approved):

PR #1203, opened 2023-08-11
fix for the https proxy with basic authn

ezhang-px:
cherry-picked from master for removing wrong setup/check for https proxy using basic auth 

Closes #

### DIFF
--- a/drivers/storage/portworx/component/telemetry.go
+++ b/drivers/storage/portworx/component/telemetry.go
@@ -1055,14 +1055,8 @@ func CanAccessArcusRegisterEndpoint(
 	}
 	client := &http.Client{}
 	if proxy != "" {
-		if strings.Contains(strings.ToLower(proxy), "@") {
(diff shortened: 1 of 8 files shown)


drivers/storage/portworx/component/telemetry.go
```diff
@@ -1055,14 +1055,8 @@ func CanAccessArcusRegisterEndpoint(
...
 	if proxy != "" {
-		if strings.Contains(strings.ToLower(proxy), "@") {
-			if !strings.HasPrefix(strings.ToLower(proxy), "https://") {
-				proxy = "https://" + proxy
-			}
-		} else {
-			if !strings.HasPrefix(strings.ToLower(proxy), "http://") {
-				proxy = "http://" + proxy
-			}
+		if !strings.HasPrefix(strings.ToLower(proxy), "http://") {
```
nikolaypopov, line 1058: We should add a condition to where if https:// was used and we do not support it, we going to error out with the explanation why we do not support it

PR #1206, opened 2023-08-14
PWX-32976: need to set telemetry port to 9029 if px >= 2.13.8

ezhang-px:
* modified px version check

We're backporting a change to 2.13.8 so that the default telemetry port changes to 9029 and porx can dynamically load and find the port value specified by operator from mounted ccm.properties file

Closes #

### DIFF
--- a/drivers/storage/portworx/component/telemetry.go
+++ b/drivers/storage/portworx/component/telemetry.go
@@ -112,11 +112,11 @@ const (
 	stagingArcusRegisterProxyURL    = "register.staging-cloud-support.purestorage.com"
 
 	// Ports for telemetry components
-	defaultCCMListeningPort          = 9024
-	defaultCCMListeningPortForPXge30 = 9029
-	defaultCollectorPort             = 10000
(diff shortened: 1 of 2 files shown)


drivers/storage/portworx/component/telemetry.go
```diff
@@ -1210,9 +1210,9 @@ func readConfigMapDataFromFile(
 func GetCCMListeningPort(cluster *corev1.StorageCluster) int {
 	defCCMPort := defaultCCMListeningPort
 
-	pxVer30, _ := version.NewVersion("3.0")
-	if pxutil.GetPortworxVersion(cluster).GreaterThanOrEqual(pxVer30) {
-		defCCMPort = defaultCCMListeningPortForPXge30
+	pxVer2138, _ := version.NewVersion("2.13.8")
```
nrevanna, line 1213: Please add a unit test for this.

zoxpx (approved):
Always happy to see more tests -- thanks Eugene

PR #1210, opened 2023-08-15
PWX-32976: need to set telemetry port to 9029 if px >= 2.13.8

ezhang-px:
* modified px version check

We're backporting a change to 2.13.8 so that the default telemetry port changes to 9029 and porx can dynamically load and find the port value specified by operator from mounted ccm.properties file

Closes #


ezhang-px:
https://purestorage.slack.com/archives/C04TD5T3YDC/p1692127037688849

PR #1213, opened 2023-08-16
PWX-32976: need to mount ccm config for 2.13.8 as well

ezhang-px:
previous [PR](https://github.com/libopenstorage/operator/pull/1204) only updated the version check for shifting the port to 9029 when px is 2.13.8, but operator also needs to allow volume mount of ccm.properties file to porx container otherwise oci-mon won't be able to find it and copy to /etc/pwx for porx' diag calls

Closes #

### DIFF
--- a/drivers/storage/portworx/deployment.go
+++ b/drivers/storage/portworx/deployment.go
@@ -49,13 +49,14 @@ const (
 )
 
 var (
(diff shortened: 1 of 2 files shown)


nikolaypopov (approved):
LGTM

PR #1224, opened 2023-08-21
Windows Airgap Support : Update windows image's tags 

nikita-bhatia:
Issue :
Currently, we are using following 3 images for pods scheduled on windows node :

1. docker.io/portworx/livenessprobe:v2.10.0-windows
2. docker.io/portworx/px-windows-csi-driver:v0.1
3. docker.io/portworx/csi-node-driver-registrar:w2.8.0-windows

where 1 and 2 are opensource image, so when downloading these images on windows nodes, there are 2 issues

1. The images are unable to download on windows node on airgap cluster, and needs to be pushed only from windows platform to custom image registry
2. The image can not be pushed with tag 'v' to custom image registry as it is being used by other pods on linux nodes

Resolution :
To solve this issue, we have retagged the images from windows env and pushed it to portworx repository with same version and postfix "-windows".

related PR : https://github.com/portworx/px-installer/pull/1859

### DIFF
--- a/deploy/crds/core_v1_storagecluster_crd.yaml
+++ b/deploy/crds/core_v1_storagecluster_crd.yaml
@@ -3831,6 +3831,9 @@ spec:
                   csiWindowsDriver:
                     type: string
                     description: Desired image for csi driver for windows.
+                  csiWindowsNodeRegistrar:
+                    type: string
+                    description: Desired image for csi driver for windows.
                   csiDriverRegistrar:
                     type: string
                     description: Desired image for CSI driver registrar.
--- a/deploy/crds/core_v1_storagecluster_crd.yaml
+++ b/deploy/crds/core_v1_storagecluster_crd.yaml
@@ -3828,7 +3828,7 @@ spec:
                   csiLivenessProbe:
                     type: string
                     description: Desired image for Liveness probe.
-                  csiDriverWin:
+                  csiWindowsDriver:
                     type: string
                     description: Desired image for csi driver for windows.
                   csiDriverRegistrar:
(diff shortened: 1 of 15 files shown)


zoxpx (approved):
Note, it should definitely be possible to add multiple binaries/arhictecures into a single container -- for example, here's one such container:

```
$ docker manifest inspect docker.io/library/httpd:latest | jq -cr '.manifests[].platform' 
{"architecture":"amd64","os":"linux"}
{"architecture":"arm","os":"linux","variant":"v5"}
{"architecture":"arm","os":"linux","variant":"v7"}
{"architecture":"arm64","os":"linux","variant":"v8"}
{"architecture":"386","os":"linux"}
{"architecture":"mips64le","os":"linux"}
{"architecture":"ppc64le","os":"linux"}
{"architecture":"s390x","os":"linux"}
```

Note, however, that by default the `docker` will be selective, and pull only the layers for its own architecture (not all architectures).
Also, we know that Linux docker simply refuses to handle Windows-images, so this might be of a limited use.

Anyway, these code-changes look OK to me..

nikita-bhatia:
Thanks @zoxpx for approving.

> Note, it should definitely be possible to add multiple binaries/arhictecures into a single container -- for example, here's one such container:
> 
> ```
> $ docker manifest inspect docker.io/library/httpd:latest | jq -cr '.manifests[].platform' 
> {"architecture":"amd64","os":"linux"}
> {"architecture":"arm","os":"linux","variant":"v5"}
> {"architecture":"arm","os":"linux","variant":"v7"}
> {"architecture":"arm64","os":"linux","variant":"v8"}
> {"architecture":"386","os":"linux"}
> {"architecture":"mips64le","os":"linux"}
> {"architecture":"ppc64le","os":"linux"}
> {"architecture":"s390x","os":"linux"}
> ```
> 
> Note, however, that by default the `docker` will be selective, and pull only the layers for its own architecture (not all architectures). Also, we know that Linux docker simply refuses to handle Windows-images, so this might be of a limited use.
> 
> Anyway, these code-changes look OK to me..

Thanks @zoxpx for approving, 
Thanks for explaining this, I can try this and test to see if it works fine on both platforms. If it works, i can target this for next release. Merging for now, also, cherry-picking to 23.7.0

PR #1236, opened 2023-09-01
Setting up 23.4.1 branch

nikita-bhatia:


deploy/olm-catalog/portworx/portworx.package.yaml
```diff
@@ -1,7 +1,7 @@
 packageName: portworx-certified
 channels:
 - name: alpha
-  currentCSV: portworx-operator.v23.4.0
+  currentCSV: portworx-operator.v23.4.1
 - name: stable
-  currentCSV: portworx-operator.v23.4.0
+  currentCSV: portworx-operator.v23.4.1
```
zoxpx, line 6: I think we had some trouble with this file in the lates builds -- we'll need to check via Jenkins builds

deploy/operator.yaml
```diff
@@ -20,7 +20,7 @@ spec:
     spec:
       containers:
       - name: portworx-operator
-        image: portworx/px-operator:1.11.0-dev
+        image: portworx/px-operator:23.4.1-dev
```
zoxpx, line 23: hmm.. we haven't changed this at all (even 23.5.1 branch still has the old operator version)

zoxpx (approved):
Just eyeballing the changes -- they look OK

deploy/operator.yaml
nikita-bhatia: do we want to keep 1.11.0-dev?

zoxpx:
We have a ..
> --- FAIL: TestMonitoringMetricsEnabled (0.04s)

.. that's because the guys have updated some files on the docs.portworx.com  (they kept breaking the px-23.7.0 build because of that..)

harsh-px:
Just comment out the TestMonitoringMetricsEnabled UT from this hotfix branch

deploy/operator.yaml
zoxpx: It may be harmless, but I'd still roll it back  (perhaps fix it with 23.7.0)?
[ it's not a good time for experiments right now.. ]

harsh-px: +1 , no experiments now.

drivers/storage/portworx/components_test.go
```diff
@@ -10763,7 +10763,7 @@ func TestMonitoringMetricsEnabled(t *testing.T) {
 	require.Len(t, serviceMonitor.OwnerReferences, 1)
 	require.Equal(t, cluster.Name, serviceMonitor.OwnerReferences[0].Name)
 	require.Equal(t, expectedServiceMonitor.Spec, serviceMonitor.Spec)
-}
+}*/
```
zoxpx, line 10766: LOL -- guess that's one way to solve this.
* another way would be to copy updated `drivers/storage/portworx/testspec/prometheusRule.yaml` file from `px-rel-23.7.0` branch

nikita-bhatia: commented it for now

deploy/operator.yaml
nikita-bhatia: reverted back to 1.11.0-dev https://github.com/libopenstorage/operator/pull/1236/commits/80dd4fa2a537c17a981a77d50e237b9b618cf20e

drivers/storage/portworx/components_test.go
zoxpx: That should be fine.. we won't carry this change into latest operator releases (23.7.0+) -- and we don't anticipate many more releases on this branch

PR #1255, opened 2023-09-07
The `-rt_opts` handling cleanup

zoxpx:

The `rt_opts` (runtime options) were first stashed into a HashMap, then collected into an array before passing them on as the options to the OCI-Monitor

ISSUE:
* since we used the HashMap -> array, the entries in the array will be UNSORTED (sorted randomly)
* this is an issue, since changing the order of parameters passed to the OCI-Monitor can cause `portworx.service` restarts

FIX:
* as a fix, we are _sorting_ the runtime-options array, before adding them into the OCI-Monitor's arguments

Closes # (no ticket)

### DIFF
--- a/drivers/storage/portworx/deployment.go
+++ b/drivers/storage/portworx/deployment.go
@@ -107,7 +107,8 @@ type template struct {
 
 func newTemplate(
 	cluster *corev1.StorageCluster,
-	nodeName string) (*template, error) {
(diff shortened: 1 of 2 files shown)


piyush-nimbalkar (approved):

zoxpx:
Thanks for the review Piyush -- merging PR into `master`

PR #1264, opened 2023-09-14
PWX-33374:Set Security context for OCP before pre-flight is launched.

jrivera-px:
Preflight running in a custom namespace (portworx) failed to start in OCP on a fresh cluster due to OCP security setting missing for the custom namespace.   The setting where missing because they where not being applied until the pre-install of the storage cluster which was done after pre-flight was run.   So now we are installing the security setting during pre-flight also to make sure they exist.
 
Closes #
PWX-33374

### DIFF
--- a/drivers/storage/portworx/portworx.go
+++ b/drivers/storage/portworx/portworx.go
@@ -98,7 +98,7 @@ func (p *portworx) Init(
 	return nil
 }
 
-func EnableSecurityContextForValidate(recorder record.EventRecorder, cluster *corev1.StorageCluster) error {
+func createSecurityContextForValidate(recorder record.EventRecorder, cluster *corev1.StorageCluster) error {
 	if secComp, exists := component.Get(component.SCCComponentName); exists {
 		if secComp.IsEnabled(cluster) {
 			err := secComp.Reconcile(cluster)
@@ -156,7 +156,7 @@ func (p *portworx) Validate(cluster *corev1.StorageCluster) error {
 		return nil
 	}
 
-	if err := EnableSecurityContextForValidate(p.recorder, cluster); err != nil {
+	if err := createSecurityContextForValidate(p.recorder, cluster); err != nil {
 		setClusterCondition(corev1.ClusterConditionStatusFailed, err.Error())
 		deletePreflight()
 		return err
--- a/drivers/storage/portworx/portworx.go
+++ b/drivers/storage/portworx/portworx.go
@@ -53,7 +53,8 @@ const (
 )
 
 var (
-	ErrNodeListEmpty = stderr.New("no nodes were available for px installation")
+	ErrNodeListEmpty  = stderr.New("no nodes were available for px installation")
+	installSecContext = true
 )
 
 type portworx struct {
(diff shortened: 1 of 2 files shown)


drivers/storage/portworx/portworx.go
```diff
@@ -53,7 +53,8 @@ const (
 )
 
 var (
-	ErrNodeListEmpty = stderr.New("no nodes were available for px installation")
+	ErrNodeListEmpty  = stderr.New("no nodes were available for px installation")
+	installSecContext = true
```
Pure-AdamuKaapan, line 57: Is the assumption that this will run every time the operator starts? Since it will be set to true on every fresh start

drivers/storage/portworx/portworx.go
```diff
@@ -98,8 +99,30 @@ func (p *portworx) Init(
...
+		}
+	}
+	return nil
+}
+
 func (p *portworx) Validate(cluster *corev1.StorageCluster) error {
 
+	if installSecContext {
+		if err := EnableSecurityContextForValidate(p.recorder, cluster); err == nil {
+			installSecContext = false // if no err, done enabling
```
Pure-AdamuKaapan, line 122: There is a small gap here - if we install a Portworx cluster, then uninstall it and install another one, all without restarting the operator, the second install will not have the security context enabled.

Can we either reset this variable on uninstall, or find a way to do the check more "atomically" without relying on an in-memory state variable?

Pure-AdamuKaapan: Also ideally add a unit test for this case :)

Pure-AdamuKaapan (changes requested):
One small gap I can see for repeat installs, but the core method makes sense.

drivers/storage/portworx/portworx.go
jrivera-px: Yes,  that is ok... This I added was more for the reconcile loop which calls this function every 20 sec until pre-flight is complete.   However I need to have it executed here and its ok, i believe if it is executed multiple time but figured i would add this to reduce the number of redundant calls while pre-flight was running.

drivers/storage/portworx/portworx.go
jrivera-px: yeah actually I can probably use the cluster object to manage this.   Like I mentioned above this is more of small optimization,  redo'ing the call would be a no-op.   But let me see if I can use a cluster status object which is better..

piyush-nimbalkar: It is okay if the reconcile of the scc component gets called on every operator reconcile loop. Most of the calls go the k8s cache and don't hit the api server so they are not expensive or time consuming.

Let's rename the function to `reconcileSecurityContextForValidate` and remove the `installSecContext` variable.

piyush-nimbalkar: What we could do is call this before we start the pre-flight on line 165. That way once the preflight is done, we don't keep calling  this as `PreInstall` will take care of the SCCs.

jrivera-px: @Pure-AdamuKaapan I slacked with Piyush about calling that code in the reconcile loop and he said it as ok.. in fact they are are already doing that indirectly in another place.    However we need it here cause that other place is done after pre-flight is run.. which is actually why this PR was filed.  ```:spiral_calendar_pad:  [1:29 PM](https://purestorage.slack.com/archives/D01NTSM12HJ/p1694723387150449)
Yes, it can be called over and over. We do that in the PreInstall too, which is called by the operator every 30 seconds anyway.
  [1:30 PM](https://purestorage.slack.com/archives/D01NTSM12HJ/p1694723411687099)
Oh ok cool... thanks forgot about that .. awesome..```

drivers/storage/portworx/portworx.go
```diff
@@ -140,6 +156,12 @@ func (p *portworx) Validate(cluster *corev1.StorageCluster) error {
 		return nil
 	}
 
+	if err := EnableSecurityContextForValidate(p.recorder, cluster); err != nil {
```
piyush-nimbalkar, line 159: Can we rename this to `reconcileSecurity...` or `createSecurity...`? We are not enabling the SCCs, but we are just creating them if they don't exist. 
Also, change the function name to lowerCamelCase as it doesn't need to be public.

piyush-nimbalkar (approved):

drivers/storage/portworx/portworx.go
jrivera-px: oh shit sorry... yeah

Pure-AdamuKaapan (approved):
Much better! LGTM

PR #1106, opened 2023-06-29
PWX-31915: kvdb-api pod cleanup

zoxpx:
* will remove the terminated kvdb-api pods

**What this PR does / why we need it**:

We got reports of 2 instances, where Kubernetes stopped recycling a dummy `kvdb-api` POD
* the `kvdb-api` POD is not really running KVDB, it's running a `pause` instead
   - we just use it to report the readyness-status of the `kvdb` service running inside Portworx
* this POD uses `restartPolicy: Always` and `terminationGracePeriodSeconds: 30` -- so it normally never terminates, nor will it reach the `Completed` or `Failed` state
   - instead, within 30 seconds it is normally garbage-collected and restarted by Kubernetes

ISSUE:
We got two reports from different environments, where `kvdb-api` POD has terminated, and was not restarted by Kubernetes -- it would just terminate, and transition into `Completed` state
* the environments reporting this were OCP 4.12, during upgrade finalization; also Rancher node after reboot, w/ nodes using [graceful-shutdown](https://kubernetes.io/blog/2021/04/21/graceful-node-shutdown-beta/) plugin

FIX:
* we were not able to reproduce this issue in our lab
* however, we have captured a detailed state of one of such `Completed` PODs from a customer's site, and we're using this state to _implement the explicit POD-cleanup_, as a part of the "KVDB reconcile" process

**Which issue(s) this PR fixes** (optional)
Closes # PWX-31915

**Special notes for your reviewer**:

PR #1279, opened 2023-10-03
PWX-31915: kvdb-api pod extended cleanup

zoxpx:
* will explicitly remove also the failed-due-node-shutdown PODs

This is the extension of https://github.com/libopenstorage/operator/pull/1106 fix

ISSUE:
* turns out we can encounter other PODs stuck in "weird states" -- that will require explicit cleanup
* if such PODs are left in the system, they can interfere with the OCP upgrades

POD-status details:

```
Status:           Failed
Reason:           NodeShutdown
Message:          Pod Pod was rejected as the node is shutting down.
```

Closes # PWX-31915

Note that we still haven't reproduced the original issue --
* we know that it is linked to the OCP upgrade, where OCP will  cordon and restart a K8s-node (also replace K8s and OS binaries/directories/kernel/...) -- and
* this can cause PODs to get "orphaned" and stop being handled / garbage-collected by the K8s API-server

### DIFF
--- a/pkg/controller/storagenode/storagenode.go
+++ b/pkg/controller/storagenode/storagenode.go
@@ -253,8 +253,9 @@ func (c *Controller) syncKVDB(
 		if p.Spec.NodeName == storageNode.Name {
 			// Let's delete the pod so that a new one will be created.
 			reason := strings.ToLower(p.Status.Reason)
-			if reason == "outofpods" || reason == "evicted" || reason == "terminated" {
-				logrus.Warningf("Found pod with %s status, will delete it: %+v", p.Status.Reason, p)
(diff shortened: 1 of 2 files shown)


harsh-px (approved):

piyush-nimbalkar (approved):

zoxpx:
Thanks for the reviews Harsh and Piyush -- merging this PR, will cherry-pick for `px-23.10.0`

zoxpx:
Woops -- messed up with the tickets --
* this PR has a dedicated ticket PWX-33831
* it is, however, an extension of the work already done w/ PWX-31915 ticket

zoxpx:
Cherry-picking into `px-rel-23.10.0`

PR #1281, opened 2023-10-05
PWX-34238 & PWX-31068 - Add gce/azure/vsphere cloud provider checks add to PX param.

jrivera-px:
Need to send the cloud provider to PX so it can be included in callhome.   This started as cherry-pick but required more changes to test and make sure that pre-flight only runs when its supposed to. 

Closes #
PWX-31068
This is actually just a cherry-pick of code which exists in master but had to be changed some for 2.10.0.


pkg/controller/storagecluster/storagecluster.go
```diff
@@ -733,6 +733,25 @@ func (c *Controller) miscCleanUp(cluster *corev1.StorageCluster) error {
 	return nil
 }
 
+func preflightShouldRun(cluster *corev1.StorageCluster) bool {
+	pxVer30, _ := version.NewVersion("3.0")
+	pxVer31, _ := version.NewVersion("3.1")
+
+	// Preflight should only run freshInstall and if the PX version is 3.0.0 and above
+	if pxutil.IsFreshInstall(cluster) {
+		if pxutil.GetPortworxVersion(cluster).GreaterThanOrEqual(pxVer30) {
```
piyush-nimbalkar, line 742: nit: Can we extract the value of GetPortworxVersion into a variable instead of calling it twice?

piyush-nimbalkar (approved):

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -733,6 +733,25 @@ func (c *Controller) miscCleanUp(cluster *corev1.StorageCluster) error {
...
+			if !pxutil.IsVsphere(cluster) {
```
nrevanna, line 743: I am missing something here.`preflightShouldRun` would return true for all platforms other than vSphere according to this if condition. What's stopping preflights from running on non AWS platforms in 3.0?

jrivera-px: That is done here in portworx.go:SetDefaultsOnStorageCluster():320  

 // Initialize the preflight check annotation                                                                                                                         
        if preflight.RequiresCheck() {
                if _, ok := toUpdate.Annotations[pxutil.AnnotationPreflightCheck]; !ok {
                        toUpdate.Annotations[pxutil.AnnotationPreflightCheck] = "true"
                }
        }

It checks ....  
// RequiresCheck returns whether a preflight check is needed based on the platform                                                                                           
func RequiresCheck() bool {
        return Instance().ProviderName() == string(cloudops.AWS) ||
                Instance().ProviderName() == string(cloudops.Vsphere)
}

This PWX is set to have this consolidated into the driver:  https://portworx.atlassian.net/browse/PWX-33218

PR #1284, opened 2023-10-06
PWX-32670: Vsphere preflight should not run on PKS

jrivera-px:
Prevent preflight from running when using Vsphere and PKS.   Also preflight should only run on Vsphere, Pure, Azure if PX image is >=3.1.0  
Closes #
PWX-32670
Added UTs however and validated this manually verified that with this change pre-flight does not run when using PKS on Vsphere.

### DIFF
--- a/pkg/preflight/utils.go
+++ b/pkg/preflight/utils.go
@@ -14,6 +14,16 @@ func IsGKE() bool {
 	return Instance().ProviderName() == string(cloudops.GCE) && Instance().K8sDistributionName() == gkeDistribution
 }
 
+// IsGKE returns whether the cloud environment is running PKS
+func IsPKS() bool {
+	return Instance().K8sDistributionName() == pksDistribution
+}
+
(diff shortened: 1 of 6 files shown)


pkg/preflight/utils.go
```diff
@@ -14,6 +14,16 @@ func IsGKE() bool {
...
 
+// IsGKE returns whether the cloud environment is running PKS
+func IsPKS() bool {
+	return Instance().K8sDistributionName() == pksDistribution
+}
+
+// IsAzure returns whether the cloud environment is running on Azure
+func IsAzure() bool {
+	return Instance().ProviderName() == string(cloudops.Azure)
+}
```
harsh-px, line 25: Do we need azure code for this PR?

jrivera-px: Added that so we can check for Azure and if we are running a PX image of 3.1.0 for pre-flight.   Using this so I don't have include cloudops in storagecluster.

harsh-px (approved):

PR #1294, opened 2023-10-12
PWX-34098: Pxstore-v2 preflight should run on azure & >= PX 3.1.0

jrivera-px:
 Pxstore-v2 preflight should run on azure & >= PX 3.1.0

Closes #
PWX-34098
Unit test exists.  But pending Azure setup for more testing.

### DIFF
--- a/drivers/storage/portworx/portworx.go
+++ b/drivers/storage/portworx/portworx.go
@@ -499,6 +499,10 @@ func (p *portworx) preflightShouldRun(toUpdate *corev1.StorageCluster) bool {
 			return true
 		}
 
+		if preflight.IsAzure() { // Preflight runs on Azure & PX 3.1.0 or above
+			return true
+		}
+
 		if pxutil.IsVsphere(toUpdate) {
 			if preflight.IsPKS() { // Don't run preflight on Vsphere w/PKS
 				return false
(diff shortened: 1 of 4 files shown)


drivers/storage/portworx/portworx.go
```diff
@@ -499,6 +499,10 @@ func (p *portworx) preflightShouldRun(toUpdate *corev1.StorageCluster) bool {
 			return true
 		}
 
+		if preflight.IsAzure() { // Preflight runs on Azure & PX 3.1.0 or above
```
harsh-px, line 502: nit: `& PX 3.1.0 or above` can be removed from the comment. It's repetitive as line 497 already covers that.

pkg/preflight/utils.go
```diff
@@ -33,7 +33,8 @@ func IsAWS() bool {
 func RequiresCheck() bool {
 	return Instance().ProviderName() == string(cloudops.AWS) ||
 		Instance().ProviderName() == string(cloudops.Vsphere) ||
-		Instance().ProviderName() == string(cloudops.Pure)
+		Instance().ProviderName() == string(cloudops.Pure) ||
+		Instance().ProviderName() == string(cloudops.Azure)
 }
 
 // RunningOnCloud checks whether portworx is running on cloud
```
harsh-px, line 40: Not related to your change but we should add IsGKE(), IsAWS(), IsAzure() under the RunningOnCloud() function.

harsh-px (approved):

pkg/preflight/utils.go
jrivera-px: Will add it to the next checkin...

PR #1297, opened 2023-10-17
PWX-31625: Turning off PVC-Controller on OpenShift

zoxpx:

We got complaints that the PX deployments on OpenShift spin up `portworx-pvc-controller-XXX` PODs
* apparently, in vSphere deployments, this results in spamming vSphere/vCenter with storage tasks

FIX:
* as a fix, we're changing the defaults in which configs will we set up PVC-controller
  - we'll SKIP setting up the PVC-controller in OpenShift distros  (unless explicitly turned on via `portworx.io/pvc-controller: "true"`)
* also adding a targeted unit-test

Closes # PWX-31625

Some manual test results:
* confirmed generic Kubernetes deployment in "portworx" namespace still NEEDS PVC-controller when using in-tree driver
   - without the PVC-controller, the PVC never gets assigned to the application POD, and POD never starts
* identical deployment tested on OCP 4.12 / vSphere -- woks OK
   - PVCs get bound+started correctly, even without the PVC-controller component

### DIFF
--- a/drivers/storage/portworx/component/pvccontroller.go
+++ b/drivers/storage/portworx/component/pvccontroller.go
@@ -12,6 +12,7 @@ import (
 	"github.com/libopenstorage/operator/pkg/constants"
 	"github.com/libopenstorage/operator/pkg/util"
 	k8sutil "github.com/libopenstorage/operator/pkg/util/k8s"
+	"github.com/sirupsen/logrus"
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


piyush-nimbalkar (approved):
Looks good. 

We should also verify this for cloud openshift distributions like - 
1. Just openshift deployment on AWS
2. ARO - Azure RedHat Openshift

The first one will probably work after this change. I am not sure about ARO.

zoxpx:
Tested OpenShift v4.10.67 deployment to Azure cloud, using `portworx` namespace and "in-tree" Portworx K8 driver (non-CSI) --> the PVCs are getting created properly.

PR #1308, opened 2023-10-24
PWX-34203: Adding Azure support

nrevanna:
- Adding Azure specific env variables to stc
    - Removing cloud_provider so that it can be moved to
      pure-px/cloud-drive-tests

Closes #

This is needed for
https://github.com/pure-px/cloud-drive-tests/pull/4
I was able to run those tests which is a validation for this


jrivera-px (approved):
LGTM....

pkg/util/test/util.go
```diff
@@ -629,6 +632,59 @@ func FindAndCopyVsphereSecretToCustomNamespace(customNamespace string) error {
...
+	var envVars []v1.EnvVar
+
+	// Get PX vSphere secret
+	_, err := coreops.Instance().GetSecret(DefaultPxAzureSecretName, namespace)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			logrus.Warnf("Azure secret %s in not found in %s namespace", DefaultPxAzureSecretName, namespace)
+			return nil, nil
+		}
+		return nil, fmt.Errorf("failed to get secret %s in %s namespace, err %v", DefaultPxVsphereSecretName, namespace, err)
```
pp511, line 646: Trying to understand the intention behind not sending error when errors.IsNotFound

nrevanna: I think we should return an error. Will raise a follow up PR

PR #1311, opened 2023-10-26
PWX-34185 preflight check for gce

chiachunlin:
preflight check for dmthin gke

Torpedo run with preflight PASSED 
https://jenkins.pwx.dev.purestorage.com/job/Users/job/ChiaChun/job/gcp-dmthin/31/

Closes #

### DIFF
--- a/drivers/storage/portworx/portworx.go
+++ b/drivers/storage/portworx/portworx.go
@@ -485,7 +485,7 @@ func (p *portworx) preflightShouldRun(toUpdate *corev1.StorageCluster) bool {
 		return check == "true"
 	}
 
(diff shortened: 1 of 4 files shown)


jrivera-px (approved):
LGTM... in porx/px-runc.go we have func `sanitizeDriveType` which is used to check the input drive format.   Need to make sure that it will work for this metadata drive type you specified.   Its possible you may need to make changes or skip validation for provider `gce`.

chiachunlin:
> LGTM... in porx/px-runc.go we have func `sanitizeDriveType` which is used to check the input drive format. Need to make sure that it will work for this metadata drive type you specified. Its possible you may need to make changes or skip validation for provider `gce`.

thanks for reminder, this has been addressed.

PR #1316, opened 2023-10-30
PWX-34145: Exposing Portworx struct

yjoshi-px:
Exposing portworx struct and getKvdbMap function

Closes #


piyush-nimbalkar (changes requested):
Just pasting the comment from our slack conversation for everyone's ref - 

Let's not treat the `portworx` struct as an util if we are going to take some of the functions from it and make them public for that purpose. That is not how we should be changing the code.
Let’s do one of the following -
- Let’s create static utility methods like suggested above.
- Let’s create a new struct which exposes these public methods. We embed that struct in the portworx struct so the operator code does not change. Your test code can use that struct directly.

drivers/storage/portworx/util/util.go
```diff
@@ -247,6 +249,10 @@ const (
 	// VsphereInstallModeLocal env value for Vsphere 'local' install
 	VsphereInstallModeLocal = "local"
 )
+const (
+	// pxEntriesKey is key which holds all the bootstrap entries
+	pxEntriesKey = "px-entries"
```
piyush-nimbalkar, line 254: Can you completely remove the duplicate `pxEntriesKey` constant from status.go and replace all occurrences with this one?

drivers/storage/portworx/util/util.go
```diff
@@ -1570,3 +1576,49 @@ func GetTLSCipherSuites(cluster *corev1.StorageCluster) (string, error) {
...
+	kvdbNodeMap := make(map[string]*kvdb_api.BootstrapEntry)
+	if cluster.Spec.Kvdb != nil && cluster.Spec.Kvdb.Internal {
+		clusterID := GetClusterID(cluster)
+		strippedClusterName := strings.ToLower(ConfigMapNameRegex.ReplaceAllString(clusterID, ""))
+		cmName := fmt.Sprintf("%s%s", InternalEtcdConfigMapPrefix, strippedClusterName)
+
+		cm := &v1.ConfigMap{}
+		err := k8sClient.Get(context.TODO(), types.NamespacedName{
+			Name:      cmName,
+			Namespace: "kube-system",
```
piyush-nimbalkar, line 1592: Let's redeclare the constant `BootstrapCloudDriveNamespace` in this file and change the existing references to it.

piyush-nimbalkar (approved):
LGTM after the requested changes.

drivers/storage/portworx/util/util.go
yjoshi-px: Updated

drivers/storage/portworx/util/util.go
yjoshi-px: Updated

PR #1318, opened 2023-10-31
PWX-34796: added scc and vol mnt

ezhang-px:
In op-23.5.0, usage of podsecuritypolicy for autopilot was removed if deployed on k8s 1.25 beyond, but in OCP in order for autopilot to access directories like /var/cores, we also need to allow its clusterrole to use portworx-restricted scc and also mount this folder from host.

Closes # PWX-34796

Error messages like 
`time="30-10-2023 17:15:34" level=error msg="failed to create debug logs: mkdir /var/cores: permission denied" file="main.go:315"
time="30-10-2023 17:15:34" level=error msg="failed to start custom signal handlers: mkdir /var/cores: permission denied" file="main.go:277"`
are gone from autopilot log on OCP with this change

### DIFF
--- a/drivers/storage/portworx/component/autopilot.go
+++ b/drivers/storage/portworx/component/autopilot.go
@@ -70,6 +70,15 @@ var (
 				},
 			},
 		},
+		{
+			Name:      "varcores",
+			MountPath: "/var/cores",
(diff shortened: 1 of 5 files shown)


zoxpx (approved):
Thanks Eugene!

PR #1325, opened 2023-11-07
PWX-28442 Use quorum member flag to count storage nodes

piyush-nimbalkar:
- Earlier we were looking at the pools in the nodes enumerate response to
determine if a node is a storage node or not. This is not reliable as sometimes
during the intialization portworx does not populate the pools. This misleads the
operator to consider it as a storageless node.
- We recently added the NonQuorumMember flag in the nodes enumerate
response. This is more reliable and will always correctly tell whether a node is
storage or storageless.

Closes # https://portworx.atlassian.net/browse/PWX-28442


zoxpx (approved):
The code-changes look OK -- but it seems like they also provoked a big /vendor update

Let's please add this into `px-rel-23.12.0` soon, so we can take advantage of test-automation.

Thanks Piyush

PR #1334, opened 2023-11-16
PWX-34360: Move csi-node-driver-registrar container from px pods to px-api daemonset

svijaykumar-px:
**What this PR does / why we need it**: Currently the csi-node-driver-registrar is a part of the px-pods (oci monitor pods). When a node drain happens the pod gets deleted. This PR moves the csi-node-driver-registrar to the portworx-api daemonset so that it remains running even during the node drain

Closes # 34360

**Special notes for your reviewer**: The portworx-api pods will also be retsarted when the storage cluster is updated. This will effect px versions 2.13 and beyond. This is to ensure the newer operator changes will be reflected on the px pods and the csi nodes will be registered.


pkg/controller/storagecluster/update.go
```diff
@@ -124,6 +144,38 @@ func (c *Controller) rollingUpdate(cluster *corev1.StorageCluster, hash string)
 	return c.syncNodes(cluster, oldPodsToDelete, []string{}, hash)
 }
 
+func getPxAPiPodForNode(nodeName string, ApiPods []*v1.Pod) *v1.Pod {
+
+	for _, pod := range ApiPods {
+		fmt.Println("PXPOD 2", pod.Spec.NodeName)
```
zoxpx, line 150: please convert to `logrus.Debugf()` or `logrus.Tracef()` if not required to always log this

pkg/controller/storagecluster/update.go
```diff
@@ -124,6 +144,38 @@ func (c *Controller) rollingUpdate(cluster *corev1.StorageCluster, hash string)
...
+		&client.ListOptions{
+			Namespace:     cluster.Namespace,
+			LabelSelector: labels.SelectorFromSet(map[string]string{"name": "portworx-api"}),
+		},
+	)
+
+	if err != nil {
+		return nil
+	}
+	result := make([]*v1.Pod, 0)
```
zoxpx, line 171: nit: since we know the array size in advance, it is better to set the array's capacity -- so we avoid resizing the arrray to "double the size"

```go
result := make([]*v1.Pod, 0, len(podList.Items))
```

pkg/controller/storagecluster/update.go
```diff
@@ -124,6 +144,38 @@ func (c *Controller) rollingUpdate(cluster *corev1.StorageCluster, hash string)
 	return c.syncNodes(cluster, oldPodsToDelete, []string{}, hash)
 }
 
+func getPxAPiPodForNode(nodeName string, ApiPods []*v1.Pod) *v1.Pod {
```
zoxpx, line 147: nit: you have a typo with `APi`

drivers/storage/portworx/deployment.go
```diff
@@ -1407,7 +1340,7 @@ func (t *template) mountsFromVolInfo(vols []volumeInfo) []v1.VolumeMount {
 func (t *template) getVolumes() []v1.Volume {
 	volumeInfoList := getDefaultVolumeInfoList(t.pxVersion)
 	extensions := []func() []volumeInfo{
-		t.getCSIVolumeInfoList,
+		//t.getCSIVolumeInfoList,
```
zoxpx, line 1343: please remove

zoxpx (approved):
I think this looks OK -- thanks Swarupa.  Some minor remarks below..

drivers/storage/portworx/component/portworx_api.go
```diff
@@ -264,6 +265,15 @@ func getPortworxAPIDaemonSetSpec(
 		},
 	}
 
+	// If CSI is enabled then run the csi-node-driver-registrar pods in the same daemonset
+	if pxutil.IsCSIEnabled(cluster) {
```
piyush-nimbalkar, line 269: Let's add the px version 2.13 check here too. We can avoid creating duplicate registrar containers that way.

pkg/controller/storagecluster/update.go
```diff
@@ -86,6 +87,16 @@ func (c *Controller) rollingUpdate(cluster *corev1.StorageCluster, hash string)
 		return err
 	}
 
+	// If px version is greater than 2.15 then delete px-api pods along with px pods
+	pxVersion := util.GetPortworxVersion(cluster)
+	supportedPxVersion, _ := version.NewVersion("2.15")
```
piyush-nimbalkar, line 92: The supported version should be 2.13 and not 2.15. Also, please extract a constant for it.

pkg/controller/storagecluster/update.go
```diff
@@ -104,6 +115,15 @@ func (c *Controller) rollingUpdate(cluster *corev1.StorageCluster, hash string)
 				numUnavailableKvdb++
 			}
 		}
+
+		// If pxversion is greater than 2.15 then delete the portworx-api pods with csi-node-registrar containers as well when updating storage cluster
```
piyush-nimbalkar, line 119: *2.13

If you are adding a comment, please explain why it needs to be deleted. Because we want to re-register the csi driver which gets removed when the registrar is removed from the px pod.

pkg/controller/storagecluster/update.go
```diff
@@ -124,6 +144,38 @@ func (c *Controller) rollingUpdate(cluster *corev1.StorageCluster, hash string)
...
+	)
+
+	if err != nil {
+		return nil
+	}
+	result := make([]*v1.Pod, 0)
+	for _, pod := range podList.Items {
+		result = append(result, pod.DeepCopy())
+	}
+	return result
```
piyush-nimbalkar, line 175: Can we just return a map of `nodeName: podName`? We don't need the entire pod object anywhere.
It will avoid the unnecessary loops over the api pods above.

piyush-nimbalkar (changes requested):

pkg/controller/storagecluster/update.go
```diff
@@ -104,6 +115,15 @@ func (c *Controller) rollingUpdate(cluster *corev1.StorageCluster, hash string)
 				numUnavailableKvdb++
 			}
 		}
+
+		// If pxversion is greater than 2.13 then delete the portworx-api pods with csi-node-registrar containers as well when updating storage cluster
+		// This is done to re-register csi driver which gets removed when csi-node-driver-registrar is removed from px pods
+		if restartPxApiPods {
+			if apiPodName, ok := pxApiPodsMap[pod.Spec.NodeName]; ok {
+				oldPodsToDelete = append(oldPodsToDelete, apiPodName)
```
piyush-nimbalkar, line 123: Can we add UTs in controller_test.go to check for portworx-api pods in the deletion list?

drivers/storage/portworx/deployment.go
```diff
@@ -49,13 +49,14 @@ const (
...
-	pxVer2_13_8, _ = version.NewVersion("2.13.8")
-	pxVer3_0_1, _  = version.NewVersion("3.0.1")
+	pxVer2_3_2, _         = version.NewVersion("2.3.2")
+	pxVer2_5_5, _         = version.NewVersion("2.5.5")
+	pxVer2_6, _           = version.NewVersion("2.6")
+	pxVer2_8, _           = version.NewVersion("2.8")
+	pxVer2_9_1, _         = version.NewVersion("2.9.1")
+	pxVer2_13_8, _        = version.NewVersion("2.13.8")
+	pxVer3_0_1, _         = version.NewVersion("3.0.1")
+	supportedPxVersion, _ = version.NewVersion("2.13")
```
piyush-nimbalkar, line 59: `supportPXVersion` is too generic. What is supported is not clear.
Let's rename it to something like `csiRegistrarRemovalPxVersion`.

drivers/storage/portworx/deployment.go
```diff
@@ -323,7 +324,8 @@ func (p *portworx) GetStoragePodSpec(
 		podSpec.DNSPolicy = v1.DNSPolicy(cluster.Annotations[pxutil.AnnotationDNSPolicy])
 	}
 
-	if pxutil.IsCSIEnabled(t.cluster) {
+	pxVersion := pxutil.GetPortworxVersion(cluster)
```
piyush-nimbalkar, line 327: We don't need to calculate this again. The version is already present in `t.pxVersion`. We can use `t.pxVersion.LessThan(...)` on the next line.

drivers/storage/portworx/deployment.go
```diff
@@ -1319,6 +1321,12 @@ func (t *template) getVolumeMounts() []v1.VolumeMount {
 		t.getBottleRocketVolumeInfoList,
 		t.GetVolumeInfoForTLSCerts,
 	}
+
+	pxVersion := pxutil.GetPortworxVersion(t.cluster)
+	if pxVersion.LessThan(supportedPxVersion) {
+		extensions = append(extensions, t.getCSIVolumeInfoList)
+	}
+
```
piyush-nimbalkar, line 1329: The CSI volumes were never part of the volume mount list. Why are we trying to add them now?

drivers/storage/portworx/deployment.go
```diff
@@ -1407,14 +1415,17 @@ func (t *template) mountsFromVolInfo(vols []volumeInfo) []v1.VolumeMount {
...
-		t.getCSIVolumeInfoList,
-		t.getTelemetryVolumeInfoList,
 		t.getK3sVolumeInfoList,
+		t.getTelemetryVolumeInfoList,
 		t.getIKSVolumeInfoList,
 		t.getPKSVolumeInfoList,
 		t.getBottleRocketVolumeInfoList,
 		t.GetVolumeInfoForTLSCerts,
 	}
+	pxVersion := pxutil.GetPortworxVersion(t.cluster)
```
piyush-nimbalkar, line 1425: Same here. Let's just use `t.pxVersion`.

drivers/storage/portworx/deployment_test.go
```diff
@@ -166,6 +169,7 @@ func TestPodSpecWithCustomKubeletDir(t *testing.T) {
 
 	driver := portworx{}
 
+	//Case 1: When portworx version is lesser than 2.13
```
piyush-nimbalkar, line 172: nit: Space after `//`

pkg/controller/storagecluster/update.go
svijaykumar-px: How to get the list of the portworx-api pods? It returns nil when I use k8sclient.

piyush-nimbalkar (approved):

PR #1338, opened 2023-11-18
PWX-35067 (23.10.1): Don't set the cloud provider from preflight provider name (PR #1339)

piyush-nimbalkar:

### Don't set the cloud provider from preflight provider name
The preflight provider name is derived from the k8s node.spec.provider
field. These values don't necessarily match with the portworx cloudops
providers. If we blindly set the preflight values to the cloud provider, it
leads to portworx crashing as it does not recognize that cloudops provider.

Reverting back to the old behavior where we only explicitly check for
vsphere. In future, we can either have common cloud providers between preflight
and cloudops or we can do the conversion somewhere before setting the cloud
provider in storagecluster spec.

### Skipping upgrading to 23.10.0 if users are on older version.

Openshift operator upgrades happen one version at a time. If users go to
23.10.0, they will hit the bug where operator is incorrectly adding the cloud
provider in the storage cluster spec.
That's why we want to skip 23.10.0 upgrade and customers can directly go from
23.7.0 to 23.10.1

Closes # https://portworx.atlassian.net/browse/PWX-35067

Testing Notes:
- Tested this manually  on an Anthos Baremetal cluster
- Tested this manually on an Anthos VMWare cluster with Pure FA CloudDrives

### DIFF
--- a/deploy/olm-catalog/portworx/23.10.1/portworx-certified.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/portworx/23.10.1/portworx-certified.clusterserviceversion.yaml
@@ -4,6 +4,7 @@ metadata:
   name: portworx-operator.v23.10.1
   namespace: placeholder
   annotations:
+    olm.skipRange: '=23.10.0'
     capabilities: Auto Pilot
     categories: "Storage"
     description: Cloud native storage solution for production workloads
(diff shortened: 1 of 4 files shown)


zoxpx (approved):
I think this looks OK -- thanks Piyush!

PR #1339, opened 2023-11-18
PWX-35067 Don't set the cloud provider from preflight provider name

piyush-nimbalkar:
The preflight provider name is derived from the k8s node.spec.provider field. These values don't necessarily match with the portworx cloudops providers. If we blindly set the preflight values to the cloud provider, it leads to portworx crashing as it does not recognize that cloudops provider.

Reverting back to the old behavior where we only explicitly check for vsphere. In future, we can either have common cloud providers between preflight and cloudops or we can do the conversion somewhere before setting the cloud provider in storagecluster spec.

Closes # https://portworx.atlassian.net/browse/PWX-35067

### DIFF
--- a/drivers/storage/portworx/deployment_test.go
+++ b/drivers/storage/portworx/deployment_test.go
@@ -9,6 +9,8 @@ import (
 	"testing"
 
 	"github.com/golang/mock/gomock"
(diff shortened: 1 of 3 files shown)


zoxpx (approved):
I think this looks OK -- thanks Piyush

PR #1347, opened 2023-11-22
PWX-34411 Allow overriding the min available value of px-storage PDB

piyush-nimbalkar:

Currently the storage PDB allows deletion of only one PX pod at any given time. This is to ensure we don't lose volume replica quorum. This allows user to knowingly change the PDB value and run the k8s upgrade faster by upgrading multiple nodes in parallel. The users who will use it understand the risk and we will document accordingly.

Closes # https://portworx.atlassian.net/browse/PWX-34411

### DIFF
--- a/drivers/storage/portworx/component/disruption_budget.go
+++ b/drivers/storage/portworx/component/disruption_budget.go
@@ -100,23 +100,34 @@ func (c *disruptionBudget) createPortworxPodDisruptionBudget(
 	cluster *corev1.StorageCluster,
 	ownerRef *metav1.OwnerReference,
 ) error {
-	var err error
-	c.sdkConn, err = pxutil.GetPortworxConn(c.sdkConn, c.k8sClient, cluster.Namespace)
+	userProvidedMinValue, err := pxutil.MinAvailableForStoragePDB(cluster)
 	if err != nil {
-		return err
+		logrus.Warnf("Invalid value for annotation %s: %v", pxutil.AnnotationStoragePodDisruptionBudget, err)
+		userProvidedMinValue = -1
 	}
 
-	storageNodesCount, err := pxutil.CountStorageNodes(cluster, c.sdkConn, c.k8sClient)
-	if err != nil {
-		c.closeSdkConn()
-		return err
+	var minAvailable int
+	if err != nil || userProvidedMinValue < 0 {
+		c.sdkConn, err = pxutil.GetPortworxConn(c.sdkConn, c.k8sClient, cluster.Namespace)
+		if err != nil {
+			return err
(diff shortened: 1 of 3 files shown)


drivers/storage/portworx/component/disruption_budget.go
```diff
@@ -100,31 +100,41 @@ func (c *disruptionBudget) createPortworxPodDisruptionBudget(
...
+	var minAvailable int
+	if err != nil || userProvidedMinValue < 0 {
+		c.sdkConn, err = pxutil.GetPortworxConn(c.sdkConn, c.k8sClient, cluster.Namespace)
+		if err != nil {
+			return err
+		}
+
+		storageNodesCount, err := pxutil.CountStorageNodes(cluster, c.sdkConn, c.k8sClient)
+		if err != nil {
+			c.closeSdkConn()
```
zoxpx, line 117: Since we called `pxutil.GetPortworxConn()` 7 lines up, shouldn't we be calling the `defer c.closeSdkConn()` unconditionally?

drivers/storage/portworx/component/disruption_budget.go
```diff
@@ -100,31 +100,41 @@ func (c *disruptionBudget) createPortworxPodDisruptionBudget(
 	cluster *corev1.StorageCluster,
 	ownerRef *metav1.OwnerReference,
 ) error {
-	var err error
-	c.sdkConn, err = pxutil.GetPortworxConn(c.sdkConn, c.k8sClient, cluster.Namespace)
+	userProvidedMinValue, err := pxutil.MinAvailableForStoragePDB(cluster)
 	if err != nil {
-		return err
+		logrus.Warnf("Invalid value for annotation %s: %v", pxutil.AnnotationStoragePodDisruptionBudget, err)
```
zoxpx, line 105: Recommend also "nullifying" the `userProvidedMinValue`, since you use it 3 lines down -- i.e.

```go
	if err != nil {
		logrus.Warnf("Invalid value for annotation %s: %v", pxutil.AnnotationStoragePodDisruptionBudget, err)
		userProvidedMinValue = -1
	}
```

zoxpx (approved):
I think this looks OK -- some minor comments listed below

drivers/storage/portworx/component/disruption_budget.go
piyush-nimbalkar: We don't close the connection on happy path as we go through this every 30s. We close it on failure because the failure could be because of the connection.
The `GetPortworxConn` method returns the same `sdkConn` object it it already exists and doesn't create a new one.

PR #1350, opened 2023-11-22
Modifying UTs for docs change in PWX-35016

svijaykumar-px:
**What this PR does / why we need it**: This PR corrects the UTs that fail due to changing the portworx docs in ticket PWX-35016

**Which issue(s) this PR fixes** :Closes # PWX-35138

**Special notes to the reviewer**: Initially, prometheus alerts - PXKvdbNodeViewUnhealthy and PXKvdbClusterViewUnhealthy used the metric label "{$labels.node_id}" to identify nodes which was not valid. The docs website was modified to change the prometheus-rules yaml to use "{$labels.node}" instead. This change has been verified in the prometheus UI as well.

### DIFF
--- a/drivers/storage/portworx/testspec/prometheusRule.yaml
+++ b/drivers/storage/portworx/testspec/prometheusRule.yaml
@@ -173,12 +173,12 @@ spec:
         scrape_target_name: "{{$labels.node}}"
     - alert: PXKvdbNodeViewUnhealthy
       annotations:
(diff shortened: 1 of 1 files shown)


nikita-bhatia (commented):
I think the ask is to change  from $labels.node._id to $labels.node.ID, can you share the doc PR?

svijaykumar-px:
> I think the ask is to change from $labels.node._id to $labels.node.ID, can you share the doc PR?

https://github.com/pure-px/portworx-documentation/pull/279 Here is the link to the docs PR. Since there is no label called nodeID, I will be using "node" as it a more human readable form

nikita-bhatia (approved):
Looks good based on the verification

PR #1352, opened 2023-11-23
PTX-21234: Fix incorrect validation for csi-node-driver-registrar inside portworx-api pods instead of oci-mon pods for older operators

svijaykumar-px:
**What this PR does / why we need it**: Fixes incorrect validation for csi-node-driver-registrar inside portworx-api pods instead of oci-mon pods for older operators. PR-1334 is breaking automation as the changes made in the integration tests do not apply to older versions of operator. Hence temporarily the test has been changed to validate  csi-node-driver-registrar inside portworx-api pods only for master version - 99.9.9. This operator version needs to be modified when the fix in PR-1334 is to be released.

Closes # PTX-21234

### DIFF
--- a/pkg/util/test/util.go
+++ b/pkg/util/test/util.go
@@ -178,6 +178,7 @@ var (
 	opVer23_5_1, _                    = version.NewVersion("23.5.1-")
 	opVer23_7, _                      = version.NewVersion("23.7.0-")
 	minOpVersionForKubeSchedConfig, _ = version.NewVersion("1.10.2-")
+	pxOperatorMasterVersion, _        = version.NewVersion("99.9.9")
 
 	// OCP Dynamic Plugin is only supported in starting with OCP 4.12+ which is k8s v1.25.0+
 	minK8sVersionForDynamicPlugin, _ = version.NewVersion("1.25.0")
@@ -2437,7 +2438,9 @@ func ValidateCsiEnabled(pxImageList map[string]string, cluster *corev1.StorageCl
 
 		var pods *v1.PodList
 		var err error
-		if pxVersion.GreaterThanOrEqual(pxVer2_13) {
+
+		// TODO: Need to change "pxOperatorMasterVersion" to the release version when released
+		if opVersion, _ := GetPxOperatorVersion(); pxVersion.LessThan(pxVer2_13) && opVersion.Equal(pxOperatorMasterVersion) {
 			pods, err = coreops.Instance().GetPods(cluster.Namespace, map[string]string{"name": "portworx-api"})
 			if err != nil {
 				return nil, true, err
@@ -2547,7 +2550,8 @@ func validateCsiContainerInPxPods(namespace string, csi bool, timeout, interval 
 				podsReady++
 			}
 
-			if pxVersion.LessThan(pxVer2_13) {
+			// TODO: Need to change "pxOperatorMasterVersion" to the release version when released
+			if opVersion, _ := GetPxOperatorVersion(); pxVersion.LessThan(pxVer2_13) && opVersion.Equal(pxOperatorMasterVersion) {
 				for _, container := range pod.Spec.Containers {
 					if container.Name == "csi-node-driver-registrar" {
 						pxPodsWithCsiContainer = append(pxPodsWithCsiContainer, pod.Name)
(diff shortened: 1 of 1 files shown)


pkg/util/test/util.go
```diff
@@ -178,6 +178,7 @@ var (
 	opVer23_5_1, _                    = version.NewVersion("23.5.1-")
 	opVer23_7, _                      = version.NewVersion("23.7.0-")
 	minOpVersionForKubeSchedConfig, _ = version.NewVersion("1.10.2-")
+	pxOperatorMasterVersion, _        = version.NewVersion("99.9.9")
```
nikolaypopov, line 181: Can we use this `PxOperatorMasterVersion` instead of "99.9.9", its already defined, just in case we change it to something else

pkg/util/test/util.go
```diff
@@ -2547,7 +2550,8 @@ func validateCsiContainerInPxPods(namespace string, csi bool, timeout, interval
 				podsReady++
 			}
 
-			if pxVersion.LessThan(pxVer2_13) {
+			// TODO: Need to change "pxOperatorMasterVersion" to the release version when released
+			if opVersion, _ := GetPxOperatorVersion(); pxVersion.LessThan(pxVer2_13) && opVersion.Equal(pxOperatorMasterVersion) {
```
nikolaypopov, line 2554: I think we still need this validation, where are we validating CSI containers inside OCI-monitor pods? We only need a condition to not validate CSI containers inside API pods, if the feature is not in the operator version we are validating against.

pkg/util/test/util.go
svijaykumar-px: You mean to say `PxOperatorMasterVersion` has already been defined elsewhere?

nikolaypopov: Yes, in same file in const

pkg/util/test/util.go
svijaykumar-px: Oh okay got it, will make the changes

pkg/util/test/util.go
```diff
@@ -2546,37 +2557,74 @@ func validateCsiContainerInPxPods(namespace string, csi bool, timeout, interval
...
-			apiPods, err := coreops.Instance().GetPods(namespace, listOptionsNew)
-			if err != nil {
-				return nil, false, err
+		if csi {
+			if len(pxPodsWithCsiContainer) != len(pods.Items) {
+				return nil, true, fmt.Errorf("failed to validate CSI containers in PX Api pods: expected %d, got %d, %d/%d Ready pods", len(pods.Items), len(pxPodsWithCsiContainer), podsReady, len(pods.Items))
 			}
+		} else {
+			if len(pxPodsWithCsiContainer) > 0 || len(pods.Items) != podsReady {
+				return nil, true, fmt.Errorf("failed to validate CSI container in PX pods Api: expected: 0, got %d, %d/%d Ready pods", len(pxPodsWithCsiContainer), podsReady, len(pods.Items))
```
nikolaypopov, line 2575: small nit
```suggestion
				return nil, true, fmt.Errorf("failed to validate CSI container in PX Api pods [portworx-api]: expected: 0, got %d, %d/%d Ready pods", len(pxPodsWithCsiContainer), podsReady, len(pods.Items))
```

pkg/util/test/util.go
```diff
@@ -2546,37 +2557,74 @@ func validateCsiContainerInPxPods(namespace string, csi bool, timeout, interval
...
 
 		}
 
-		if pxVersion.GreaterThanOrEqual(pxVer2_13) {
-			apiPods, err := coreops.Instance().GetPods(namespace, listOptionsNew)
-			if err != nil {
-				return nil, false, err
+		if csi {
+			if len(pxPodsWithCsiContainer) != len(pods.Items) {
+				return nil, true, fmt.Errorf("failed to validate CSI containers in PX Api pods: expected %d, got %d, %d/%d Ready pods", len(pods.Items), len(pxPodsWithCsiContainer), podsReady, len(pods.Items))
```
nikolaypopov, line 2571: small nit
```suggestion
				return nil, true, fmt.Errorf("failed to validate CSI containers in PX Api pods [portworx-api]: expected %d, got %d, %d/%d Ready pods", len(pods.Items), len(pxPodsWithCsiContainer), podsReady, len(pods.Items))
```

nikolaypopov (approved):
Small nits, just to have a clear distinction in the errors, before PX pods were just meant OCI pods, but technically all of them are PX pods :) can we just add more verbosity.

I cannot comment on a code that you did not touch, but can you also please modify same errors in the `validateCsiContainerInPxPods()` func to say `failed to validate CSI container in PX OCI monitor pods [portworx]: expected...`

That should do it. Thank you for making this change!

PR #1355, opened 2023-11-28
PWX-30050: Add PriorityClass field to the STC CDR that the operator will just pass to the px-storage pods

svijaykumar-px:
**What this PR does / why we need it**: Solves a customer ask where customer wants to  add a PriorityClass to the StorageCluster so the portworx server pods (px-storage) can get this PriorityClass and be scheduled accordingly.

Closes # PWX-30050


nikita-bhatia (approved):
LGTM, please add testing notes in the comment.

PR #1356, opened 2023-11-29
PWX-31340: Add capability to handle fatal pre-flight error which will…

jrivera-px:
… prevent PX startup.

For AWS io2 disk are not supported for dmthin or btrfs.    Usually operator will start PX with btrfs if any pre-flight check fails.  However if a "fatal" result is returned operator will not start PX.  

Closes #
PWX-31340
This PX PR handles the 'io2' fatal error returned:  https://github.com/portworx/porx/pull/12663 and we need to vendor in APIs into px-installer after this merge is made.

### DIFF
--- a/deploy/crds/core_v1_storagenode_crd.yaml
+++ b/deploy/crds/core_v1_storagenode_crd.yaml
@@ -148,6 +148,9 @@ spec:
                     success:
                       type: boolean
                       description: If true, the check was successful
+                    result:
+                      type: string
+                      description: Result of the success or failure of the check
               geography:
                 type: object
                 description: Contains topology information for the storage node.
--- a/deploy/crds/core_v1_storagenode_crd.yaml
+++ b/deploy/crds/core_v1_storagenode_crd.yaml
@@ -150,7 +150,7 @@ spec:
                       description: If true, the check was successful
                     result:
                       type: string
-                      description: Result of the check: fatal, warning, success
+                      description: Result of the check fatal, warning, success
               geography:
                 type: object
                 description: Contains topology information for the storage node.
--- a/deploy/crds/core_v1_storagenode_crd.yaml
+++ b/deploy/crds/core_v1_storagenode_crd.yaml
@@ -150,7 +150,7 @@ spec:
                       description: If true, the check was successful
                     result:
                       type: string
-                      description: Result of the success or failure of the check
+                      description: Result of the check: fatal, warning, success
               geography:
                 type: object
                 description: Contains topology information for the storage node.
--- a/drivers/storage/portworx/preflight.go
+++ b/drivers/storage/portworx/preflight.go
@@ -323,6 +323,8 @@ func (u *preFlightPortworx) processNodesChecks(recorder record.EventRecorder, st
 	}
 
 	passed := true
+	isFatal := false
+	fmsgs := []string{}
 	for _, node := range storageNodes {
 		// Process storageNode checks list for failures. Also make sure the "status" check entry
 		// exists, this indicates all the checks were submitted from the pre-flight pod.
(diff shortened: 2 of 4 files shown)


drivers/storage/portworx/preflight.go
```diff
@@ -369,6 +380,13 @@ func (u *preFlightPortworx) processNodesChecks(recorder record.EventRecorder, st
 		}
 	}
 
+	if isFatal { // Add events for fatal checks
```
dahuang-purestorage, line 383: curious why this block of code cannot be moved in the `if` block for line 366.

drivers/storage/portworx/preflight.go
```diff
@@ -358,9 +360,18 @@ func (u *preFlightPortworx) processNodesChecks(recorder record.EventRecorder, st
...
 				continue
 			}
+
+			passed = false // pre-flight status check failed, keep going for logging
+
+			if check.Result == "fatal" {
+				// collect fatal msgs for output at end.
+				fmsgs = append(fmsgs, check.Reason)
+				isFatal = true
+				continue
```
alicelyu-px, line 370: Why do we `continue` instead of letting line 373 and 374 to log the failure out?

drivers/storage/portworx/preflight.go
jrivera-px: The loop will process checks looking for any failures.  However with this PR there is a difference between a failure and "fatal" failure.    So I wanted to make sure the fatal events are printed after all the other failed events.     So I collected the fatal msgs so they can be sent all at once instead of being intermixed with other failed events.

drivers/storage/portworx/preflight.go
```diff
@@ -369,6 +380,13 @@ func (u *preFlightPortworx) processNodesChecks(recorder record.EventRecorder, st
 		}
 	}
 
+	if isFatal { // Add events for fatal checks
+		for _, fmsg := range fmsgs {
+			k8sutil.WarningEvent(recorder, u.cluster, util.FailedPreFlight, fmsg)
+		}
```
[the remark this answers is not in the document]
jrivera-px: The loop will process checks looking for any failures. However with this PR there is a difference between a failure and "fatal" failure. So I wanted to make sure the fatal events are printed after all the other failed events. So I collected the fatal msgs so they can be sent all at once instead of being intermixed with other failed events.

drivers/storage/portworx/preflight.go
dahuang-purestorage: that makes sense! this definitely worth a comment

deploy/crds/core_v1_storagenode_crd.yaml
```diff
@@ -148,6 +148,9 @@ spec:
                     success:
                       type: boolean
                       description: If true, the check was successful
+                    result:
+                      type: string
+                      description: Result of the success or failure of the check
```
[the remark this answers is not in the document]
jrivera-px: Yes, thank you @alicelyu-px .   Definitely will add the proposed values in the description.    I just need to find out what they are ;) ...

drivers/storage/portworx/preflight.go
jrivera-px: Want to make sure that all the fatal events are logged at once.. will add a better comment.

drivers/storage/portworx/preflight.go
```diff
@@ -358,9 +360,18 @@ func (u *preFlightPortworx) processNodesChecks(recorder record.EventRecorder, st
 				k8sutil.InfoEvent(recorder, u.cluster, util.PassPreFlight, msg)
 				continue
 			}
+
+			passed = false // pre-flight status check failed, keep going for logging
+
+			if check.Result == "fatal" {
```
piyush-nimbalkar, line 366: nit: Let's use a capitalized constant. Also, can we declare this constant above?

pkg/apis/core/v1/storagenode.go
```diff
@@ -100,6 +100,8 @@ type CheckResult struct {
 	Reason string `json:"reason,omitempty"`
 	// Success indicates if the check was successful or failed
 	Success bool `json:"success,omitempty"`
+	// Result of the success or failure
```
piyush-nimbalkar, line 103: *Result of the success of failure of the check

drivers/storage/portworx/preflight.go
piyush-nimbalkar: @jrivera-px Can we have a `Fatal: ` prefix for the fatal errors and remove the `failed: ` prefix from the non-fatal errors. The reason already says "PreflightFailed".

piyush-nimbalkar (approved):

dahuang-purestorage (approved):

PR #1357, opened 2023-11-29
PWX-32111: Disable security when running pre-flight.

jrivera-px:
Secrets are not setup when pre-flight is run so if security is enabled disable it for pre-flight pod so pre-flight checks can be run.   
Closes #
PWX-32111
Tested by enabling security on my k8s cluster and pre-flight ran fine.   Will retest on Rohits setup before merging in.

### DIFF
--- a/drivers/storage/portworx/portworx.go
+++ b/drivers/storage/portworx/portworx.go
@@ -127,7 +127,9 @@ func (p *portworx) Validate(cluster *corev1.StorageCluster) error {
 		util.UpdateStorageClusterCondition(cluster, condition)
 	}
 
-	podSpec, err := p.GetStoragePodSpec(cluster, "")
+	preFltCluster := GetPreFlightStorageCluster(cluster)
+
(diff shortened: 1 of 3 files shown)


drivers/storage/portworx/portworx.go
```diff
@@ -127,7 +127,9 @@ func (p *portworx) Validate(cluster *corev1.StorageCluster) error {
 		util.UpdateStorageClusterCondition(cluster, condition)
 	}
 
-	podSpec, err := p.GetStoragePodSpec(cluster, "")
+	preFltCluster := GetPreFlightStorageCluster(cluster)
```
piyush-nimbalkar, line 130: It is fine to do it this way for now. In future we would ideally want to send the exact same parameters to the preflight that we would send to the actual px/oci-monitor pod. That way it can run all the prechecks.

piyush-nimbalkar (approved):

PR #1360, opened 2023-11-30
PWX-35203 : Disable dmthin preflight check for all environments except for EKS

nikita-bhatia:

Tested based on UTs

### DIFF
--- a/drivers/storage/portworx/portworx.go
+++ b/drivers/storage/portworx/portworx.go
@@ -127,7 +127,9 @@ func (p *portworx) Validate(cluster *corev1.StorageCluster) error {
 		util.UpdateStorageClusterCondition(cluster, condition)
 	}
 
-	podSpec, err := p.GetStoragePodSpec(cluster, "")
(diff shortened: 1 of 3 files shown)


drivers/storage/portworx/portworx.go
```diff
@@ -149,6 +149,11 @@ func (p *portworx) Validate(cluster *corev1.StorageCluster) error {
 		}
 	}
 
+	if !pxutil.IsEKS(cluster) {
+		logrus.Warnf("pre-flight is not enabled on this environment")
+		return nil
+	}
+
```
jrivera-px, line 156: I am not so sure this is the right place to do this?  We have a function which determines if pre-flight is supposed to run:  https://github.com/libopenstorage/operator/blob/35b110c2aa33a3f0773ec6e4e75e30d3be835238/drivers/storage/portworx/portworx.go#L477.  I think its better to disable there in master.  

But in 12.10.2 branch its a bit different portworx.go:preflightShouldRun() at line ~295.

PR #1361, opened 2023-11-30
PWX-35203 : Disable dmthin preflight check for all environments excep…

nikita-bhatia:

Tested based on UTs

### DIFF
--- a/drivers/storage/portworx/portworx.go
+++ b/drivers/storage/portworx/portworx.go
@@ -308,29 +308,34 @@ func (p *portworx) preflightShouldRun(toUpdate *corev1.StorageCluster) bool {
 		return check == "true"
 	}
 
-	if !preflight.RequiresCheck() { // Preflight is only supported on AWS, VSPHERE & Pure
+	/*if !preflight.RequiresCheck() { // Preflight is only supported on AWS, VSPHERE & Pure
 		return false
 	}
+	*/
 
+	// PWX-35203 : Preflight should run only on AWS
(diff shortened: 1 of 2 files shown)


drivers/storage/portworx/portworx.go
```diff
@@ -310,29 +310,33 @@ func (p *portworx) preflightShouldRun(toUpdate *corev1.StorageCluster) bool {
 		return check == "true"
 	}
 
-	if !preflight.RequiresCheck() { // Preflight is only supported on AWS, VSPHERE & Pure
+	/*if !preflight.RequiresCheck() { // Preflight is only supported on AWS, VSPHERE & Pure
```
piyush-nimbalkar, line 313: Let's remove the commented code. We can always revert this commit to get the deleted changes.

piyush-nimbalkar (approved):
I understand the code has diverged a bit, but ideally we should always send PRs to master first and then cherrypick to release branches.

nikita-bhatia:
> > I understand the code has diverged a bit, but ideally we should always send PRs to master first and then cherrypick to release branches.
> 
> do we need to add this to master? CC : @jrivera-px

https://github.com/libopenstorage/operator/pull/1367

jrivera-px:
@piyush-nimbalkar @nikita-bhatia I don't think we should remove this from master.   Adding this change to master will just have to be reverted the resize issue can be tested fully when they fix that.   I am not sure we want to remove that.    Cause then they we will just have to add it back in?

PR #1367, opened 2023-12-04
[master] PWX-35203 : Disable dmthin preflight check for all environments except eks

nikita-bhatia:
Changes to master for PR https://github.com/libopenstorage/operator/pull/1361

Tested based on UTs


piyush-nimbalkar:
We have decided to revert this only from the release branch. We will leave it as is in master.

PR #1368, opened 2023-12-05
PWX-35255: add AR finalizers to clusterrole for autopilot on OCP

ezhang-px:
Currently autopilot fails to generate AROs on OCP due to the lack of update permission of AR finalizer in its own clusterrole definition. Solution is to add such when the cluster is found to be on OCP. 

Closes #

After manually editing autopilot clusterrole to add "update" on "autopilotrules/finalizers", the error message of "_failed to insert watcher event due to: autopilotruleobjects.autopilot.libopenstorage.org \"pvc-e836c495-0eab-4a63-b9e8-1aa4481bee37-volume-resize\" is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on: , <nil>" file="statehandlers.go:696_" is gone and ARO shows up correctly

`[root@users-ezhang-ezhang-k8s-1-4-3 PWX-35255]# kubectl get aro -A`
`NAMESPACE   NAME                                                     AGE`
`test        pvc-435108c1-effa-4213-8d75-be4735fb8aaa-volume-resize   28s`
`test        pvc-a5cd77a2-545f-44dc-84dc-e9a229398c12-volume-resize   28s`

### DIFF
--- a/drivers/storage/portworx/component/autopilot.go
+++ b/drivers/storage/portworx/component/autopilot.go
@@ -122,7 +122,7 @@ func (c *autopilot) Reconcile(cluster *corev1.StorageCluster) error {
 	if err := c.createServiceAccount(cluster.Namespace, ownerRef); err != nil {
 		return err
 	}
-	if err := c.createClusterRole(cluster); err != nil {
+	if err := c.createClusterRole(); err != nil {
 		return err
 	}
 	if err := c.createClusterRoleBinding(cluster.Namespace); err != nil {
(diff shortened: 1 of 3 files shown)


drivers/storage/portworx/component/autopilot.go
```diff
@@ -316,6 +316,14 @@ func (c *autopilot) createClusterRole() error {
 		},
 		)
 	}
+	if pxutil.IsOpenshift(cluster) {
```
piyush-nimbalkar, line 319: Let's add this irrespective of whether this is OCP or not. Some other env in future might need this.

piyush-nimbalkar:
We will have to fix the UT too.

drivers/storage/portworx/component/autopilot.go
ezhang-px: Sure, thanks. I added this resource to the group where other autopilot resources stay in and fixed the UT yamls

piyush-nimbalkar (approved):

PR #1375, opened 2023-12-11
PWX-34825: Do not continue with PX installation if PX versions configmap or URL are not present/reachable

svijaykumar-px:
**What this PR does / why we need it**: Currently when versions configmap is not present or the URL to fetch it is not reachable, we fall back on default images. This behavior is changed and instead of using default images, the storage cluster installation will fail with an appropriate error.

**Which issue(s) this PR fixes**:
Closes # PWX-34825

**Testing Notes**: Tested changes on anthos cluster of 3 nodes
1) Deploy operator image with the changes
2) Ensure there is no versions manifests
3) Add the following env variable to spec of storage cluster yaml file
```
env:
    - name: PX_RELEASE_MANIFEST_URL
      value: "https://install.portworx.com/invalid/version"
```
4) Apply the modified yaml file
Expected results: The storage cluster does not get installed and enters into degraded state. On describing the stc, the following warning is seen
`  Warning  FailedComponent  6m57s (x13 over 7m10s)  storagecluster-controller  StorageCluster reconciliation paused as versions from px-versions ConfigMap cannot be fetched and URL is unreachable due to: release not found`

Positive Test:
1) Deploy operator with changes 
2) Ensure versions manifest does not exist and the URL "https://edge-install.portworx.com/3.1.0/version" is reachable.
3) Add the env variable to storage cluster yaml file :
```
env:
    - name: PX_RELEASE_MANIFEST_URL
      value: "https://edge-install.portworx.com/3.1.0/version"
```
This step can also be skipped. In that case the default URL will be used.
4) Apply the yaml file. Storage cluster installation will proceed without any issues as the versions will be fetched from the URL
During Uninstall, node wiper pods will continue to use default image if versions manifest cannot be fetched. This is done so that storage cluster should never be stuck in a stage where it cannot be uninstalled. Steps to test this, continue from previous test steps
5) Edit the stc yaml file to the incorrect manifest URL again, add uninstall strategy and save changes
6) Delete the storage cluster and watch the logs of the operator. There will be a log stating:

`level=warning msg="Failed to get release versions as versions from px-versions ConfigMap cannot be fetched and URL is unreachable due to: release not found. Using default NodeWiper image" file="uninstall.go:189""`


drivers/storage/portworx/manifest/manifest.go
```diff
@@ -207,16 +207,16 @@ func (m *manifest) GetVersions(
...
 	}
 
 	// Bug: if it fails due to temporarily network issue, we should retry.
 	rel, err := provider.Get()
 	if err != nil {
-		msg := fmt.Sprintf("Using default version due to: %v", err)
+		msg := fmt.Sprintf("Portworx install stopped as versions from Configmap cannot be fetched and URL unreachable due to: %v", err)
 		logrus.Error(msg)
 		m.recorder.Event(cluster, v1.EventTypeWarning, util.FailedComponentReason, msg)
-		return defaultRelease(m.k8sVersion)
```
nikita-bhatia, line 219: Let' remove func defaultRelease if we are not using it anymore anywhere in the code.

nikita-bhatia (approved):
LGTM, please update positive tests also that has been performed.

drivers/storage/portworx/uninstall.go
```diff
@@ -183,7 +183,10 @@ func (u *uninstallPortworx) RunNodeWiper(
 
 	wiperImage := k8sutil.GetValueFromEnv(envKeyNodeWiperImage, u.cluster.Spec.Env)
 	if len(wiperImage) == 0 {
-		release := manifest.Instance().GetVersions(u.cluster, true)
+		release, err := manifest.Instance().GetVersions(u.cluster, true)
+		if err != nil {
+			logrus.Warnf("Failed to get release versions: %v", err)
+		}
 		wiperImage = release.Components.NodeWiper
```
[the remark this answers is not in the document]
svijaykumar-px: But isn't the main essence of the ticket to remove the usage of default images? That leads to another question. If only few images are not present should we fail or fill in those with default images?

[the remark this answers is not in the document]
svijaykumar-px: Partially filled versions manifest will be addressed in a seperate ticket as discussed

drivers/storage/portworx/manifest/manifest.go
```diff
@@ -207,16 +207,17 @@ func (m *manifest) GetVersions(
...
 	rel, err := provider.Get()
 	if err != nil {
-		msg := fmt.Sprintf("Using default version due to: %v", err)
-		logrus.Error(msg)
-		m.recorder.Event(cluster, v1.EventTypeWarning, util.FailedComponentReason, msg)
-		return defaultRelease(m.k8sVersion)
+		msg := fmt.Sprintf("versions from px-versions ConfigMap cannot be fetched and URL is unreachable due to: %v", err)
+		errmsg := fmt.Sprintf("StorageCluster reconciliation paused as %v", msg)
+		logrus.Error(errmsg)
+		m.recorder.Event(cluster, v1.EventTypeWarning, util.FailedComponentReason, errmsg)
```
piyush-nimbalkar, line 219: Now that we are returning an error and raising a warning from the caller, should we remove this?
Can you check if it raises the event twice?

svijaykumar-px: Yes this event comes up twice, one from the "SetDefaultsOnStorageCluster" where it reports defaults cannot be set on the stc due to error . another as a failed component warning. I thought we should let the failed component warning remain as it is a better representation of the issue. Should I remove it ?

piyush-nimbalkar: If we fail here, we won't even proceed to install the portworx pods. The failed component warning may not be relevant in that case. Let's just return the error here and let the caller raise the event.

svijaykumar-px: Okay I will remove both the logrus and recorder.event

PR #1377, opened 2023-12-11
PWX-35330: fixed gitops spec definition

ezhang-px:
[PR830](https://github.com/libopenstorage/operator/pull/830) was supposed to add gitops support to autopilot through stc configuration. But it fails to do so due to the fact that wrong type of defaultReviewers parameter is processed (expected array but got string). This bug will cause operator to go down and couldn't generate anticipated autopilot configmap with gitops config definition

Closes #

Before the change, operator breaks down if in stc spec we have definition of gitops like:
`    autopilot:
      args:
        log-level: debug
        min_poll_interval: "10"
      enabled: true
      gitops:
        params:
          baseUrl: https://bitbucket-staging.dev.purestorage.com
          branch: master
          defaultReviewers:
          - svc-cnbu-devops
          - user1
          folder: test
          projectKey: PXAUT
          repo: autopilot-eugene-test
          user: svc-cnbu-devops
        type: bitbucket-scm
      providers:
      - name: default
        params:
          url: http://px-prometheus:9090
        type: prometheus`

with error 

> W1211 20:41:26.727165       1 reflector.go:424] sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:262: failed to list *v1.StorageCluster: json: cannot unmarshal array into Go struct field GitOpsSpec.items.spec.autopilot.gitops.params of type string
E1211 20:41:26.727232       1 reflector.go:140] sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:262: Failed to watch *v1.StorageCluster: failed to list *v1.StorageCluster: json: cannot unmarshal array into Go struct field GitOpsSpec.items.spec.autopilot.gitops.params of type string

Therefore no gitops definition was found in autopilot-config cm.

With this change, operator now runs without error/restart and autopilot-config cm now contains the correct content for gitops
`[root@users-ezhang-ezhang-k8s-1-4-3 autopilot]# kp get cm  autopilot-config -oyaml
apiVersion: v1
data:
  config.yaml: "providers:\n- name: default\n  type: prometheus\n  params: url=http://px-prometheus:9090\ngitops:\n
    \ name: \n  type: bitbucket-scm\n  params:\n    baseUrl: https://bitbucket-staging.dev.purestorage.com\n
    \   branch: master\n    defaultReviewers:\n      - \"svc-cnbu-devops\"\n      -
    \"user1\"\n    folder: test\n    projectKey: PXAUT\n    repo: autopilot-eugene-test\n
    \   user: svc-cnbu-devops\nmin_poll_interval: 10"
kind: ConfigMap`

### DIFF
--- a/drivers/storage/portworx/component/autopilot.go
+++ b/drivers/storage/portworx/component/autopilot.go
@@ -205,11 +205,11 @@ func (c *autopilot) createConfigMap(
 			if key == AutopilotDefaultReviewersKey {
 				params += fmt.Sprintf(`
     %s:`, AutopilotDefaultReviewersKey)
-				values := strings.Split(val, ",")
-				for _, v := range values {
-					v = strings.TrimSpace(v)
+				var reviewers []interface{} = val.([]interface{})
+				for _, r := range reviewers {
+					r = strings.TrimSpace(r.(string))
 					params += fmt.Sprintf(`
-      - "%s"`, v)
+      - "%s"`, r)
 				}
 				continue
 			}
--- a/drivers/storage/portworx/component/autopilot.go
+++ b/drivers/storage/portworx/component/autopilot.go
@@ -206,10 +206,12 @@ func (c *autopilot) createConfigMap(
 				params += fmt.Sprintf(`
     %s:`, AutopilotDefaultReviewersKey)
 				var reviewers []interface{} = val.([]interface{})
-				for _, r := range reviewers {
-					r = strings.TrimSpace(r.(string))
-					params += fmt.Sprintf(`
+				for _, reviewer := range reviewers {
+					if r, ok := reviewer.(string); ok {
+						r = strings.TrimSpace(r)
+						params += fmt.Sprintf(`
       - "%s"`, r)
+					}
 				}
 				continue
 			}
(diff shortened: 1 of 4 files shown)


drivers/storage/portworx/component/autopilot.go
```diff
@@ -205,11 +205,11 @@ func (c *autopilot) createConfigMap(
 			if key == AutopilotDefaultReviewersKey {
 				params += fmt.Sprintf(`
     %s:`, AutopilotDefaultReviewersKey)
-				values := strings.Split(val, ",")
-				for _, v := range values {
-					v = strings.TrimSpace(v)
+				var reviewers []interface{} = val.([]interface{})
+				for _, r := range reviewers {
+					r = strings.TrimSpace(r.(string))
```
zoxpx, line 210: Could we please add some type-checking here?  I.e.

```go
s := r.(string)       // this throws PANIC if r was not string-type
s, ok := r.(string)   // always works, will set `ok = false` if r was not string-type
```

So, the current code is susceptible to "panic's"

zoxpx (approved):
Thanks Eugene

PR #1384, opened 2023-12-21
PTX-21563 Add check for plugin resource

nikolaypopov:
Add check for plugin resource. This is needed for Dynamic Plugin validation decision.


jrivera-px (approved):
LGTM...

PR #1387, opened 2024-01-10
PTX-21676 Set correct config.json path in case of PKS

nikolaypopov:
Set correct config.json path in case of PKS


mborodin-px (approved):
LGTM

PR #1388, opened 2024-01-14
PWX-34256: node-wiper fix for custom-dir installs

zoxpx:
Adding custom user-volumes to the node-wiper spec

The portworx-install can be customized using user-volumes
* i.e. ~/opt/pwx~ -> `/usr/local/portworx` by adding `spec.Volume` -override

  ```yaml
  kind: StorageCluster
  spec:
    volumes:
    - name: custom-opt
      mountPath: /opt/pwx
      hostPath:
        path: /usr/local/portworx
  ```

However, in such customized installs, for the cleanup to work correctly, we have to provide the same custom user-volumes to the node-wiper image
* note, the new node-wiper is smart enough to detect the install-dir overrides, and can perform the cleanup/wipe correctly --  but it'll need the correct user-volumes setup

Closes # PWX-34256

Note, the customized px-install (~/opt/pwx~ -> `/usr/local/portworx`) is being node-wiped correctly after this fix.


zoxpx:
Thanks Piyush -- merging this PR, will cherry-pick to `px-rel-23.10.2` via separate PR.

PR #1390, opened 2024-01-17
PWX-35711: Update dynamic-plugin ImagePullPolicy to Always

nikita-bhatia:
**What this PR does / why we need it**: The ImagePullPolicy of dynamic-plugin pod is set ti 'IfNotPresent, an upgrade in operator will not pull latest image with same tag for dynamic plugin (same tag might have security fixes) 

Closes # https://portworx.atlassian.net/browse/PWX-35771


zoxpx (approved):
Looks harmless enough -- thanks Nikita

piyush-nimbalkar (approved):

PR #1391, opened 2024-01-17
PWX-35704: The pks distribution string was incorrect.  Change it from  'gke' to 'pks'

jrivera-px:
The pks distribution string was incorrect and triggered incorrect mounts to be added to the telemetry spec when running on gke install.  Change the the string be 'pks'.

Closes #
PWX-35704


zoxpx (approved):
Looks good -- thanks Jose

piyush-nimbalkar (approved):

PR #1398, opened 2024-01-24
PTX-21947 - Change master version to 23.10.3 for changes made in integration test

svijaykumar-px:
**What this PR does / why we need it**: In this ticket - https://portworx.atlassian.net/browse/PTX-21234 the test for checking if csi-node-driver registrar container in px-api pod was only in master version of operator, but since the change has been tied to a release 23.10.3, the integration test has been modified to check if the container is in px-api pod for all versions greater than release version

Closes # PTX-21947


bhavana-ar (approved):

nikita-bhatia (approved):
LGTM

PR #1399, opened 2024-01-24
Add additional properties to the crd schema definitions

GabrielCFigueira:
**What this PR does / why we need it**: When using [kubeconform](https://github.com/yannh/kubeconform) to validate CRDs, the StorageCluster CRD was being reported as invalid due to the CRD schema definition not possessing the keywords 'kind', 'metadata' and 'apiVersion'. This PR would make the CRD schema definition more complete and allow a more strict validation of the StorageCluster resource.

Closes #

### DIFF
--- a/deploy/crds/core_v1_storagecluster_crd.yaml
+++ b/deploy/crds/core_v1_storagecluster_crd.yaml
@@ -39,6 +39,18 @@ spec:
       openAPIV3Schema:
         type: object
         properties:
+          apiVersion:
+            type: string
+            description: 'APIVersion defines the versioned schema of this representation
+              of an object. Servers should convert recognized schemas to the latest
+              internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+          kind:
+            type: string
+            description: 'Kind is a string value representing the REST resource this
+              object represents. Servers may infer this from the endpoint the client
+              submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+          metadata:
+            type: object
           spec:
             type: object
             description: The desired behavior of the storage cluster.
--- a/deploy/crds/core_v1_storagenode_crd.yaml
+++ b/deploy/crds/core_v1_storagenode_crd.yaml
@@ -39,6 +39,18 @@ spec:
       openAPIV3Schema:
         type: object
         properties:
+          apiVersion:
+            type: string
+            description: 'APIVersion defines the versioned schema of this representation
+              of an object. Servers should convert recognized schemas to the latest
+              internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+          kind:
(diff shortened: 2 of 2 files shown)


GabrielCFigueira:
Hey @svijaykumar-px @nikita-bhatia @nikolaypopov, may I get a review?

GabrielCFigueira:
@piyush-nimbalkar @svijaykumar-px bump

deploy/crds/core_v1_storagecluster_crd.yaml
```diff
@@ -39,6 +39,18 @@ spec:
...
+            type: string
+            description: 'APIVersion defines the versioned schema of this representation
+              of an object. Servers should convert recognized schemas to the latest
+              internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+          kind:
+            type: string
+            description: 'Kind is a string value representing the REST resource this
+              object represents. Servers may infer this from the endpoint the client
+              submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+          metadata:
```
piyush-nimbalkar, line 52: In case of a stricter validation, won't it complain for the fields under metadata (name, namespace, resourceVersion, etc)?

GabrielCFigueira: So I checked the CRDs for some other popular projects and they seem to not enforce this as well. For example:
- https://github.com/cert-manager/cert-manager/blob/master/deploy/crds/crd-challenges.yaml#L66
- https://github.com/metallb/metallb/blob/main/charts/metallb/charts/crds/templates/crds.yaml#L41

The reason it does not complain has to do with a design choice in the script used to convert a CRD into a json schema definition: https://github.com/yannh/kubeconform/blob/master/scripts/openapi2jsonschema.py#L23 . There is a boolean field called "additionalProperties" which determines whether you can have additional fields (in this case, in metadata) beyond those that are defined. However, if there are no "properties" to begin with, then the field is not added (as False). I guess the strict validation is only done when there is a "properties" and/or "required" field.

If I were to add those fields, it would look something like this:
https://github.com/yannh/kubernetes-json-schema/blob/master/v1.27.6-standalone-strict/pod-v1.json#L24-L270 (but in yaml instead of json)
Based on the docs: https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/object-meta/

piyush-nimbalkar (approved):

zoxpx:
Ping @piyush-nimbalkar -- if you like this PR, I think you'll need to merge it
* (i.e. I don't think Gabriel has user-rights to be able to merge PRs on this github project)

piyush-nimbalkar:
> @piyush-nimbalkar -- if you like this PR, I think you'll need to merge it
> 
> * (i.e. I don't think Gabriel has user-rights to be able to merge PRs on this github project)

@zoxpx I tried to that earlier this week, but the branch is out-of-date and I can't merged it.
@GabrielCFigueira can you rebase the branch?

GabrielCFigueira:
@piyush-nimbalkar done! But it seems to be failing CI

PR #1409, opened 2024-01-30
PWX-35418 : Wrong value of min available in PDB for px-storage during surge upgrade

svijaykumar-px:
**What this PR does / why we need it**:  PDB value for normal clusters is being calculated by using total storage nodes -1. Currently, the storage node count is obtained by counting portworx nodes that have been scheduled to a kubernetes node. This causes an issue during k8s upgrades that happen as surge upgrades. The total storage nodes count decreases when a portworx node does not have a corresponding k8s node when that node is being upgraded. This causes incorrect value of PDB. To change this, the total storage nodes count has been set to number of portworx nodes only. 
As this case will not allow individual clusters to upgrade separately in DR setup, the logic is only used for non DR cluster.

Closes # PWX-35418


drivers/storage/portworx/components_test.go
```diff
@@ -11647,8 +11647,6 @@ func TestPodDisruptionBudgetEnabled(t *testing.T) {
 		// Node that does not return a scheduler node name, does not get counted towards quorum calculation as we cannot
 		// determine if it is part of the current k8s cluster or not.
 		{NonQuorumMember: false, NodeLabels: map[string]string{pxutil.NodeLabelPortworxVersion: "3.1.0"}},
-		// Node that is not part of the cluster, does not get counted towards quorum calculation.
```
piyush-nimbalkar, line 11650: Didn't we have to mock the ClusterDomainServer in this test? It should always be calling the enumerate API, right?

drivers/storage/portworx/components_test.go
```diff
@@ -11647,8 +11647,6 @@ func TestPodDisruptionBudgetEnabled(t *testing.T) {
...
-		{NonQuorumMember: false, SchedulerNodeName: "node6", NodeLabels: map[string]string{pxutil.NodeLabelPortworxVersion: "3.1.0"}},
```
piyush-nimbalkar, line 11651: We should actually keep the case and just update the comment above "Node that is not part of the k8s cluster, still gets counted towards quorum calculation".
We will have to change the required value below to 3 instead of 2.

This will test your changes in this PR.

Makefile
```diff
@@ -301,7 +301,7 @@ get-release-manifest: clean-release-manifest
 
 mockgen: $(GOPATH)/bin/mockgen
 	$(GOPATH)/bin/mockgen -destination=pkg/mock/portworxsdk.mock.go -package=mock github.com/libopenstorage/operator/api/px PortworxServiceServer
-	$(GOPATH)/bin/mockgen -destination=pkg/mock/openstoragesdk.mock.go -package=mock github.com/libopenstorage/openstorage/api OpenStorageRoleServer,OpenStorageNodeServer,OpenStorageClusterServer,OpenStorageNodeClient,OpenStorageVolumeServer
+	$(GOPATH)/bin/mockgen -destination=pkg/mock/openstoragesdk.mock.go -package=mock github.com/libopenstorage/openstorage/api OpenStorageRoleServer,OpenStorageNodeServer,OpenStorageClusterServer,OpenStorageNodeClient,OpenStorageVolumeServer,OpenStorageClusterDomainsServer,OpenStorageClusterDomainsClient
```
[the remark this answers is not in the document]
svijaykumar-px: My bad, i thought we needed is as nodeclient was used

drivers/storage/portworx/components_test.go
svijaykumar-px: I have added it now

PR #1414, opened 2024-02-01
PTX-21748 Adding check for clusterVersion object status when deployed on OCP

nikolaypopov:
Adding check for clusterVersion object status when deployed on OCP


pkg/util/test/util.go
```diff
@@ -812,6 +816,37 @@ func ValidateStorageCluster(
...
+	// NOTE: For now only checking this for Openshift, will add for other providers in future when its needed
+	if isOpenshift(cluster) {
+		t := func() (interface{}, bool, error) {
+			logrus.Debug("This is Openshift cluster, checking ClusterVersion object status..")
+			clusterVersion, err := openshiftops.Instance().GetClusterVersion("version")
+			if err != nil {
+				return nil, false, fmt.Errorf("failed to get Openshift ClusterVersion object, Err: %v", err)
+			}
+			logrus.Debugf("ClusterVersion object list of all conditions:\n%s", clusterVersion.Status.Conditions)
+			lastCondition := clusterVersion.Status.Conditions[len(clusterVersion.Status.Conditions)-1]
```
piyush-nimbalkar, line 838: I am not sure if this is the best way to get the desired condition. We should filter out the condition with the type we want and then look at its message.

nikolaypopov: The string I am looking for is what is considered a healthy status message, for the prometheus issue, yes there is a specific status message about monitoring, but we don't know what else we could potentially cause in the OCP in future, hence why I am only checking for the healthy status message

piyush-nimbalkar (approved):
I don't think this is full proof. But let's work with what we have for now and keep an eye out for any modifications.

nikolaypopov:
> I don't think this is full proof. But let's work with what we have for now and keep an eye out for any modifications.

I agree, this is something to start tracking and look for a better way to get this health status check

PR #1410, opened 2024-01-30
PWX-35477 : Support Openshift Prometheus for portworx monitoring on OCP 4.14

nikita-bhatia:
Issue : OCP upgrade from 4.13.x to 4.14.5 is blocked due to PX prometheus. From OCP 4.14, PX Prometheus stopped working.

Resolution : From OCP 4.14, Portworx deployment on OCP should should use Openshift prometheus stack for monitoring Portworx metrics instead of deploying PX-Prometheus. We need to expose Portworx metrics to Openshift.
To do this, user-workload-monitoring has to be enabled on OCP cluster to use openshift's monitoring and alerting for portworx.

**What this PR does / why we need it**:
1. Default autopilot URL changed from  http://px-prometheus:9090/ to https://$THANOS_QUERIER_HOST 
2. Creating autopilot-auth-token-secret  to store ocp token and ca.cert
3. Mount ocp token and ca.cert as file in autopilot pod with file names ca-certificates.crt and token. 

**Which issue(s) this PR fixes** (optional)
Closes # https://portworx.atlassian.net/browse/PWX-35477

**Special notes for your reviewer**:
TBD

**Testing done**
TBD

PR #1415, opened 2024-02-01
OCP 4.14 : Fix Mount volume to autopilot pod only if secret is created

nikita-bhatia:
**What this PR does / why we need it**: We should mount token and sertificates to autopilot pod only if autopilot-prometheus-auth secret is created, otherwise autopilot pod will start failing to come up and mounting error.

Extension of PR https://github.com/libopenstorage/operator/pull/1410

### DIFF
--- a/drivers/storage/portworx/component/autopilot.go
+++ b/drivers/storage/portworx/component/autopilot.go
@@ -757,7 +757,7 @@ func (c *autopilot) isAutopilotSecretCreated(namespace string) bool {
 		secret,
 	)
 
-	if err == nil && secret.Name == AutopilotSecretName {
+	if err == nil {
 		return true
 	}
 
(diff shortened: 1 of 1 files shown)


piyush-nimbalkar (approved):

drivers/storage/portworx/component/autopilot.go
```diff
@@ -745,6 +745,30 @@ func (c *autopilot) getDesiredVolumesAndMounts(
...
+	err := c.k8sClient.Get(
+		context.TODO(),
+		types.NamespacedName{
+			Name:      AutopilotSecretName,
+			Namespace: namespace,
+		},
+		secret,
+	)
+
+	if err == nil && secret.Name == AutopilotSecretName {
```
piyush-nimbalkar, line 760: nit: We don't need the name check here as we got the object using the name.

PR #1419, opened 2024-02-02
PWX-35945 : Create autopilot deployment if secret is not created

nikita-bhatia:

Autopilot pod was not getting created because of two issues.
1. If secret is not created (because user workload monitoring was not enabled), the autopilot deployment was not created because failure of secret exits the code and doesn’t proceed with autopilot deployment creation. Removed returning error and added loggoing in case of secret creation failure,

2. Now once user-workload monitoring is enabled, and creation of secret is completed, so operator’s reconcilation should be able to create autopilot deployment, but it is failing with ‘duplicate volume error’ because operator is trying to mount same volume more than once. This was happening when fresh install is done more than once with same operator. 
Install operator → install portworx (goes fine) → unsinstall portworx(goes fine) → install portworx (issue with volume)
The list autopilotDeploymentVolumes already had elements of openshiftDeploymentVolume from previous install, and new install just adds it again, because of which creation of deployment fails with error "duplicate volumes".
As a fix, removed adding openshiftDeploymentVolume into autopilotDeploymentVolumes.

Closes # https://portworx.atlassian.net/browse/PWX-35945

1. Tested fresh install with new operator + new portworx
2. Tested multiple times fresh install of portworx without uninstalling operator


piyush-nimbalkar (approved):

drivers/storage/portworx/component/autopilot.go
```diff
@@ -746,6 +750,10 @@ func (c *autopilot) getDesiredVolumesAndMounts(
 }
 
 func (c *autopilot) isAutopilotSecretCreated(namespace string) bool {
+	if c.isSecretCreated != nil {
```
piyush-nimbalkar, line 753: I don't think we need this flag at all because we anyway read the secret from the cache when we execute the `k8sClient.Get`.

PR #1422, opened 2024-02-05
Update plugin deployment PullPolicy to Always regardless of STC's PullPolicy

nikita-bhatia:
**What this PR does / why we need it**: Update imagePullPolicy for px-plugin pods to always, regardless of STC's ImagePullPolicy, to fetch latest security/other fixes on existing tag.
The imagePullPolicy for the pod is now being set to Always, after StorageClusterSetting is applied, so that STC's PullPolicy doesn't override.

Verified that ImagePullPolicy of plugin pod is set to 'Always' if
- ImagePullPolicy in Storagecluster pod is set to 'IfNotPresent'
- ImagePullPolicy in Storagecluster pod is set to 'Always'


Glitchfix (approved):
LGTM

PR #1429, opened 2024-02-07
[Cherrypick] PWX-36015 Disable PX-prometheus for OCP 4.14+ on new installs

piyush-nimbalkar:
- Starting OCP 4.14 we recommend using OCP user workload prometheus to be used for Portworx and Autopilot. That is why we will disable prometheus from STC spec if the user enables it on this environment. This will be done only on first/fresh installs. We will not change it again if the user overwrites that back to "enabled".
- We also raise a warning when disabling it so the user knows why we did that.
- Unrelated change: Checking openshift version only if the autopilot providers are not set. That way we don't end up querying the K8s API server repeatedly.

Closes # https://portworx.atlassian.net/browse/PWX-36015
 
UTs added

### DIFF
--- a/drivers/storage/portworx/portworx.go
+++ b/drivers/storage/portworx/portworx.go
@@ -344,6 +344,8 @@ func (p *portworx) SetDefaultsOnStorageCluster(toUpdate *corev1.StorageCluster) 
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


nikita-bhatia (approved):
LGTM

PR #1432, opened 2024-02-08
[PWX-36029] Update certificate's mountpath in autopilot pod

nikita-bhatia:
Our container-mount for Destination:/etc/ssl/certs/ca-certificates.crt was overriding (removing) the system-CA bundle /etc/ssl/certs/ca-bundle.crt (/etc/pki/tls/certs/ca-bundle.crt) inside the container.

since the https://thanos-querier-openshift-monitoring… SSL cert was signed by the issuer: C=US; O=Let's Encrypt; CN=R3 – we needed the original system-CA bundle to verify this SSL certificate chain

We need to change the path where  operator is mounting the certificate currently, as that ovverides existing certificate bundle.

With this change, operator mounts custom-certs at /etc/ssl/px-custom/1  (can also do /etc/ssl/custom/2 , 3, 4 -- if needed)

Closes # https://portworx.atlassian.net/browse/PWX-36029

### DIFF
--- a/drivers/storage/portworx/component/autopilot.go
+++ b/drivers/storage/portworx/component/autopilot.go
@@ -106,7 +106,7 @@ var (
 		},
 		{
 			Name:      "ca-cert-volume",
-			MountPath: "/etc/ssl/certs",
+			MountPath: "/etc/ssl/px-custom/1",
 			ReadOnly:  true,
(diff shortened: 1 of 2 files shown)


zoxpx (approved):
Thanks Nikita

PR #1449, opened 2024-02-20
PWX-35633 : Don't allow overriding storage PDB value to more than number of storagenodes

svijaykumar-px:
**What this PR does / why we need it**: As a part of PWX-34411, PDB value for px-storage can be overridden using an annotation, but when the value is greater than or equal to total storage nodes, it prevents upgrades from happening as allowed disruptions becomes 0. This PR ensures than if the annotation value is greater than or equal to total storage nodes, then the default calculation for pdb is used.

Closes # PWX-35633

### DIFF
--- a/drivers/storage/portworx/component/disruption_budget.go
+++ b/drivers/storage/portworx/component/disruption_budget.go
@@ -107,17 +107,21 @@ func (c *disruptionBudget) createPortworxPodDisruptionBudget(
 	}
 
 	var minAvailable int
-	if err != nil || userProvidedMinValue < 0 {
-		c.sdkConn, err = pxutil.GetPortworxConn(c.sdkConn, c.k8sClient, cluster.Namespace)
-		if err != nil {
-			return err
-		}
+	c.sdkConn, err = pxutil.GetPortworxConn(c.sdkConn, c.k8sClient, cluster.Namespace)
+	if err != nil {
+		return err
+	}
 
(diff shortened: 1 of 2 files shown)


drivers/storage/portworx/component/disruption_budget.go
```diff
@@ -107,17 +107,21 @@ func (c *disruptionBudget) createPortworxPodDisruptionBudget(
...
+	storageNodesCount, err := pxutil.CountStorageNodes(cluster, c.sdkConn, c.k8sClient)
+	if err != nil {
+		c.closeSdkConn()
+		return err
+	}
+
+	// Set minAvailable value to storagenodes-1 if no value is provided,
+	// or if the user provided value is lesser than 0
+	// or greater than or equal to the number of storage nodes.
+	if userProvidedMinValue < 0 || userProvidedMinValue >= storageNodesCount {
```
nikita-bhatia, line 124: We should log if we are not honouring user provided value with reason

nikita-bhatia (approved):
LGTM

piyush-nimbalkar:
The ticket here is to avoid quorum loss. The minAvailable number should not be lower than `n/2 + 1`.

svijaykumar-px:
> The ticket here is to avoid quorum loss. The minAvailable number should not be lower than `n/2 + 1`.

The ticket mentions that the minAvailable value should not be greater than quorum number of nodes. Should I include the check to ensure it does not go below n/2+1 as well?

PR #1463, opened 2024-02-28
PWX-34410 : Consider cluster domain when calculating the PDB numbers

nikita-bhatia:
**What this PR does / why we need it**:  In case of a DR cluster, calculation the PDB number is done per k8s cluster instead of looking at the entire portworx cluster. The portworx cluster size is always going to be more than the k8s cluster. So to allow the k8s upgrades to go through we need to set the PDB’s minAvailable value for that cluster domain to px nodes count in cluster domain - 1.

Changes done  : 
1. The storagenodecount in case of DR cluster is calculated by checking if storagenode is part of current cluster domain by inspecting current node. 
2. Remove checking if node is a k8s node, and count node for storage even if it is not k8s node. (While upgrade, there are few insstances where k8s node doesn't return hostname, and thus operator thinks it is non-k8s node, where its was actually k8s node.)
4. The node will not be part of cluster domain if it in  decomission state. ( previously, we were checking if k8s is running on node or not, we were never ignoring node based on Decomission state previously. Now the second condition not being there,  have aded explicit check for ignoring decomissioned node)
5.  Remove checking k8sNodesStoragePodCouldRun, as we are not checking if node is a k8s node anymore.

Closes # https://portworx.atlassian.net/browse/PWX-34410


drivers/storage/portworx/util/util.go
```diff
@@ -1472,7 +1480,9 @@ func CountStorageNodes(
 				storageNodesCount++
 			} else {
 				if _, ok := k8sNodesStoragePodCouldRun[node.SchedulerNodeName]; ok {
```
svijaykumar-px, line 1482: The reason we changed the logic to calculate PDB is because during upgrades, at some point there exists a node where portworx is being run, but the k8s api doesnt not return the k8s node corresponding to the px node. So this px node wasn't being considered in the PDB calculation. 
I think we should remove mapping the px node to k8s node . We can directly increment storagenode count if its the cluster domain

nikita-bhatia: Done, not checking if node is a k8s node and directly moving ahead with cluster domain checks.

svijaykumar-px (approved):

drivers/storage/portworx/util/util.go
```diff
@@ -1402,30 +1402,12 @@ func CountStorageNodes(
...
-			k8sNodesStoragePodCouldRun[node.Name] = true
-		}
-	}
-
 	// Use the quorum member flag from the node enumerate response if all the nodes are upgraded to the
 	// newer version. The Enumerate response could be coming from any node and we want to make sure that
 	// we are not talking to an old node when enumerating.
 	useQuorumFlag := true
 	for _, node := range nodeEnumerateResponse.Nodes {
-		if node.Status == api.Status_STATUS_DECOMMISSION {
```
piyush-nimbalkar, line 1426: We should still check that the node is not decommissioned when figuring out whether to use the non_quorum_member flag or not. For some reason, if the decommissioned node is still part of the cluster and is the only node on an older version, then we should still go ahead and use the quorum flag for counting storage nodes.

Without this check we will end up setting useQuorumFlag = false.

drivers/storage/portworx/util/util.go
```diff
@@ -1471,19 +1454,18 @@ func CountStorageNodes(
 			if !isDRSetup {
 				storageNodesCount++
 			} else {
-				if _, ok := k8sNodesStoragePodCouldRun[node.SchedulerNodeName]; ok {
+				if node.ClusterDomain == currentClusterDomain {
 					storageNodesCount++
 				} else {
-					logrus.Debugf("node %s should not run portworx", node.SchedulerNodeName)
+					logrus.Debugf("node %s is not part of cluster domain %s ", node.SchedulerNodeName, currentClusterDomain)
```
piyush-nimbalkar, line 1460: nit: Let's print the node_name/node_id as part of the output in case the scheduler node name is not present.

drivers/storage/portworx/util/util.go
```diff
@@ -1471,19 +1454,18 @@ func CountStorageNodes(
 			if !isDRSetup {
 				storageNodesCount++
 			} else {
-				if _, ok := k8sNodesStoragePodCouldRun[node.SchedulerNodeName]; ok {
```
piyush-nimbalkar, line 1474: I just realized that we will still need to keep this check for pre-3.1.1 PX versions. Before 3.1.1, the enumerate response would not have the cluster_domain value populated.

So, we would end up counting all the nodes in both the domains and the upgrade will never proceed. Thus to avoid breaking the existing behavior for <3.1.1 DR clusters, let's calculate using the k8sNodeStoragePodCouldRun map, else we compare the cluster domains.

Also, ensure that all the nodes are on 3.1.1 before we can use the cluster domain field (just like how we do for quorum flag above).

drivers/storage/portworx/component/disruption_budget.go
```diff
@@ -151,9 +151,13 @@ func (c *disruptionBudget) createPortworxPodDisruptionBudget(
 				},
 			},
 		}
-		return k8sutil.CreateOrUpdatePodDisruptionBudget(c.k8sClient, pdb, ownerRef)
+
+		err = k8sutil.CreateOrUpdatePodDisruptionBudget(c.k8sClient, pdb, ownerRef)
+		if err == nil {
+			logrus.Infof("Value for px-storage pod disruption budget is set to %d ", pdb.Spec.MinAvailable.IntValue())
```
piyush-nimbalkar, line 157: This will print every time even if the value is not changing. Let's move this print to https://github.com/libopenstorage/operator/blob/master/pkg/util/k8s/k8s.go#L1675 and print it only when the existing and new min_available values don't match.

PR #1472, opened 2024-03-06
PTX-22712 Add method to check if fresh PX installation

nikolaypopov:
Adding a method to check if fresh PX installation, this will be eventually used for validating components that behave different based on if its fresh install or not. For now just doing a print

### DIFF
--- a/pkg/util/test/util.go
+++ b/pkg/util/test/util.go
@@ -24,7 +24,6 @@ import (
 	"github.com/golang/mock/gomock"
 	"github.com/hashicorp/go-version"
 	"github.com/libopenstorage/openstorage/api"
-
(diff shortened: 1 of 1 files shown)


jrivera-px:
This looks ok .. your looping until the cluster is status changes.   So this can be run during upgrade also correct?

jrivera-px (approved):

nikolaypopov:
> This looks ok .. your looping until the cluster is status changes. So this can be run during upgrade also correct?

Actually I might not even need this, I added this cuz I was getting nil pointer exception and I thought its because of this, but it was different issue
```
		if cluster.Status.Phase == "" {
			return nil, true, fmt.Errorf("failed to get cluster status")
		}
```

PR #1484, opened 2024-03-13
PWX-36298 - Set node tolerations before maxStorageNodesPerZone

svijaykumar-px:
**What this PR does / why we need it**: Previously, default value of maxStorageNodesPerZone was being set before the tolerations were added to the storagecluster. This caused a problem when the annotation to run on master was set. The toleration to noSchedule on master was not added to the storagecluster and thus when setting maxStorageNodesPerZone, it was considered that the master node is not schedulable and hence not counted for setting the value. 

Closes # PWX-36298

**Special notes for your reviewer**: This PR cannot be merged to master as it has diverged

**Testing steps** : 
Testing to see all nodes run portworx
1) In OCP cluster, deploy the operator with this change. 
2) In the storage cluster stc, add annotation - ` portworx.io/run-on-master: "true"`
3) Apply the spec file and wait till storage cluster is running. There will be 1 storage pod created for every node including master nodes.
4) Check the yaml file of the stc - the field `maxStorageNodesPerZone` will  have a value equal number of worker nodes + master nodes
5) Exec into a storage pod and run `pxctl status` - All nodes will be shown to have storage including master nodes

Testing to see only worker nodes run portworx
1) In an OCP cluster, deploy operator and storagecluster without the run-on-master annotation.
2) There will be no storage pods scheduled on master nodes
3) MaxStorageNodesPerZone will be number of worker nodes
4) Exec into storage pod and run pxctl status - all worker nodes will have storage

### DIFF
--- a/pkg/controller/storagecluster/controller_test.go
+++ b/pkg/controller/storagecluster/controller_test.go
@@ -2351,7 +2351,7 @@ func TestStoragePodGetsScheduledWithCustomNodeSpecs(t *testing.T) {
 	driver.EXPECT().GetKVDBMembers(gomock.Any()).Return(nil, nil).AnyTimes()
 	driver.EXPECT().SetDefaultsOnStorageCluster(gomock.Any()).Return(nil).AnyTimes()
 	gomock.InOrder(
-	driver.EXPECT().GetStoragePodSpec(gomock.Any(), "k8s-node-1").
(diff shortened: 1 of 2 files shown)


nikita-bhatia:
@svijaykumar-px can you add in description what testing has been done ?

svijaykumar-px:
> @svijaykumar-px can you add in description what testing has been done ?

I have added the test done. Please review it, thanks!

piyush-nimbalkar (approved):

PR #1488, opened 2024-03-19
PWX-36389: Adding changes for metroDR PDB for px <3.1.1

svijaykumar-px:
**What this PR does / why we need it**:  Addressing the comments of [this ](https://github.com/libopenstorage/operator/pull/1463) PR, changes have been made to ensure that pdb calculation is valid for DR clusters that use nodes with portworx version lesser than 3.1.1. 

Closes # PWX-36389


drivers/storage/portworx/util/util.go
```diff
@@ -1403,12 +1407,34 @@ func CountStorageNodes(
 		return -1, fmt.Errorf("failed to enumerate nodes: %v", err)
 	}
 
+	// We need to find kubernetes nodes that can run storage pods only for DR clusters that use px versions lesser than 3.1.1
+	k8sNodesStoragePodCouldRun := make(map[string]bool)
+	if pxVersion.LessThan(MinimumPxVersionCurrentDomain) && isDRSetup {
```
piyush-nimbalkar, line 1412: Should we move this after line 1455 and update the condition to ` isDRSetup && !useClusterDomain `?

This way we would have compared the px versions for all nodes. During a PX upgrade, until all PX nodes are not upgraded to 3.1.1, we should continue to calculate PDB value as before. 

`pxVersion` is the desired version from StorageCluster, it does not mean all the nodes are on that version.

drivers/storage/portworx/util/util.go
```diff
@@ -1449,17 +1491,20 @@ func CountStorageNodes(
 		}
 
 		// In case of non metro-DR setup, all portworx nodes that contain backend storage in the enumerate response are storage nodes
-		// In the case of metro DR setup, temporarily using existing logic to determine storage nodes count
-		// TODO: Need to update this logic in metro DR setup to use portworx nodes in the current domain
+		// In the case of metro DR setup, for PX versions 3.1.1 onwards, px nodes of that cluster are counted as storage nodes
```
piyush-nimbalkar, line 1494: ... of that cluster domain are counted ...

drivers/storage/portworx/util/util.go
```diff
@@ -1449,17 +1491,20 @@ func CountStorageNodes(
...
 			if !isDRSetup {
 				storageNodesCount++
-			} else {
+			} else if useClusterDomain {
+
 				if node.ClusterDomain == currentClusterDomain {
 					storageNodesCount++
 				} else {
-					logrus.Debugf("node %s is not part of cluster domain %s ", node.SchedulerNodeName, currentClusterDomain)
+					logrus.Debugf("node %s is not part of cluster domain %s ", node.Id, currentClusterDomain)
```
piyush-nimbalkar, line 1504: nit: Can we print both id and name?
```
logrus.Debugf("node %s (%s) is not part of the cluster domain %s ", node.SchedulerNodeName, node.Id, currentClusterDomain)
```

piyush-nimbalkar (changes requested):
Can you UTs for the new cases?

pkg/util/k8s/k8s.go
```diff
@@ -1672,7 +1672,11 @@ func CreateOrUpdatePodDisruptionBudget(
 
 	if modified || len(pdb.OwnerReferences) > len(existingPDB.OwnerReferences) {
 		pdb.ResourceVersion = existingPDB.ResourceVersion
-		logrus.Infof("Updating PodDisruptionBudget %s/%s", pdb.Namespace, pdb.Name)
+		if pdb.Spec.MinAvailable == nil {
+			logrus.Infof("Updating PodDisruptionBudget %s/%s", pdb.Namespace, pdb.Name)
+		} else {
+			logrus.Infof("Updating value for PodDisruptionBudget of px-storage to %d ", pdb.Spec.MinAvailable.IntValue())
```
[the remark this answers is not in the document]
svijaykumar-px: I will make it pdb.Namespace/pdb.name as that seems more intuitive

drivers/storage/portworx/util/util.go
svijaykumar-px: Sometimes the node will not have a schedulername right, will that be fine ?

piyush-nimbalkar: Yes, it is fine if it empty. We are making a change in porx to always have the scheduler name anyway.

pkg/util/k8s/k8s.go
```diff
@@ -1672,7 +1672,11 @@ func CreateOrUpdatePodDisruptionBudget(
 
 	if modified || len(pdb.OwnerReferences) > len(existingPDB.OwnerReferences) {
 		pdb.ResourceVersion = existingPDB.ResourceVersion
-		logrus.Infof("Updating PodDisruptionBudget %s/%s", pdb.Namespace, pdb.Name)
+		if pdb.Spec.MinAvailable == nil {
+			logrus.Infof("Updating PodDisruptionBudget %s/%s", pdb.Namespace, pdb.Name)
+		} else {
+			logrus.Infof("Updating value for PodDisruptionBudget of %s/%s to %d ", pdb.Namespace, pdb.Name, pdb.Spec.MinAvailable.IntValue())
```
piyush-nimbalkar, line 1678: nit: Remove "of"
Let's also say minAvailable to indicate what we are updating. `value` could mean anything.

```
Updating minAvailable for PodDisruptionBudget %s/%s to %d
```

drivers/storage/portworx/components_test.go
```diff
@@ -12815,20 +12813,15 @@ func TestPodDisruptionBudgetWithMetroDR(t *testing.T) {
...
-			{Pools: []*osdapi.StoragePool{{ID: 6}}, SchedulerNodeName: "metro-dr-node1", ClusterDomain: "domain 2"},
-			{Pools: []*osdapi.StoragePool{{ID: 7}}, SchedulerNodeName: "metro-dr-node2", ClusterDomain: "domain 2"},
-			{Pools: []*osdapi.StoragePool{}},
-			{},
+			{Pools: []*osdapi.StoragePool{{ID: 1}}, SchedulerNodeName: "node1", ClusterDomain: "domain 1", NodeLabels: map[string]string{pxutil.NodeLabelPortworxVersion: "3.1.1"}},
+			{Pools: []*osdapi.StoragePool{{ID: 2}}, SchedulerNodeName: "node2", ClusterDomain: "domain 1", NodeLabels: map[string]string{pxutil.NodeLabelPortworxVersion: "3.1.1"}},
+			{Pools: []*osdapi.StoragePool{{ID: 3}}, SchedulerNodeName: "", Hostname: "node3", ClusterDomain: "domain 1", NodeLabels: map[string]string{pxutil.NodeLabelPortworxVersion: "3.1.1"}},
+			{Pools: []*osdapi.StoragePool{{ID: 4}}, SchedulerNodeName: "node4", ClusterDomain: "domain 2", NodeLabels: map[string]string{pxutil.NodeLabelPortworxVersion: "3.1.1"}},
+			{Pools: []*osdapi.StoragePool{{ID: 5}}, SchedulerNodeName: "node5", ClusterDomain: "domain 2", NodeLabels: map[string]string{pxutil.NodeLabelPortworxVersion: "3.1.1"}},
```
piyush-nimbalkar, line 12824: Can we add a decommissioned node with an older version here, so we can test the case that we ignore such nodes?

drivers/storage/portworx/components_test.go
```diff
@@ -12911,6 +12904,26 @@ func TestPodDisruptionBudgetWithMetroDR(t *testing.T) {
 	err = testutil.Get(k8sClient, storagePDB, component.StoragePodDisruptionBudgetName, cluster.Namespace)
 	require.NoError(t, err)
 	require.Equal(t, 2, storagePDB.Spec.MinAvailable.IntValue())
+
+	// When portworx version is lesser than 3.1.1 , cluster domain field will not be available, schedulerNodeName will be used to decide the nodes in current domain
```
piyush-nimbalkar, line 12908: Let's duplicate this test case and test scenario
- when NodeLabelPortworxVersion has invalid version string and the parsing fails
- when the version is greater than or equal 3.1.0 and less than 3.1.1

Basically we should try to cover all the branches in our code for good UT coverage.

piyush-nimbalkar (approved):
LGTM after adding more UTs.

PR #1491, opened 2024-03-20
PWX-36460 : Enforce portworx SCC on portworx pods

nikita-bhatia:
On OCP 4.15, seccompProfiles runtime/default is applied on node-exporter SCC, and openshift applies node-exporter SCC to pods which doesn't have SCC applied. The default seccomp profile runtime/default is applied to these pods which can be used to limit the process running in a container to only using a subset of the available system calls.
This limits portworx pod to make k8s update.

**What this PR does / why we need it**: Enforcing portworx pods to always use portworx SCC so that openshift doesn't apply random SCC which fits the pod.

Closes # https://portworx.atlassian.net/browse/PWX-36460

- Tested on OCP 4.15
- Tested on OCO 4.14
- Tested OCP upgrade from 4.14 to 4.15


pkg/controller/storagecluster/storagecluster.go
```diff
@@ -1450,6 +1451,10 @@ func (c *Controller) CreatePodTemplate(
 		}
 	}
 
+	if pxutil.IsOpenshift(cluster) {
+		newTemplate.Annotations[component.OpenshiftRequiredSCCAnnotation] = component.PxSCCName
```
piyush-nimbalkar, line 1455: Can you write a unit test to validate that the pod created when is-openshift set is true has the required-scc annotation?

piyush-nimbalkar (approved):
LGTM after UT.

In the testing and release notes we should mention that the operator needs to be upgraded first before openshift is upgraded. That way when the px pods restart after the ocp upgrade, they get the correct SCC.

pkg/controller/storagecluster/storagecluster.go
nikita-bhatia: Will add in seperate PR as discussed

PR #1496, opened 2024-03-22
PWX-36649 Keep the required-scc annotation if added during creation

piyush-nimbalkar:
- The `openshift.io/required-scc` was added to the portworx pod during creation. But it was getting removed during updating the custom annotations on the px pod. Fixed this by adding it to known pod annotation so we don't skip it.
- Added UTs for all required-scc changes.

Closes # https://portworx.atlassian.net/browse/PWX-36649


nikita-bhatia (approved):
LGTM

PR #1503, opened 2024-03-27
PWX-36170: Adding k8s node name as label to all px pods

svijaykumar-px:
**What this PR does / why we need it**:This change is to help in the parallel upgrade feature and is a part of [this](https://purestorage.atlassian.net/browse/PWX-25414) epic. A unique label, in this case kubernetes node name that the portworx pod is running in is added to all the portworx pods.

Closes #  PWX-36170

### DIFF
--- a/pkg/constants/metadata.go
+++ b/pkg/constants/metadata.go
@@ -44,6 +44,9 @@ const (
 	OperatorLabelManagedByKey = OperatorPrefix + "/managed-by"
 	// OperatorLabelManagedByValue indicates that the object is managed by portworx.
 	OperatorLabelManagedByValue = "portworx"
+	// K8sNodeRunningPxPod hold the value of the kubernetes node on whcih the portworx pod is running
+    // Used for creating a Node PDB
+    K8sNodeRunningPxPod = "k8s-node-running-px-pod"
 )
 
 const (
--- a/pkg/constants/metadata.go
+++ b/pkg/constants/metadata.go
@@ -45,8 +45,8 @@ const (
 	// OperatorLabelManagedByValue indicates that the object is managed by portworx.
 	OperatorLabelManagedByValue = "portworx"
 	// K8sNodeRunningPxPod hold the value of the kubernetes node on whcih the portworx pod is running
-    // Used for creating a Node PDB
-    K8sNodeRunningPxPod = "k8s-node-running-px-pod"
+	// Used for creating a Node PDB
+	K8sNodeRunningPxPod = "k8s-node-running-px-pod"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


pkg/constants/metadata.go
```diff
@@ -44,6 +44,9 @@ const (
 	OperatorLabelManagedByKey = OperatorPrefix + "/managed-by"
 	// OperatorLabelManagedByValue indicates that the object is managed by portworx.
 	OperatorLabelManagedByValue = "portworx"
+	// K8sNodeRunningPxPod hold the value of the kubernetes node on whcih the portworx pod is running
+	// Used for creating a Node PDB
+	K8sNodeRunningPxPod = "k8s-node-running-px-pod"
```
piyush-nimbalkar, line 49: OperatorLabelNodeNameKey  = OperatorPrefix + "/node-name"

pkg/controller/storagecluster/controller_test.go
```diff
@@ -1443,8 +1443,10 @@ func TestStoragePodGetsScheduled(t *testing.T) {
 	require.Len(t, podControl.Templates, 2)
 	podTemplate1 := expectedPodTemplate.DeepCopy()
 	podTemplate1.Spec.NodeName = "k8s-node-1"
+	podTemplate1.Labels["k8s-node-running-px-pod"] = "k8s-node-1"
```
piyush-nimbalkar, line 1446: nit: Use the constant here so we don't have to change code everywhere if we end up changing the key.

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -1441,6 +1441,7 @@ func (c *Controller) CreatePodTemplate(
 	podSpec.NodeName = node.Name
 	labels := c.StorageClusterSelectorLabels(cluster)
 	labels[constants.OperatorLabelManagedByKey] = constants.OperatorLabelManagedByValue
+	labels[constants.K8sNodeRunningPxPod] = node.Name
```
piyush-nimbalkar, line 1444: I was initially thinking that we should update the label for existing pods as well (without portworx update), but looks like we will support the upgrade feature starting portworx 3.1.2/3.2.0. So users will anyway have to upgrade portworx and will end up re-creating portworx pods.

piyush-nimbalkar (approved):

pkg/controller/storagecluster/storagecluster.go
svijaykumar-px: Right so this change should go along with pxe release

piyush-nimbalkar: Yes, before the pxe release.

PR #1504, opened 2024-03-27
PTX-16508 Aetos Wrap for Operator Integration Test

nilanto07-px:
This purpose of this PR is to integrate Aétos with operator integration tests.

main_test.go is modified to init the Aetos Dashboard (functions defined under `test/integration_test/utils/aetos`)
run time flags are also modified as constants to beautify the code. No change to parsing logic for these flags.

basic_test.go has 2 additions: testCase `BasicInstallAetosDemo` and test `TestBasicInstall`
We would need to update other testCases as per `BasicInstallAetosDemo` and test `TestStorageClusterBasic` as per `TestBasicInstall` for correct reporting into Aétos. 

storagecluster.go has 2 additions as well: functions `DeployAndValidateStorageClusterAetosDemo` and `UninstallAndValidateStorageClusterAetosDemo`
The reason for putting these Demo functions is that the current implementation for testSteps have `require.NoError(t, err)` checks instead of returning err to caller functions in basic_test.go -> this makes it hard/impossible to catch the error at testCase level. 

The reason for adding new functions is to not cause any regression and keep these as placeholder for future changes. 
This would require a major revamp of the integration test code to integrate all of the tests to Aétos.

### DIFF
--- a/test/integration_test/basic_test.go
+++ b/test/integration_test/basic_test.go
@@ -17,6 +17,7 @@ import (
 	testutil "github.com/libopenstorage/operator/pkg/util/test"
 	"github.com/libopenstorage/operator/test/integration_test/types"
 	ci_utils "github.com/libopenstorage/operator/test/integration_test/utils"
+	aetos "github.com/libopenstorage/operator/test/integration_test/utils/aetos"
 	coreops "github.com/portworx/sched-ops/k8s/core"
 	"github.com/portworx/sched-ops/k8s/operator"
 	"github.com/sirupsen/logrus"
@@ -304,6 +305,49 @@ func BasicInstall(tc *types.TestCase) func(*testing.T) {
 
 		// Deploy PX and validate
 		cluster = ci_utils.DeployAndValidateStorageCluster(cluster, ci_utils.PxSpecImages, t)
+	}
+}
+
+func TestBasicInstall(t *testing.T) {
+	testCase := &types.TestCase{
+		TestName: "BasicInstall",
+		TestSpec: func(t *testing.T) interface{} {
+			objects, err := ci_utils.ParseSpecs("storagecluster/storagecluster-with-all-components.yaml")
+			require.NoError(t, err)
+			cluster, ok := objects[0].(*corev1.StorageCluster)
+			require.True(t, ok)
+			cluster.Name = "test-basic-stc"
+			cluster.Namespace = ci_utils.PxNamespace
+			return cluster
+		},
+		TestFunc: BasicInstallAetosDemo,
+	}
+	BasicInstallAetosDemo(testCase)(t)
+}
+
+func BasicInstallAetosDemo(tc *types.TestCase) func(*testing.T) {
+	return func(t *testing.T) {
+		testSpec := tc.TestSpec(t)
+		cluster, ok := testSpec.(*corev1.StorageCluster)
+		dash.VerifyFatal(t, ok, true, fmt.Sprintf("Get StorageCluster from test spec"))
+
+		defer dash.TestCaseEnd()
+		dash.TestCaseBegin("BasicInstall", "Deploying and validating StorageCluster", "", nil)
+
+		// Construct StorageCluster
+		aetos.Infof("Constructing StorageCluster %s", cluster.Name)
+		err := ci_utils.ConstructStorageCluster(cluster, ci_utils.PxSpecGenURL, ci_utils.PxSpecImages)
+		dash.VerifyFatal(t, err, nil, fmt.Sprintf("Constructed StorageCluster %s? %v", cluster.Name, err))
+
+		// Deploy PX and validate
+		aetos.Infof("Deploying & Validating StorageCluster %s", cluster.Name)
+		_, err = ci_utils.DeployAndValidateStorageClusterAetosDemo(cluster, ci_utils.PxSpecImages)
+		dash.VerifyFatal(t, err, nil, fmt.Sprintf("Deployed and validated StorageCluster %s? %v", cluster.Name, err))
+
+		// Uninstall and validate
+		aetos.Infof("Uninstalling & Validating StorageCluster %s", cluster.Name)
+		err = ci_utils.UninstallAndValidateStorageClusterAetosDemo(cluster)
+		dash.VerifyFatal(t, err, nil, fmt.Sprintf("Uninstalled and validate StorageCluster %s? %v", cluster.Name, err))
 	}
 }
 
--- a/test/integration_test/main_test.go
+++ b/test/integration_test/main_test.go
@@ -4,28 +4,72 @@
 package integrationtest
 
 import (
+	"bufio"
+	"crypto/tls"
+	"encoding/csv"
 	"flag"
 	"fmt"
+	"net/http"
 	"os"
+	"strconv"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/sirupsen/logrus"
 
 	test_util "github.com/libopenstorage/operator/pkg/util/test"
 	"github.com/libopenstorage/operator/test/integration_test/types"
 	ci_utils "github.com/libopenstorage/operator/test/integration_test/utils"
+	aetosutil "github.com/libopenstorage/operator/test/integration_test/utils/aetos"
 	appsv1 "k8s.io/api/apps/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
+
+const (
+	logLevelFlag            = "log-level"
+	vSphereUsernameFlag     = "portworx-vsphere-username"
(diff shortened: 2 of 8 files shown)


nilanto07-px:
**Test Pass scenario:**
Ran Locally
_Aétos_ - https://aetos.pwx.purestorage.com/resultSet/testSetID/593894/testCaseID/3627526

**Test Failure scenario:** 
https://jenkins.pwx.dev.purestorage.com/job/Users/job/nbhattacharya/job/nilanto-ocp-job/33/console 
_Aétos_ - https://aetos.pwx.purestorage.com/resultSet/testSetID/593789

_TODO_: If we pass `-e DASH_UID=`ls ${WORKSPACE}/spawn/build.properties >> /dev/null 2>&1 && cat ${WORKSPACE}/spawn/build.properties | grep -i DASH_UID= | grep -Eo '[0-9]+'|| echo 0` \
` to docker run command for operator integration tests for it to re-use the existing DASH_UID created during spawn provision or px install, it still creates a new testSetID. Need to debug and make integration tests re-use the same.

test/integration_test/basic_test.go
```diff
@@ -307,6 +308,61 @@ func BasicInstall(tc *types.TestCase) func(*testing.T) {
...
+		dash.VerifySafely(err == nil, true,
+			fmt.Sprintf("Constructed StorageCluster %s? %v", cluster.Name, err))
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		// Deploy PX and validate
+		aetos.LogInfoD("Deploying & Validating StorageCluster %s", cluster.Name)
+		_, err = ci_utils.DeployAndValidateStorageClusterAetosDemo(cluster, ci_utils.PxSpecImages)
+		dash.VerifySafely(err == nil, true,
```
kphalgun-px, line 347: How about? So, we need not include an error message in the description.

```go
dash.VerifySafely(err, nil, fmt.Sprintf("Deployed and validated StorageCluster %s?", cluster.Name))
```

test/integration_test/basic_test.go
```diff
@@ -307,6 +308,61 @@ func BasicInstall(tc *types.TestCase) func(*testing.T) {
...
+		aetos.LogInfoD("Uninstalling & Validating StorageCluster %s", cluster.Name)
+		err = ci_utils.UninstallAndValidateStorageClusterAetosDemo(cluster)
+		dash.VerifySafely(err == nil, true,
+			fmt.Sprintf("Uninstalled and validate StorageCluster %s? %v", cluster.Name, err))
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		defer dash.TestCaseEnd()
+		defer dash.TestSetEnd()
```
kphalgun-px, line 362: How about adding `defer` statements immediately after `TestCaseBegin` and `TestSetBegin`, or next to their respective counterparts? This ensures that if an error triggers `t.Fatal(err)`, causing the test to fail and halt immediately, the `defer` functions will still execute, as they would have been set prior to the point of failure.

test/integration_test/basic_test.go
nilanto07-px: We would want the error to be logged to Aétos.

test/integration_test/basic_test.go
nilanto07-px: main_test.go already has TestSetEnd() called. This defer is to defer ending the testSet in case there are other tests in the Focus list. I will remove the defer for TestCaseEnd, good catch!

kphalgun-px (approved):
LGTM

test/integration_test/basic_test.go
```diff
@@ -307,6 +308,61 @@ func BasicInstall(tc *types.TestCase) func(*testing.T) {
...
+
+func BasicInstallAetosDemo(tc *types.TestCase) func(*testing.T) {
+	return func(t *testing.T) {
+		testSpec := tc.TestSpec(t)
+		cluster, ok := testSpec.(*corev1.StorageCluster)
+		require.True(t, ok)
+		dash.TestCaseBegin("BasicInstall", "Deploying and validating StorageCluster", "", nil)
+
+		// Construct StorageCluster
+		aetos.LogInfoD("Constructing StorageCluster %s", cluster.Name)
```
aghodke1312, line 336: change this to aetos.InfoD instead of calling log everytime.

test/integration_test/basic_test.go
```diff
@@ -307,6 +308,61 @@ func BasicInstall(tc *types.TestCase) func(*testing.T) {
...
+		err := ci_utils.ConstructStorageCluster(cluster, ci_utils.PxSpecGenURL, ci_utils.PxSpecImages)
+		dash.VerifySafely(err == nil, true,
+			fmt.Sprintf("Constructed StorageCluster %s? %v", cluster.Name, err))
+		if err != nil {
+			t.Fatal(err)
```
aghodke1312, line 341: if you want to log the error and fail, use verifyFatal method.

aghodke1312 (changes requested):

test/integration_test/basic_test.go
nilanto07-px: Done.

test/integration_test/basic_test.go
```diff
@@ -307,6 +308,61 @@ func BasicInstall(tc *types.TestCase) func(*testing.T) {
...
+			require.NoError(t, err)
+			cluster, ok := objects[0].(*corev1.StorageCluster)
+			require.True(t, ok)
+			cluster.Name = "test-basic-stc"
+			cluster.Namespace = ci_utils.PxNamespace
+			return cluster
+		},
+		TestFunc: BasicInstallAetosDemo,
+	}
+	BasicInstallAetosDemo(testCase)(t)
```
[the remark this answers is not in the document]
nilanto07-px: TestSetBegin called at main_test.go itself. No reference in testcases.

test/integration_test/main_test.go
```diff
@@ -170,3 +252,125 @@ func setup() error {
...
+	}
+
+	dash.IsEnabled = enableDash
+	testSet := aetosutil.TestSet{
+		User:        user,
+		Product:     testProduct,
+		Description: testDescription,
+		Branch:      testBranch,
+		TestType:    testType,
+		Tags:        make(map[string]string),
```
aghodke1312, line 272: create a tags map for string : string. Pass operator-test:"true" for now.

test/integration_test/main_test.go
```diff
@@ -35,91 +79,126 @@ func setup() error {
 	var operatorUpgradeHopsImages string
 	var logLevel string
 	var err error
+	var enableDash bool
+	var user, testBranch, testProduct, testType, testDescription, testTags string
+	var testsetID int
 
 	flag.StringVar(&ci_utils.PxDockerUsername,
-		"portworx-docker-username",
+		dockerUsernameFlag,
```
aghodke1312, line 87: why are we changing these flags? The existing tests will fail.

aghodke1312 (changes requested):
Along with these, you will need to update. https://github.com/libopenstorage/operator/blob/master/test/integration_test/test-deploy.sh
and 
https://github.com/libopenstorage/operator/blob/master/test/integration_test/operator-test-pod-template.yaml

Please refer to: https://github.com/libopenstorage/stork/pull/1636
These will be required when we call the test from jenkins.

test/integration_test/main_test.go
nilanto07-px: > why are we changing these flags? The existing tests will fail.

If you look at the commit message it says clearly that these are moved to variables for easy code maintenance and readability.

test/integration_test/main_test.go
nilanto07-px: That is already part of TestCaseBegin call. In case it's user provided, we have handled the same in deploy.sh

test/integration_test/basic_test.go
```diff
@@ -307,6 +308,49 @@ func BasicInstall(tc *types.TestCase) func(*testing.T) {
...
+}
+
+func BasicInstallAetosDemo(tc *types.TestCase) func(*testing.T) {
+	return func(t *testing.T) {
+		testSpec := tc.TestSpec(t)
+		cluster, ok := testSpec.(*corev1.StorageCluster)
+		dash.VerifyFatal(t, ok, true, fmt.Sprintf("Get StorageCluster from test spec"))
+
+		defer dash.TestCaseEnd()
```
lsrinivas-pure, line 334: below code should be the first step . so that no logs will be missed to push to aetos
```
defer dash.TestCaseEnd()
dash.TestCaseBegin("BasicInstall", "Deploying and validating StorageCluster", "", nil)

```

lsrinivas-pure (approved):
Minor comment plz have a look

PR #1508, opened 2024-04-02
PWX-36475 : Creating node PDB for parallel upgrades

svijaykumar-px:
**What this PR does / why we need it**: Creating Node PDB for every node in the cluster which is a part of quorum. This will be used in parallel upgrades. The PDB for every node is created with the name px-storage-<nodeName> and matches with the node name label of every portworx pod

Closes # PWX-36475

**Special notes for your reviewer**: This per node PDB will only be created when portworx version is greater than or equal to 3.1.2


drivers/storage/portworx/util/util.go
```diff
@@ -1902,3 +1905,97 @@ func isVersionSupported(current, target string) bool {
...
+	err = k8sClient.List(context.TODO(), k8sNodeList)
+	if err != nil {
+		return nil, err
+	}
+	for _, node := range k8sNodeList.Items {
+		shouldRun, shouldContinueRunning, err := k8sutil.CheckPredicatesForStoragePod(&node, cluster, nil)
+		if err != nil {
+			return nil, err
+		}
+		if shouldRun || shouldContinueRunning {
```
piyush-nimbalkar, line 1938: Why do we need to check if that k8s node can run a portworx pod or not? Don't we need to only check if the portworx node is part of the current cluster or not? Basically this can be a simple map - 
```
for _, node := range k8sNodeList.Items {
  k8sNodes[node.Name] = true
}
```

drivers/storage/portworx/util/util.go
```diff
@@ -1902,3 +1905,97 @@ func isVersionSupported(current, target string) bool {
...
+// Get list of storagenodes that are a part of the current cluster that need a node PDB
+func NodesNeedingPDB(cluster *corev1.StorageCluster, sdkConn *grpc.ClientConn, k8sClient client.Client) ([]string, error) {
+	// Get the list of storage nodes
+	nodeClient := api.NewOpenStorageNodeClient(sdkConn)
+	ctx, err := SetupContextWithToken(context.Background(), cluster, k8sClient)
+	if err != nil {
+		return nil, err
+	}
+
+	nodeEnumerateResponse, err := nodeClient.EnumerateWithFilters(
```
piyush-nimbalkar, line 1918: We are making this call here, in CountStorageNodes and in ClusterSupportsParallelUpgrade. To avoid making 2 portworx calls to get the same data, can we call it once from the pdb component and then pass the output to these functions?

drivers/storage/portworx/components_test.go
```diff
@@ -13411,6 +13411,113 @@ func TestDisablePodDisruptionBudgets(t *testing.T) {
...
+	component.DeregisterAllComponents()
+	component.RegisterDisruptionBudgetComponent()
+
+	driver := portworx{}
+	err = driver.Init(k8sClient, runtime.NewScheme(), record.NewFakeRecorder(0))
+	require.NoError(t, err)
+	err = driver.PreInstall(cluster)
+	require.NoError(t, err)
+
+	// PDB is created for 2 storage nodes and kvdb
```
piyush-nimbalkar, line 13488: We are not setting the `spec.kvdb.internal` field in the storage cluster. Then why is the kvdb pdb getting created here?

drivers/storage/portworx/components_test.go
```diff
@@ -13411,6 +13411,113 @@ func TestDisablePodDisruptionBudgets(t *testing.T) {
...
+	require.Len(t, pdbList.Items, 3)
+
+	storagePDB := &policyv1.PodDisruptionBudget{}
+	err = testutil.Get(k8sClient, storagePDB, component.StoragePodDisruptionBudgetName+"-node1", cluster.Namespace)
+	require.NoError(t, err)
+	require.Equal(t, 1, storagePDB.Spec.MinAvailable.IntValue())
+	require.Equal(t, "node1", storagePDB.Spec.Selector.MatchLabels[constants.OperatorLabelNodeNameKey])
+
+	err = testutil.Get(k8sClient, storagePDB, component.StoragePodDisruptionBudgetName+"-node3", cluster.Namespace)
+	fmt.Println(err)
```
piyush-nimbalkar, line 13501: nit: Remove print

drivers/storage/portworx/components_test.go
```diff
@@ -13411,6 +13411,113 @@ func TestDisablePodDisruptionBudgets(t *testing.T) {
...
+	require.Error(t, err, fmt.Errorf("poddisruptionbudgets.policy %s-node3 not found", component.StoragePodDisruptionBudgetName))
```
piyush-nimbalkar, line 13502: nit: Alternatively you can validate this like this -
```
require.True(t, errors.IsNotFound(err))
```

drivers/storage/portworx/components_test.go
```diff
@@ -13411,6 +13411,113 @@ func TestDisablePodDisruptionBudgets(t *testing.T) {
...
+		Spec: corev1.StorageClusterSpec{
+			Image: "portworx/oci-monitor:3.1.2",
+		},
+		Status: corev1.StorageClusterStatus{
+			Phase: string(corev1.ClusterStateRunning),
+		},
+	}
+	fakeK8sNodes := &v1.NodeList{Items: []v1.Node{
+		{ObjectMeta: metav1.ObjectMeta{Name: "node1"}},
+		{ObjectMeta: metav1.ObjectMeta{Name: "node2"}},
```
piyush-nimbalkar, line 13458: You need to add `node3` as well here. A PDB is not getting created on this node also because it is not even part of the k8s cluster.

drivers/storage/portworx/components_test.go
```diff
@@ -13411,6 +13411,113 @@ func TestDisablePodDisruptionBudgets(t *testing.T) {
...
+
+	testutil.SetupEtcHosts(t, sdkServerIP, pxutil.PortworxServiceName+".kube-test")
+	defer testutil.RestoreEtcHosts(t)
+
+	// PDB should not be created for non quorum members
+	expectedNodeEnumerateResp := &osdapi.SdkNodeEnumerateWithFiltersResponse{
+		Nodes: []*osdapi.StorageNode{
+			{SchedulerNodeName: "node1", NodeLabels: map[string]string{pxutil.NodeLabelPortworxVersion: "3.1.2"}},
+			{SchedulerNodeName: "node2", NodeLabels: map[string]string{pxutil.NodeLabelPortworxVersion: "3.1.2"}},
+			{NonQuorumMember: true, SchedulerNodeName: "node3", NodeLabels: map[string]string{pxutil.NodeLabelPortworxVersion: "3.1.2"}},
```
piyush-nimbalkar, line 13436: We should add a decommissioned node and a node that is not part of the cluster (don't create it below as part of the fakeK8sNodes).

Let's just ensure that we have a case that covers and validates every if condition/branch in the code we are touching.

drivers/storage/portworx/component/disruption_budget.go
```diff
@@ -157,6 +167,40 @@ func (c *disruptionBudget) createPortworxPodDisruptionBudget(
...
+				MinAvailable: &minAvailable,
+				Selector: &metav1.LabelSelector{
+					MatchLabels: map[string]string{
+						constants.LabelKeyClusterName:      cluster.Name,
+						constants.OperatorLabelNodeNameKey: node,
+					},
+				},
+			},
+		}
+		err = k8sutil.CreateOrUpdatePodDisruptionBudget(c.k8sClient, pdb, ownerRef)
```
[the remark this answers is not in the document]
svijaykumar-px: If we aggregate the errors in a map/list then we will have to change the function signatures of all the calling functions. Instead I will do this - print the errors as they appear, and then return when the first error occurs while trying to create node PDBs. Or should i instead wait till node PDBs are created for all nodes and return the first error in the end?

drivers/storage/portworx/util/util.go
svijaykumar-px: It was mentioned that we will create PDB only for k8s nodes that can run px right? So during surge upgrade when a node is being removed from the cluster, it shouldn't run px and we wont create a node PDB for it .

drivers/storage/portworx/components_test.go
svijaykumar-px: It is creating kvdb pdb as kvdb spec itself is nil, so it doesnt check if we set internal or not

svijaykumar-px: `if cluster.Spec.Kvdb != nil && !cluster.Spec.Kvdb.Internal ` the first condition fails and thus creates kvdb pdb

drivers/storage/portworx/util/util.go
svijaykumar-px: For now I will make the change suggested, but lets discuss why we need to create node PDB for k8s nodes that should not run px pod

drivers/storage/portworx/component/disruption_budget.go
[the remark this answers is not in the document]
piyush-nimbalkar: I wasn't asking to return an array of errors, but an aggregated error.
You can use the error aggregator from kubernetes to do that. See example - https://github.com/libopenstorage/operator/blob/master/pkg/controller/storagecluster/storagecluster.go#L1170.

drivers/storage/portworx/util/util.go
piyush-nimbalkar: If a portworx node has scheduler name which is part of the cluster, then we should protect that k8s node from upgrading irrespective of whether currently the k8s node should run portworx or not, right?

Let's discuss more on this offline to make sure we are covering all cases correctly.

svijaykumar-px: I assumed a kubernetes node when it goes to the state of should not run portworx, then px will be deleted from that node soon anyways, so why should we create pdb for it only to delete it in sometime?

drivers/storage/portworx/component/disruption_budget.go
```diff
@@ -71,9 +75,40 @@ func (c *disruptionBudget) Reconcile(cluster *corev1.StorageCluster) error {
...
+	}
+	nodeEnumerateResponse, err := nodeClient.EnumerateWithFilters(
+		ctx,
+		&api.SdkNodeEnumerateWithFiltersRequest{},
+	)
+	if err != nil {
+		return fmt.Errorf("failed to enumerate nodes: %v", err)
+	}
+
+	if pxutil.ClusterSupportsParallelUpgrade(nodeEnumerateResponse) {
```
nikita-bhatia, line 99: If parallel upgrade is supported, calling k8sClient.List() can be a heavy operation depending on the size of the list it's retrieving. Can we instead fetch the list once and passing it as a parameter below?

nikita-bhatia (approved):
LGTM

drivers/storage/portworx/component/disruption_budget.go
```diff
@@ -178,6 +215,73 @@ func (c *disruptionBudget) createPortworxPodDisruptionBudget(
...
+	k8sNodeList *v1.NodeList,
+) error {
+	nodesNeedingPDB, err := pxutil.NodesNeedingPDB(c.k8sClient, nodeEnumerateResponse, k8sNodeList)
+	if err != nil {
+		return err
+	}
+	errors := []error{}
+	for _, node := range nodesNeedingPDB {
+		minAvailable := intstr.FromInt(1)
+		PDBName := "px-" + node
```
piyush-nimbalkar, line 231: nit: internal variables should have lower case.

PR #1510, opened 2024-04-03
PWX-36159: Use different port for PVC deployment on k3s

nikita-bhatia:
**What this PR does / why we need it**: PVC controller is in CrashLoopBackOff when deployed on k3s master node because of  following error.
**failed to create listener: failed to listen on 0.0.0.0:10257: listen tcp 0.0.0.0:10257: bind: address already in use"**

The default ports 10252 and 10257 for PVC are already in use on k3s clusters. This PR chenges the default config, and uses port numbers 10260 and 10261 on k3s deployment.

Closes # https://purestorage.atlassian.net/browse/PWX-36159

- Verified on k3s cluster, with pvc annotation enabled.
-  All 3 PVCs pods are running fine
I0403 12:46:30.728036       1 secure_serving.go:213] Serving securely on 0.0.0.0:10261

### DIFF
--- a/drivers/storage/portworx/component/pvccontroller.go
+++ b/drivers/storage/portworx/component/pvccontroller.go
@@ -3,17 +3,12 @@ package component
 import (
 	"context"
 	"fmt"
-	"reflect"
-	"strconv"
-	"strings"
-
 	"github.com/hashicorp/go-version"
 	pxutil "github.com/libopenstorage/operator/drivers/storage/portworx/util"
 	corev1 "github.com/libopenstorage/operator/pkg/apis/core/v1"
 	"github.com/libopenstorage/operator/pkg/constants"
 	"github.com/libopenstorage/operator/pkg/util"
 	k8sutil "github.com/libopenstorage/operator/pkg/util/k8s"
-	coreops "github.com/portworx/sched-ops/k8s/core"
 	"github.com/sirupsen/logrus"
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
@@ -25,7 +20,9 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/client-go/tools/record"
+	"reflect"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"strconv"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


drivers/storage/portworx/component/pvccontroller.go
```diff
@@ -551,6 +568,31 @@ func (c *pvcController) getPVCControllerDeploymentSpec(
...
+		}
+
+		// Iterate over nodes and check their versions for "k3s"
+		for _, node := range nodes.Items {
+			if strings.Contains(node.Status.NodeInfo.KubeletVersion, "k3s") {
+				c.isK3s = boolPtr(true)
+				return true
+			}
+		}
+		c.isK3s = boolPtr(false)
```
piyush-nimbalkar, line 590: This is not incorrect, but we already have a way in the operator code to detect a k3s cluster from the k8s server version string. https://github.com/libopenstorage/operator/blob/master/drivers/storage/portworx/deployment.go#L1951

Let's keep it consistent and use the same. I would recommend modifying the `isK3sClusterExt` method and extracting out the `strings.HasPrefix(ext[1:], "k3s")` part into `portworx/util.go`. Then call the same from here.

drivers/storage/portworx/component/pvccontroller.go
```diff
@@ -49,6 +51,12 @@ const (
 	AksPVCControllerInsecurePort = "10260"
 	// AksPVCControllerSecurePort is the PVC controller secure port.
 	AksPVCControllerSecurePort = "10261"
+
+	// K3sPVCControllerInsecurePort is the PVC controller port and health check port, due to the default port
+	// is already used on K3S we use different port for K3S.
+	K3sPVCControllerInsecurePort = "10260"
+	// K3SPVCControllerSecurePort is the PVC controller secure port.
+	K3sPVCControllerSecurePort = "10261"
```
piyush-nimbalkar, line 59: Instead of re-declaring the constants, let's just rename them and use one constant for the same ports.
May be `CustomPVCControllerSecurePort` and `CustomPVCControllerInsecurePort`.

drivers/storage/portworx/component/pvccontroller.go
```diff
@@ -333,13 +342,17 @@ func (c *pvcController) createDeployment(
 			command = append(command, "--port="+port)
 		} else if pxutil.IsAKS(cluster) {
 			command = append(command, "--port="+AksPVCControllerInsecurePort)
+		} else if c.isK3sDeployment() {
```
piyush-nimbalkar, line 345: Once we use the same constant name, we can club these two conditions into one condition with an OR operator.

piyush-nimbalkar (approved):

PR #1514, opened 2024-04-05
PWX-36663 : Remove px-prometheus user from Portworx SCC for OCP 4.12+

nikita-bhatia:
**What this PR does / why we need it**: Since we stopped support for px-prometheus for OCP version >= 4.12, we don’t need the User system:serviceaccount:portworx:px-prometheus in the Security Context Constraint portworx anymore. This PR adds the px-prometheus user in Portworx SCC, only if current platform is not Openshift 4.12+. For other platforms, it will remain same.

Closes # https://purestorage.atlassian.net/browse/PWX-36663

- Tested on Openshift 4.15 and 4.14, Portworx SCC doesn't have px-prometheus User


drivers/storage/portworx/component/securitycontextconstraints.go
```diff
@@ -185,95 +185,101 @@ func (s *scc) getSCCs(cluster *opcorev1.StorageCluster) []ocp_secv1.SecurityCont
...
+			fmt.Sprintf("system:serviceaccount:%s:%s", cluster.Namespace, pxutil.PortworxServiceAccountName(cluster)),
+			fmt.Sprintf("system:serviceaccount:%s:%s", cluster.Namespace, CollectorServiceAccountName),
+			fmt.Sprintf("system:serviceaccount:%s:%s", cluster.Namespace, ServiceAccountNameTelemetry),
+			fmt.Sprintf("system:serviceaccount:%s:%s", cluster.Namespace, "px-node-wiper"),
+			fmt.Sprintf("system:serviceaccount:%s:%s", cluster.Namespace, "stork-scheduler"),
 		},
 	}
+
+	if gte_ocp_412, err := pxutil.IsSupportedOCPVersion(s.k8sClient, pxutil.OpenshiftPrometheusSupportedVersion); err == nil {
+		if !gte_ocp_412 {
```
piyush-nimbalkar, line 230: Can we override this behavior if prometheus is explicitly enabled using `spec.monitoring.prometheus.enabled`? That way we still leave a backdoor for the px deployed prometheus in case there is a remote possibility that we recommend deploying our own prometheus.

nikita-bhatia: Updated : Adding px-prometheus user to Portworx SCC if
- either openshift version is less than 4.12
- or px prometheus is enabled

piyush-nimbalkar (approved):

drivers/storage/portworx/component/securitycontextconstraints.go
```diff
@@ -185,95 +185,102 @@ func (s *scc) getSCCs(cluster *opcorev1.StorageCluster) []ocp_secv1.SecurityCont
...
+		Users: []string{
+			fmt.Sprintf("system:serviceaccount:%s:%s", cluster.Namespace, pxutil.PortworxServiceAccountName(cluster)),
+			fmt.Sprintf("system:serviceaccount:%s:%s", cluster.Namespace, CollectorServiceAccountName),
+			fmt.Sprintf("system:serviceaccount:%s:%s", cluster.Namespace, ServiceAccountNameTelemetry),
+			fmt.Sprintf("system:serviceaccount:%s:%s", cluster.Namespace, "px-node-wiper"),
+			fmt.Sprintf("system:serviceaccount:%s:%s", cluster.Namespace, "stork-scheduler"),
 		},
 	}
+
+	if gte_ocp_412, err := pxutil.IsSupportedOCPVersion(s.k8sClient, pxutil.OpenshiftPrometheusSupportedVersion); err == nil {
```
diptiranjanpx, line 229: In error case, we are going with the default scc list ?

PR #1520, opened 2024-04-10
Remove image and bundle building from travis

piyush-nimbalkar:

Closes #

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,19 +3,12 @@ dist: focal
 language: go
 go:
   - 1.21.4
-before_install:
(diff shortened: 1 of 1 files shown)


mborodin-px (commented):
I guess we can remove Travis integration completely as we run the same on Jenkins

mborodin-px (approved):

piyush-nimbalkar:
> I guess we can remove Travis integration completely as we run the same on Jenkins

It is useful now to track the code coverage and block PRs if the coverage is not enough. To calculate the code coverage, we need to run the UTs. That is all we are doing in Travis.

PR #1522, opened 2024-04-15
PTX-23350 validatePxPodsAnnotations for StorageNode should be called when op >= 23.10.0

nilanto07-px:
Some `storageNode.Status.NodeAttributes` like `nodeAttributes` were introduced in op 23.10.0
This PR is to check for these attributes only when opVersion is higher than expected. 

Closes # PTX-23350

### DIFF
--- a/pkg/util/test/util.go
+++ b/pkg/util/test/util.go
@@ -196,7 +196,7 @@ const (
 var TestSpecPath = "testspec"
 
 var (
-	opVer1_5_0, _   = version.NewVersion("1.5.0-")
+	// opVer1_5_0, _   = version.NewVersion("1.5.0-")
 	opVer1_9_1, _   = version.NewVersion("1.9.1-")
 	opVer1_10, _    = version.NewVersion("1.10.0-")
 	opVer23_3, _    = version.NewVersion("23.3.0-")
@@ -204,6 +204,7 @@ var (
 	opVer23_5_1, _  = version.NewVersion("23.5.1-")
 	opVer23_7, _    = version.NewVersion("23.7.0-")
 	opVer23_8, _    = version.NewVersion("23.8.0-")
+	opVer23_10, _   = version.NewVersion("23.10.0-")
 	OpVer23_10_3, _ = version.NewVersion("23.10.3-")
 	opVer24_1_0, _  = version.NewVersion("24.1.0-")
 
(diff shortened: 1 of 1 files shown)


pkg/util/test/util.go
```diff
@@ -894,7 +899,7 @@ func validatePxPodsAnnotations(cluster *corev1.StorageCluster, timeout, interval
 			// This should validate the annotation cluster-autoscaler.kubernetes.io/safe-to-evict value,
 			// when node is Storage, annotation value should be set to false,
 			// when node is Storageless, annotation value should be set to true
-			if storageNode.Status.NodeAttributes.Storage != nil && *storageNode.Status.NodeAttributes.Storage != !safeToEvict {
+			if opVersion.GreaterThanOrEqual(OpVer23_10_3) && storageNode.Status.NodeAttributes.Storage != nil && *storageNode.Status.NodeAttributes.Storage != !safeToEvict {
```
nikolaypopov, line 902: This is not gonna work, we should either find a different way to validate this or skip the whole thing all together for old operator and just print something like PX Operator version is too old and we do not support validating this

nikolaypopov: Also next debug print after this statement will not be useful as the older PX Operators do not have `*storageNode.Status.NodeAttributes.Storage`

nilanto07-px: Addressed the comments.

nikolaypopov (approved):

pkg/util/test/util.go
```diff
@@ -867,6 +868,15 @@ func ValidateStorageCluster(
 
 func validatePxPodsAnnotations(cluster *corev1.StorageCluster, timeout, interval time.Duration) error {
 	logrus.Debug("Validate PX pods..")
+	opVersion, err := GetPxOperatorVersion()
+	if err != nil {
+		logrus.Errorf("Operator version not found!")
+		return err
+	}
+	if opVersion.GreaterThanOrEqual(opVer23_10) {
```
nikolaypopov, line 876: This is wrong statement, it needs to be lower then 23.10 and only then we skip it, 23.10.0 or higher is good and we should not skip it

nikolaypopov (changes requested):

pkg/util/test/util.go
nilanto07-px: Ah. Updated.

nilanto07-px:
https://jenkins.pwx.dev.purestorage.com/job/Torpedo/view/Torpedo%20NextPX/job/tp-ocp-op-px-upgrade/83/console 
Fixed worked. 
check around: 
`15:32:10 time="2024-04-16T10:03:16Z"
15:32:32 time="2024-04-16T10:03:39Z"`

nikolaypopov (approved):
LGTM

PR #1524, opened 2024-04-19
PTX-23462: Correcting PDB annotation in integration test

svijaykumar-px:

Closes #

### DIFF
--- a/test/integration_test/pdb_test.go
+++ b/test/integration_test/pdb_test.go
@@ -105,7 +105,9 @@ func OverridePDBUsingValidAnnotation(tc *types.TestCase) func(*testing.T) {
 		k8snodecount = k8snodecount - 1
 
 		// Override PDB value with (number of k8s nodes -2) to check if allowed disruptions is 2
-		cluster.Annotations = make(map[string]string)
+		if cluster.Annotations == nil {
+			cluster.Annotations = make(map[string]string)
+		}
 		logrus.Infof("Validating PDB using minAvailable value: %d", k8snodecount-2)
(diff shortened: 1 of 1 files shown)


test/integration_test/pdb_test.go
```diff
@@ -101,11 +101,13 @@ func OverridePDBUsingValidAnnotation(tc *types.TestCase) func(*testing.T) {
 
 		k8snodecount, err := ci_utils.GetK8sNodeCount()
 		require.NoError(t, err)
-		// Assuming 1 node is dedicated control plane node
-		k8snodecount = k8snodecount - 1
+		// Assuming 3 nodes are dedicated control plane nodes
```
nikolaypopov, line 104: We shouldn't assume there are 3 control plane nodes, why not just get the count of master nodes?

svijaykumar-px: Is there a way to get the actual count?

svijaykumar-px: @nikolaypopov made the requested changes and tested it, here is the link - https://jenkins.pwx.dev.purestorage.com/job/Operator/job/op-pdb-ocp/11/console

nikolaypopov: Yes, there is, you can loop through nodes and find which are the master nodes by using this `sched-ops` method https://github.com/portworx/sched-ops/blob/master/k8s/core/nodes.go#L151

svijaykumar-px: Yes I have used this same function. The run passed with these changes

nikolaypopov (approved):

PR #1530, opened 2024-04-23
PWX-36838 : Failed to validate PX node after StorageCluster update, Access denied without authentication token 

nikita-bhatia:
What this PR does / why we need it:
When security is enabled and guest access is disabled, operator fails to make API call to PX service to get KVDB members info with following error : Access denied without authentication token

As a fix, we are using context that uses token for authenticating the request with the SDK server.
Closes # https://purestorage.atlassian.net/browse/PWX-36838

### DIFF
--- a/drivers/storage/portworx/portworx.go
+++ b/drivers/storage/portworx/portworx.go
@@ -990,7 +990,9 @@ func (p *portworx) GetKVDBMembers(cluster *corev1.StorageCluster) (map[string]bo
 		return nil, fmt.Errorf("error in connecting to portworx sdk server: %s", err)
 	}
 	serviceClient := pxapi.NewPortworxServiceClient(p.sdkConn)
-	ctx, err := pxutil.SetupContextWithToken(context.Background(), cluster, p.k8sClient)
+	ctx, cancel := context.WithTimeout(context.Background(), 20*time.Second)
+	defer cancel()
+	ctx, err = pxutil.SetupContextWithToken(ctx, cluster, p.k8sClient)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


drivers/storage/portworx/portworx.go
```diff
@@ -990,8 +990,10 @@ func (p *portworx) GetKVDBMembers(cluster *corev1.StorageCluster) (map[string]bo
 		return nil, fmt.Errorf("error in connecting to portworx sdk server: %s", err)
 	}
 	serviceClient := pxapi.NewPortworxServiceClient(p.sdkConn)
-	ctx, cancel := context.WithTimeout(context.Background(), 20*time.Second)
-	defer cancel()
+	ctx, err := pxutil.SetupContextWithToken(context.Background(), cluster, p.k8sClient)
```
piyush-nimbalkar, line 993: Why are we removing the timeout context? We can still create that and pass it to SetupContextWithToken instead of passing a Background context.

drivers/storage/portworx/portworx_test.go
```diff
@@ -11701,6 +11701,105 @@ func TestGetKVDBMembers(t *testing.T) {
...
+		},
+		Status: corev1.StorageClusterStatus{
+			Phase: string(corev1.ClusterStateInit),
+		},
+		Spec: corev1.StorageClusterSpec{
+			Security: &corev1.SecuritySpec{
+				Enabled: true, // enable security
+				Auth: &corev1.AuthSpec{
+					GuestAccess: guestAccessTypePtr(corev1.GuestRoleDisabled), // guest access disabled
+				},
```
piyush-nimbalkar, line 11729: It doesn't matter if guest access is enabled or not. We are always using token if security is enabled.

nikita-bhatia: That's correct, the test would still remain same. Added guest access to test with the requirement.

drivers/storage/portworx/portworx.go
nikita-bhatia: Passing the context to create token only if security is enabled.

svijaykumar-px (approved):

PR #1533, opened 2024-04-30
PWX-34271: live migrate KubeVirt VMs before deleting the pod during PX rolling update

pureneelesh:

    In order to avoid downtime for KubeVirt VMs during PX rolling update, the operator now
    evicts the VMs before deleting PX pod. Live migration is used for VM eviction.

    PX pod is deleted only when there are no virt-launcher pods running on the node.

    There is a chance that a VM could land on the node after operator has checked for the VMs
    and proceeded with the deletion of PX pod. Operator adds "unschedulable" annotation on the
    node to tell Stork that it should not schedule pods on the node. However, Stork changes
    are needed for the annotation to take effect.

    We ignore the VMs that are not live migratable. If VM eviction fails, PX upadte is paused
    on the node and an event is generated.

    Signed-off-by: Neelesh Thakur <neelesh.thakur@purestorage.com>

PWX-34271


pp511:
Need to take a look at the test code. Will do it in another pass.

pp511 (approved):
lgtm

PR #1542, opened 2024-05-09
PWX-37204 , PWX- 37178 : Resolving gaps in KVDB fix

svijaykumar-px:
**What this PR does / why we need it**: This PR contains the fixes for 2 tickets : 

1. [PWX-37178](https://purestorage.atlassian.net/browse/PWX-37178) - Ensuring that atleast 3 kvdb nodes are present to upgrade those portworx nodes which have internal kvdb running in it. This is necessary as if a kvdb node is down for more than 2 minutes it gets removed from the list of kvdb nodes thus leading to confusion that kvdb cluster is working correctly. Adding extra check to ensure 3 nodes are present along with all 3 being healthy to upgrade kvdb nodes.
2. [PWX-37204](https://purestorage.atlassian.net/browse/PWX-37204) - When portworx is installed freshly, the portworx sdk server is not up for a while. This causes the getKVDBNodeAvailability function to return error and pause the rolling update and make cluster to enter degraded state till portworx sdk starts. The fix ensures that this error is ignored in the case of fresh install of portworx .

Closes # PWX-37204, PWX-37178

**Testing notes**: 
Tested fresh install, cluster no longer goes to degraded state
Tested with 5 node cluster, kvdb down on one node and wait for 3 minutes, then start the upgrade. The other 2 kvdb nodes do not get upgraded


pkg/controller/storagecluster/update.go
```diff
@@ -82,7 +82,7 @@ func (c *Controller) rollingUpdate(cluster *corev1.StorageCluster, hash string)
 
 	// get the number of kvdb members which are unavailable in case of internal kvdb
 	numUnavailableKvdb, kvdbNodes, err := c.getKVDBNodeAvailability(cluster)
```
nikita-bhatia, line 84: do we even need to check KVDB node's availability in case of fresh install then?

pkg/controller/storagecluster/update.go
```diff
@@ -94,9 +94,10 @@ func (c *Controller) rollingUpdate(cluster *corev1.StorageCluster, hash string)
...
 
 		// check if pod is running in a node which has internal kvdb running in it
-		if _, isKvdbNode := kvdbNodes[pod.Spec.NodeName]; cluster.Spec.Kvdb != nil && cluster.Spec.Kvdb.Internal && isKvdbNode {
-			// if number of unavailable kvdb nodes is greater than or equal to 1, then dont restart portworx on this node
-			if numUnavailableKvdb > 0 {
+		if _, isKvdbNode := kvdbNodes[pod.Spec.NodeName]; err == nil && cluster.Spec.Kvdb != nil && cluster.Spec.Kvdb.Internal && isKvdbNode {
+			// if number of unavailable kvdb nodes is greater than or equal to 1, or lesser than 3 entries are present in the kvdb map
+			// then dont restart portworx on this node
+			if numUnavailableKvdb > 0 || len(kvdbNodes) < 3 {
 				logrus.Infof("Number of unavaliable KVDB members exceeds 1, temporarily skipping update for this node to prevent KVDB from going out of quorum ")
```
nikita-bhatia, line 101: update the comment here based on which usecase it is, numUnavailableKvdb > 0 or len(kvdbNodes) < 3

nikita-bhatia (approved):
LGTM

PR #1544, opened 2024-05-11
PWX-34271: pick the pods on unschedulable nodes for update first

pureneelesh:

If we have marked the node unschedulable in the previous Reconcile(), we have to pick that node first. Otherwise, we may end up making too many nodes unschedulable.

Also, retry when add/remove of node label fails with conflict error so we don't wait for another 30 seconds.

PWX-34271

### DIFF
--- a/pkg/controller/storagecluster/update.go
+++ b/pkg/controller/storagecluster/update.go
@@ -51,7 +51,7 @@ import (
 
 // rollingUpdate deletes old storage cluster pods making sure that no more than
 // cluster.Spec.UpdateStrategy.RollingUpdate.MaxUnavailable pods are unavailable
-func (c *Controller) rollingUpdate(cluster *corev1.StorageCluster, hash string) error {
+func (c *Controller) rollingUpdate(cluster *corev1.StorageCluster, hash string, nodeList *v1.NodeList) error {
 	logrus.Debug("Perform rolling update")
 	nodeToStoragePods, err := c.getNodeToStoragePods(cluster)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/controller/storagecluster/update.go
```diff
@@ -213,15 +227,30 @@ func (c *Controller) rollingUpdate(cluster *corev1.StorageCluster, hash string)
...
-	return storkUnschedulableAnnotationHelper(c.client, nodeName, false)
+func (c *Controller) removeNodeUnschedulableLabel(nodeName string) error {
+	return nodeUnschedulableAnnotationHelperWithRetries(c.client, nodeName, false)
+}
+
+func nodeUnschedulableAnnotationHelperWithRetries(
+	k8sClient client.Client, nodeName string, unschedulable bool,
+) error {
+	var err error
+	for i := 0; i < 5; i++ {
```
pp511, line 242: nit..We could define a global var for num of retries here

pkg/controller/storagecluster/update.go
```diff
@@ -233,28 +262,28 @@ func storkUnschedulableAnnotationHelper(k8sClient client.Client, nodeName string
...
 	}
 	if !needsUpdate {
 		return nil
 	}
 	if unschedulable {
-		logrus.Infof("Adding annotation %s to node %s", constants.StorkAnnotationUnschedulable, nodeName)
+		logrus.Infof("Adding annotation %s to node %s", constants.AnnotationUnschedulable, nodeName)
 	} else {
-		logrus.Infof("Removing annotation %s from node %s", constants.StorkAnnotationUnschedulable, nodeName)
+		logrus.Infof("Removing annotation %s from node %s", constants.AnnotationUnschedulable, nodeName)
```
pp511, line 282: nit..We can consider moving these print statements inside the above if/else blocks. 
Logically it's correct since we are updating needsUpdate inside the if/else and returning if do no need update.
But from the readability standpoint, it's a little confusing as it suggests that we are printing "Adding annotation" if unschedulable is true(and the only condition) and vice versa which is not true.

pp511 (approved):

PR #1548, opened 2024-05-16
PWX-37274 : Adding events for setting PDB using annotation

svijaykumar-px:
**What this PR does / why we need it**: As suggested during the TOI presentation, we will add an event in the storagecluster when the user sets the value of minAvailable for px-storage PDB using annotation. An event will be logged if this value is valid, invalid, and everytime the annotated value is changed. It also gets raised once everytime an incorrect value is either changed to default or correct value is given. Here is the list of scenarios and its expected events raised

| Storage cluster state                                           | Event             |
|-----------------------------------------------------------------|-------------------|
| Fresh px install , default PDB                                  | No event          |
| Fresh px install, correct PDB annotation                        | Event raised once |
| Fresh install, incorrect PDB annotation                         | Event raised once |
| Using default PDB - &gt; override with correct PDB              | Event raised once |
| Using default PDB -&gt; Override with incorrect PDB             | Event raised once |
| Using correct PDB annotation -&gt; Remove annotation            | Event raised once |
| Using incorrect PDB annotation -&gt; Remove annotation          | Event raised once |
| Using correct PDB annotation-&gt; Override incorrect annotation | Event raised once |
| Using incorrect annotation -&gt; Override correct annotation    | Event raised once |

Closes # PWX-37274

### DIFF
--- a/drivers/storage/portworx/component/disruption_budget.go
+++ b/drivers/storage/portworx/component/disruption_budget.go
@@ -134,7 +134,7 @@ func (c *disruptionBudget) createPortworxPodDisruptionBudget(
 
 		// If the user provided value is invalid and it is different from the previously given value, raise an event in storagecluster
 		if cluster.Annotations[pxutil.AnnotationStoragePodDisruptionBudget] != "" && userProvidedMinValue != c.annotatedMinAvailable {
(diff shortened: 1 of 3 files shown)


nikita-bhatia:
@svijaykumar-px can you please add UTs?

svijaykumar-px:
> @svijaykumar-px can you please add UTs?

The UTs for this change would just be the existing ones with checks for events in the recorder.  I have added them

nikita-bhatia (approved):
LGTM

PR #1552, opened 2024-05-21
PWX-37266 : Upgrade should not continue when PDB value is set to 1 and there is already a PX node down

nikita-bhatia:
**What this PR does / why we need it:**
The PDB value for px-storage is not changing if PX service is stopped on a node by labelling with px/service=stop label.

Reason :
This was happening because when we label our node with px/service=stop , the label storage=true was getting removed from PX pods, since to decide whether node can be storagenode or not, operator checks only status of the storagenode and not rely on any parameter.
The PR fixes this behaviour and uses NonQuorumMember flag to decide storage/storageless node.

- a node will be considered storage node if we PX version is >= 3.1.0 , and the node is a quorum member after Node is not in Initializing state
- a node will be considered storageless node if we PX version is >= 3.1.0 , and the node is not a quorum member after Node is not in Initializing state
- a node will be considered storage node if we PX version is < 3.1.0 and nodepools are not empty after Node is not in Initializing state
- a node will be considered storage node if we PX version is < 3.1.0 and nodepools are empty after Node is not in Initializing state

**Which issue(s) this PR fixes (optional)**
Closes # PWX-37266

**Testing Done:**

- Fresh install and upgrade PX
- labeling node with px/service= stop and start
- Install with Security enabled/disabled

### DIFF
--- a/pkg/controller/storagenode/storagenode.go
+++ b/pkg/controller/storagenode/storagenode.go
@@ -528,7 +528,7 @@ func (c *Controller) isStorageNode(storageNode *corev1.StorageNode, cluster *cor
 			} else {
 				isQuorumMember = len(n.Pools) > 0 && n.Pools[0] != nil
 			}
-			logrus.Infof("shouldUseQuorumMembe for node %v: %v , isQuorumMember? %v", shouldUseQuorumMember, storageNode.Name, isQuorumMember)
+			logrus.Infof("shouldUseQuorumMembe for node %v: %v , isQuorumMember? %v", storageNode.Name, shouldUseQuorumMember, isQuorumMember)
 		}
 	}
 
--- a/pkg/controller/storagenode/storagenode.go
+++ b/pkg/controller/storagenode/storagenode.go
@@ -509,7 +509,13 @@ func (c *Controller) isStorageNode(storageNode *corev1.StorageNode, cluster *cor
 	}
 
 	nodeClient := api.NewOpenStorageNodeClient(c.sdkConn)
-	nodeResp, err := nodeClient.EnumerateWithFilters(context.Background(), &api.SdkNodeEnumerateWithFiltersRequest{})
+	ctx, err := pxutil.SetupContextWithToken(context.Background(), cluster, c.client, false)
+	if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/controller/storagenode/storagenode.go
```diff
@@ -474,10 +479,10 @@ func canNodeServeStorage(storagenode *corev1.StorageNode) bool {
...
 		if cond.Type == corev1.NodeStateCondition {
-			if cond.Status == corev1.NodeOnlineStatus ||
-				cond.Status == corev1.NodeMaintenanceStatus ||
-				cond.Status == corev1.NodeDegradedStatus {
-				return true
+			if cond.Status != corev1.NodeInitStatus {
+				if ok, err := c.isStorageNode(storagenode, cluster); err == nil {
+					return ok
+				}
 			}
```
svijaykumar-px, line 486: Previously we were assuming that a node in maintenance state as well as degraded state can serve storage, but that will not happen now right? What will be the impact of this?
Will is be an issue if node goes into maintenance and storage label will be removed from it

nikita-bhatia: @svijaykumar-px now we are assuming a node in any state apart from INITIALIZING can serve storage, since storage node can also be in any state.
CC : @vinayakshnd @adityadani @harsh-px @pp511

vinayakshnd: yes @nikita-bhatia we should  consider nodes in maintainence and degreaded state to be part of cluster quorum. If OpenStorage Enumerate node API is returning list of nodes as part of quorum even if those nodes are in maintenance mode or in degraded state then we are good here.

nikita-bhatia: That's correct. Thanls for the confirmation @vinayakshnd

svijaykumar-px (approved):

nikita-bhatia:
> I'm not the expert in this code (@piyush-nimbalkar is) -- but after skimming through the content, it looks OK to me (couldn't spot any issues).
> 
> Since @piyush-nimbalkar probably won't be able to review this in time -- let me unblock this PR.

Thanks @zoxpx for approving

PR #1554, opened 2024-05-22
PTX-23958: Check custom secure port in k3s for pvcController in tests

svijaykumar-px:
**What this PR does / why we need it**: This is a fix for the integration tests where a custom secure port is checked for pvc controller as well for k3s clusters

Closes # PTX-23958

### DIFF
--- a/pkg/util/test/util.go
+++ b/pkg/util/test/util.go
@@ -3155,7 +3155,7 @@ func validatePvcControllerPorts(cluster *corev1.StorageCluster, pvcControllerDep
 						for _, containerCommand := range container.Command {
 							if strings.Contains(containerCommand, "--secure-port") {
 								if len(pvcSecurePort) == 0 {
-									if isAKS(cluster) {
+									if isAKS(cluster) || IsK3sCluster() {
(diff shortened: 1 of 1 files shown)


nikita-bhatia (approved):
LGTM

pkg/util/test/util.go
```diff
@@ -3155,7 +3155,7 @@ func validatePvcControllerPorts(cluster *corev1.StorageCluster, pvcControllerDep
 						for _, containerCommand := range container.Command {
 							if strings.Contains(containerCommand, "--secure-port") {
 								if len(pvcSecurePort) == 0 {
-									if isAKS(cluster) {
+									if isAKS(cluster) || IsK3sCluster() {
 										if strings.Split(containerCommand, "=")[1] != AksPVCControllerSecurePort {
```
nikolaypopov, line 3159: Now that it seems this is behavior not only for AKS, can we rename `AksPVCControllerSecurePort` to something like `defaultPvcControllerSecurePort` instead please

nikolaypopov (approved):

pkg/util/test/util.go
svijaykumar-px: Sure I will make the change

PR #1555, opened 2024-05-23
PWX-37366: Add function to check if k3s deployment or rke2

svijaykumar-px:
**What this PR does / why we need it**: Similar to [#1510](https://github.com/libopenstorage/operator/pull/1510) , same issue is seen in rke2 cluster. Solution is to check if cluster is k3s or rke2 and add a custom pvccontroller port

Closes # PWX-37366


nikita-bhatia (approved):
LGTM

PR #1561, opened 2024-05-31
[cherry-pick] PWX-37480: avoid starting extra live-migraions for the same VM (#1560)

pureneelesh:

When deciding whether a running virt-launcher pod needs to be evicted, we now also check if the VMI is active on the same node. If VMI says that it is running on a different node and there is no live-migration in progress, we skip the eviction for that pod since the pod should go into a completed state on its own.

Also, we now call shouldEvictPod only for the virt-launcher pods on the subset of nodes due for the update in the current Reconcile() iteration to avoid invoking k8s APIs unnecessarily.

When starting live-migration, we now have an additional check to see if the live-migration succeeded for the same VM in the same upgrade cycle. If yes, we don't start an additional live-migration and generate an event instead. We were already doing this for a failed migration.

PWX-37480


pp511 (approved):
lgtm

PR #1562, opened 2024-06-05
PWX-35416 : Fail components when they are not found in the versions configmap

svijaykumar-px:
**What this PR does / why we need it**: After [PWX-34825](https://purestorage.atlassian.net/browse/PWX-34825) , storagecluster fails to install if versions configmap is not found and endpoint is not reachable. But it till used defaults for images that were not found in versions cm. This PR aims to change this behavior and henceforth any image that is not present in the cm will be failed with a warning in the stc.

Closes # PWX-35416


nikita-bhatia (approved):
LGTM

PR #1563, opened 2024-06-05
PWX-37573: tighten the checks for avoiding unnecessary VM live migration

pureneelesh:

If the live-migration is in progress for a VM that we want to evict, do not create another live migration for that VM in the same Reconcile() cycle.

Just before starting a live-migration, check the VMI one more time to verify that the VMI is pointing to the same node as the pod being evicted.

These two extra checks reduce the window in which PX and operator may try to live-migrate the same VM out of the same node. Also, it handles any other unexpected live-migrations that might start.

PWX-37573

### DIFF
--- a/pkg/controller/storagecluster/controller_test.go
+++ b/pkg/controller/storagecluster/controller_test.go
@@ -2695,9 +2695,10 @@ func TestKubevirtVMsDuringUpgrade(t *testing.T) {
 		k8sNodes[1].Name: true,
 		k8sNodes[2].Name: true,
 	}
+	evictions := getVMPodEvictions(t, vmPods)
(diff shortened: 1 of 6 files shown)


pp511 (approved):
Just one nit.
Also Maybe VMPodToBeEvicted, VMPodToEvict are better alternatives to VMPodEvictions ? But I am ok to keep it as it is too.

pureneelesh:
> Just one nit. Also Maybe VMPodToBeEvicted, VMPodToEvict are better alternatives to VMPodEvictions ? But I am ok to keep it as it is too.

Leaving this as it is since we may want to add additional eviction-related fields to that structure in future.

PR #1564, opened 2024-06-06
StorageCluster schema changes to support parallel portworx upgrades

hitesh-wani-px:
This PR will add a new field to control the PX upgrade behavior. The default behavior would be to have a non-disruptive upgrade. But users should be able to disable that and go back to the old behavior of upgrading nodes in parallel without worrying about the disruption. 

Closes # https://purestorage.atlassian.net/browse/PWX-36509

### DIFF
--- a/pkg/apis/core/v1/storagecluster.go
+++ b/pkg/apis/core/v1/storagecluster.go
@@ -313,11 +313,11 @@ type RollingUpdateStorageCluster struct {
 	// The default behavior is non-disruptive upgrades. This setting disables the default
 	// non-disruptive upgrades and reverts to the previous behavior of upgrading nodes in
 	// parallel without worrying about disruption.
-	Disruption DisruptionSpec `json:"disruption,omitempty"`
+	Disruption Disruption `json:"disruption,omitempty"`
 }
 
-// DisruptionSpec contains disruption details
-type DisruptionSpec struct {
+// Disruption contains configuration for disruption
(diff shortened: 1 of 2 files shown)


pkg/apis/core/v1/storagecluster.go
```diff
@@ -309,6 +309,17 @@ type RollingUpdateStorageCluster struct {
...
 	MinReadySeconds int32 `json:"minReadySeconds,omitempty"`
+
+	// The default behavior is non-disruptive upgrades. This setting disables the default
+	// non-disruptive upgrades and reverts to the previous behavior of upgrading nodes in
+	// parallel without worrying about disruption.
+	Disruption DisruptionSpec `json:"disruption,omitempty"`
+}
+
+// DisruptionSpec contains disruption details
+type DisruptionSpec struct {
```
nikita-bhatia, line 320: let's change this to Disruption instead of DisruptionSpec as we just have single field allow for that

hitesh-wani-px: Done @nikita-bhatia

nikita-bhatia (approved):
LGTM, please cherrypick the change to feature release branch **PWX-25414-release-24.2.0**

hitesh-wani-px:
> LGTM, please cherrypick the change to feature release branch **PWX-25414-release-24.2.0**

okay I will create cherry-pick PR for this.

PR #1571, opened 2024-06-13
PWX-37620 Vendor in latest sched ops

ssz1997:
Operator needs to create service account token with the help from sched-ops. Thus vendor in the latest sched ops.
Corresponding sched PR: https://github.com/portworx/sched-ops/pull/394


pp511 (approved):
lgtm

pureneelesh (approved):

PR #1576, opened 2024-06-16
[cherry-pick] PWX-36509 : StorageCluster schema changes to support parallel portworx upgrades

hitesh-wani-px:
This PR will add a new field to control the PX upgrade behavior. The default behavior would be to have a non-disruptive upgrade. But users should be able to disable that and go back to the old behavior of upgrading nodes in parallel without worrying about the disruption. 

Closes # https://purestorage.atlassian.net/browse/PWX-36509


svijaykumar-px (approved):

svijaykumar-px (commented):
LGTM

PR #1582, opened 2024-06-17
[PWX-37620] Create/Refresh token and secret and mount to oci-mon pod

ssz1997:
1. Create service account token and put into a secret
2. Refresh the token in the secret every 6 hours
3. Mount the secret into oci-monitor pod.

Oci-monitor pod will mount the secret to portworx runc container for it to communicate with k8s api server. This token/secret is still valid when oci-monitor pod is down and default token is invalidate during k8s node drain. 

For some reason the `createToken()` doesn't work with `fake.Clientset` - it errors out saying the serviceaccount "" does not exist, even though it is correctly passed in. Therefore I mock the CreateToken() with schedOps, which embeds the `fake.Clientset`so maintaining exists tests can be easier.


ssz1997:
Test plan:
1. Update `TestBasicComponentsInstall` to make sure the secret is created during cluster creation
2. Update 'TestBasicInstallWithPortworxDisabled' to make sure the secret is not created because the serviceaccount is not created.
3. Update `TestPortworxWithCustomServiceAccount` to check the secret is created
4. Update `TestDisablePortworx` to make sure the secret is deleted after portworx is disabled because the serviceaccount is deleted.
5. Add `TestServiceAccountTokenRefreshOnExpire` by creating the secret, monkey patch the `time.Now()` method to be after expiry time, and verify the token in the secret is refreshed. 
6. Add `TestPxAPIDaemonSetTemplate` to check the secret is correctly mounted into the oci-monitor pod by checking the daemonset config.

pureneelesh:
@nikita-bhatia / @svijaykumar-px : can one or both of you take a look at this PR?

drivers/storage/portworx/component/portworx_basic.go
```diff
@@ -110,6 +123,9 @@ func (c *portworxBasic) Delete(cluster *corev1.StorageCluster) error {
 	if err := k8sutil.DeleteService(c.k8sClient, pxutil.PortworxKVDBServiceName, cluster.Namespace, *ownerRef); err != nil {
 		return err
 	}
+	if err := k8sutil.DeleteSecret(c.k8sClient, pxutil.PortworxServiceAccountTokenSecretName, cluster.Namespace, *ownerRef); err != nil {
```
pureneelesh, line 126: should we ignore "not found" error?

drivers/storage/portworx/component/portworx_basic.go
```diff
@@ -521,6 +537,118 @@ func getPortworxServiceSpec(
 	return newService
 }
 
+func (c *portworxBasic) createAndMaintainPxSaTokenSecret(cluster *corev1.StorageCluster, ownerRef *metav1.OwnerReference) error {
+	if err := c.createTokenSecretIfNotExist(cluster, ownerRef); err != nil {
+		return err
+	}
+	if err := c.maintainTokenSecret(cluster, ownerRef); err != nil {
+		return fmt.Errorf("failed to maintain the token secret for px container. %v", err)
```
pureneelesh, line 545: with fmt.Errorf, use `": %w", err`

drivers/storage/portworx/component/portworx_basic.go
```diff
@@ -521,6 +537,118 @@ func getPortworxServiceSpec(
...
+			Name:      pxutil.PortworxServiceAccountTokenSecretName,
+			Namespace: cluster.Namespace,
+		}, secret)
+	if err != nil {
+		if errors.IsNotFound(err) || len(secret.Data) == 0 || len(secret.Data[v1.ServiceAccountTokenKey]) == 0 {
+			token, err := generateToken(cluster)
+			if err != nil {
+				return err
+			}
+			rootCaCrt, _ := os.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/ca.crt")
```
pureneelesh, line 563: don't ignore error

drivers/storage/portworx/component/portworx_basic.go
```diff
@@ -521,6 +537,118 @@ func getPortworxServiceSpec(
...
+			curTime := time.Now()
```
pureneelesh, line 564: nit: curTime is not needed. Use time.Now() directly?

drivers/storage/portworx/component/portworx_basic.go
```diff
@@ -521,6 +537,118 @@ func getPortworxServiceSpec(
...
+			tokenRefreshTimeBytes, err := curTime.Add(time.Duration(tokenExpirationSeconds / 2)).MarshalBinary()
```
pureneelesh, line 565: need to multiply tokeExpirationSeconds by time.Second?

drivers/storage/portworx/component/portworx_basic.go
```diff
@@ -521,6 +537,118 @@ func getPortworxServiceSpec(
...
+			if err != nil {
+				return fmt.Errorf("error marshalling current time to bytes. %v", err)
```
pureneelesh, line 567: check the usage of %v in fmt.Errorf everywhere

drivers/storage/portworx/component/portworx_basic.go
```diff
@@ -521,6 +537,118 @@ func getPortworxServiceSpec(
...
+			}
+			secret := &v1.Secret{
+				ObjectMeta: metav1.ObjectMeta{
```
pureneelesh, line 570: we should add "created by operator" label/annotation. Check other objects that operator creates.

drivers/storage/portworx/component/portworx_basic.go
```diff
@@ -521,6 +537,118 @@ func getPortworxServiceSpec(
...
+
+func (c *portworxBasic) maintainTokenSecret(cluster *corev1.StorageCluster, ownerRef *metav1.OwnerReference) error {
+	secret := &v1.Secret{}
+	err := c.k8sClient.Get(context.TODO(),
+		types.NamespacedName{
+			Name:      pxutil.PortworxServiceAccountTokenSecretName,
+			Namespace: cluster.Namespace,
+		}, secret)
+	if err != nil {
+		return fmt.Errorf("error getting the secret containing token for px container. %v", err)
```
pureneelesh, line 601: we should probably handle accidental or intentional deletion of the token as well. Maybe "maintain" function should be stateless and do everything (create if not exist, update if stale)? We should probably create an empty secret and then use "refresh" logic to populate it. That way, "refresh" logic gets exercised every time we create the token giving us more coverage from the existing tests. Also, we can avoid some code duplication seen below.

pureneelesh (commented):
Have not reviewed the tests yet. Will do that next.

drivers/storage/portworx/component/portworx_basic.go
ssz1997: I think `ownerReference` serves the purpose?

drivers/storage/portworx/component/portworx_basic.go
ssz1997: How shall we handle the error if the file doesn't exist? If it's in an actual k8s cluster, the file is supposed to be there because it's part of the service account secret. But it doesn't exist in UTs and breaking all the tests

pureneelesh: If the file does not exist, we don't need to add it to the token. But what if it is some other error? Should we handle that?

drivers/storage/portworx/component/portworx_basic.go
pureneelesh: Yes, but I thought we added the label/annotation as well. I have asked this question on https://github.com/libopenstorage/operator/pull/1585

drivers/storage/portworx/components_test.go
```diff
@@ -50,11 +52,41 @@ import (
...
+			},
+		}, nil).
+		AnyTimes()
+	mockCoreOps.EXPECT().
+		GetVersion().
+		Return(clientset.Discovery().ServerVersion()).
+		AnyTimes()
+	mockCoreOps.EXPECT().
+		ResourceExists(gomock.Any()).
+		Return(simpleClientset.ResourceExists(schema.GroupVersionKind{
```
pureneelesh, line 82: This is fine. But just curious. If this test file was calling more k8s functions, we would have to write an EXPECT() for each of them to redirect that call to simpleClientSet, right?

drivers/storage/portworx/components_test.go
```diff
@@ -321,6 +354,19 @@ func TestBasicComponentsInstall(t *testing.T) {
...
+	expectedPXSaTokenSecret := testutil.GetExpectedSecret(t, "portworxServiceAccountTokenSecret.yaml")
+	saTokenSecret := &v1.Secret{}
+	err = testutil.Get(k8sClient, saTokenSecret, pxutil.PortworxServiceAccountTokenSecretName, cluster.Namespace)
+	require.NoError(t, err)
+	fmt.Println(saTokenSecret)
+	require.Equal(t, expectedPXSaTokenSecret.Name, saTokenSecret.Name)
+	require.Equal(t, expectedPXSaTokenSecret.Namespace, saTokenSecret.Namespace)
+	require.Len(t, saTokenSecret.OwnerReferences, 1)
+	require.Equal(t, cluster.Name, saTokenSecret.OwnerReferences[0].Name)
+	require.Equal(t, len(expectedPXSaTokenSecret.Data), len(saTokenSecret.Data))
```
pureneelesh, line 367: we could also return a specific token from the mock and verify that it is set in the secret here?

pureneelesh (approved):

drivers/storage/portworx/components_test.go
```diff
@@ -105,16 +137,15 @@ func TestOrderOfComponents(t *testing.T) {
 	require.Equal(t, int32(0), components[1].Priority())
 	require.Equal(t, int32(0), components[2].Priority())
 	require.Equal(t, int32(0), components[3].Priority())
-	for _, comp := range components[4:] {
-		require.Equal(t, component.DefaultComponentPriority, comp.Priority())
-	}
```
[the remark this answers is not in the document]
ssz1997: Oops unintentional

drivers/storage/portworx/components_test.go
```diff
@@ -321,6 +354,19 @@ func TestBasicComponentsInstall(t *testing.T) {
 	require.Equal(t, expectedPXService.Labels, pxService.Labels)
 	require.Equal(t, expectedPXService.Spec, pxService.Spec)
 
+	// Portworx Service Account Token Secret
+	expectedPXSaTokenSecret := testutil.GetExpectedSecret(t, "portworxServiceAccountTokenSecret.yaml")
+	saTokenSecret := &v1.Secret{}
+	err = testutil.Get(k8sClient, saTokenSecret, pxutil.PortworxServiceAccountTokenSecretName, cluster.Namespace)
+	require.NoError(t, err)
+	fmt.Println(saTokenSecret)
```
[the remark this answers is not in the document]
ssz1997: Actually we don't need it

drivers/storage/portworx/components_test.go
ssz1997: Yes. This is done in the file `status_test.go`

drivers/storage/portworx/component/portworx_basic.go
```diff
@@ -521,6 +537,102 @@ func getPortworxServiceSpec(
...
+	return secret, nil
+}
+
+func (c *portworxBasic) refreshTokenSecret(secret *v1.Secret, cluster *corev1.StorageCluster, ownerRef *metav1.OwnerReference) error {
+	newToken, err := generatePxSaToken(cluster)
+	if err != nil {
+		return err
+	}
+	secret.Data[core.ServiceAccountTokenKey] = newToken
+	secret.StringData[PxSaTokenRefreshTimeKey] = time.Now().Add(time.Duration(pxSaTokenExpirationSeconds/2) * time.Second).Format(time.RFC3339)
```
pureneelesh, line 599: Use `time.Now().UTC().Add...` here to stay away from the timezone issues.

drivers/storage/portworx/util/util.go
```diff
@@ -179,6 +181,9 @@ const (
 	// EnvKeyPortworxServiceAccount key for the env var which tells custom Portworx
 	// service account
 	EnvKeyPortworxServiceAccount = "PX_SERVICE_ACCOUNT"
+	// EnvKeyPortworxServiceAccountTokenRefreshIntervalHours key for the env var which tells how often
+	// the token associated with the Px service account gets refreshed
+	EnvKeyPortworxServiceAccountTokenRefreshIntervalHours = "PX_SERVICE_ACCOUNT_TOKEN_REFRESH_INTERVAL_HOURS"
```
ssz1997, line 186: Putting "hours" here limits the interval to be at least 1 hour. But it's also probably not a good idea for user to put 30 seconds here I assume? What would be a lower bound of this interval?

pureneelesh: Can we use "Minutes"? I was hoping to use this setting from a functional test to configure lower during for testing.

ssz1997: Oh yeah right

pureneelesh (approved):

ssz1997:
@pureneelesh Any blocker for this PR? Otherwise maybe we can merge this?

pureneelesh:
> @pureneelesh Any blocker for this PR? Otherwise maybe we can merge this?

We can merge it. Thanks @ssz1997

PR #1585, opened 2024-06-18
PWX-37681 : Use service account for prometheus endpoint authentication for OCP 4.16

nikita-bhatia:
On OCP 4.16, the thanos-ruler-token secret, which was used to fetch ca.crt and token for prometheus endpoint authentication is removed and the authentication is current user's authentication token is used for the same.
From operator 24.1.1 and OCP 4.16, operator will not use Openshift's secret, rather do following to authenticater Openshift prometheus endopoint.
1. Create **autopilot-prometheus** service account
2.  Create **autopilot-promethues-binding** cluster role binding to created service account with **cluster-monitoring-view** cluster role.
3.  In **autopilot-prometheus-auth** secret, which is already created today, add following to use the secret's ca.crt and token with above permissions.
```
metadata:  
  annotations:
      kubernetes.io/service-account.name: autopilot-prometheus
 type: kubernetes.io/service-account-token
```
4. The token will expire every 30 days
5. Till token expiry time is reached half way, the token is refreshed and autopilot pod is killed to restart new pod with updated token

- The above mentioned steps are done only for OCP 4.16+ versions.
- For 4.15 and lesser version, no change will be there 

Closes # https://purestorage.atlassian.net/browse/PWX-37681

**Testing done**:
- Tested on OCP 4.16.0-rc.4 build, autopilot is successfully able to register to prometheus endpoint

### DIFF
--- a/drivers/storage/portworx/component/autopilot.go
+++ b/drivers/storage/portworx/component/autopilot.go
@@ -199,7 +199,7 @@ func (c *autopilot) Reconcile(cluster *corev1.StorageCluster) error {
 				if err := c.createServiceAccountForOCP(cluster.Namespace, ownerRef); err != nil {
 					return err
 				}
-				if err := c.createClusterRoleBindingForOCP(cluster.Namespace, ownerRef); err != nil {
+				if err := c.createClusterRoleBindingForOCP(cluster.Namespace); err != nil {
 					return err
 				}
 			}
@@ -325,7 +325,7 @@ func (c *autopilot) createServiceAccountForOCP(namespace string, ownerRef *metav
 // createClusterRoleBindingForOCP creates cluster role binding for OCP Prometheus
 // autopilot to access Prometheus metrics requires cluster-monitoring-view role
 // This is required for OCP 4.16 and above
-func (c *autopilot) createClusterRoleBindingForOCP(namespace string, ownerRef *metav1.OwnerReference) error {
+func (c *autopilot) createClusterRoleBindingForOCP(namespace string) error {
 	return k8sutil.CreateOrUpdateClusterRoleBinding(
 		c.k8sClient,
 		&rbacv1.ClusterRoleBinding{
@@ -363,61 +363,47 @@ func (c *autopilot) createSecret(clusterNamespace string, ownerRef *metav1.Owner
 		},
 	}
 
-	secretExist := true
-
 	if ocp416Plus {
-
 		err = k8sutil.GetSecret(c.k8sClient, AutopilotSecretName, clusterNamespace, secret)
 		if err != nil {
 			return fmt.Errorf("Error during getting secret %v ", err)
 		}
 
-		// if secret is created, check expiry of token
 		if secret.Data != nil {
-
-			if secret.Type != v1.SecretTypeServiceAccountToken {
+			// if secret exists, check secret type
+			if secret.Type == v1.SecretTypeServiceAccountToken {
+				// if secret type is service account token, check if token is expired
+				ok, err := isTokenRefreshRequired(secret)
+				if err != nil {
+					return fmt.Errorf("Error during checking token refresh %v ", err)
+				}
+				if ok {
+					// refresh the token if it is expired
+					return c.refreshTokenSecret(secret, clusterNamespace, ownerRef)
+				}
+			} else {
 				// if secret type is not service account token, delete the secret
 				// this is to handle first time creation of secret on OCP 4.16+
 				err := k8sutil.DeleteSecret(c.k8sClient, AutopilotSecretName, clusterNamespace, *ownerRef)
 				if err != nil {
 					return fmt.Errorf("Error during deleting secret %v ", err)
 				}
-				secretExist = false
 			}
-
-			if secretExist {
-				// check if token is expired
-				ok, err := isTokenRefreshRequired(secret)
-				if err != nil {
-					return fmt.Errorf("Error during checking token refresh %v ", err)
-				}
-				if ok {
-					// refresh the token
-					if err = c.refreshTokenSecret(secret, clusterNamespace, ownerRef); err != nil {
-						return fmt.Errorf("Error during refreshing token %v ", err)
-					}
-				}
-				return nil
-			}
-		} else {
-			secretExist = false
 		}
 
-		if !secretExist {
-			// if secret is not created, create the secret
-			// create secret with service account token
-			secret.Type = v1.SecretTypeServiceAccountToken
-			secret.Annotations = map[string]string{
-				"kubernetes.io/service-account.name": AutopilotPrometheusServiceAccountName,
-			}
-
-			token, err := generateAPSaToken(clusterNamespace, defaultAutoPilotSaTokenExpirationSeconds)
-			if err != nil {
-				return fmt.Errorf("Error during generating token %v ", err)
-			}
-			secret.Data = make(map[string][]byte)
-			secret.Data["token"] = token
+		// if secret is not created, create the secret
+		// create secret with service account token
+		secret.Type = v1.SecretTypeServiceAccountToken
+		secret.Annotations = map[string]string{
+			"kubernetes.io/service-account.name": AutopilotPrometheusServiceAccountName,
 		}
+
+		token, err := generateAPSaToken(clusterNamespace, defaultAutoPilotSaTokenExpirationSeconds)
+		if err != nil {
+			return fmt.Errorf("Error during generating token %v ", err)
+		}
+		secret.Data = make(map[string][]byte)
+		secret.Data["token"] = token
 
 	} else {
 		// OCP 4.15 and below, secret is created if user workload monitoring is enabled
--- a/drivers/storage/portworx/component/autopilot.go
+++ b/drivers/storage/portworx/component/autopilot.go
@@ -191,7 +191,7 @@ func (c *autopilot) Reconcile(cluster *corev1.StorageCluster) error {
 		ocp416plus, err := pxutil.IsSupportedOCPVersion(c.k8sClient, pxutil.Openshift_4_16_version)
 
 		if err != nil {
-			logrus.Errorf("error during checking OCP version %w ", err)
+			logrus.Errorf("error during checking OCP version %v ", err)
 		} else {
 			if ocp416plus {
 				// on OCP 4.16 and above, create service account and cluster role binding for OCP Prometheus by default
@@ -207,7 +207,7 @@ func (c *autopilot) Reconcile(cluster *corev1.StorageCluster) error {
 			if err := c.createSecret(cluster.Namespace, ownerRef, ocp416plus); err != nil {
 				// log the error and proceed for deployment creation
 				// if secret is created in next reconcilation loop successfully, deployment will be updated with volume mounts
-				logrus.Errorf("error during creating secret %w ", err)
+				logrus.Errorf("error during creating secret %v ", err)
 			}
 		}
 	}
--- a/drivers/storage/portworx/component/autopilot.go
+++ b/drivers/storage/portworx/component/autopilot.go
@@ -188,29 +188,30 @@ func (c *autopilot) Reconcile(cluster *corev1.StorageCluster) error {
 		return err
 	}
 	if c.isOCPUserWorkloadSupported() {
-		ok, err := pxutil.IsSupportedOCPVersion(c.k8sClient, pxutil.Openshift_4_16_version)
+		ocp416plus, err := pxutil.IsSupportedOCPVersion(c.k8sClient, pxutil.Openshift_4_16_version)
 
 		if err != nil {
 			logrus.Errorf("Error during checking OCP version %v ", err)
-		}
-
-		if err == nil && ok {
-			// on OCP 4.16 and above, create service account and cluster role binding for OCP Prometheus by default
-			// autopilot requires to access Prometheus metrics
-			if err := c.createServiceAccountForOCP(cluster.Namespace, ownerRef); err != nil {
-				return err
+		} else {
+			if ocp416plus {
+				// on OCP 4.16 and above, create service account and cluster role binding for OCP Prometheus by default
+				// autopilot requires to access Prometheus metrics
+				if err := c.createServiceAccountForOCP(cluster.Namespace, ownerRef); err != nil {
+					return err
+				}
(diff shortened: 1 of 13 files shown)


drivers/storage/portworx/component/autopilot.go
```diff
@@ -265,26 +291,109 @@ func (c *autopilot) createConfigMap(
...
+		// check if secret already exists, this is first time creation
+		// or secret is created in last reconcilation loop
+		err := k8sutil.GetSecret(c.k8sClient, secret.Name, secret.Namespace, secret)
+		if err != nil {
+			return err
+		}
+
+		// if secret already exists, return
+		// once secret is created, it should not be updated
+		// if secret is deleted, it will be created if operator is restarted
```
pureneelesh, line 368: why should we require operator restart when secret is deleted? Can we not just make this stateless and create the secret if it does not exist?

drivers/storage/portworx/component/autopilot.go
```diff
@@ -265,26 +291,109 @@ func (c *autopilot) createConfigMap(
...
+		if secret.Data != nil {
+			c.isSecretCreated = true
+			return nil
+		}
+
+		// create secret with service account token
```
pureneelesh, line 374: "create secret" -> "update secret" since the secret exists already

drivers/storage/portworx/component/autopilot.go
```diff
@@ -265,26 +291,109 @@ func (c *autopilot) createConfigMap(
...
+		secret.Type = v1.SecretTypeServiceAccountToken
```
pureneelesh, line 375: You can't change the secret type of the existing secret. I tried that and got this error
```
# * type: Invalid value: "kubernetes.io/service-account-token": field is immutable
```
I think it will be cleaner to just delete the existing secret and create a brand new one.

pkg/util/k8s/k8s.go
```diff
@@ -1992,6 +1992,31 @@ func CreateOrUpdateSecret(
...
+		context.TODO(),
+		types.NamespacedName{
+			Name:      name,
+			Namespace: namsespace,
+		},
+		secret,
+	)
+
+	if errors.IsNotFound(err) {
+		return nil
```
pureneelesh, line 2012: Is it a common practice in Operator to swallow NotFound error? I would find this very misleading. If GetSecret() returns err==nil, I would expect that it has found the secret I asked it to get.

pureneelesh:
Question for my own understanding: I thought that the operator adds "created by operator" label/annotation to all the objects that it creates. I don't see that being added to the secret. Is that not needed?

drivers/storage/portworx/component/autopilot.go
pureneelesh: IIUC, this will create a permanent token that is valid forever. Ideally, we should create a limited duration token and keep updating it before the token expires. In PR https://github.com/libopenstorage/operator/pull/1582 , we are using a CreateToken API to create an expiring token.

drivers/storage/portworx/component/autopilot.go
```diff
@@ -174,6 +181,18 @@ func (c *autopilot) Reconcile(cluster *corev1.StorageCluster) error {
 		return err
 	}
 	if c.isOCPUserWorkloadSupported() {
+		ok, err := pxutil.IsSupportedOCPVersion(c.k8sClient, pxutil.Openshift_4_16_version)
+		if err == nil && ok {
```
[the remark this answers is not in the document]
nikita-bhatia: We do not want to stop if err != nil, since deployment should still be created. Will add the error logging here.

drivers/storage/portworx/component/autopilot.go
nikita-bhatia: we can create the secret if it is deleted, for that we will have to make API call in every 15 seconds to kube-api to check the existence of the secret.

drivers/storage/portworx/component/autopilot.go
```diff
@@ -207,6 +226,12 @@ func (c *autopilot) Delete(cluster *corev1.StorageCluster) error {
 		if err := k8sutil.DeleteSecret(c.k8sClient, AutopilotSecretName, cluster.Namespace, *ownerRef); err != nil {
 			return err
 		}
+		if err := k8sutil.DeleteClusterRoleBinding(c.k8sClient, AutopilotClusterRoleBindingName); err != nil {
```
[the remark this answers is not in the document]
nikita-bhatia: Actually we are not using ownerref during clusterrolebinding creation, it is added in parameters by mistake. removed the instance from params.

drivers/storage/portworx/component/autopilot.go
```diff
@@ -265,26 +291,109 @@ func (c *autopilot) createConfigMap(
...
+	if ocp_416 {
+		// OCP 4.16 and above, secret is created by default if autopilot is enabled
+		// if secret is already created, return
+		if c.isCreated {
+			return nil
+		}
+
+		// check if secret already exists, this is first time creation
+		// or secret is created in last reconcilation loop
+		err := k8sutil.GetSecret(c.k8sClient, secret.Name, secret.Namespace, secret)
```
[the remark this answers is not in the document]
nikita-bhatia: Updated the code.
If secret doesn't exist : create
if secret exist -> check the type
    if type != kubernetes.io/service-account-token -> delete and create again
    if type ==  kubernetes.io/service-account-token -> refresh token if required

drivers/storage/portworx/component/autopilot.go
nikita-bhatia: This will be create only since this piece of code will run only if secret doesn't exist.

drivers/storage/portworx/component/autopilot.go
nikita-bhatia: updated

drivers/storage/portworx/component/autopilot.go
```diff
@@ -807,6 +959,51 @@ func (c *autopilot) isOCPUserWorkloadSupported() bool {
...
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func generateAPSaToken(clusterNamespace string, expirationSeconds int64) ([]byte, error) {
+	tokenRequest := &authv1.TokenRequest{
+		Spec: authv1.TokenRequestSpec{
+			ExpirationSeconds: &expirationSeconds,
```
pureneelesh, line 983: Kubernetes allows limiting this value using a param to kube-apiserver.
https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/
```
--service-account-max-token-expiration duration
 The maximum validity duration of a token created by the service account token issuer. If an otherwise valid TokenRequest with a validity duration larger than this value is requested, a token will be issued with a validity duration of this value.
```
We need to read the actual expiration duration from the tokenResp and set that in the secret instead of relying on the value that we requested.

drivers/storage/portworx/component/autopilot.go
```diff
@@ -265,26 +305,138 @@ func (c *autopilot) createConfigMap(
...
+			"kubernetes.io/service-account.name": AutopilotPrometheusServiceAccountName,
+		}
+
+		token, err := generateAPSaToken(clusterNamespace, defaultAutoPilotSaTokenExpirationSeconds)
+		if err != nil {
+			return fmt.Errorf("Error during generating token %v ", err)
+		}
+		secret.Data = make(map[string][]byte)
+		secret.Data["token"] = token
+		secret.Data[AutopilotSaTokenRefreshTimeKey] = []byte(time.Now().UTC().Add(time.Duration(defaultAutoPilotSaTokenExpirationSeconds/2) * time.Second).Format(time.RFC3339))
```
pureneelesh, line 421: we need to read the actual token expiration duration instead of relying on `defaultAutoPilotSaTokenExpirationSeconds`. See my comment below.

drivers/storage/portworx/component/autopilot.go
```diff
@@ -265,26 +305,138 @@ func (c *autopilot) createConfigMap(
...
+		secret.Annotations = map[string]string{
+			"kubernetes.io/service-account.name": AutopilotPrometheusServiceAccountName,
+		}
+
+		token, err := generateAPSaToken(clusterNamespace, defaultAutoPilotSaTokenExpirationSeconds)
+		if err != nil {
+			return fmt.Errorf("Error during generating token %v ", err)
+		}
+		secret.Data = make(map[string][]byte)
+		secret.Data["token"] = token
```
pureneelesh, line 420: do we need to add the `ca.crt` key as well?

ssz1997 (commented):
Left some small comments

drivers/storage/portworx/component/autopilot.go
```diff
@@ -265,26 +305,138 @@ func (c *autopilot) createConfigMap(
...
+				ok, err := isTokenRefreshRequired(secret)
+				if err != nil {
+					return fmt.Errorf("Error during checking token refresh %v ", err)
+				}
+				if ok {
+					// refresh the token if it is expired
+					logrus.Infof("refreshing token %s in namespace %s", AutopilotSecretName, clusterNamespace)
+					err := c.refreshTokenSecret(secret, clusterNamespace, ownerRef)
+					if err != nil {
+						return fmt.Errorf("Error during refreshing token %v ", err)
```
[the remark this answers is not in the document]
nikita-bhatia: updated all the fmts

drivers/storage/portworx/component/autopilot.go
```diff
@@ -265,26 +305,138 @@ func (c *autopilot) createConfigMap(
...
+					}
+
+					// delete autopilot deployment to sync with new token
+					logrus.Info("deleting autopilot pod to sync with new token")
+					err = k8sutil.DeletePodsByLabel(c.k8sClient, map[string]string{"name": "autopilot"}, clusterNamespace)
```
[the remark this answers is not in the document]
nikita-bhatia: No, the secret is not mounted inside the pod, instead it is used as volume mount for token and cert, so it doesn't update the token after it is refreshed. Hence this is required.

drivers/storage/portworx/testspec/autopilotPrometheusClusterRoleBinding.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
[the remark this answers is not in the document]
nikita-bhatia: The cluster role binding is created to provide cluster-monitoring-view role to the created service account, which is required for monitoring.

drivers/storage/portworx/component/autopilot.go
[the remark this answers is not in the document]
nikita-bhatia: This can be handled using **ProjectedVolumeSource** as ColumeSource in autopilot deployment, it will update the token inside the pod without restart.
@zoxpx does autopilot needs to do anything everytime token is updated inisde the pod and pod is not restarted?

drivers/storage/portworx/component/autopilot.go
nikita-bhatia: this is taken care with updated code

drivers/storage/portworx/component/autopilot.go
```diff
@@ -265,26 +305,138 @@ func (c *autopilot) createConfigMap(
...
+				logrus.Infof("Deleted secret to recreate %s in namespace %s", AutopilotSecretName, clusterNamespace)
+				if err != nil {
+					return fmt.Errorf("Error during deleting secret %v ", err)
+				}
+			}
+		}
+
+		// if secret is not created, create the secret
+		// create secret with service account token
+		secret.Type = v1.SecretTypeServiceAccountToken
```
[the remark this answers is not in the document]
nikita-bhatia: changed this to opaque type with using ca.crt of certificate mounted inside the pod.

drivers/storage/portworx/component/autopilot.go
```diff
@@ -174,12 +189,30 @@ func (c *autopilot) Reconcile(cluster *corev1.StorageCluster) error {
...
 	}
 	if c.isOCPUserWorkloadSupported() {
-		if err := c.createSecret(cluster.Namespace, ownerRef); err != nil {
-			// log the error and proceed for deployment creation
-			// if secret is created in next reconcilation loop successfully, deployment will be updated with volume mounts
-			logrus.Errorf("Error during creating secret %v ", err)
+		ocp416plus, err := pxutil.IsSupportedOCPVersion(c.k8sClient, pxutil.Openshift_4_16_version)
+
+		if err != nil {
+			logrus.Errorf("error during checking OCP version %v ", err)
```
pureneelesh, line 195: nit: sorry to nit-pick on this again
```
logrus.Errorf("Error during checking OCP version: %v", err)
```

drivers/storage/portworx/component/autopilot.go
```diff
@@ -174,12 +189,30 @@ func (c *autopilot) Reconcile(cluster *corev1.StorageCluster) error {
...
+				}
+				if err := c.createClusterRoleBindingForOCP(cluster.Namespace); err != nil {
+					return err
+				}
+			}
+
+			if err := c.createSecret(cluster.Namespace, ownerRef, ocp416plus); err != nil {
+				// log the error and proceed for deployment creation
+				// if secret is created in next reconcilation loop successfully, deployment will be updated with volume mounts
+				logrus.Errorf("error during creating secret %v ", err)
```
pureneelesh, line 211: nit: `": %v", err`

drivers/storage/portworx/component/autopilot.go
```diff
@@ -265,26 +306,115 @@ func (c *autopilot) createConfigMap(
...
+		err := k8sutil.GetSecret(c.k8sClient, AutopilotSecretName, clusterNamespace, secret)
+		if err != nil {
+			return fmt.Errorf("error during getting secret %w ", err)
+		}
+
+		if secret.Data != nil {
+			// if secret exists, check if token is expired
+			refreshNeeded, err := isTokenRefreshRequired(secret)
+			if err != nil {
+				return fmt.Errorf("error during checking token refresh %w ", err)
```
pureneelesh, line 372: nit: `refresh: %w", err`

pureneelesh (approved):
Did we do any testing on non-openshift deployment to ensure that we did not break anything there? Otherwise, LGTM.

nikita-bhatia:
> Did we do any testing on non-openshift deployment to ensure that we did not break anything there? Otherwise, LGTM.

Yes it was verified on vanilla k8s cluster as well.

PR #1499, opened 2024-03-25
PWX-36551: Update plugin image with change in versions if autoupdatecomponents is used

svijaykumar-px:
**What this PR does / why we need it**: Previously, the image versions for dynamic plugin and dynamic plugin proxy where not updated when versions configmap / URL endpoint was changed even though autoUpdateComponents was set to Once or Always. This PR makes sure the images of both get updated in such a case , i.e; when autoUpdateComponents is Once/Always

**Which issue(s) this PR fixes** (optional)
Closes # PWX-36551

**Testing notes** :

1.  Using versions configmap

 a) When autoUpdateComponents is set to Once
   - Create versions configmap using `kubectl -n kube-system create configmap px-versions --from-file=versions.yaml`. 
       The versions.yaml file should contain the component image: `dynamicPluginProxy: nginxinc/nginx-unprivileged:1.23`
   - Create a storage cluster with any portworx version
   - Verify with the stc yaml , desired images field should contain the image `dynamicPluginProxy: nginxinc/nginx- 
        unprivileged:1.23`
   - Edit the stc and add the field - `autoUpdateComponents: Once` and save. Wait till the change is applied and px pods 
      retsrart
   - Edit the px-versions cm and change image of dynamicPluginProxy to `dynamicPluginProxy: nginxinc/nginx- 
      unprivileged:1.25`.
   - Wait for a minute and check the stc yaml. This time the desired images of dynamicPluginProxy should be 1.25 instead 
      of 1.23

 b) When autoUpdateComponents is set to Always
   - Create versions cm as before
   - Create a storage cluster with field - `autoUpdateComponents: Always`
   - Once the stc is running, check the yaml to verify desired images has  `dynamicPluginProxy: nginxinc/nginx- 
       unprivileged:1.23`
   - Edit the px-versions cm and change image of dynamicPluginProxy to `dynamicPluginProxy: nginxinc/nginx- 
         unprivileged:1.25`.
   - Verify yaml of stc has desired images `dynamicPluginProxy: nginxinc/nginx-unprivileged:1.25`.

2. Using URL endpoint for versions
 
 a) When autoUpdateComponents is set to Once
  - Create a http server which hosts a custom px-versions file in an endpoint. This versions file should use the older 
     version-  `dynamicPluginProxy: nginxinc/nginx-unprivileged:1.23`
  - Create a stc with env variable 
   `- name: PX_RELEASE_MANIFEST_URL
      value: "http://<IP>:<PORT>"`
[...]

PR #1587, opened 2024-06-20
PWX-37760: Update component images when autoUpdateComponents is set

svijaykumar-px:
**What this PR does / why we need it**: For components like prometheus, alert manager and grafana they do not update desiredImages in the stc when autoUpdateComponents is set to Once or Always. This change will ensure that the component images get updated

Closes # PWX-37760, PWX-37545

**Special notes for your reviewer**: Similar issue was found for dynamic Plugin images and the [fix](https://github.com/libopenstorage/operator/pull/1499) was similar to this PR


nikolaypopov (approved):

nilanto07-px (approved):
LGTM.

PR #1595, opened 2024-07-01
PWX-37733 Add pod name in system metadata device for dmthin on ocp facd

ssz1997:
https://purestorage.atlassian.net/browse/PWX-37733

Along with https://github.com/pure-px/porx/pull/13643 I successfully run a px cluster with dmthin with ocp facd

### DIFF
--- a/drivers/storage/portworx/preflight.go
+++ b/drivers/storage/portworx/preflight.go
@@ -410,7 +410,19 @@ func (u *preFlightPortworx) processPassedChecks(recorder record.EventRecorder) {
 		if pxutil.IsVsphere(u.cluster) {
 			cmetaData = DefCmetaVsphere
 		} else if pxutil.IsPure(u.cluster) {
-			cmetaData = DefCmetaFACD
+			var podName string
+		specsItr:
+			for _, spec := range *u.cluster.Spec.CloudStorage.DeviceSpecs {
+				options := strings.Split(spec, ",")
+				for _, option := range options {
+					keyValPair := strings.Split(option, "=")
+					if len(keyValPair) == 2 && keyValPair[0] == "pod" {
+						podName = option
+						break specsItr
+					}
+				}
+			}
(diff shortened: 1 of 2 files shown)


drivers/storage/portworx/preflight.go
```diff
@@ -410,7 +410,19 @@ func (u *preFlightPortworx) processPassedChecks(recorder record.EventRecorder) {
...
 		} else if pxutil.IsPure(u.cluster) {
-			cmetaData = DefCmetaFACD
+			var podName string
+		specsItr:
+			for _, spec := range *u.cluster.Spec.CloudStorage.DeviceSpecs {
+				options := strings.Split(spec, ",")
+				for _, option := range options {
+					keyValPair := strings.Split(option, "=")
+					if len(keyValPair) == 2 && keyValPair[0] == "pod" {
+						podName = option
```
pp511, line 420: Let's not name the variable name as podName here. It appears as if it's the name of the pod while it actually is the key, value pair. Maybe let's just suffix it with "Option".

jrivera-px: We actually have a drive format validation check in px-runc pre-flight.   So we need to make sure that the check in px-runc will accept this format `"size=49,pod=<name>"`

sudo /opt/pwx/bin/px-runc pre-flight -c .... -s size=200,pod=test ... -T px-storeV2 ... --cloud_provider pure ...

ssz1997: People did some changes in `px-runc.go` that will be cherry-picked into the px release branch, and I believe this commit serves the purpose of skipping the validation: https://github.com/pure-px/porx/commit/e1d44d49737daf6f11c083a5c228a74c58913579

ssz1997: I also cherry-picked this commit to my local test and px didn't complain with the format

pp511 (approved):
lgtm

PR #1596, opened 2024-07-02
PWX-37982 Use the expiration time returned by k8s api server for px sa token

ssz1997:
There's a global `maxExpirationSeconds` which can be smaller than the token expiration time we set. If this is the case, k8s will silently set the returned token expiration time to the `maxExpirationSeconds`, which will expire earlier than we expected, brining px not having a valid token to talk to k8s.

### DIFF
--- a/drivers/storage/portworx/component/portworx_basic.go
+++ b/drivers/storage/portworx/component/portworx_basic.go
@@ -597,12 +597,12 @@ func (c *portworxBasic) refreshTokenSecret(secret *v1.Secret, cluster *corev1.St
 	if err != nil {
 		return err
 	}
-	secret.Data[PxSaTokenRefreshTimeKey] = []byte(time.Now().UTC().Add(time.Duration(expirationSeconds/2) * time.Second).Format(time.RFC3339))
 	newToken, err := generatePxSaToken(cluster, expirationSeconds)
 	if err != nil {
 		return err
 	}
-	secret.Data[core.ServiceAccountTokenKey] = newToken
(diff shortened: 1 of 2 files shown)


drivers/storage/portworx/component/portworx_basic.go
```diff
@@ -597,20 +597,20 @@ func (c *portworxBasic) refreshTokenSecret(secret *v1.Secret, cluster *corev1.St
...
 		return err
 	}
-	secret.Data[PxSaTokenRefreshTimeKey] = []byte(time.Now().UTC().Add(time.Duration(expirationSeconds/2) * time.Second).Format(time.RFC3339))
 	newToken, err := generatePxSaToken(cluster, expirationSeconds)
 	if err != nil {
 		return err
 	}
-	secret.Data[core.ServiceAccountTokenKey] = newToken
+	secret.Data[core.ServiceAccountTokenKey] = []byte(newToken.Status.Token)
+	secret.Data[PxSaTokenRefreshTimeKey] = []byte(time.Now().UTC().Add(time.Duration(*newToken.Spec.ExpirationSeconds/2) * time.Second).Format(time.RFC3339))
```
pureneelesh, line 605: Let's add a comment here to indicate why we are doing this. It is a bit subtle.

pureneelesh (approved):

drivers/storage/portworx/component/portworx_basic.go
ssz1997: Agree

PR #1597, opened 2024-07-03
PWX-38004 : Update ca.crt in autopilot secret when upgrading from OCP from 4.15 to 4.16

nikita-bhatia:
Issue :
Issue with the code was that ca.crt would not get updated during upgrade case of OCP cluster, since secret was already available. The change in the code updates ca.crt everytime token is refreshed. This handles case when pod certidicate is also updated on runtime. The new code flow is like following during every reconcilation loop : 
1. Fetch the secret
2. check if secret requires update, if secret is not present, present but expiry is not set or expiry is set but token is expired, we will create/update the secret.
3. If step 2 says update is required, certificare is fetched from path /var/run/secrets/kubernetes.io/serviceaccount/ca.crt within the pod, and updated in the secret.

This handles fresh install and upgrade case as well.

Closes # https://purestorage.atlassian.net/browse/PWX-38004

- Verified on OCP 4.15 and OCP 4.16
- Added UTs for OCP upgrade case.

### DIFF
--- a/drivers/storage/portworx/component/autopilot.go
+++ b/drivers/storage/portworx/component/autopilot.go
@@ -68,6 +68,8 @@ const (
 	OpenshiftClusterViewRoleName = "cluster-monitoring-view"
 	// AutopilotSaTokenRefreshTimeKey time to refresh the service account token
 	AutopilotSaTokenRefreshTimeKey = "autopilotSaTokenRefreshTime"
+	// CaCertPath path to ca.crt in the pod
+	CaCertPath = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
 )
 
 var (
@@ -370,14 +372,14 @@ func (c *autopilot) createSecret(clusterNamespace string, ownerRef *metav1.Owner
 		// or refresh the token if it is expired
 		updateRequired, err := isSecretUpdateRequired(secret)
 		if err != nil {
-			return fmt.Errorf("error during checking token refresh: %w ", err)
+			return fmt.Errorf("error during checking token refresh: %w", err)
 		}
 
 		if updateRequired {
 			logrus.Infof("refreshing token for secret %s/%s", clusterNamespace, AutopilotSecretName)
 			err := c.updateSecretTokenAndCert(secret, clusterNamespace, ownerRef)
 			if err != nil {
-				return fmt.Errorf("failed to check token in secret %s/%s: %w", clusterNamespace, AutopilotSecretName, err)
+				return fmt.Errorf("failed to update secret %s/%s: %w", clusterNamespace, AutopilotSecretName, err)
 			}
 		}
 		return nil
@@ -932,7 +934,7 @@ func (c *autopilot) updateSecretTokenAndCert(secret *v1.Secret, clusterNamespace
 	// get the root ca.crt from the secret mounted inside the pod
 	// this is required while upgrading from OCP 4.15 to 4.16
 	// since the ca.crt is changed in 4.16 and not managed by openshift
-	rootCaCrt, err := getCertForSecret()
+	rootCaCrt, err := getRootCACert()
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


drivers/storage/portworx/component/autopilot.go
```diff
@@ -937,22 +920,41 @@ func (c *autopilot) isOCPUserWorkloadSupported() bool {
...
 	}
 
-	// update the secret with new token
-	secret.Data[core.ServiceAccountTokenKey] = []byte(newToken.Status.Token)
-	secret.Data[AutopilotSaTokenRefreshTimeKey] = []byte(time.Now().UTC().Add(time.Duration(*newToken.Spec.ExpirationSeconds/2) * time.Second).Format(time.RFC3339))
-	err = k8sutil.CreateOrUpdateSecret(c.k8sClient, secret, ownerRef)
+	// get the root ca.crt from the secret mounted inside the pod
+	// this is required while upgrading from OCP 4.15 to 4.16
+	// since the ca.crt is changed in 4.16 and not managed by openshift
+	rootCaCrt, err := getCertForSecret()
```
pureneelesh, line 935: nit: name getCertForSecret is too generic. How about renaming it to getRootCACert?

drivers/storage/portworx/component/autopilot.go
```diff
@@ -937,22 +920,41 @@ func (c *autopilot) isOCPUserWorkloadSupported() bool {
...
+	// set the token refresh time to half the way of token expiration
+	secret.Data[AutopilotSaTokenRefreshTimeKey] = []byte(time.Now().UTC().Add(time.Duration(*newToken.Spec.ExpirationSeconds/2) * time.Second).Format(time.RFC3339))
+	return k8sutil.CreateOrUpdateSecret(c.k8sClient, secret, ownerRef)
+}
+
+// get the root ca.crt from the secret mounted inside the pod
+func getCertForSecret() ([]byte, error) {
+	rootCaCrt, err := os.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/ca.crt")
+	if err != nil && !os.IsNotExist(err) {
+		return nil, fmt.Errorf("error reading k8s cluster certificate located inside the pod at /var/run/secrets/kubernetes.io/serviceaccount/ca.crt: %w", err)
```
pureneelesh, line 955: nit: since the `/var/run/secrets/.../ca.crt` path is mentioned multiple times, we should put it in a const

drivers/storage/portworx/component/autopilot.go
```diff
@@ -968,8 +970,12 @@ func generateAPSaToken(clusterNamespace string, expirationSeconds int64) (*authv
 	return tokenResp, nil
 }
 
-func isTokenRefreshRequired(secret *v1.Secret) (bool, error) {
-	if len(secret.Data) == 0 || len(secret.Data[v1.ServiceAccountTokenKey]) == 0 {
+// isSecretUpdateRequired returns true if secret is not present
+// isSecretUpdateRequired returns true if token expiry is not set
+// isSecretUpdateRequired returns true if token is expired
+// isSecretUpdateRequired returns false if token is not expired
```
pureneelesh, line 976: nit: avoid explaining the whole logic in the comment since it will be difficult to keep it up-to-date as the function changes. We should just explain the background that is not obvious from reading the code.

drivers/storage/portworx/components_test.go
```diff
@@ -5067,6 +5067,185 @@ func TestAutopilotInstallAndUninstallOnOpenshift416(t *testing.T) {
...
+	// on OCP 4.15 , thanos-ruler-token secret is used for certificate and token
+	// this secret is created by openshift
+	secret := &v1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "thanos-ruler-token-abcd",
+			Namespace: "openshift-user-workload-monitoring",
+		},
+		Data: map[string][]byte{
+			"token":  []byte("dG9rZW4tdmFsdWU="),
+			"ca.crt": []byte("Y2VydGlmaWNhdGUtdmFsdWU="),
```
pureneelesh, line 5173: will the key "ca.crt" be different in the old secret? Should we mimic that here?

pureneelesh (approved):

drivers/storage/portworx/components_test.go
nikita-bhatia: I tried to mock os.Readfile, but i was unable to do so as i was getting permission denied error while creating the file from the test temporarily.

PR #1598, opened 2024-07-03
PWX-37884 Add portworx serviceaccount token refresh integration test

ssz1997:
Add an integrartion test to verify the serviceaccount token refresh mechanism for portworx. 

PWX-37884

### DIFF
--- a/test/integration_test/basic_test.go
+++ b/test/integration_test/basic_test.go
@@ -466,52 +466,46 @@ func BasicInstallWithNodeAffinity(tc *types.TestCase) func(*testing.T) {
 }
 
 // 1. Deploy PX and verify the secret containing the token for px stored in the k8s secret correctly mounted inside px runc container.
-// 2. The cluster spec set the token expiration time to be 3 min. So the token is supposed to refresh after 1.5min.
-// 3. Wait for 1.5min. Verify the token is refreshed and correctly mounted inside px runc container.
-// 4. Delete the secret. Wait for 90sec. Verify the token is refreshed and correctly mounted inside px runc container.
+// 2. The cluster spec set the token expiration time to be 10 min, which is the minimum allowed token expiration time. The token should get refreshed after 5min.
+// 3. Wait for 5min. Verify the token is refreshed and correctly mounted inside px runc container.
+// 4. Delete the secret. Wait for 2min. Verify the token is refreshed and correctly mounted inside px runc container.
 func BasicInstallWithPxSaTokenRefresh(tc *types.TestCase) func(*testing.T) {
 	return func(t *testing.T) {
 		testSpec := tc.TestSpec(t)
 		cluster, ok := testSpec.(*corev1.StorageCluster)
 		require.True(t, ok)
 
-		cluster = ci_utils.DeployAndValidateStorageCluster(cluster, ci_utils.PxSpecImages, t)
-		pxSaSecret, err := coreops.Instance().GetSecret(pxutil.PortworxServiceAccountTokenSecretName, cluster.Namespace)
-		require.NoError(t, err)
-		expectedToken := pxSaSecret.Data[core.ServiceAccountTokenKey]
 		cmd := "runc exec portworx cat /var/run/secrets/kubernetes.io/serviceaccount/token"
-		actualToken, stderr, err := ci_utils.RunPxCmd(cmd)
-		require.Empty(t, stderr)
-		require.Nil(t, err)
-		require.Equal(t, actualToken, expectedToken)
-		logrus.Infof("First token created and verified: %s", actualToken)
+		verifyTokenFunc := func() string {
+			pxSaSecret, err := coreops.Instance().GetSecret(pxutil.PortworxServiceAccountTokenSecretName, cluster.Namespace)
+			require.NoError(t, err)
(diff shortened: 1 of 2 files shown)


test/integration_test/basic_test.go
```diff
@@ -139,6 +141,29 @@ var testStorageClusterBasicCases = []types.TestCase{
 		}),
 		TestFunc: BasicInstallWithNodeAffinity,
 	},
+	{
+		TestName:        "BasicInstallWithPxSaTokenRefresh",
+		TestrailCaseIDs: []string{},
```
ssz1997, line 146: What is this Case ID for? Couldn't figure out what ids I should put here

test/integration_test/basic_test.go
```diff
@@ -440,6 +465,37 @@ func BasicInstallWithNodeAffinity(tc *types.TestCase) func(*testing.T) {
...
+		expectedToken := pxSaSecret.Data[core.ServiceAccountTokenKey]
+		cmd := "runc exec portworx cat /var/run/secrets/kubernetes.io/serviceaccount/token"
+		actualToken, stderr, err := ci_utils.RunPxCmd(cmd)
+		require.Empty(t, stderr)
+		require.Nil(t, err)
+		require.Equal(t, actualToken, expectedToken)
+
+		time.Sleep(time.Duration(90) * time.Second)
+
+		pxSaSecret, err = coreops.Instance().GetSecret(pxutil.PortworxServiceAccountTokenSecretName, cluster.Namespace)
```
pureneelesh, line 489: Can we verify that the token here is different from the token we saw initially? Also, in k8s code, I see that they enforce minimum token duration of 10min. We may have to account for that if "3 min" duration is being silently overridden.

test/integration_test/basic_test.go
pureneelesh: We have to create a testcase in TestRail and use that ID.

test/integration_test/basic_test.go
```diff
@@ -139,6 +141,29 @@ var testStorageClusterBasicCases = []types.TestCase{
...
 		TestFunc: BasicInstallWithNodeAffinity,
 	},
+	{
+		TestName:        "BasicInstallWithPxSaTokenRefresh",
+		TestrailCaseIDs: []string{},
+		TestSpec: ci_utils.CreateStorageClusterTestSpecFunc(&corev1.StorageCluster{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "test-stc",
+				Annotations: map[string]string{
+					"portworx.io/host-pid": "true",
```
pureneelesh, line 151: add a comment to indicate why we are adding this annotation

test/integration_test/basic_test.go
ssz1997: Any instruction on how I can do that?

test/integration_test/basic_test.go
ssz1997: If we set it to a smaller value (<10min), the token used by px will still be refreshed according to our configuration. It's just there will be multiple valid tokens hanging around. For example, if we refresh the token every 1min, there will be 10 valid tokens before the first one expires. If we are not enforcing user behavior, I think it's ok to set it short?

test/integration_test/basic_test.go
```diff
@@ -139,6 +141,29 @@ var testStorageClusterBasicCases = []types.TestCase{
...
+				Annotations: map[string]string{
+					"portworx.io/host-pid": "true", // for running commands inside px runc container from oci-mon container
+				},
+			},
+			Spec: corev1.StorageClusterSpec{
+				CommonConfig: corev1.CommonConfig{
+					Env: []v1.EnvVar{
+						{
+							Name:  pxutil.EnvKeyPortworxServiceAccountTokenExpirationMinutes,
+							Value: "3",
```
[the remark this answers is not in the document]
ssz1997: In fact it is not ok because the test is failing. We have to change it to 10min

test/integration_test/basic_test.go
```diff
@@ -440,6 +465,56 @@ func BasicInstallWithNodeAffinity(tc *types.TestCase) func(*testing.T) {
...
+		time.Sleep(time.Duration(90) * time.Second)
+
+		pxSaSecret, err = coreops.Instance().GetSecret(pxutil.PortworxServiceAccountTokenSecretName, cluster.Namespace)
+		require.NoError(t, err)
+		secondRefreshedExpectedToken := pxSaSecret.Data[core.ServiceAccountTokenKey]
+		secondRefreshedActualToken, stderr, err := ci_utils.RunPxCmd(cmd)
+		require.Empty(t, stderr)
+		require.Nil(t, err)
+		require.Equal(t, secondRefreshedActualToken, secondRefreshedExpectedToken)
+		require.NotEqual(t, secondRefreshedActualToken, secondRefreshedActualToken)
```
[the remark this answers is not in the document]
ssz1997: Yeah typo...

test/integration_test/basic_test.go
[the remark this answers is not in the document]
ssz1997: More specifically, the token request call returns error saying the minimum is 10min, just like if we try to create a token with less than 10min expiration time via command line.

pureneelesh (approved):

PR #1600, opened 2024-07-09
PWX-36514: Integration test to validate node pdb

svijaykumar-px:
**What this PR does / why we need it**: This PR adds integration tests to validate Node PDBs as a part of Smart and Parallel Upgrades in 24.2.0

Closes # PWX-36514, PWX-37561

**Special notes for your reviewer**: The tests covered by this PR are mentioned in this [testrail](https://portworx.testrail.net/index.php?/suites/view/1&group_by=cases:section_id&group_id=31616&group_order=asc&display_deleted_cases=0) link 
[This](https://jenkins.pwx.dev.purestorage.com/job/Users/job/svijaykumar/job/dev-nodePDB-test/6/) is the link for jenkins job of the test

### DIFF
--- a/pkg/util/test/util.go
+++ b/pkg/util/test/util.go
@@ -5362,28 +5362,30 @@ func ValidatePodDisruptionBudget(cluster *corev1.StorageCluster, timeout, interv
 			if err != nil {
 				return nil, true, fmt.Errorf("failed to get storage nodes, Err: %v", err)
 			}
-			availablenodes := 0
+			availableNodes := 0
 			for _, node := range nodes.Items {
-				if *node.Status.NodeAttributes.Storage && node.Status.Phase == "Online" {
-					availablenodes++
-				} else {
-					logrus.Infof("Node [%s] in state [%s] is not Online, PDB might be incorrect", node.Name, node.Status.Phase)
+				if *node.Status.NodeAttributes.Storage {
+					if node.Status.Phase == "Online" {
+						availableNodes++
+					} else {
+						logrus.Infof("Node %s is in state [%s], PDB might be incorrect", node.Name, node.Status.Phase)
+					}
 				}
 			}
 			pdbs, err := policyops.Instance().ListPodDisruptionBudget(cluster.Namespace)
 			if err != nil {
 				return nil, true, fmt.Errorf("failed to list all poddisruptionbudgets, Err: %v", err)
 			}
-			actualNodePDBcount := 0
+			actualNodePDBCount := 0
 			for _, pdb := range pdbs.Items {
 				if strings.HasPrefix(pdb.Name, "px-") && pdb.Name != "px-kvdb" {
-					actualNodePDBcount++
+					actualNodePDBCount++
 				}
 			}
-			if actualNodePDBcount == availablenodes {
+			if actualNodePDBCount == availableNodes {
 				return nil, false, nil
 			}
-			return nil, true, fmt.Errorf("incorrect node PDB count. Expected node PDB count [%d], Actual node PDB count [%d]", availablenodes, actualNodePDBcount)
+			return nil, true, fmt.Errorf("incorrect node PDB count. Expected node PDB count [%d], Actual node PDB count [%d]", availableNodes, actualNodePDBCount)
 
 		}
 		if _, err := task.DoRetryWithTimeout(t, timeout, interval); err != nil {
@@ -5400,16 +5402,16 @@ func ValidatePodDisruptionBudget(cluster *corev1.StorageCluster, timeout, interv
 			}
 
 			nodeslen := 0
-			availablenodes := 0
+			availableNodes := 0
 			for _, node := range nodes.Items {
 				if *node.Status.NodeAttributes.Storage {
 					nodeslen++
 					if node.Status.Phase == "Online" {
-						availablenodes++
+						availableNodes++
 					}
 				}
 			}
-			nodesUnavailable := nodeslen - availablenodes
+			nodesUnavailable := nodeslen - availableNodes
 			// Skip PDB validation for px-storage if number of storage nodes is lesser than or equal to 2
 			if nodeslen <= 2 {
 				logrus.Infof("Storage PDB does not exist for storage nodes lesser than or equal to 2, skipping PDB validattion")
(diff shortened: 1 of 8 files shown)


test/integration_test/node_pdb_test.go
```diff
@@ -0,0 +1,205 @@
...
+			k8snodecount, err := ci_utils.GetNonMasterK8sNodeCount()
+			require.NoError(t, err)
+			cluster, err = operator.Instance().GetStorageCluster(cluster.Name, cluster.Namespace)
+			require.NoError(t, err)
+			cluster.Annotations["portworx.io/storage-pdb-min-available"] = fmt.Sprintf("%d", k8snodecount-1)
+			cluster, err = ci_utils.UpdateStorageCluster(cluster)
+			require.NoError(t, err)
+
+			logrus.Infof("Validating number of nodes ready for upgrade with minAvailable annotation %d", k8snodecount-1)
+			err = testutil.ValidateNodesSelectedForUpgrade(cluster, k8snodecount-1, ci_utils.DefaultValidateUpgradeTimeout, ci_utils.DefaultValidateUpgradeRetryInterval)
```
piyush-nimbalkar, line 155: In the `ValidateNodesSelectedForUpgrade` function, we are using the `k8snodecount` as a quorum count. Which means we will try to validate upgradable nodes count against `k8snodecount/2 + 1` instead of `totalStorageNodes - k8snodecount`.

test/integration_test/node_pdb_test.go
```diff
@@ -0,0 +1,205 @@
...
+			require.NoError(t, err)
+
+			err = ci_utils.UncordonNodes()
+			require.NoError(t, err)
+		}
+		ci_utils.UninstallAndValidateStorageCluster(cluster, t)
+
+	}
+}
+func TestNodePDB(t *testing.T) {
```
piyush-nimbalkar, line 201: nit: newline above this function.

test/integration_test/node_pdb_test.go
```diff
@@ -0,0 +1,205 @@
...
+			require.NoError(t, err)
+
+			err = ci_utils.UncordonNodes()
+			require.NoError(t, err)
+		}
+		ci_utils.UninstallAndValidateStorageCluster(cluster, t)
+	}
+}
+
+func NodePDBDisablingParallelUpgrade(tc *types.TestCase) func(*testing.T) {
```
piyush-nimbalkar, line 165: What are we trying to test here?  Are we just checking that the number is less that k8snodecount - 1? This would mean that we will wait for the entire upgrade timeout to validate that. That would be a long running test.

pkg/util/test/util.go
```diff
@@ -5948,3 +5987,72 @@ func RestoreEtcHosts(t *testing.T) {
...
+					return nil, true, fmt.Errorf("incorrect PDB minAvailable value for node %s. Expected PDB [%d], Actual PDB [%d]", strings.TrimPrefix(pdb.Name, "px-"), 1, pdb.Spec.MinAvailable.IntValue())
+				}
+			}
+		}
+
+		for _, node := range nodes.Items {
+			if coreops.Instance().IsNodeMaster(node) {
+				continue
+			}
+			if _, ok := pxNodesMap["px-"+node.Name]; !ok {
```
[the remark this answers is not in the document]
svijaykumar-px: In that parent validation function, right now I only validate that there are correct number of node PDBs in it . That should be enough right or should I check for the names as well?

test/integration_test/node_pdb_test.go
svijaykumar-px: So when we pass k8snodecount-1 as quorum, we directly use it as the minAvailable value. I realised the variable name is confusing. I will change it to minAvailable. But yeah, I did realise the issue here. I should check if that value is lesser than totalcount-quorum and not just quorum

test/integration_test/node_pdb_test.go
svijaykumar-px: All the tests in the operator do not actually perform the k8s upgrade. This just checks the nodes that are selected for upgrade. Since the operator can only control this and the nodes actually selected are not under its control.

pkg/util/test/util.go
```diff
@@ -5348,10 +5350,49 @@ func ValidatePodDisruptionBudget(cluster *corev1.StorageCluster, timeout, interv
...
+	pxVersion := GetPortworxVersion(cluster)
 
 	// PodDisruptionBudget is supported for k8s version greater than or equal to 1.21 and operator version greater than or equal to 1.5.0
 	// Changing opVersion to 23.10.0 for PTX-23350 | TODO: add better logic with PTX-23407
-	if k8sVersion.GreaterThanOrEqual(minSupportedK8sVersionForPdb) && opVersion.GreaterThanOrEqual(opVer23_10) {
+
+	// Smart and parallel upgrades is supported from px version 3.1.2 and operator version 24.2.0
+	if k8sVersion.GreaterThanOrEqual(minSupportedK8sVersionForPdb) && opVersion.GreaterThanOrEqual(opVer24_2_0) && pxVersion.GreaterThanOrEqual(pxVer3_1_2) {
+		t := func() (interface{}, bool, error) {
+			nodes, err := operatorops.Instance().ListStorageNodes(cluster.Namespace)
```
nikita-bhatia, line 5361: are these going to be PX StorageNodes (kubectl get sn) or kubernetes nodes (kubectl get nodes)?
what will be the output of this ?

pkg/util/test/util.go
```diff
@@ -5354,7 +5354,44 @@ func ValidatePodDisruptionBudget(cluster *corev1.StorageCluster, timeout, interv
...
+			if err != nil {
+				return nil, true, fmt.Errorf("failed to list all poddisruptionbudgets, Err: %v", err)
+			}
+			actualNodePDBCount := 0
+			for _, pdb := range pdbs.Items {
+				if strings.HasPrefix(pdb.Name, "px-") && pdb.Name != "px-kvdb" {
+					actualNodePDBCount++
+				}
+			}
+			if actualNodePDBCount == availableNodes {
```
nikita-bhatia, line 5384: In case of cluster with storageless nodes, this condition will fail right since PDBs will not be created for storageless nodes?

svijaykumar-px: Yeah PDB for storageless nodes will not be created, and that is why i only increment availableNodes when storagenode field is true. I had actually tested this with storagelessnodes as well and the test passed

nikita-bhatia (approved):
LGTM

PR #1601, opened 2024-07-09
PTX-25083 Add validation for pure-telemetry-certs secret OwnerReference

nikolaypopov:
Add validation for pure-telemetry-certs secret OwnerReference, which should solve the issue of wiping the cluster before it gets the OwnerReference and leaving secret behind


zoxpx (approved):
Looks good to me..

btw, you have a small typo in the comment: ~referense~ -> `reference`

nikolaypopov:
This won't work, as we only set ownerReference when we set deleteStrategy, I will wait for the new implementation of how we create this secret before improving this flow

PR #1553, opened 2024-05-21
PWX-36928: telemetry SSL-cert validation

zoxpx:
* adding code to validate (and refresh) Pure1 Telemetry's SSL certificate

**What this PR does / why we need it**:

ISSUE:
* customer did not properly recycle the px-cluster, so the SSL certificate in Telemetry's secret was pointing to a STALE ClusterUUID

FIX:
* as a fix, operator is now explicitly checking the Telemetry's SSL certificate
* if certificate is invalid, it'll remove the Telemetry's secret, and restart the telemetry-registration POD

**Which issue(s) this PR fixes** (optional)
Closes # PWX-36928

**Special notes for your reviewer**:

Adding dedicated Unit-test

Also, manually changed Telemetry's secret with a dummy SSL certificate -- and the PODs are restarting:

```
# kubectl get pods -o wide

NAME                                             READY   STATUS        RESTARTS   AGE     IP            NODE   NOMINATED NODE   READINESS GATES
...
portworx-operator-6bd847f88f-n8ssm               1/1     Running       0          45s     10.244.2.22   foo2   <none>           <none>
...
px-telemetry-phonehome-fxcqb                     2/2     Running       0          5d15h   10.244.1.15   foo3   <none>           <none>
px-telemetry-registration-69ccd685f7-bgb8c       2/2     Running       0          24s     10.13.1.64    foo4   <none>           <none>
px-telemetry-registration-69ccd685f7-hmgp7       2/2     Terminating   0          7d14h   10.13.1.59    foo2   <none>           <none>
```

Logs:
```
time="21-05-2024 20:11:45" level=info msg="Using ccm-go version: 1.2.2" file="util.go:1282"
time="21-05-2024 20:11:45" level=debug msg="Inspecting secret portworx/pure-telemetry-certs for SSL cert" file="telemetry.go:1024"
time="21-05-2024 20:11:45" level=error msg="SSL cert Pseudonym not found" file="telemetry.go:1063"
time="21-05-2024 20:11:45" level=warning msg="refreshing telemetry SSL cert" file="telemetry.go:843"
time="21-05-2024 20:11:45" level=warning msg="refreshTelemetrySSLCert - deleting telemetry SSL cert secret portworx/pure-telemetry-certs" file="telemetry.go:1081"
time="21-05-2024 20:11:45" level=info msg="Deleting portworx/pure-telemetry-certs Secret" file="k8s.go:2021"
[...]

PR #1603, opened 2024-07-10
PWX-38007: Telemetry cert-secret refresh fix

zoxpx:
* the telemetry needs to force-delete the invalid pure-telemetry-certs secret  (no owner's check)

This is a follow-up fix for #1553

**ISSUE**:
* it turns out the `secret/pure-telemetry-certs` may be stored _WITHOUT_ the `metatata.ownerReference`
* note that the px-operator's "normal"  `DeleteSecret()` will remove the secret _only_ if it has required object owners
  - this is not a problem for most of operator-created secrets, since it's the operator itself that creates the `secret` object, and it'll populate the correct `ownerReference`
  - however, with `secret/pure-telemetry-certs` the secret is created by telemetry POD (`ccm-go` container), but owner gets populated at later point in time  (i.e. this particular secret follows 2-step-creation process)

**FIX**:
* as a fix, when we decide we need to remove invalid `secret/pure-telemetry-certs` -- we'll use UNCONDITIONAL `DeleteSecret()` which does not check the object's owners

Closes # PWX-38007

Manual test:
1. set up INVALID `secret/pure-telemetry-certs` -- which we picked up from a different cluster
   - note, this carries SSL-cert, which has a wrong ClusterUUID embedded
2. made sure to remove the `meradata.ownerReference` on the `secret/pure-telemetry-certs`
3. the `px-operator` now correctly removes the `secret/pure-telemetry-certs` and restarts `pod/px-telemetry-registration-XX`
   - note, BEFORE this fix, the `secret/pure-telemetry-certs` could not get deleted -- it was error-looping since the object's OWNER could not be matched

### DIFF
--- a/drivers/storage/portworx/component/telemetry.go
+++ b/drivers/storage/portworx/component/telemetry.go
@@ -1,6 +1,7 @@
 package component
 
 import (
+	"context"
 	cryptoTls "crypto/tls"
 	"crypto/x509"
 	"encoding/asn1"
@@ -842,13 +843,12 @@ func (t *telemetry) createDeploymentTelemetryRegistration(
 
 	// have a valid cluster UUID?  lets validate the Telemetry SSL cert
 	if cuuid := cluster.Status.ClusterUID; cuuid != "" {
-		if certBytes, err := t.getTelemetrySSLCert(deployment.Namespace); err != nil {
-			logrus.WithError(err).Errorf("failed to get telemetry SSL cert")
-		} else if err2 := t.validateTelemetrySSLCert(certBytes, cuuid); err2 == errInvalidTelemetryCert {
+		sec, err := t.validateTelemetrySSLCert(deployment.Namespace, cuuid)
+		if err == errInvalidTelemetryCert {
 			logrus.Warn("refreshing telemetry SSL cert")
-			t.refreshTelemetrySSLCert(deployment)
-		} else if err2 != nil {
-			logrus.WithError(err2).Errorf("failed to validate telemetry SSL cert")
+			t.refreshTelemetrySSLCert(sec)
+		} else if err != nil {
+			logrus.WithError(err).Errorf("failed to validate telemetry SSL cert")
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


drivers/storage/portworx/component/telemetry.go
```diff
@@ -1059,38 +1061,43 @@ func (t *telemetry) validateTelemetrySSLCert(certBytes []byte, cuuid string) err
...
+	if sec == nil {
+		return
+	} else if sec.Name != pxutil.TelemetryCertName {
+		logrus.Errorf("invalid secret name %s/%s  (expected %s)", sec.Namespace, sec.Name, pxutil.TelemetryCertName)
 		return
 	}
-	logrus.Warnf("refreshTelemetrySSLCert - deleting telemetry SSL cert secret %s/%s", dep.Namespace, pxutil.TelemetryCertName)
-	err := k8sutil.DeleteSecret(t.k8sClient, pxutil.TelemetryCertName, dep.Namespace, dep.OwnerReferences...)
+
+	logrus.Warnf("refreshTelemetrySSLCert - deleting telemetry SSL cert secret %s/%s", sec.Namespace, sec.Name)
```
nikita-bhatia, line 1092: are we only deleting telemetry cert for refresh? and wit for next reconcilation loop to create it back with owner reference?

nikita-bhatia:
Can we add UTs to cover deletion with different owner reference in this case?

drivers/storage/portworx/component/telemetry.go
zoxpx: The secret is created by `px-telemetry-registration` POD, while the owner-reference is supposed to set up by `px-operator`.  So it's not an "atomic create", and things can break and get interrupted.
* also, in normal circumstances, `px-operator` refuses to delete secrets that don't have the `StorageCluster` set as the owner, probably for added security
* some opportunities where things can get interrupted --
  1. by "install PX" quickly followed by "delete PX"  (<< IMO, common when people install PX for the 1st time) -- or
  2. by cycling `spec.telemetry.enabled: true <-> false`

This particular `refreshTelemetrySSLCert` is ran only when we confirmed that the `secret/pure-telemetry-certs` has the wrong ClusterUUID embedded in the SSL certificate -- so we should absolutely NOT use it with Telemetry
* to make things worse, I think Telemetry backend won't try to match the ClusterUUID from the client SSL-cert, with the actual ClusterUUID that's inside the Telemetry data-payload -- and it'll just accept it, and store it against the WRONG CLUSTER!!  (wrong ClusterUUID)
* so, I have no issues calling "unconditional delete" on the `secret/pure-telemetry-certs` -- when we _know_ it contains WRONG DATA

BTW, once we remove the secret, we'll bounce (remove) the `pod/portworx-telemetry-registration` -- so we don't have to wait for the next reconcilation loop as the deployment will restart the POD.

zoxpx:
> Can we add UTs to cover deletion with different owner reference in this case?

I think it would be an overkill to create mock to emulate broken telemetry registration, especially for this trivial bugfix.
* if you take a look at the https://github.com/libopenstorage/operator/pull/1603/files -- the unit-test coverage gaps are ALL in the error-checking blocks  (i.e. null-checks and such)

nikita-bhatia (approved):
LGTM, based on discussion, existing UTs are covering main testing areas.

zoxpx:
Thanks for the review Nikita -- merging the PR

PR #1605, opened 2024-07-11
PWX-38035 : Remove ca.crt mount volume from autopilot pod

nikita-bhatia:
This PR is required in addition to autopilot changes done here https://github.com/pure-px/autopilot/pull/262
Since autopilot will read ca.crt from /var/run/secrets/kubernetes.io/serviceaccount/ca.crt within the autopilot pod instead of /etc/ssl/px-custom/ that operator mounts, operator no longer is required to mount the certificate to authenticate prometheus URL.

Closes # https://purestorage.atlassian.net/browse/PWX-38035

- This is specifically for OCP 4.16
- For OCP 4.15 and below, there will be no change
-  E2E testing with Autopilot + operator on OCP 4.16 and OCP 4.15

### DIFF
--- a/drivers/storage/portworx/component/autopilot.go
+++ b/drivers/storage/portworx/component/autopilot.go
@@ -814,12 +814,13 @@ func (c *autopilot) getDesiredVolumesAndMounts(
 	if c.isOCPUserWorkloadSupported() && c.isAutopilotSecretCreated(cluster.Namespace) {
 		for _, v := range openshiftDeploymentVolume {
 			// skip adding ca-cert-volume if OCP 4.16 and above
-			// autopilot pod will use kuberntes crt /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
+			// autopilot pod will use kubernetes crt /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
 			// to access prometheus metrics
-			if v.Name != "ca-cert-volume" || !ocp416plus {
-				vCopy := v.DeepCopy()
-				volumeSpecs = append(volumeSpecs, *vCopy)
+			if v.Name == "ca-cert-volume" && ocp416plus {
+				continue
 			}
+			vCopy := v.DeepCopy()
+			volumeSpecs = append(volumeSpecs, *vCopy)
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


zoxpx (approved):
I think the changes are OK -- but I do have one remark

Since we are dealing with the OCP installs quite a lot, I don't expect this will be the last "conditional install"
* could we introduce the `StorageClusterSpec.KubePlatform string` and populate it as `ocp-` + the OCP version we normally get via `IsSupportedOCPVersion()`
* that way, any piece of code can inspect `StorageClusterSpec.KubePlatform` to react on OCP-version, similar how they currently react on `StorageClusterSpec.Version` to get PX-versions and add version-customized code


Also note that we currently require `portworx.io/is-openshift: "true"` to add OCP-specific modifications, which is not ideal.
* we should rely or auto-detection of the platform, and stuff like `StorageClusterSpec.KubePlatform`

drivers/storage/portworx/component/autopilot.go
```diff
@@ -805,13 +807,19 @@ func (c *autopilot) getDesiredAutopilotImage(cluster *corev1.StorageCluster) str
...
+	ocp416plus bool,
 ) ([]v1.Volume, []v1.VolumeMount) {
 	volumeSpecs := make([]corev1.VolumeSpec, 0)
 
 	if c.isOCPUserWorkloadSupported() && c.isAutopilotSecretCreated(cluster.Namespace) {
 		for _, v := range openshiftDeploymentVolume {
-			vCopy := v.DeepCopy()
-			volumeSpecs = append(volumeSpecs, *vCopy)
+			// skip adding ca-cert-volume if OCP 4.16 and above
+			// autopilot pod will use kuberntes crt /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
```
diptiranjanpx, line 817: nit: kubernetes

drivers/storage/portworx/component/autopilot.go
```diff
@@ -805,13 +807,19 @@ func (c *autopilot) getDesiredAutopilotImage(cluster *corev1.StorageCluster) str
...
+			// to access prometheus metrics
+			if v.Name != "ca-cert-volume" || !ocp416plus {
```
diptiranjanpx, line 819: nit :
if v.Name == "ca-cert-volume" && ocp416plus {
   continue
}

diptiranjanpx (approved):

PR #1606, opened 2024-07-11
PWX-38064 Update ca.crt in the PX ServiceAccount Secret if updated

ssz1997:
The root certificate file ca.crt is also rsync’d to px runc container. In the rare case of this certificate is modified/replaced, we need to inform the px runc container as well. Thus update the file in the secret from opertator.


drivers/storage/portworx/components_test.go
```diff
@@ -1248,12 +1250,6 @@ func TestServiceAccountTokenRefreshOnExpire(t *testing.T) {
 			Name:      "px-cluster",
 			Namespace: "kube-test",
 		},
-		Spec: corev1.StorageClusterSpec{
```
ssz1997, line 1251: Not necessary

drivers/storage/portworx/components_test.go
```diff
@@ -1277,6 +1273,49 @@ func TestServiceAccountTokenRefreshOnExpire(t *testing.T) {
...
+	}
+	err = driver.PreInstall(cluster)
+	require.NoError(t, err)
+
+	saTokenSecret := &v1.Secret{}
+	err = testutil.Get(k8sClient, saTokenSecret, pxutil.PortworxServiceAccountTokenSecretName, cluster.Namespace)
+	require.NoError(t, err)
+	oldCaCrt := saTokenSecret.Data[v1.ServiceAccountRootCAKey]
+
+	err = os.WriteFile(caCrtPath, []byte("test"), 0644)
```
ssz1997, line 1307: If the test is ran in a k8s pod, this is effectively removing the root certificate, which I think is very dangerous.
Shall we just skip the test if this file exists?
Another solution would be introduce the monkey patch module into operator repo and patch the `os.ReadFile`.

drivers/storage/portworx/component/portworx_basic.go
```diff
@@ -557,23 +557,28 @@ func (c *portworxBasic) createAndMaintainPxSaTokenSecret(cluster *corev1.Storage
...
 		}
 	}
-	needRefresh, err := isTokenRefreshRequired(secret)
+	caCrtUpdated, err := updateCaCrtIfNeeded(secret)
 	if err != nil {
 		return err
 	}
-	if needRefresh {
-		if err := c.refreshTokenSecret(secret, cluster, ownerRef); err != nil {
+	needRefreshToken, err := isTokenRefreshRequired(secret)
```
pureneelesh, line 564: nit: add a updateTokenIfNeeded() wrapper to be consistent with updateCaCrtIfNeeded?

drivers/storage/portworx/component/portworx_basic.go
```diff
@@ -592,7 +596,33 @@ func (c *portworxBasic) createTokenSecret(cluster *corev1.StorageCluster, ownerR
 	return secret, nil
 }
 
-func (c *portworxBasic) refreshTokenSecret(secret *v1.Secret, cluster *corev1.StorageCluster, ownerRef *metav1.OwnerReference) error {
+func updateCaCrtIfNeeded(secret *v1.Secret) (bool, error) {
+	rootCaCrt, err := os.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/ca.crt")
+	if err != nil && !os.IsNotExist(err) {
+		return false, fmt.Errorf("error reading k8s cluster certificate located inside the pod at /var/run/secrets/kubernetes.io/serviceaccount/ca.crt: %w", err)
+	}
+	if len(secret.Data) == 0 || len(secret.Data[v1.ServiceAccountRootCAKey]) == 0 || !bytes.Equal(secret.Data[v1.ServiceAccountRootCAKey], rootCaCrt) {
```
pureneelesh, line 604: nit: are the len() checks really needed? bytes.Equal() should probably handle it?

drivers/storage/portworx/components_test.go
pureneelesh: How about making caCrtPath a variable and allow the test to set a non-default value for this unit test?

pureneelesh (approved):

drivers/storage/portworx/component/portworx_basic.go
ssz1997: yeah the equal should be able to

drivers/storage/portworx/component/portworx_basic.go
```diff
@@ -46,6 +46,8 @@ const (
 
 var (
 	defaultPxSaTokenExpirationSeconds = int64(12 * 60 * 60)
+	// Accessible for testing purpose
+	RootCaCrtPath = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
```
pureneelesh, line 50: Slightly cleaner way is to keep this variable private and add a function `setRootCaCrtPathForTesting(path string)` to set it.

pureneelesh (approved):

pureneelesh (approved):

PR #1607, opened 2024-07-12
PWX-38083 Fix metadata device spec format when FACD pod name doesn't exist

ssz1997:
When FACD pod name is not specified, for example, FACD with dmthin without multitenancy, the auto-added metadata device spec should not have extra comma at the end.

https://purestorage.atlassian.net/browse/PWX-38083


pp511 (approved):
lgtm

PR #1608, opened 2024-07-16
PWX-31632: Add telemetry secret expiry check

puremaxm:

Extending our `validateTelemetrySSLCert` func to check the expiry of the certificate too. Created two helper methods, one for doing the `pseudonym` check (this logic was already there) and one for doing the expiry check.

If the current time of the pod is within 12 hours of the expiry time of the certificate, we will return an error that results in `refreshTelemetrySSLCert` being called

If the current time of the pod is within 60 days of the expiry time, we will log a warning in the operators logs.

Closes PWX-31632

<I'm still investigating how to run tests>


puremaxm:
Closing as we decided this check was not necessary.

PR #1610, opened 2024-07-24
PWX-38202 : Fix NTP server issue by changing token issue time in past

nikita-bhatia:
The token authentication issued by operator fails on PX side sometimes due to future issued time with the error : Unable to authenticate token. Token used before issued.

The fix is done to adjust the issued time 10 minutes back from the current time, which is issued for next 24 hours. This will handle the case of a minor time drift.

Closes  https://purestorage.atlassian.net/browse/PWX-38202


diptiranjanpx (approved):

piyush-nimbalkar:
@nikita-bhatia can you cherrypick this to 24.2.0 as well?

PR #1619, opened 2024-08-01
PWX-38387: Disabling smart and parallel upgrade by default

svijaykumar-px:
**What this PR does / why we need it**: We will disable the non disruptive upgrades for kubernetes and portworx by default until more testing is done
1. **Portworx upgrades:**
By default `disruptions` field of RollingUpdate will be set to true. To enable the feature, user needs to set the value to false

2.  **Kubernetes Upgrades:**
By default no annotation will be added, and will be considered as the feature is disabled. The user can also add annotation `"portworx.io/disable-non-disruptive-upgrade":true` to disable the feature. To enable it, the user will need to add the annotation with the value false

### DIFF
--- a/drivers/storage/portworx/component/disruption_budget.go
+++ b/drivers/storage/portworx/component/disruption_budget.go
@@ -412,8 +412,12 @@ func (c *disruptionBudget) updateMinAvailableForNodePDB(cluster *corev1.StorageC
 	// Calculate the minimum number of nodes that should be available
 	quorumValue := int(math.Floor(float64(storageNodesCount)/2) + 1)
 	calculatedMinAvailable := quorumValue
+
+	// TODO: Once the feature has been tested well enough, change behavior to enable feature by default
+	// For now, setting default behavior to disable non-disruptive upgrades
+
 	// When non disruptive upgrades is disabled
-	if cluster.Annotations != nil && cluster.Annotations[pxutil.AnnotationsDisableNonDisruptiveUpgrade] == "true" {
+	if cluster.Annotations == nil || cluster.Annotations[pxutil.AnnotationsDisableNonDisruptiveUpgrade] == "" || cluster.Annotations[pxutil.AnnotationsDisableNonDisruptiveUpgrade] == "true" {
 		if cluster.Annotations[pxutil.AnnotationStoragePodDisruptionBudget] == "" {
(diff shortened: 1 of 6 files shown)


nikita-bhatia:
@piyush-nimbalkar @svijaykumar-px 
Since we have decided to disable the feqature by default, should we change it to make it simple as follow? Since it is always better to keep values as false by default and avoid too much of negations. Also, to keep it consistent with PX and k8s upgrade.

1. Portworx upgrades:
By default non-disruptions field of RollingUpdate will be set to false, which means it will not be smart upgrade by default. To enable the feature, user needs to set the value to true

```
spec:
  updateStrategy:
    type: RollingUpdate:
         non-disruptive:
           allow: false
```

2. Kubernetes Upgrades:
By default no annotation will be added, and will be considered as the feature is disabled. 
The user can add annotation **"portworx.io/non-disruptive-upgrade":true** to enable the feature. 

`portworx.io/non-disruptive-upgrade:false `

piyush-nimbalkar:
> @piyush-nimbalkar @svijaykumar-px Since we have decided to disable the feature by default, should we change it to make it simple as follow? Since it is always better to keep values as false by default and avoid too much of negations. Also, to keep it consistent with PX and k8s upgrade.
> 
> 1. Portworx upgrades:
>    By default non-disruptive field of RollingUpdate will be set to false, which means it will not be smart upgrade by default. To enable the feature, user needs to set the value to true
> 
> ```
> spec:
>   updateStrategy:
>     type: RollingUpdate:
>          non-disruptive:
>            allow: false
> ```
> 
> 2. Kubernetes Upgrades:
>    By default no annotation will be added, and will be considered as the feature is disabled.
>    The user can add annotation **"portworx.io/non-disruptive-upgrade":true** to enable the feature.
> 
> `portworx.io/non-disruptive-upgrade:false `

Even though we are planning to disable this feature, we will keep this feature enabled by default eventually. In that case, we will always have to set the flag to true in the spec. Also, it feels wrong to ask customers to set 'non-disruptive' to true. We should always be doing non disruptive upgrades.

piyush-nimbalkar:
@svijaykumar-px let's wait on merging this until Monday.

svijaykumar-px:
Closing the PR as it has been decided to enable the feature by default

PR #1621, opened 2024-08-05
PX Resource Gateway

dgoel-px:

Closes #


Makefile
```diff
@@ -298,3 +315,41 @@ clean: clean-release-manifest clean-bundle
 	@go clean -i $(PKGS)
 	@echo "Deleting image "$(OPERATOR_IMG)
 	@docker rmi -f $(OPERATOR_IMG) registry.access.redhat.com/ubi9-minimal:latest
+
+px-resource-gateway-proto-build-docker-proto:
+	docker build -t pure-artifactory.dev.purestorage.com/px-docker-dev-virtual/dgoel/px-resource-gateway-proto --network=host -f Dockerfile.proto .
```
lpabon, line 320: Remove

Makefile
```diff
@@ -298,3 +315,41 @@ clean: clean-release-manifest clean-bundle
...
+
+px-resource-gateway-docker-proto:
+	docker run \
+		--privileged --rm \
+		-v $(shell pwd):/go/src/github.com/libopenstorage/operator \
+		-e "GOPATH=/go" \
+		-e "DOCKER_PROTO=yes" \
+		-e "PATH=/bin:/usr/bin:/usr/local/bin:/go/bin:/usr/local/go/bin" \
+		pure-artifactory.dev.purestorage.com/px-docker-dev-virtual/dgoel/px-resource-gateway-proto \
```
lpabon, line 329: Remove

Makefile
```diff
@@ -298,3 +315,41 @@ clean: clean-release-manifest clean-bundle
 	@go clean -i $(PKGS)
 	@echo "Deleting image "$(OPERATOR_IMG)
 	@docker rmi -f $(OPERATOR_IMG) registry.access.redhat.com/ubi9-minimal:latest
+
+px-resource-gateway-proto-build-docker-proto:
+	docker build -t pure-artifactory.dev.purestorage.com/px-docker-dev-virtual/dgoel/px-resource-gateway-proto --network=host -f Dockerfile.proto .
+
+px-resource-gateway-docker-proto:
```
lpabon, line 322: Move this to a Makefile in the `proto` folder instead. Change it to a `proto:` target then have a `$(MAKE) -C proto` to execute it

lpabon (changes requested):
still reviewing...

Makefile
```diff
@@ -298,3 +315,41 @@ clean: clean-release-manifest clean-bundle
...
+	docker run \
+		--privileged --rm \
+		-v $(shell pwd):/go/src/github.com/libopenstorage/operator \
+		-e "GOPATH=/go" \
+		-e "DOCKER_PROTO=yes" \
+		-e "PATH=/bin:/usr/bin:/usr/local/bin:/go/bin:/usr/local/go/bin" \
+		pure-artifactory.dev.purestorage.com/px-docker-dev-virtual/dgoel/px-resource-gateway-proto \
+			make px-resource-gateway-proto
+
+px-resource-gateway-proto: $(GOPATH)/bin/protoc-gen-go $(GOPATH)/bin/protoc-gen-grpc-gateway $(GOPATH)/bin/protoc-gen-swagger
```
lpabon, line 332: Nice, but move to the Makefile in the `proto` folder

Makefile
lpabon, line 320: Use a name for the container like `px-resource-gateway-proto:dev` so that it is never pushed.

cmd/px-resource-gateway/px_resource_gateway.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+func run(c *cli.Context) {
+	logrus.SetLevel(logrus.DebugLevel)
+	semaphoreServer := server.NewServer()
+
+	// TODO: add security
+	config := &grpcServer.ServerConfig{
+		Name:         "px-resource-gate",
+		Address:      "127.0.0.1:9009",
```
lpabon, line 50: I'm not sure if this will work. Listening on 127.0.0.1 will force it to not accept connections from any other system and only from the local system. You want `:9009` instead.

cmd/px-resource-gateway/px_resource_gateway.go
```diff
@@ -0,0 +1,101 @@
...
+		fmt.Printf("Unable to start server: %v", err)
+		os.Exit(1)
+	}
+
+	// Wait. The signal handler will exit cleanly
+	logrus.Info("Px gRPC server running")
+	select {}
+}
+
+func gRPCInterceptor(
```
lpabon, line 92: Remove this unless you need it.

lpabon (changes requested):
continuing...

pkg/px-resource-gateway/priority_queue.go
```diff
@@ -0,0 +1,82 @@
+package server
```
lpabon, line 1: although not incorrect, it is confusing to have the package not be the name of the directory. I recommend creating a `server` directory under `px-resource-gateway` and placing these files there.

pkg/px-resource-gateway/priority_queue.go
```diff
@@ -0,0 +1,82 @@
...
+	return &priorityQueue{
+		queues: queues,
+	}
+}
+
+func (pq *priorityQueue) Enqueue(clientId string, priority pb.SemaphoreAccessPriority_Type) {
+	pq.Lock()
+	defer pq.Unlock()
+
+	fmt.Printf("Enqueueing client %v with priority %v\n", clientId, priority)
```
lpabon, line 46: Remove all Printfs

pkg/px-resource-gateway/priority_queue.go
```diff
@@ -0,0 +1,82 @@
...
+
+type PriorityQueue interface {
+	Enqueue(clientId string, priority pb.SemaphoreAccessPriority_Type)
+	Dequeue(priority pb.SemaphoreAccessPriority_Type)
+	Front() (string, pb.SemaphoreAccessPriority_Type)
+	Remove(clientId string)
+}
+
+type priorityQueue struct {
+	sync.Mutex
```
lpabon, line 27: interesting style also. Never seen it. Normally it is a variable in the struct and not an embedding. Ok, by me, just haven't seen it.

pkg/px-resource-gateway/priority_queue.go
```diff
@@ -0,0 +1,82 @@
...
+	// using list.List for effecicient implementation of queues
+	queues map[pb.SemaphoreAccessPriority_Type]*list.List
+}
+
+func NewPriorityQueue() PriorityQueue {
+	queues := map[pb.SemaphoreAccessPriority_Type]*list.List{}
```
lpabon, line 33: style: normally `queues := make(map[pb.SemaphoreAccessPriority_Type]*list.List)`, but what you have works.

pkg/px-resource-gateway/px_resource_gateway.go
```diff
@@ -0,0 +1,84 @@
...
+		k8s:         core.Instance(),
+		semaphorePQ: NewSemaphorePriorityQueue(),
+	}
+	return s
+}
+
+func (s *server) AcquireLock(ctx context.Context, req *pb.AcquireLockRequest) (*pb.AcquireLockResponse, error) {
+	// validate request
+	if req.ResourceId == "" {
+		return &pb.AcquireLockResponse{}, status.Error(codes.InvalidArgument, "Resource ID is required")
```
lpabon, line 33: Nicely done. You can return `nil` instead of an empty struct.

pkg/px-resource-gateway/px_resource_gateway.go
```diff
@@ -0,0 +1,84 @@
...
+func NewServer() *server {
+	s := &server{
+		k8s:         core.Instance(),
+		semaphorePQ: NewSemaphorePriorityQueue(),
+	}
+	return s
+}
+
+func (s *server) AcquireLock(ctx context.Context, req *pb.AcquireLockRequest) (*pb.AcquireLockResponse, error) {
+	// validate request
```
lpabon, line 31: check if `req` is nil

pkg/px-resource-gateway/px_resource_gateway.go
```diff
@@ -0,0 +1,84 @@
...
+	if req.ResourceId == "" {
+		return &pb.AcquireLockResponse{}, status.Error(codes.InvalidArgument, "Resource ID is required")
+	}
+	if req.ClientId == "" {
+		return &pb.AcquireLockResponse{}, status.Error(codes.InvalidArgument, "Client ID is required")
+	}
+	accessPriority := req.AccessPriority
+	if accessPriority == pb.SemaphoreAccessPriority_TYPE_UNSPECIFIED {
```
lpabon, line 39: I would highly discourage default values for priority. Make the client be clear and if they dont pass a priority return an error that the priority must be set

proto/pxresourcegateway.proto
```diff
@@ -0,0 +1,105 @@
...
+  // ReleaseLock releases the semaphore lock on the resource
+  rpc ReleaseLock (ReleaseLockRequest)
+  returns (google.protobuf.Empty) {
+    option (google.api.http) = {
+        post : "/v1/pxe/pxresourcegateway:releaseLock"
+        body : "*"
+    };
+  }
+
+  // KeepAlive sends a heartbeat to the semaphore service to keep the lock alive
```
lpabon, line 37: This needs a timeout. Keep alive for "how long"? for example 5 secs? etc.

pkg/px-resource-gateway/px_resource_gateway.go
```diff
@@ -0,0 +1,84 @@
...
+
+	// process request to release lock
+	err := s.semaphorePQ.ReleaseLock(req.ClientId)
+	if err != nil {
+		return &emptypb.Empty{}, status.Error(codes.Internal, err.Error())
+	}
+	return &emptypb.Empty{}, nil
+}
+
+func (s *server) KeepAlive(ctx context.Context, req *pb.KeepAliveRequest) (*emptypb.Empty, error) {
```
lpabon, line 72: This function is confusing on the side effects and expectations. I added a comment to the `.proto` file on this rpc. Maybe add more comments on how this function works and expectations as a comment on the .proto file

pkg/px-resource-gateway/semaphore.go
```diff
@@ -0,0 +1,310 @@
...
+	lockId := s.availableLocks[0]
+	s.availableLocks = s.availableLocks[1:]
+	return lockId
+}
+
+func (s *semaphorePriorityQueue) tryLock(clientId string) bool {
+	// check if the node is at the front of the queue
+	nextResouceId, priority := s.priorityQ.Front()
+	if nextResouceId == "" {
+		panic("Queue is empty")
```
lpabon, line 191: Can we do something else other than panic here? I would suggest handling this situation

pkg/px-resource-gateway/semaphore.go
```diff
@@ -0,0 +1,310 @@
+package server
```
lpabon, line 1: Please explain the algorithm as comment header in this file to help maintainers (and me as the reviewer)

lpabon: I'm skipping this file until the algorithm is included.

lpabon (changes requested):
Please also add unit tests to all of these files. You also need e2e tests: create small grpc servers and have a client send requests to tests the full paths.

pkg/px-resource-gateway/px_resource_gateway.go
lpabon: actually. instead use all the `GetXXX()` accessors from the gRPC proto output. They already manage `nil`. All the values have a `GetXXX()` function. For example use `req.GetResourceId()`, etc.

drivers/storage/portworx/component/px_resource_gateway.go
```diff
@@ -0,0 +1,444 @@
...
+	rules := []rbacv1.PolicyRule{
+		{
+			APIGroups: []string{""},
+			Resources: []string{"configmaps"},
+			Verbs:     []string{"create", "get", "list", "patch", "update", "watch"},
+		},
+	}
+	return k8sutil.CreateOrUpdateClusterRole(
+		p.k8sClient,
+		&rbacv1.ClusterRole{
```
piyush-nimbalkar, line 162: Is there is a need to use cluster role instead of role? We should restrict the permissions to our components whenever possible. At least the new ones that we are adding now.

I believe we can create all the configmaps in the namespace of the storage cluster (where all px components are deployed). We should not even try to pollute other namespaces.

drivers/storage/portworx/component/px_resource_gateway.go
```diff
@@ -0,0 +1,444 @@
...
+	maxUnavailable := intstr.FromInt(1)
+	maxSurge := intstr.FromInt(1)
+	imagePullPolicy := pxutil.ImagePullPolicy(cluster)
+
+	deployment := &appsv1.Deployment{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      PxResourceGatewayDeploymentName,
+			Namespace: cluster.Namespace,
+			Annotations: map[string]string{
+				"scheduler.alpha.kubernetes.io/critical-pod": "",
```
piyush-nimbalkar, line 345: We don't need this. K8s doesn't look at this anymore. Instead they use priority class.

drivers/storage/portworx/component/px_resource_gateway.go
```diff
@@ -0,0 +1,444 @@
...
+	}
+	targetCPUQuantity, err := resource.ParseQuantity(targetCPU)
+	if err != nil {
+		return err
+	}
+
+	targetCPULimitQuantity, err := resource.ParseQuantity(defaultPxResourceGatewayCPULimit)
+	if err != nil {
+		return err
+	}
```
piyush-nimbalkar, line 230: Let's remove this. We can directly use the resources from the spec. We don't need the annotations. This was an old way that we used for some of the components.

drivers/storage/portworx/component/px_resource_gateway.go
```diff
@@ -0,0 +1,444 @@
...
+
+	args := map[string]string{}
+	for k, v := range cluster.Spec.PxResourceGateway.Args {
+		if _, exists := autopilotConfigParams[k]; exists {
```
piyush-nimbalkar, line 234: We don't need this. It seems to be copy pasted.

drivers/storage/portworx/component/px_resource_gateway.go
```diff
@@ -0,0 +1,444 @@
...
+			break
+		}
+	}
+
+	targetDeployment := p.getPxResourceGatewayDeploymentSpec(cluster, ownerRef, imageName,
+		command, targetCPUQuantity, targetCPULimitQuantity)
+
+	// Check if the deployment has changed
+	modified := existingImage != imageName ||
+		!reflect.DeepEqual(existingCommand, command) ||
```
piyush-nimbalkar, line 270: Instead of comparing each and every part of the deployment spec, we made some recent changes to compare the pod template directly. 
Check this out for reference - https://github.com/libopenstorage/operator/blob/release-24.1.2-drw/drivers/storage/portworx/component/csi.go#L462

piyush-nimbalkar (changes requested):
Please add UTs to the resource gateway installation and the actual server logic too before you merge the PR.

piyush-nimbalkar:
I have looked at only the deployment logic. I am assuming @nrevanna and @lpabon are looking at the actual server implementation.

pkg/px-resource-gateway/semaphore.go
```diff
@@ -0,0 +1,302 @@
...
+				return true
+			}
+		}
+		return false
+	}
+
+	data := map[string]string{} // reset the data
+	for clientId, lockId := range s.locks {
+		data[clientId] = fmt.Sprintf("%d", lockId)
+	}
```
sangleganesh, line 227: isn't s.locks protected with the mutex ?

drivers/storage/portworx/component/px_resource_gateway.go
dgoel-px: Keeping args because we have the debug flag and might add more later.

cmd/px-resource-gateway/px_resource_gateway.go
```diff
@@ -0,0 +1,111 @@
...
+	signal_handler := util.NewSigIntManager(func() {
+		pxResourceGatewayServer.Stop()
+		os.Remove(PxResourceGatewayServerSocket)
+		os.Exit(0)
+	})
+	err = signal_handler.Start()
+	if err != nil {
+		fmt.Printf("Unable to start signal handler: %v", err)
+		os.Exit(1)
+	}
```
adityadani: Also setup memory and cpu profiling, and then you can add a signal handler for SIGUSR1 to dump the heap and stack profile - https://github.com/libopenstorage/operator/blob/303cc3735024cf33d14956236142dfe215bc4577/cmd/operator/operator.go#L148

drivers/storage/portworx/util/util.go
```diff
@@ -109,6 +109,8 @@ const (
 	AnnotationPVCControllerSecurePort = pxAnnotationPrefix + "/pvc-controller-secure-port"
 	// AnnotationAutopilotCPU annotation for overriding the default CPU for Autopilot
 	AnnotationAutopilotCPU = pxAnnotationPrefix + "/autopilot-cpu"
+	// AnnotationPxResourceGatewayCPU annotation for overriding the default CPU for PxResourceGateway
+	AnnotationPxResourceGatewayCPU = pxAnnotationPrefix + "/px-resource-gateway-cpu"
```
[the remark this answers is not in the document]
adityadani: Do we need this annotation? I see a couple of valid comments which are resolved w/o a comment. If you had an offline conversation can you mention them here.

pkg/px-resource-gateway/px_resource_gateway.go
```diff
@@ -0,0 +1,79 @@
...
+
+type semaphoreServer struct {
+	k8s         core.Ops
+	semaphorePQ SemaphorePriorityQueue
+}
+
+// TODO implement semaphore map for different resource ids
+func NewSemaphoreServer() *semaphoreServer {
+	s := &semaphoreServer{
+		k8s:         core.Instance(),
```
adityadani: No need of storing the core.Instance() object since every Instance() execution will check the validity of the k8s client.

pkg/px-resource-gateway/px_resource_gateway.go
```diff
@@ -0,0 +1,79 @@
...
+		semaphorePQ: NewSemaphorePriorityQueue(),
+	}
+	return s
+}
+
+func (s *semaphoreServer) AcquireLock(ctx context.Context, req *pb.AcquireLockRequest) (*pb.AcquireLockResponse, error) {
+	// validate request
+	if req.GetResourceId() == "" {
```
adityadani: Is resourceId not used anywhere?

pkg/px-resource-gateway/semaphore.go
```diff
@@ -0,0 +1,302 @@
...
+	pb "github.com/libopenstorage/operator/proto"
+	"github.com/portworx/sched-ops/k8s/core"
+	"github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	K8sNamespace         = "kube-system"
```
adityadani: We have a backlog item to move all our config maps out of kube-system namespace. Can you instead create this new config map in the namespace where Portworx is deployed

pkg/px-resource-gateway/semaphore.go
```diff
@@ -0,0 +1,302 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      ConfigMapName,
+			Namespace: K8sNamespace,
+			Labels:    labels,
+		},
+		Data: map[string]string{},
+	}
+	_, err := core.Instance().CreateConfigMap(configMap)
+	if err != nil && !k8s_errors.IsAlreadyExists(err) {
+		panic(fmt.Sprintf("Failed to create configmap %s in namespace %s: %v", ConfigMapName, K8sNamespace, err))
```
adityadani: Instead of panic, can you return an error here and let the main() exit out gracefully?

pkg/px-resource-gateway/semaphore.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+
+	// get the latest copy of configmap
+	configMap, err = core.Instance().GetConfigMap(ConfigMapName, K8sNamespace)
+	if err != nil {
+		panic(fmt.Sprintf("Failed to get configmap %s in namespace %s: %v", ConfigMapName, K8sNamespace, err))
```
adityadani: same here

pkg/px-resource-gateway/semaphore.go
```diff
@@ -0,0 +1,302 @@
...
+
+const (
+	K8sNamespace         = "kube-system"
+	ConfigMapName        = "px-bringup-configmap"
+	NULL_LOCK     uint32 = 0
+)
+
+var (
+	NLocks                  = 10
+	ConfigMapUpdateInterval = time.Second * 1
```
adityadani: IMO 1 sec is still too aggressive for a single client to invoke a k8s api. In comparison, our default operator reconcilor will wake up every 30s and then make a bunch of k8s apis if required. 

If the goal is to just persist the locks so that if this pod crashes we can recover then isn't 30s good enough?

pkg/px-resource-gateway/semaphore.go
```diff
@@ -0,0 +1,302 @@
...
+		return NULL_LOCK
+	}
+	lockId := s.availableLocks[0]
+	s.availableLocks = s.availableLocks[1:]
+	return lockId
+}
+
+func (s *semaphorePriorityQueue) tryLock(clientId string) bool {
+	// check if the node is at the front of the queue
+	nextResouceId, priority := s.priorityQ.Front()
```
adityadani: Rename this to nextClientId ?

pkg/px-resource-gateway/semaphore.go
```diff
@@ -0,0 +1,302 @@
...
+		}
+		return false
+	}
+
+	data := map[string]string{} // reset the data
+	for clientId, lockId := range s.locks {
+		data[clientId] = fmt.Sprintf("%d", lockId)
+	}
+
+	if !isUpdateRequired(data, s.configMap.Data) {
```
adityadani: Is the goal to re-initialize the s.locks from s.configMap.Data when this process starts? 
Right now, I don't see this config map's contents being used anywhere

pkg/px-resource-gateway/px_resource_gateway.go
```diff
@@ -0,0 +1,79 @@
...
+		return &pb.KeepAliveResponse{}, status.Error(codes.InvalidArgument, "Resource ID is required")
+	}
+	if req.GetClientId() == "" {
+		return &pb.KeepAliveResponse{}, status.Error(codes.InvalidArgument, "Client ID is required")
+	}
+
+	// process request to keep alive
+	s.semaphorePQ.KeepAlive(req.ClientId)
+	return &pb.KeepAliveResponse{}, nil
```
adityadani: Should we return a response here indicating whether the client still holds the lock? In this way the clients will get to know if they still have the lock. Otherwise we will have many clients assuming they have the lock even though they did not send a keep alive.

pkg/px-resource-gateway/semaphore.go
```diff
@@ -0,0 +1,302 @@
...
+			s.release(clientId, lockId)
+			removedNode[clientId] = ""
+		}
+	}
+
+	for _, clientId := range deadNodes {
+		if _, ok := removedNode[clientId]; ok { // already removed
+			continue
+		}
+		s.priorityQ.Remove(clientId)
```
adityadani: I am missing something here, why not remove a dead node unconditionally from the priorityQueue if it was part of s.locks ?

drivers/storage/portworx/util/util.go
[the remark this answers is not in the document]
dgoel-px: I resolved the comments because I had addressed them locally. I have pushed the changes now.

pkg/px-resource-gateway/px_resource_gateway.go
dgoel-px: Addressed

pkg/px-resource-gateway/px_resource_gateway.go
dgoel-px: Addressed

pkg/px-resource-gateway/semaphore.go
dgoel-px: Addressed

pkg/px-resource-gateway/semaphore.go
dgoel-px: Addressed

pkg/px-resource-gateway/px_resource_gateway.go
dgoel-px: Addressed

PR #1622, opened 2024-08-06
PWX-38103 Ignore unhealthy nodes if a healthy one is found

piyush-nimbalkar:
for the same k8s node. If one healthy node is found then there is a PX node which is healthy on that k8s node. The unhealthy ones are present because they were previously running on that k8s node before. Such nodes will either find a new k8s node or will get auto-decommissioned in case of a storageless nodes.

Closes # https://purestorage.atlassian.net/browse/PWX-38103

- UTs added
- Manually tested with a local mode vsphere setup where storageless nodes become new storageless nodes on restart.

### DIFF
--- a/pkg/controller/storagecluster/update.go
+++ b/pkg/controller/storagecluster/update.go
@@ -661,7 +661,19 @@ func (c *Controller) getPodAvailability(
 			if len(storageNode.SchedulerNodeName) == 0 {
 				nonK8sStorageNodes = append(nonK8sStorageNodes, storageNode)
 			} else {
-				storageNodeMap[storageNode.SchedulerNodeName] = storageNode
+				if existing, ok := storageNodeMap[storageNode.SchedulerNodeName]; ok {
+					// During a k8s upgrade, it may happen that a new node is started on the same k8s host using
+					// a scheduler node name that was previously used by a different node. In this case, if we have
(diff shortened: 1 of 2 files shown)


pkg/controller/storagecluster/update.go
```diff
@@ -661,7 +661,17 @@ func (c *Controller) getPodAvailability(
 			if len(storageNode.SchedulerNodeName) == 0 {
 				nonK8sStorageNodes = append(nonK8sStorageNodes, storageNode)
 			} else {
-				storageNodeMap[storageNode.SchedulerNodeName] = storageNode
+				if existing, ok := storageNodeMap[storageNode.SchedulerNodeName]; ok {
+					// During a k8s upgrade, it may happen that a new node is started on the same k8s host using
```
nrevanna, line 665: Shouldn't we verify that it's storageless and not do this for storage node?

nrevanna (approved):
LGTM

pkg/controller/storagecluster/update.go
piyush-nimbalkar: Thanks for pointing this out @nrevanna .

After the offline discussion, I have added a comment and added more details for the scenario in https://purestorage.atlassian.net/browse/PWX-38505. We will handle this later if we find out this scenario is possible and we can reproduce it.

PR #1635, opened 2024-08-09
PWX-38512 Skip token refresh verification if host-pid not enabled

ssz1997:
host-pid is required to run nsenter with runc. nsenter without runc is fine is the reason why other verifications are fine.

Closes PWX-38512

A full set of integration test is being run at https://jenkins.pwx.dev.purestorage.com/job/Operator/view/Operator%20Release%20Dashboard/job/op-basic-hotfix/

### DIFF
--- a/pkg/util/test/util.go
+++ b/pkg/util/test/util.go
@@ -1847,6 +1847,13 @@ func validatePortworxTokenRefresh(cluster *corev1.StorageCluster, timeout, inter
 	}
 	pidEnabled, err := strconv.ParseBool(cluster.Annotations["portworx.io/host-pid"])
 	if err != nil || !pidEnabled {
+		pxSaSecret, err := coreops.Instance().GetSecret(pxSaTokenSecretName, cluster.Namespace)
+		if err != nil {
+			return fmt.Errorf("px serviceaccount token validation failed. Unable to get px serviceaccount secret. Err: %w", err)
+		}
+		if len(pxSaSecret.Data[core.ServiceAccountTokenKey]) == 0 {
+			return fmt.Errorf("px serviceaccount token validation failed. Token doesn't exist or length is 0")
+		}
 		logrus.Infof("Annotation `host-pid: true` is required for verifying token refresh because we need to run command inside px runc container. Thus Skipping verification.")
 		return nil
 	}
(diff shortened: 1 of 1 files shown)


nikolaypopov:
Do you think we should maybe check for annotation and not validate the actual token, if the annotation is missing, but maybe we should still validate up to the point that the secret exists, at least we will be sure that it gets created?

pkg/util/test/util.go
```diff
@@ -1845,6 +1845,18 @@ func validatePortworxTokenRefresh(cluster *corev1.StorageCluster, timeout, inter
 		logrus.Infof("pxVersion: %v, opVersion: %v. Skip verification because px token refresh is not supported with these versions.", pxVersion, opVersion)
 		return nil
 	}
+	pidEnabled, err := strconv.ParseBool(cluster.Annotations["portworx.io/host-pid"])
+	if err != nil || !pidEnabled {
+		pxSaSecret, err := coreops.Instance().GetSecret(pxSaTokenSecretName, cluster.Namespace)
+		if err != nil {
+			return fmt.Errorf("px serviceaccount token validation failed. Unable to get px serviceaccount secret. Err: %w", err)
```
nikolaypopov, line 1852: I would suggest to print errors like this, from my experience, it helps later on to debug simple things just from logs
```suggestion
			return fmt.Errorf("failed to get px serviceaccount secret [%s] in namespace [%s]. Err: %w", pxSaTokenSecretName, cluster.Namespace, err)
```

pkg/util/test/util.go
```diff
@@ -1845,6 +1845,18 @@ func validatePortworxTokenRefresh(cluster *corev1.StorageCluster, timeout, inter
...
+		}
+		if len(pxSaSecret.Data[core.ServiceAccountTokenKey]) == 0 {
```
nikolaypopov, line 1854: Also we can probably just do this, less heavier operation
```suggestion
		if pxSaSecret.Data[core.ServiceAccountTokenKey] == "" {
```

nikolaypopov (approved):
LGTM with few nit comments

pkg/util/test/util.go
ssz1997: It's a byte array, not a string, so checking the length would be better in this case?

nikita-bhatia (approved):
LGTM

PR #1639, opened 2024-08-12
[ Draft ] [ToBeDeleted] Release 24.1.2 drw changes

sshreyas-px:

Closes #


deploy/olm-catalog/portworx/portworx.package.yaml
```diff
@@ -1,7 +1,7 @@
 packageName: portworx-certified
 channels:
 - name: alpha
-  currentCSV: portworx-operator.v24.1.1
+  currentCSV: portworx-operator.v24.1.2-drw
```
sshreyas-px, line 4: @nikita-bhatia , I believe this should be changed into `portworx-operator.v24.1.3` or when the release is happening. 
Since this is part of olm, can you suggest what should be the value in the PR

PR #1643, opened 2024-08-14
PWX-38583 : fix nil pointer derefernce in operator pod

nikita-bhatia:
**What this PR does / why we need it**: Operator pod crashes with nil pointer dereference error while upgrading PX from 3.1.1 to 3.1.2 
because, cluster.Spec.UpdateStrategy.RollingUpdate.Disruption.Allow, becuase when Allow is *bool and the value was not set, it was nil and trying to dereference a nil pointer resulted in a nil pointer dereference error.

As a fix, changed  Allow  bool instead of *bool, since the value can be either true/false.

Closes # https://purestorage.atlassian.net/browse/PWX-38583

Verified on vanilla k8s


svijaykumar-px (approved):
LGTM. This works, tested on upgrade job

PR #1646, opened 2024-08-15
PWX-38118: Health Checks

lpabon:
Adds environment health checks when a new StorageCluster is installed.

Closes #

There are two major sections in this PR:
1. The health check framework is under `px/px-health-check`. This framework will be removed from this repo and moved to its own repo at a later date so that other programs can include it easily. You may ignore it for the review if you want.
2. The health checks and integration with the Operator. When the reconciler starts, it will check if the StorageCluster object has the `portworx.io/health-check` annotation:

* if the annotation is set to `skip`, then the Health Checks will not run.
* if the annotation is missing, then the Health Checks will execute. It will then set the annotation to `false` if it failed, or `passed` if it worked. If it fails the test, it will also return failure disallowing PxE to be installed.
* if the annotation is set to `false`, it will again return an error disallowing PxE to be installed.
* if the annotation is set to `passed`, it will not run.

Ignore the test coverage of `px/px-health-check` since that will be moved to a new repo after.


lpabon:
# Testing

## Positive Testing
1. I used the file `pxe.yml` attached to make sure it can install without error. 
2. Also asserted that the annotation `portworx.io/health-check` is set to `passed`.

## Negative Testing
1. Remove and uninstalled the previous installation
2. I have a k8s with three nodes and a control node. I labeled one of the nodes with `px/enabled=false`.
3. I used the file `pxe-affinity.yml` to create a cluster.
4. Assert that PxE is not installed
5. Assert that describing the StorageCluster provides information about the failure and provides a link to the documentation: https://docs.portworx.com/codes/pxe/1004

lpabon:
[test-storageclusters.zip](https://github.com/user-attachments/files/16627192/test-storageclusters.zip)
These are the test yamls

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -788,6 +804,111 @@ func (c *Controller) miscCleanUp(cluster *corev1.StorageCluster) error {
 	return nil
 }
 
+func (c *Controller) preInstallHealthChecks(cluster *corev1.StorageCluster) error {
+
+	// Determine if we should run the health checks
+	// If not set, then run the health checks
+	check := cluster.Annotations[pxutil.AnnotationHealthCheck]
+	check = strings.TrimSpace(strings.ToLower(check))
+	if check == "skip" || check == "passed" {
```
sshreyas-px: strings.Equalfold can be used here

sshreyas-px: also can values be a constant and  this be part of health check framework

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -788,6 +804,111 @@ func (c *Controller) miscCleanUp(cluster *corev1.StorageCluster) error {
...
+	}
+
+	errorMsg := fmt.Errorf(
+		"preflight health check failed for StorageCluster %v/%v. "+
+			"Please rectify the issue then remove annotation %s to retry the tests",
+		cluster.Namespace,
+		cluster.Name,
+		pxutil.AnnotationHealthCheck,
+	)
+	if check == "failed" {
```
sshreyas-px: Question: 
if healthCheck fail once, should we always keep failing ?  
Additionally, asking the user to remove the annotation would be a manual intervention and would not fall in line with Helm chart approaches. 
Also, the instructions of removing the annotation , would it be documented ? or is it expected that the user goes through the logs or would it be part of the STC status ? 

Events will be cleaned up after few mins

pkg/controller/storagecluster/storagecluster.go
```diff
@@ -807,7 +928,14 @@ func (c *Controller) syncStorageCluster(
 			cluster.Namespace, cluster.Name, err)
 	}
 
-	// If preflight failed, or previous check failed, reconcile would stop here until issues got resolved
+	// Run healthchecks
+	// 1. On first install, run HC in same context
+	// 2. Fail if HC fails
+	// 3. An annotation is created to save the results of the health check
+	if err := c.preInstallHealthChecks(cluster); err != nil {
```
sshreyas-px: SyncStorageCluster is a function which is called as part of the reconciler , adding healthChecks as part of a reconcile loop would add to the latency. 
Would it be possible to decouple this ?

I believe pre-checks should be done before the controllers are started. Please correct me if my understanding is wrong

pkg/controller/storagecluster/storagecluster.go
lpabon: `strings.Equalfold can be used here` Didn't know about this call! Nice catch

PR #1648, opened 2024-08-16
PWX-38596 : After disabling PX Security STC goes into DEGRADED state due to rolling update failed

nikita-bhatia:
After disabling PX Security STC goes into DEGRADED state due to rolling update failed: cannot upgrade nodes in parallel: failed to get non overlapping nodes: rpc error: code = PermissionDenied desc = Access denied without authentication token

Reason : When security was disabled from previously enabled value, the update on storagenodes is still not done at the point when API call is made to PX to get storage nodes, so even thought security is disabled from STC, storagenode is still not in sync and expects token to be passed by the operator during GRPC call.

Note that `skipSecurityCheck' variable is removed which was previously added in https://github.com/libopenstorage/operator/pull/1539 for improvement of logic as discussed in PR.

**Fix**
The code change does following to decide if token is required or not to for API calls.
- fetches the secret
- if secret is not found and security is disabled on STC, proceeds without token.
- if secret is not found and security is enabled on STC, returns error because secret is expected in this case

Closes # 
https://purestorage.atlassian.net/browse/PWX-38596
https://purestorage.atlassian.net/browse/PWX-37232

- verified on vanilla k8s cluster, by updating the operator image with this fix, the STC comes out of Degraded state and all nodes were upgraded

### DIFF
--- a/drivers/storage/portworx/util/util.go
+++ b/drivers/storage/portworx/util/util.go
@@ -79,8 +79,6 @@ const (
 	EssentialsOSBEndpointKey = "px-osb-endpoint"
 	// EnvKeyKubeletDir env var to set custom kubelet directory
 	EnvKeyKubeletDir = "KUBELET_DIR"
-	// Dummy Secret value for authentication when Security is disabled
-	DummySecretValue = "dummy-secret-value"
 
 	// AnnotationIsPKS annotation indicating whether it is a PKS cluster
 	AnnotationIsPKS = pxAnnotationPrefix + "/is-pks"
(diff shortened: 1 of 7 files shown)


drivers/storage/portworx/util/util.go
```diff
@@ -955,29 +953,26 @@ func SecurityEnabled(cluster *corev1.StorageCluster) bool {
...
+	if err != nil && !errors.IsNotFound(err) {
+		return nil, err
 	}
 
-	if pxAppsSecret == "" {
+	// if secret is not found, check if security is enabled
+	if errors.IsNotFound(err) {
+		if SecurityEnabled(cluster) {
+			// when security is enabled in STC, and secret is not found, return error
+			return nil, fmt.Errorf("failed to get portworx apps secret: %v", err.Error())
```
pureneelesh, line 969: nit: include namespace/name of the secret in the error message
Also, use `("...: %w", err)`

pureneelesh (approved):

pureneelesh:
If they disable security AND delete the secret, we will still run into an issue?

drivers/storage/portworx/util/util.go
nikolaypopov: Just wondering why is this secret called apps secret? Shouldn't it just be a security token secret or security context secret?

nikita-bhatia: Since we are specifically trying to fetch **apps-secret** key here inside **px-system-secrets** secret, so the error message is specifying that.

nikita-bhatia:
> If they disable security AND delete the secret, we will still run into an issue?

Ideally No, because 
- since when security is disabled, automatically secret is not deleted and if it is deleted manually then the workaround would be to delete enable the security and disable again since it is not a valid scenario.
> I tried deleting the secret manually but all the nodes update went fine. This might depend upon the condition what is happening first.

So to conclude, in a valid cluster state, this would not create problem.

PR #1663, opened 2025-02-24
Add topology spread constraint to kvdb for specific distribution of kv store

jmcshane:
While kvdb labeling can enable us to force the kvdb pods onto specific nodes,
this creates an issue when those nodes fail. To resolve this, topology spread constraints allow
Kubernetes users to spread these pods across specific failure domains, whether they are
zones, racks, or other user defined topologies.

Add topologySpreadConstraint to kvdb configuration


jmcshane:
The scheduling control is handled earlier in the operator, so this does not do anything. Closing

